PR #193, opened 2016-09-01
Documenting options in launch.json

AndrewBrianHall:
@pieandcakes @delmyers @jacdavis 

Next attempt at documenting the launch.json.  Took feedback from Pierson about trying to organize it more logically and providing some samples

### DIFF
--- a/launch.md
+++ b/launch.md
@@ -79,7 +79,7 @@ You can change the behavior of GDB or LLDB by setting the following options.
 
 **MIMode**: Indicates the debugger that VS Code will connect to. Must be set to `gdb` or `lldb`.  This is pre-configured on a per-operating system basis and can be changed as needed.
 
-**miDebuggerPath**: The path to the debugger (such as gdb). When unspecified, it will search the operating system's PATH variable first for the debugger. default": "/usr/bin/gdb".
+**miDebuggerPath**: The path to the debugger (such as gdb). When unspecified, it will search the operating system's PATH variable for a debugger (GDB on Linux and Windows, LLDB on OS X).
 
 **stopAtEntry**: If set to true, the debugger should stop at the entrypoint of the target (ignored on attach). Default value is `false`.
 
@@ -122,7 +122,8 @@ The C/C++ extension enables debugging dump files on Windows and core dump files 
 
 **dumpPath**: If you want to debug a Windows dump file, set this to the path to the dump file to start debugging in the `launch` configuration.
 
-**coreDumpPath**: Full path to a core dump file to debug for the specified program.  Set this to the path to the core dump file to start debugging in the `launch` configureation. 
+**coreDumpPath**: Full path to a core dump file to debug for the specified program.  Set this to the path to the core dump file to start debugging in the `launch` configuration. 
+_Note: core dump debuging is not supported with MinGw._  
 
 ## Remote debugging or debugging with a local debugger server
 
--- a/launch.md
+++ b/launch.md
@@ -1,0 +1,150 @@
+# Configuring launch.json for C/C++ debugging
+
+The `launch.json` file is used to configure the debugger in Visual Studio Code. 
+
+Visual Studio Code generates a `launch.json` with almost all of the required information.  To get started
+debugging you need to fill in the `program` field with the path to the executable you plan to debug. This must be specified for 
+both the launch and attach (if you plan to attach to a running instance at any point) configurations.
+
+The generated file contains two sections.  One that configures debugging for launch and a second that configures debugging for attach.
+
+# Configure VS Code's debugging behavior
+
+Set or change the following options to control VS Code's behavior during debugging
+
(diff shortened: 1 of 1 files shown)


launch.md
```diff
@@ -0,0 +1,150 @@
+# Configuring launch.json for C/C++ debugging
+
+The `launch.json` file is used to configure the debugger in Visual Studio Code. 
+
+Visual Studio Code generates a `launch.json` with almost all of the required information.  To get started
+debugging you need to fill in the `program` field with the path to the executable you plan to debug. This must be specified for 
+both the launch and attach (if you plan to attach to a running instance at any point) configurations.
```
jacdavis, line 7: This is actually only true for gdb/llb in the attach case. Of course, right now, we have a bug that concord is requiring it for attach.

launch.md
```diff
@@ -0,0 +1,150 @@
...
+}
+```
+
+## Customizing GDB or LLDB 
+
+You can change the behavior of GDB or LLDB by setting the following options.
+
+**MIMode**: Indicates the debugger that VS Code will connect to. Must be set to `gdb` or `lldb`.  This is pre-configured on a per-operating system basis and can be changed as needed.
+
+**miDebuggerPath**: The path to the debugger (such as gdb). When unspecified, it will search the operating system's PATH variable first for the debugger. default": "/usr/bin/gdb".
```
jacdavis: since we search the path, the default really isn't /usr/bin/gdb any more. I wonder if we should just remove the default on this one? In the schema, this is actually the generated value if the user decides to add the field.

jacdavis:
Only had two comments. Rest looks good to me.

launch.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## Debugging dump files
+
+The C/C++ extension enables debugging dump files on Windows and core dump files Linux and OS X.
+
+**dumpPath**: If you want to debug a Windows dump file, set this to the path to the dump file to start debugging in the `launch` configuration.
+
+**coreDumpPath**: Full path to a core dump file to debug for the specified program.  Set this to the path to the core dump file to start debugging in the `launch` configureation. 
+
```
delmyers, line 127: Note that you can't debug dump files with mingw?

launch.md
```diff
@@ -0,0 +1,150 @@
...
+**serverLaunchTimeout**: Time in milliseconds, for the debugger to wait for the debugServer to start up. Default is 10000.
+
+## Additional properties
+
+**processId**: Defaults to `${command.pickProcess}` which will display a list of available processes the debugger can attach to.  It is recommended to leave this default, but the property can be explicitly set to a specific process ID for the debugger to attach to.
+
+**request**: Indicates whether the configuration section is intended to `launch` the program or `attach` to an already running instance.
+
+**targetArchitecture**: `Deprecated` This option is no longer needed as the target architecture is automatically detected.
```
delmyers, line 146: @jacdavis, was this correctly fixed? I know that we had to remove some of Paul's code that did the architecture detection because it was blocking multi-threaded apps.

delmyers:
I had just a couple of minor things. Otherwise, approved.

launch.md
AndrewBrianHall: @jacdavis Does it look for either gdb or lldb on the path, or just gdb?

jacdavis: on Linux, gdb. On mac, lldb. The actual "default" being mentioned here is the default generated value when the user types this field in. I think we should just not document the default.

launch.md
jacdavis: Yes, it is fixed. I didn't remove paul's code, I fixed the nasty race condition.

launch.md
AndrewBrianHall: @jacdavis So it sounds like we should leave it like this until we fix that?

AndrewBrianHall:
Adding documentation for launch.json

launch.md
```diff
@@ -0,0 +1,151 @@
...
+
+You can change the behavior of GDB or LLDB by setting the following options.
+
+**MIMode**: Indicates the debugger that VS Code will connect to. Must be set to `gdb` or `lldb`.  This is pre-configured on a per-operating system basis and can be changed as needed.
+
+**miDebuggerPath**: The path to the debugger (such as gdb). When unspecified, it will search the operating system's PATH variable for a debugger (GDB on Linux and Windows, LLDB on OS X).
+
+**stopAtEntry**: If set to true, the debugger should stop at the entrypoint of the target (ignored on attach). Default value is `false`.
+
+**setupCommands**: JSON array of commands to execute in order to setup the GDB or LLDB. Example: `"customLaunchSetupCommands": [ { "text": "target-run", "description": "run target", "ignoreFailures": false }]`.
```
abelenky, line 86: Experimentation shows me that the element is NOT named "customLaunchSetupCommands", as shown here, but just "setupCommands".  This example (and again down a few lines) is wrong.

pieandcakes: @abelenky Thank you. I'll get it fixed.

PR #1132, opened 2017-10-14
Update MinGW paths to 6.3.0.

Jaxkr:
As the latest recommended version of MinGW is 6.3.0 and MinGW strongly discourages using old versions, I think it's appropriate to update the version numbers for this file and add a notice informing users that they may have to change the MinGW file path to match their version.


Documentation/LanguageServer/MinGW.md
```diff
@@ -1,5 +1,7 @@
 For developers using MinGW on Windows, we recommend you start with the following **c_cpp_properties.json** template.  Select "C/Cpp: Edit Configurations" from the command palette to create this file if you haven't already.
 
+Note that you may have to change the MinGW version number to match what you have installed. Eg. `C:/MinGW/lib/gcc/mingw32/5.3.0/` to `C:/MinGW/lib/gcc/mingw32/x.x.x/`.
```
bobbrow: can you remove the second path with the x.x.x?  If they have 5.3.0 installed it doesn't make sense to change it to something else.

PR #1179, opened 2017-10-28
Fix syntax error in JSON

russelltg:
The `c_cpp_properties.json` example file was missing a comma in the reference for `c_cpp_properties.json`


bobbrow:
Thank you for spotting this.

PR #1387, opened 2017-12-29
Moving from debugProxy to AdapterExecutableCommand

WardenGnaw:
For the move to remove the cpptools extension from reloading VsCode, we are removing program, runtime, and the debugProxy. Instead we will use the adapterExecutableCommand. 

This command is registered at the extension activation and waits for the dependency downloads to finish (indicated by install.lock) before actually executing the adapter.


Extension/src/main.ts
```diff
@@ -452,20 +455,57 @@ function rewriteManifest(installBlob: InstallBlob): Promise<void> {
...
-
-    // Remove the entry for cppvsdbg's proxy stub and replace it with the real debugger binary for Windows only.
-    if (os.platform() === 'win32') {
-        util.packageJson.contributes.debuggers[1].runtime = undefined;
-        util.packageJson.contributes.debuggers[1].program = './debugAdapters/vsdbg/bin/vsdbg.exe';
-    }
+// Registers adapterExecutableCommands for cppdbg and cppvsdbg. If it is not ready, it will prompt waiting for the download.
+// 
+// Note: util.extensionContext.extensionPath is needed for the commands because VsCode does not support relative paths for adapterExecutableComand
+function registerAdapterExecutableCommands(): void {
```
pieandcakes: I feel like since this is debugger related, we should do it in the debugAdapter and not here.

Extension/src/main.ts
```diff
@@ -452,20 +455,57 @@ function rewriteManifest(installBlob: InstallBlob): Promise<void> {
...
+            vscode.window.showErrorMessage("Debugger type 'cppvsdbg' is not avaliable for non-Windows machines.");
+            return null;
+        }
+        else {
+            return util.checkInstallLockFile().then(ready => {
+                if (ready)
+                {
+                    return {
+                        command: path.join(util.extensionContext.extensionPath,'./debugAdapters/vsdbg/bin/vsdbg.exe'),
+                        args: ['--interpreter=vscode']
```
pieandcakes: Is it possible to still specify `"args"` in package.json or it has to be here now? This is the way we specify pauseForDebugger, so it'd be good to know.

Extension/src/main.ts
```diff
@@ -452,20 +455,57 @@ function rewriteManifest(installBlob: InstallBlob): Promise<void> {
...
+                }
 
-    if (util.packageJson.extensionFolderPath.includes(".vscode-insiders"))
-        util.packageJson.contributes.configuration.properties["C_Cpp.intelliSenseEngine"].default = "Default";
+                return {
+                    command: command
+                }
+            }
+            else {
+                util.showReloadOrWaitPromptOnce();
```
pieandcakes: Are you supposed to throw an error instead of returning null then?

Extension/src/main.ts
```diff
@@ -427,7 +430,7 @@ function checkDistro(channel: vscode.OutputChannel, platformInfo: PlatformInform
     }
 }
 
-function rewriteManifest(installBlob: InstallBlob): Promise<void> {
+function rewriteManifest(installBlob: InstallBlob): void {
```
pieandcakes, line 430: Did you mean to make this change? I don't see the corresponding change on the caller.

Extension/src/main.ts
```diff
@@ -110,6 +110,9 @@ export function activate(context: vscode.ExtensionContext) {
     // Activate Configuration Provider and Process Picker Commands.
     DebuggerExtension.activate();
 
+    // Activate Adapter Commands 
+    registerAdapterExecutableCommands();
```
pieandcakes: Could we put this in the debuggerextension.activate?

pieandcakes (changes requested):

Extension/src/main.ts
WardenGnaw: Since I'm not returning anything now, this change is needed to compile. The caller still waits for the function to complete.

Extension/src/main.ts
WardenGnaw: Has to be here now. adaterExecutableCommand replaces runtime, args, and program.

Extension/src/main.ts
WardenGnaw: Tested in C# extension and did not work, they are currently adding a fix for this and will tell use how to properly handle errors.

pieandcakes (approved):

PR #1403, opened 2018-01-03
Adding tslint and gulp

WardenGnaw:
Adding tslint and the associated fixes that it complains about.

Adding gulp to do tests, tslint, and more.

Including package-lock.json that it keeps complaining about.

### DIFF
--- a/Extension/gulpfile.js
+++ b/Extension/gulpfile.js
@@ -1,0 +1,46 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+'use strict';
+
+const gulp = require('gulp');
+const tslint = require('gulp-tslint');
+const mocha = require('gulp-mocha');
+
+gulp.task('unitTest', () => {
+    gulp.src('./out/test/unitTests', {read: false}).pipe(
+        mocha({
+            ui: "tdd"
+        })
+    );
+});
+
+/// Misc Tasks
+const allTypeScript = [
+    'src/**/*.ts',
+    '!**/*.d.ts',
+    '!**/typings**'
+];
+
+const lintReporter = (output, file, options) => {
+    //emits: src/helloWorld.c:5:3: warning: implicit declaration of function ‘prinft’
+    var relativeBase = file.base.substring(file.cwd.length + 1).replace('\\', '/');
+    output.forEach(e => {
+        var message = relativeBase + e.name + ':' + (e.startPosition.line + 1) + ':' + (e.startPosition.character + 1) + ': ' + e.failure;
+        console.log('[tslint] ' + message);
+    });
+};
+
+gulp.task('tslint', () => {
+    gulp.src(allTypeScript)
+        .pipe(tslint({
+            program: require('tslint').Linter.createProgram("./tsconfig.json"),
+            configuration: "./tslint.json"
+        }))
+        .pipe(tslint.report(lintReporter, {
+            summarizeFailureOutput: false,
+            emitError: false
+        }))
+});
--- a/Extension/src/Debugger/attachToProcess.ts
+++ b/Extension/src/Debugger/attachToProcess.ts
@@ -62,7 +62,7 @@ export class RemoteAttachPicker {
             let pipeProgram: string = pipeTransport.pipeProgram;
             let pipeArgs: string[] = pipeTransport.pipeArgs;
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
+            let argList: string = RemoteAttachPicker.createArgumentList(pipeArgs);
 
             let pipeCmd: string = `"${pipeProgram}" ${argList}`;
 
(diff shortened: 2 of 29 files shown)


Extension/src/Debugger/attachToProcess.ts
```diff
@@ -63,7 +62,7 @@ export class RemoteAttachPicker {
             let pipeProgram: string = pipeTransport.pipeProgram;
             let pipeArgs: string[] = pipeTransport.pipeArgs;
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
+            let argList: string = RemoteAttachPicker.createArgumentList(pipeArgs);
```
bobbrow, line 65: Do we really need to add a rule to complain about not being explicit everywhere?  A lot of these typings are obvious based on the code (which is probably why they were left out), and VS Code IntelliSense already lets you know the type when the appropriate typings are found.  I find this rule to be a bit too cumbersome.

bobbrow: (BTW, I meant to put this comment on line 87 which makes a little more sense than putting it here - sorry!)

Extension/src/main.ts
```diff
@@ -176,25 +179,25 @@ export function deactivate(): Thenable<void> {
...
 function removePotentialPII(str: string): string {
-    let words = str.split(" ");
-    let result = "";
+    let words: string[] = str.split(" ");
+    let result: string = "";
     for (let word of words) {
         if (word.indexOf(".") == -1 && word.indexOf("/") == -1 && word.indexOf("\\") == -1 && word.indexOf(":") == -1) {
             result += word + " ";
         }
         else {
```
bobbrow: is there a rule to enforce correct formatting of if-else blocks?  The "else" should be on the line above next to the `{`

Extension/src/Debugger/attachToProcess.ts
sean-mcmanus: Are always using explicit types recommended typescript practice? I assume you could omit the type when the compiler is able to determine it, like with "auto" in C++. I think making the type explicit is good for readability when it's unclear, but often times it doesn't really help and just adds extra text.

WardenGnaw: Yes, explicit types is recommended Typescript practice. If the type is a dictionary, feel free to use ```any```. However most of the time, the type should be defined.

On the rare chance that this is too much (e.g. https://github.com/WardenGnaw/vscode-cpptools/blob/5c231a544c283a0ef6fb053e703846eff76e350f/Extension/src/LanguageServer/protocolFilter.ts#L13) then disabling the linter is fine.

Extension/gulpfile.js
```diff
@@ -0,0 +1,46 @@
...
+/// Misc Tasks
+const allTypeScript = [
+    'src/**/*.ts',
+    '!**/*.d.ts',
+    '!**/typings**'
+];
+
+const lintReporter = (output, file, options) => {
+    //emits: src/helloWorld.c:5:3: warning: implicit declaration of function ‘prinft’
+    var relativeBase = file.base.substring(file.cwd.length + 1).replace('\\', '/');
```
rajkumar42, line 29: see if you could get source navigation from the lint output to code, from vscode build output.

Extension/src/packageManager.ts
```diff
@@ -150,27 +148,36 @@ export class PackageManager {
...
         })
             .then((tmpResult) => {
                 pkg.tmpFile = tmpResult;
 
-                let lastError = null;
-                let retryCount = 0;
-                var handleDownloadFailure = (num, error) => {
+                let lastError: any = null;
+                let retryCount: number = 0;
+                let handleDownloadFailure: (num: any, error: any) => void = (num, error) => {
```
rajkumar42, line 159: num: number

Extension/src/packageManager.ts
```diff
@@ -150,27 +148,36 @@ export class PackageManager {
...
                 };
                 // Retry the download at most 5 times with 2-32 seconds delay.
-                return this.DownloadFile(pkg.url, pkg, 0).catch((error) => { handleDownloadFailure(1, error);
-                return this.DownloadFile(pkg.url, pkg, 1).catch((error) => { handleDownloadFailure(2, error);
-                return this.DownloadFile(pkg.url, pkg, 2).catch((error) => { handleDownloadFailure(3, error);
-                return this.DownloadFile(pkg.url, pkg, 3).catch((error) => { handleDownloadFailure(4, error);
-                return this.DownloadFile(pkg.url, pkg, 4).catch((error) => { handleDownloadFailure(5, error);
-                return this.DownloadFile(pkg.url, pkg, 5); // Last try, don't catch the error.
-                })})})})}).then(() => {
+                return this.DownloadFile(pkg.url, pkg, 0).catch((error) => {
```
rajkumar42, line 165: add todo to fix later.

rajkumar42:
👍

rajkumar42 (approved):

bobbrow (approved):

Extension/src/packageManager.ts
WardenGnaw: Will add to next PR

Extension/src/packageManager.ts
WardenGnaw: Will add to next PR

PR #1425, opened 2018-01-09
Fix install on Mac.

sean-mcmanus:
Fix install on Mac.

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -306,7 +306,7 @@ function makeOfflineBinariesExecutable(installBlob: InstallBlob): Thenable<void>
     packages.forEach(p => {
         if (p.binaries && p.binaries.length > 0 &&
             p.platforms.findIndex(plat => plat === installBlob.info.platform) !== -1 &&
-            (p.architectures == undefined || p.architectures.findIndex(arch => arch === installBlob.info.architecture) !== - 1)) {
+            (p.architectures === undefined || p.architectures.findIndex(arch => arch === installBlob.info.architecture) !== - 1)) {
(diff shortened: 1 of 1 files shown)


Extension/src/main.ts
```diff
@@ -306,7 +306,7 @@ function makeOfflineBinariesExecutable(installBlob: InstallBlob): Thenable<void>
     packages.forEach(p => {
         if (p.binaries && p.binaries.length > 0 &&
             p.platforms.findIndex(plat => plat === installBlob.info.platform) !== -1 &&
-            p.architectures.findIndex(arch => arch === installBlob.info.architecture) !== - 1) {
+            (p.architectures == undefined || p.architectures.findIndex(arch => arch === installBlob.info.architecture) !== - 1)) {
```
bobbrow: always use `===` over `==`.  Is there a linting rule for this?

bobbrow (changes requested):

Extension/src/main.ts
WardenGnaw: Yep. ```"triple-equals": true```

sean-mcmanus: Ok.

sean-mcmanus: So should I add triple-equal to tslint and fix all the non-triple equals (a lot)?

sean-mcmanus: I'll do that....

WardenGnaw: Talked with @sean-mcmanus, delaying the triple-equal tslint to another PR.

sean-mcmanus: I just changed this 1 instance to ===.

WardenGnaw (approved):

bobbrow (approved):

PR #1428, opened 2018-01-10
Cleanup reload dialog and processRuntimeDependencies flow.

WardenGnaw:
Removing reload dialog and file watchers. 

Adding isExtensionReady that checks to see if install.lock and debugAdapters folder exists. 

processRuntimeDependencies took in a weird call back that was pass along. Refactored to make it easier to read.

Refactored InstallBlob to be a singleton with setStage function. 

Created TempCommandsRegistrar to handle temp commands and activate language service. 
Created cpptoolsJsonUtils to handle cpptoolsJson handling. 

With these changes, debugging works for windows and mac


Extension/src/extensionActivationInformation.ts
```diff
@@ -0,0 +1,44 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+export enum InstallBlobStage {
```
bobbrow: minor: I would prefer "InstallationStage" over "InstallBlobStage".  "InstallBlob" was a quick name that wasn't meant to be cannonized into a "thing" :)

Extension/src/extensionActivationInformation.ts
```diff
@@ -0,0 +1,44 @@
...
+    hasError: boolean;
+    telemetryProperties: { [key: string]: string };
+
+    constructor() {
+        this.hasError = false;
+        this.telemetryProperties = {};
+    }
+}
+
+let installBlob: InstallBlob;
```
bobbrow: I am of the opinion that the installation state should be managed within a single module, not decentralized across multiple modules.  Can you please keep the object inside main.ts and remove these exports?  Or was there something that made it hard to call setInstallBlobStage from within the flow of events in main.ts?

bobbrow (changes requested):

Extension/src/extensionActivationInformation.ts
WardenGnaw: I personally felt that it made sense that the function that was being called would set the InstallBlobStage rather than the caller. E.g.

```
installBlob.stage = "downloadPackages"; 
packageManager.DownloadPackages(); 
...
installBlob.stage = "installPackages"; 
packageManager.InstallPackages(); 
```

sean-mcmanus (approved):
I'm fine with this, assuming it was tested. The offline install scenario still works, right? I'll test more with it anyway later.

Extension/src/main.ts
```diff
@@ -368,37 +173,38 @@ function handleError(installBlob: InstallBlob, error: any): void {
...
     // Show the actual message and not the sanitized one
-    if (installBlob.stage == "downloadPackages") {
+    if (installBlob.stage == InstallBlobStage.downloadPackages) {
         channel.appendLine("");
     }
-    channel.appendLine(`Failed at stage: ${installBlob.stage}`);
+    channel.appendLine(`Failed at stage: ${InstallBlobStage[installBlob.stage]}`);
     channel.appendLine(errorMessage);
     channel.appendLine("");
     channel.appendLine(`If you work in an offline environment or repeatedly see this error, try downloading a version of the extension with all the dependencies pre-included from https://github.com/Microsoft/vscode-cpptools/releases, then use the "Install from VSIX" command in VS Code to install it.`);
```
sean-mcmanus: It would be nice if it would omit this message if the error occurred during offline installation...I was going to fix that, but I didn't really think it was worth it since we don't expect the offline install to fail.

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -48,38 +50,39 @@ export class RemoteAttachPicker {
...
+                let pipeCmd: string = `"${pipeProgram}" ${argList}`;
 
-            return this.getRemoteOSAndProcesses(pipeCmd)
-                .then(processes => {
-                    let attachPickOptions: vscode.QuickPickOptions = {
-                        matchOnDetail: true,
-                        matchOnDescription: true,
-                        placeHolder: "Select the process to attach to"
-                    };
+                return this.getRemoteOSAndProcesses(pipeCmd).then(processes => {
```
pieandcakes: I had this on the next line to line up with the closing brace. can you move it back. you have a few of these.

Extension/src/commands.ts
```diff
@@ -0,0 +1,62 @@
...
+        this.tempCommands = [];
+
+        LanguageServer.activate(this.delayedCommandsToExecute.size > 0);
+        
+        this.delayedCommandsToExecute.forEach((command) => {
+            vscode.commands.executeCommand(command);
+        });
+        this.delayedCommandsToExecute.clear();
+    }
+}
```
pieandcakes, line 72: newline

Extension/src/common.ts
```diff
@@ -434,3 +392,24 @@ export function allowExecution(file: string): Promise<void> {
...
+        if (word.indexOf(".") == -1 && word.indexOf("/") == -1 && word.indexOf("\\") == -1 && word.indexOf(":") == -1) {
+            result += word + " ";
+        } else {
+            result += "? ";
+        }
+    }
+    return result;
+}
+
+export function checkDistro(platformInfo: PlatformInformation): void {
```
pieandcakes, line 396: can we add a check for bitness? we haven't technically tested arm/arm64 either.

Extension/src/extensionActivationInformation.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+export enum InstallationStage {
+    getPlatformInfo,
+    downloadPackages,
+    installPackages,
+    makeBinariesExecutable,
+    makeOfflineBinariesExecutable,
```
pieandcakes: isn't this the same as the one above? why is it separate?

Extension/src/extensionActivationInformation.ts
pieandcakes: I think this might be heavier for what we need it since we use it once only.

Extension/src/extensionActivationInformation.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+let installBlob: InstallBlob;
+
+export function initializeInstallBlob(): void {
+    installBlob = new InstallBlob();
+}
+
+export function getInstallBlob(): InstallBlob {
```
pieandcakes: add "instance"

Extension/src/extensionActivationInformation.ts
pieandcakes: callee should set stage please

Extension/src/main.ts
```diff
@@ -11,330 +11,135 @@ import * as util from './common';
 import * as Telemetry from './telemetry';
 import * as LanguageServer from './LanguageServer/extension';
 import * as DebuggerExtension from './Debugger/extension';
+import * as commands from './commands';
```
pieandcakes: why import *?

sean-mcmanus (changes requested):
downloadCpptoolsJsonPkg needs to be called after activation. We shipped a bug in 0.14.5 that caused language service loading to be delayed by the time it takes to download the cpptools.json (i.e. 20 seconds for slow connections): https://github.com/Microsoft/vscode-cpptools/issues/1429 . I could make this change afterwards if you want.

Extension/src/cpptoolsJsonUtils.ts
```diff
@@ -0,0 +1,95 @@
+/* --------------------------------------------------------------------------------------------
```
bobbrow: The purpose of the "cpptools.json" file is for A/B testing.  This module would be better named "abTesting.ts" or similar.

bobbrow: I have some ideas for a refactor here that I won't burden you with in this review.

Extension/src/extensionActivationInformation.ts
```diff
@@ -0,0 +1,44 @@
+/* --------------------------------------------------------------------------------------------
```
bobbrow: This module isn't really about activation, so it would be better named "installationInformation.ts"

Extension/src/extensionActivationInformation.ts
bobbrow: The caller sets the stage in every other installation step.  I think that any possible confusion is caused by the fact that two "stages" occur in `downloadAndInstallPackages` instead of splitting this function into two (e.g. `downloadPackages` and `installPackages`) and chaining them.  If you did that then the pattern would look consistent across the entire file.

Extension/src/main.ts
```diff
@@ -11,330 +11,135 @@ import * as util from './common';
...
-                    .then(() => rewriteManifest(installBlob))
-                    .then(() => touchInstallLockFile(installBlob))
-                    .catch(error => handleError(installBlob, error))
-                    .then(() => statusItem.dispose())
-                    .then(() => postInstall(installBlob))
-                    .then(() => activateExtensions());
-            }
-        });
-    });
+    let makeBinariesExecutablePromise: Promise<void> = makeBinariesExecutable();
```
bobbrow: allowing the tasks to execute in arbitrary order breaks the usefulness of the installation "stage" telemetry.  The concept of adding an installation stage to the telemetry requires a deterministic execution order so we can know what has succeeded prior to the extension getting stuck.  (this applies to both online and offline installations)

Extension/src/main.ts
```diff
@@ -11,330 +11,135 @@ import * as util from './common';
...
-    for (let word of words) {
-        if (word.indexOf(".") == -1 && word.indexOf("/") == -1 && word.indexOf("\\") == -1 && word.indexOf(":") == -1) {
-            result += word + " ";
-        } else {
-            result += "? ";
+async function processRuntimeDependencies(): Promise<void> {
+    const installLockExists: boolean = await util.checkInstallLockFile();
+
+    try {
+        const info: PlatformInformation = await PlatformInformation.GetPlatformInformation();
```
bobbrow: `GetPlatformInformation` doesn't need to be called when we aren't installing the extension, so perhaps this call should be made independently in both `offlineInstallation` and `onlineInstallation`.

WardenGnaw:
@sean-mcmanus Go ahead and make that change in another PR. I'll handle the merge conflicts.

sean-mcmanus:
I'll make the change in a new branch based off yours and then try to PR to this one. Where is your branch? I don't seem to see it for some reason. Oh, you did some sort of fork? I guess I'll just try to get the change into master.

WardenGnaw:
Different

sean-mcmanus (approved):

sean-mcmanus:
I have a pull request with the change (and another) at https://github.com/Microsoft/vscode-cpptools/pull/1433 .

bobbrow (approved):
:shipit:

Extension/src/common.ts
```diff
@@ -79,6 +31,29 @@ export function getPackageJsonString(): string {
     return JSON.stringify(packageJson, null, 2);
 }
 
+// Extension is ready if install.lock exists and debugAdapters folder exist.
+export async function isExtensionReady(): Promise<boolean> {
+    const doesDebugAdapterFolderExist: boolean = await checkFolderExists(getExtensionFilePath("debugAdapters"));
```
pieandcakes: >onst doesDebugAdapterFolderExist: boolean = await checkFolderExists(getExtensionFilePath("debugAdapters")); [](start = 5, length = 107)

Publicly this will always be true. Why are we checking for it?

Extension/src/common.ts
```diff
@@ -79,6 +31,29 @@ export function getPackageJsonString(): string {
...
+
+export function displayExtensionNotReadyPrompt(): void {
+
+    if (!isExtensionNotReadyPromptDisplayed) {
+        isExtensionNotReadyPromptDisplayed = true; 
+
+        vscode.window.showInformationMessage(extensionNotReadyString).then(
+            () => { isExtensionNotReadyPromptDisplayed = false; },
+            () => { isExtensionNotReadyPromptDisplayed = false; }
+        );
```
pieandcakes, line 53: can we toggle the output window if this gets shown?

Extension/src/common.ts
```diff
@@ -268,17 +232,24 @@ export function checkFileExists(filePath: string): Promise<boolean> {
     });
 }
 
+/** Test whether a folder exists */
+export function checkFolderExists(filePath: string): Promise<boolean> {
```
pieandcakes: >checkFolderExists [](start = 16, length = 17)

checkDirectoryExists

Extension/src/common.ts
```diff
@@ -268,17 +232,24 @@ export function checkFileExists(filePath: string): Promise<boolean> {
     });
 }
 
+/** Test whether a folder exists */
```
pieandcakes: >folder [](start = 19, length = 6)

directory

Extension/src/installationInformation.ts
```diff
@@ -0,0 +1,32 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+export class InstallBlob {
```
pieandcakes: >InstallBlob [](start = 13, length = 11)

If this file is called installationInformation can we rename the class?

Extension/src/main.ts
```diff
@@ -11,331 +11,150 @@ import * as util from './common';
...
+import * as commands from './commands';
 import { PlatformInformation } from './platform';
 import { PackageManager, PackageManagerError, PackageManagerWebResponseError, IPackage } from './packageManager';
 import { PersistentState } from './LanguageServer/persistentState';
-import * as url from 'url';
-import * as https from 'https';
-import { extensionContext } from './common';
-import { ClientRequest } from 'http';
+import {initializeInstallBlob, getInstallBlob, InstallBlob, setInstallationStage } from './installationInformation';
+import * as cpptoolsJsonUtils from './abTesting';
```
pieandcakes: reorder to put the * at the top

Extension/src/main.ts
```diff
@@ -11,331 +11,150 @@ import * as util from './common';
...
-                // It'll be used after the extension reloads.
-                downloadCpptoolsJsonPkg();
-            });
-    });
-
-    setInterval(() => {
-        // Redownload occasionally to prevent an extra reload during long sessions.
-        downloadCpptoolsJsonPkg();
-    }, 30 * 60 * 1000); // 30 minutes.
+    return processRuntimeDependencies();
```
pieandcakes, line 35: >processRuntimeDependencies [](start = 11, length = 26)

should this be awaited?

Extension/src/main.ts
```diff
@@ -11,331 +11,150 @@ import * as util from './common';
...
-    for (let word of words) {
-        if (word.indexOf(".") == -1 && word.indexOf("/") == -1 && word.indexOf("\\") == -1 && word.indexOf(":") == -1) {
-            result += word + " ";
-        } else {
-            result += "? ";
+async function processRuntimeDependencies(): Promise<void> {
+    const installLockExists: boolean = await util.checkInstallLockFile();
+
+    // Offline Scenario: Lock file exists but package.json has not had its activationEvents rewritten.
+    if (installLockExists && util.packageJson.activationEvents && util.packageJson.activationEvents.length == 1) {
```
pieandcakes: if (! exists) {
   do a;
}
else
{
 if(second condition)
{ do b; }
}

Extension/src/main.ts
WardenGnaw: I don't think we want vscode to hang on this activation?

pieandcakes:
>         let packageError: PackageManagerError = error;

if it fails the cast, what happens?

---
Refers to: Extension/src/main.ts:175 in 39f4e91. [](commit_id = 39f4e9193276e40f3637ce090ea68debec7f8dea, deletion_comment = False)

Extension/src/main.ts
```diff
@@ -371,15 +190,15 @@ function handleError(installBlob: InstallBlob, error: any): void {
...
-            installBlob.telemetryProperties['error.errorCode'] = removePotentialPII(packageError.errorCode);
+            installBlob.telemetryProperties['error.errorCode'] = util.removePotentialPII(packageError.errorCode);
         }
     } else {
         errorMessage = error.toString();
-        installBlob.telemetryProperties['error.toString'] = removePotentialPII(errorMessage);
+        installBlob.telemetryProperties['error.toString'] = util.removePotentialPII(errorMessage);
     }
 
     // Show the actual message and not the sanitized one
```
pieandcakes: >// Show the actual message and not the sanitized one [](start = 4, length = 52)

this comment seems wrong

WardenGnaw: I have no idea what was supposed to go there.

WardenGnaw:
@pieandcakes ``` if (error instanceof PackageManagerError) {``` is a couple lines above it.

Extension/src/main.ts
pieandcakes: is it the channel.appendline stuff with the errorMessage? i feel like it is that. i don't know why we are addin ga line here. 

---
In reply to: [160840925](https://github.com/Microsoft/vscode-cpptools/pull/1428#discussion_r160840925) [](ancestors = 160840925)

Extension/src/main.ts
```diff
@@ -413,33 +228,50 @@ function postInstall(installBlob: InstallBlob): Thenable<void> {
...
+                const cpptoolsString: string = await util.readFileText(cpptoolsJsonFile);
+                await cpptoolsJsonUtils.processCpptoolsJson(cpptoolsString);
+            }
+        } catch (error) {
+            // Ignore any cpptoolsJsonFile errors
+        }
 
-    return Promise.resolve();
-}
+        tempCommandRegistrar.activateLanguageServer();
```
pieandcakes: > tempCommandRegistrar.activateLanguageServer(); [](start = 7, length = 47)

should you await this before downloading the new file?

Extension/src/Debugger/attachToProcess.ts
pieandcakes: this still needs to be fixed. 

---
In reply to: [160575603](https://github.com/Microsoft/vscode-cpptools/pull/1428#discussion_r160575603) [](ancestors = 160575603)

Extension/src/main.ts
WardenGnaw: Not a promise

pieandcakes (changes requested):
:clock1:

pieandcakes (approved):
:shipit:

PR #1433, opened 2018-01-10
Fix delay in language service activation caused by cpptools.json downloading.

sean-mcmanus:
Also, changed the default workspaceParsingPriority to "highest" and renamed "normal" to "medium".

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -138,27 +138,30 @@ export function activate(context: vscode.ExtensionContext): void {
     registerTempCommand("C_Cpp.TakeSurvey");
 
     processRuntimeDependencies(() => {
-        downloadCpptoolsJsonPkg().then(() => {
-            util.readFileText(util.getExtensionFilePath("cpptools.json"))
-                .then((cpptoolsString) => {
-                    processCpptoolsJson(cpptoolsString);
-                })
-                .catch((error) => {
-                    // We already log telemetry if cpptools.json fails to download.
-                })
-                .then(() => {
-                    // Main activation code.
-                    tempCommands.forEach((command) => {
-                        command.dispose();
-                    });
-                    tempCommands = [];
-                    LanguageServer.activate(delayedCommandsToExecute);
-                    delayedCommandsToExecute.forEach((command) => {
-                        vscode.commands.executeCommand(command);
-                    });
-                    delayedCommandsToExecute.clear();
+        util.readFileText(util.getExtensionFilePath("cpptools.json"))
+            .then((cpptoolsString) => {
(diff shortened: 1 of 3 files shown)


sean-mcmanus:
The "normal" enum change is safe without an engine change because it never checks for "normal", but I'll change the engine references to "normal" later.

Extension/src/main.ts
```diff
@@ -138,27 +138,30 @@ export function activate(context: vscode.ExtensionContext): void {
...
-                    tempCommands.forEach((command) => {
-                        command.dispose();
-                    });
-                    tempCommands = [];
-                    LanguageServer.activate(delayedCommandsToExecute);
-                    delayedCommandsToExecute.forEach((command) => {
-                        vscode.commands.executeCommand(command);
-                    });
-                    delayedCommandsToExecute.clear();
+        util.readFileText(util.getExtensionFilePath("cpptools.json"))
```
pieandcakes, line 141: Can you check and make sure the file exists before you read it?

pieandcakes (approved):
Please make sure you squash when you merge. Thanks!

Extension/src/main.ts
```diff
@@ -138,27 +138,30 @@ export function activate(context: vscode.ExtensionContext): void {
...
+            .then((cpptoolsString) => {
+                processCpptoolsJson(cpptoolsString);
+            })
+            .catch((error) => {
+                // We already log telemetry if cpptools.json fails to download.
```
bobbrow, line 146: minor: update comment.  This code path does not download anything anymore.

sean-mcmanus: I don't understand. What code path? You mean the comment is misleading?

Extension/src/main.ts
sean-mcmanus: That shouldn't be necessary. The file should always be there unless someone manually deletes it or something else impossible happens and we catch any exceptions in that case too.

pieandcakes: But we shouldn't fail activation if the file isn't there, we should activate with baseline values. why should the file missing make activation stop?

sean-mcmanus: We don't. There's a catch, right?

Extension/src/main.ts
bobbrow: I take it back.  I made this comment through the web UI and didn't have the full context.

bobbrow (approved):

Extension/src/main.ts
sean-mcmanus: If Andrew wants to check the file's existence in his refactor pull request, I'd be fine with that.

PR #1460, opened 2018-01-17
support wildcards in the path (but only 1 per path). This is not mean…

chai3819:
…t to be used for recursive directory inclusion.


bobbrow:
This branch is not ready to merge yet.

PR #1462, opened 2018-01-17
Fix rename error and its message and linux build

WardenGnaw:
Rename was always outputting an error message since the callback is
always called. Included and error check. For offline install, if
opendebug.exe.config exists, rename needs to happen too. Added function
call for offline installation.

'use strict' was erroring on ubuntu build in copyScript.ts

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -151,7 +151,7 @@ function removeUnnecessaryFile(): Promise<void> {
         if (fs.existsSync(sourcePath)) {
             fs.rename(sourcePath, util.getDebugAdaptersPath("bin/OpenDebugAD7.exe.config.unused"), (err: NodeJS.ErrnoException) => {
                 if (err) {
-                    getOutputChannelLogger().appendLine(`ERROR: fs.rename failed with "${err.message}". Delete ${sourcePath} manually.`);
+                    getOutputChannelLogger().appendLine(`ERROR: fs.rename failed with "${err.message}". Delete ${sourcePath} manually to enable debugging.`);
                 }
             });
         }
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -78,6 +78,9 @@ async function offlineInstallation(): Promise<void> {
     setInstallationStage('makeOfflineBinariesExecutable');
     await makeOfflineBinariesExecutable(info);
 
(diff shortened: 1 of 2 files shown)


WardenGnaw:
😭 https://www.traviscistatus.com/incidents/p1g70p3qrq4z

However, [Twitter](https://twitter.com/traviscistatus?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Eauthor) says otherwise...

Extension/src/main.ts
```diff
@@ -147,7 +150,9 @@ function removeUnnecessaryFile(): Promise<void> {
         let sourcePath: string = util.getDebugAdaptersPath("bin/OpenDebugAD7.exe.config");
         if (fs.existsSync(sourcePath)) {
             fs.rename(sourcePath, util.getDebugAdaptersPath("bin/OpenDebugAD7.exe.config.unused"), (err: NodeJS.ErrnoException) => {
-                getOutputChannelLogger().appendLine("removeUnnecessaryFile: fs.rename failed");
+                if (err) {
+                    getOutputChannelLogger().appendLine(`ERROR: fs.rename failed with "${err.message}". Delete ${sourcePath} manually.`);
```
pieandcakes: maybe `Delete ${sourcePath} manually to enable debugging.`

pieandcakes (approved):

sean-mcmanus (approved):

sean-mcmanus:
Can you get the fix into the "release" branch so I can make the release vsix?

pieandcakes:
This fix should be shipped because it has to do with offline installation issues for any non-Windows release + a fix for the error messaging for all releases.

Extension/src/main.ts
```diff
@@ -78,6 +78,9 @@ async function offlineInstallation(): Promise<void> {
     setInstallationStage('makeOfflineBinariesExecutable');
     await makeOfflineBinariesExecutable(info);
 
+    setInstallationStage('removeUnnecessaryFile');
+    await removeUnnecessaryFile();
```
bobbrow, line 82: No.  This is wrong.  There are no unnecessary files in the offline package.  If there are, there is a bug with our Build Definition.

bobbrow (commented):
You should remove the removeUnnecessaryFile call from the offline install path.

PR #1470, opened 2018-01-19
fix wrong URL (msvscode -> vscode)

shsong:


sean-mcmanus (approved):

sean-mcmanus:
Thanks for finding/fixing this. Are you able to merge or should I? I'm not sure so I'll wait for @bobbrow .

PR #1480, opened 2018-01-22
Adding triple-equals to tslint

WardenGnaw:
What reminded me: https://stackoverflow.com/questions/48270127/can-a-1-a-2-a-3-ever-evaluate-to-true


bobbrow (approved):

bobbrow:
I thought we had added that rule already...

WardenGnaw:
I told sean to delay adding this rule since my large refactor PR was in progress then and I would handle it.

PR #1482, opened 2018-01-23
Inactive region highlight

grdowns:
Prepare for inactive region highlighting feature by adding a notification to client.ts for text decoration. Also fix the deprecated changes in launch.json + add some auto-generated changes which are seemingly necessary.


Extension/package.json
```diff
@@ -1115,8 +3759,8 @@
         "darwin"
       ],
       "binaries": [
-        "./bin/Microsoft.VSCode.CPP.Extension.darwin",
-        "./bin/Microsoft.VSCode.CPP.IntelliSense.Msvc.darwin"
+        "/home/griffin/vs/vscode-cpptools/Extension/bin/Microsoft.VSCode.CPP.Extension.darwin",
```
sean-mcmanus: This change to package.json needs to be reverted. Before making any changes to package.json, you need to make sure to undo the changes that occur when re-write the package.json with the runtime version of the file. The make your changes and stage the changes before running the extension again.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +627,16 @@ class DefaultClient implements Client {
         this.model.tagParserStatus.Value = notificationBody.status;
     }
 
+    private updateInactiveRegions(params: InactiveRegionParams): void {
+        let renderOptions: vscode.DecorationRenderOptions = {
+            color: "rgba(255,0,255,0.5)"
+        };
+        let textEditorDecoration = vscode.window.createTextEditorDecorationType(renderOptions);
+        
+        let editor = vscode.window.visibleTextEditors.find(e => e.document.uri.toString() == params.uri); // TODO change to document.uri == params.uri
```
sean-mcmanus: What does this TODO mean?

sean-mcmanus (commented):
Not sure why GitHub has me doing 2 reviews.

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +627,16 @@ class DefaultClient implements Client {
         this.model.tagParserStatus.Value = notificationBody.status;
     }
 
+    private updateInactiveRegions(params: InactiveRegionParams): void {
+        let renderOptions: vscode.DecorationRenderOptions = {
+            color: "rgba(255,0,255,0.5)"
```
bobbrow: >rgba(255,0,255,0.5) [](start = 20, length = 19)

it looks like we still have the test color in here.  We need to remember to make a light-themed and dark-themed gray color here,

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +627,16 @@ class DefaultClient implements Client {
...
+        };
+        let textEditorDecoration = vscode.window.createTextEditorDecorationType(renderOptions);
```
bobbrow: >let textEditorDecoration = vscode.window.createTextEditorDecorationType(renderOptions); [](start = 8, length = 87)

We probably only need to create this object once when the class is first initialized.

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +627,16 @@ class DefaultClient implements Client {
...
+        
+        let editor = vscode.window.visibleTextEditors.find(e => e.document.uri.toString() == params.uri); // TODO change to document.uri == params.uri
+        editor.setDecorations(textEditorDecoration, params.ranges);
```
bobbrow: >editor.setDecorations [](start = 8, length = 21)

with each notification, I think we'll need to save the ranges that come in.  This way we can clear out the old decorations before setting the new ones.

bobbrow: It looks like you might also be able to call dispose on the TextEditorDecorationType to remove all decorations associated with that object.  If that's the case, then my previous comment about creating the TextEditorDecorationType only once would be invalid.  You might be able to just keep a reference to the previous TextEditorDecorationType and dispose it before creating a new one and assigning it to the new ranges.

---
In reply to: [163304083](https://github.com/Microsoft/vscode-cpptools/pull/1482#discussion_r163304083) [](ancestors = 163304083)

grdowns: That sounds like a fairly tidy way to manage the decoration types. I will make sure to investigate that possibility first!

Extension/src/LanguageServer/client.ts
grdowns: Most definitely. I will add different coloration based on light/dark, and I will keep note of it as a discussion point when we talk as a group, and when we find out more about the possibility of opacity.

Extension/package.json
grdowns: Thanks for catching that! Is there any way we can automate this process for the future? I would like to investigate the possibility if it's worthwhile.

Extension/src/LanguageServer/client.ts
grdowns: This is a TODO that I should have investigated earlier. Most operations within client.ts operate via vscode.Uri, whereas this method only compares the absolute file path via string. @bobbrow opinion on the matter?

Extension/package.json
bobbrow: I'm going to add a new option to our CMakeLists.txt for the language server that will point to your local clone of this instance and copy it into a disposable location so you don't have to run the extension on top of the local repository.  (this has been bugging me as well) :)

grdowns: Excellent! I will revert the changes as per Sean's comment and keep my eyes peeled for that change in the future.

Extension/src/LanguageServer/client.ts
bobbrow: You should probably send the uri in the JSON message instead of the filename.

Extension/ThirdPartyNotices.Mono.txt
```diff
@@ -0,0 +1,170 @@
+
```
WardenGnaw: This is only needed for OSX. Please do not include this in the check-in. 

I think it would be ok to add this to the .gitignore. @pieandcakes?

Extension/cpptools.json
```diff
@@ -1,3 +1,3 @@
-{
-    "intelliSenseEngine_default_percentage": 100
+{
```
WardenGnaw: git diff says this is a whitespace difference, specifically a carriage return (^M). Undo.

Extension/bin/msvc.64.darwin.json
```diff
@@ -2,8 +2,7 @@
     "defaults": [
       "--clang",
       "--pack_alignment",
-      "8",
-      "--framework_include_directory=/System/Library/Frameworks"
+      "8"
```
WardenGnaw, line 5: Is this a required change?

WardenGnaw (dismissed):

Extension/bin/msvc.64.darwin.json
bobbrow: This should go in at some point though it's not required to be part of this change. This repo is out of sync with the language server.

Extension/cpptools.json
grdowns: I will make sure to undo this change. Do we have any tools for Git to automatically ignore carriage return changes? I feel that checking for carriage returns manually with each commit is wasteful.

WardenGnaw: You could try ```git diff -w | git apply --cached --ignore-whitespace```. However, I would do a quick glance at```git diff``` before creating the PR. 

The common things to look out for with this repo is any changes that you did not make or files that you did not add should not be included with your changes.

grdowns: Thanks! I appreciate the advice + tip. You are totally right about the automatically generated changes and I will make sure to be diligent in scrutinizing these diffs pre-PR.

Extension/ThirdPartyNotices.Mono.txt
pieandcakes: @grdowns Was this added on purpose because you have a requirement for Mono now?

Extension/package.json
```diff
@@ -38,7 +38,25 @@
     "Linters"
   ],
```
pieandcakes: I think you need to undo this file change. I dont' think we should be changing our activation from *.

pieandcakes (changes requested):

Extension/ThirdPartyNotices.Mono.txt
sean-mcmanus: No, I'm pretty sure we don't have a requirement for Mono with this change.

grdowns: This license was auto-generated. I believe Sean is correct and it will be removed.

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +627,26 @@ class DefaultClient implements Client {
         this.model.tagParserStatus.Value = notificationBody.status;
     }
 
+    private decoration : vscode.TextEditorDecorationType;
```
bobbrow: Please move this up to the top of the class definition with the other member variables.  You might also want to rename it something more specific so we don't lose track of it.  e.g. inactiveRegionsDecoration.

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +628,25 @@ class DefaultClient implements Client {
...
 
+    private updateInactiveRegions(params: InactiveRegionParams): void {
+        let renderOptions: vscode.DecorationRenderOptions = {
+            light: { color: "rgba(125,125,125,1.0)" },
+            dark: { color: "rgba(155,155,155,1.0)" }
+        };
+
+        // Recycle the active text decorations when we receive a new set of inactive regions
+        if (this.inactiveRegionsDecoration !== undefined) {
+            this.inactiveRegionsDecoration.dispose();
```
bobbrow: Thinking about this a bit more...  Since the inactiveRegionsDecorarion is linked to the Client, this will end up accidentally deleting highlights from files when you have more than one file open in the editor.  You're going to want to have one of these for each uri that you encounter.

And then you will want to add disposal code to DefaultClient.dispose

bobbrow (changes requested):

Extension/.vscode/launch.json
```diff
@@ -10,7 +10,7 @@
 			"args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outDir": "${workspaceRoot}/out/src",
+			"outFiles": ["${workspaceRoot}/out/src"],
```
pieandcakes, line 13: Reading the helptext on outFiles, looks like its a glob pattern to specify js files. do we want to set this to `
"${workspaceRoot}/out/src/**/*.js"` ?

pieandcakes (approved):

bobbrow (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -198,6 +205,7 @@ class DefaultClient implements Client {
     private crashTimes: number[] = [];
     private failureMessageShown = new PersistentState<boolean>("DefaultClient.failureMessageShown", false);
     private isSupported: boolean = true;
+    private inactiveRegionsDecorations = new Map<string, [vscode.TextEditorDecorationType, vscode.Range[]]>();
```
WardenGnaw: nit: Can you make this an interface? 
```valuePair[0].dispose();``` is not as readable as ```valuePair.decoration.dispose();```
and ```valuePair[1] = params.ranges; ``` vs ```valuePair.ranges = params.ranges; ```

**Review by WardenGnaw — APPROVED** [review]

File: Extension/src/LanguageServer/client.ts (hunk 13 above)
Comment (grdowns) [reply Extension/src/LanguageServer/client.ts]: Absolutely! I considered doing so, but decided against it as it didn't seem worth polluting the file + namespace. I think that someone else requesting an interface is enough to rationalize doing it, however. I'll put it through!

File: Extension/src/LanguageServer/client.ts [hunk 14]
```diff
@@ -382,6 +390,14 @@ class DefaultClient implements Client {
         }
     }
 
+    public onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
+        //Apply text decorations to inactive regions
+        for (let e of editors) {
+            let valuePair: [vscode.TextEditorDecorationType, vscode.Range[]] = this.inactiveRegionsDecorations.get(e.document.uri.toString());
+            e.setDecorations(valuePair[0], valuePair[1]); // VSCode clears the decorations when the text editor becomes invisible
```
bobbrow: need to guard against valuePair being undefined here.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +637,32 @@ class DefaultClient implements Client {
         this.model.tagParserStatus.Value = notificationBody.status;
     }
 
+    private updateInactiveRegions(params: InactiveRegionParams): void {
+        let renderOptions: vscode.DecorationRenderOptions = {
+            light: { color: "rgba(125,125,125,1.0)" },
```
bobbrow: nit: personal preference (being a Light user 😄): I built this change and am playing with it.  This is a bit too dark to easily distinguish the inactive regions for me.  Can you bump it up a bit?  175,175,175 looks a bit better to me.

grdowns: Definitely! I don't expect these numbers to survive until the feature is complete

Extension/src/LanguageServer/client.ts
```diff
@@ -186,6 +194,11 @@ export function createNullClient(): Client {
     return new NullClient();
 }
 
+interface decorationRangesPair {
```
WardenGnaw: nit: Please move it closer to the other group of interfaces.

WardenGnaw: Captitalize

Extension/src/LanguageServer/client.ts
```diff
@@ -382,6 +396,16 @@ class DefaultClient implements Client {
         }
     }
 
+    public onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
+        //Apply text decorations to inactive regions
+        for (let e of editors) {
+            let valuePair: decorationRangesPair = this.inactiveRegionsDecorations.get(e.document.uri.toString());
+            //if (valuePair !== undefined) {
```
WardenGnaw: Why is this commented out?

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +645,36 @@ class DefaultClient implements Client {
...
+        let decoration: vscode.TextEditorDecorationType = vscode.window.createTextEditorDecorationType(renderOptions);
+
+        // Recycle the active text decorations when we receive a new set of inactive regions
+        let valuePair: decorationRangesPair = this.inactiveRegionsDecorations.get(params.uri);
+        if (valuePair !== undefined) {
+            // Disposing of and resetting the decoration will undo previously applied text decorations
+            valuePair.decoration.dispose();
+            valuePair.decoration = decoration;
+
+            valuePair.ranges = params.ranges; // As vscode.TextEditor.setDecorations only applies to visible editors, we must cache the range for when another editor becomes visible
```
WardenGnaw: nit: move comment above

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +645,36 @@ class DefaultClient implements Client {
...
+    private updateInactiveRegions(params: InactiveRegionParams): void {
+        let renderOptions: vscode.DecorationRenderOptions = {
+            light: { color: "rgba(175,175,175,1.0)" },
+            dark: { color: "rgba(155,155,155,1.0)" }
+        };
+        let decoration: vscode.TextEditorDecorationType = vscode.window.createTextEditorDecorationType(renderOptions);
+
+        // Recycle the active text decorations when we receive a new set of inactive regions
+        let valuePair: decorationRangesPair = this.inactiveRegionsDecorations.get(params.uri);
+        if (valuePair !== undefined) {
```
WardenGnaw: Either do ```!valuePair``` or ```this.inactiveRegionsDecorations.has(params.uri)```

Extension/src/LanguageServer/client.ts
```diff
@@ -21,6 +21,7 @@ import { createProtocolFilter } from './protocolFilter';
 import { DataBinding } from './dataBinding';
 import minimatch = require("minimatch");
 import * as logger from '../logger';
+import { deactivate } from './extension';
```
bobbrow: This should not be imported here.  Only vscode should ever call this method.

Extension/src/LanguageServer/client.ts
bobbrow: it should be uncommented.

Extension/src/LanguageServer/client.ts
grdowns: `!valuePair` changes the meaning of the statement, and `this.inactiveRegionsDecorations.has(params.uri)` is (assumedly) as expensive as calling `this.inactiveRegionsDecorations.get(params.uri)` which while minor it would be best to avoid it. Is that the case?

Extension/src/LanguageServer/client.ts
grdowns: @bobbrow is correct. Apologies while I get used to the code review cycle. All tips and advice and nitpicks are welcome. :smile:

Extension/src/LanguageServer/client.ts
bobbrow: no need to call `.has`.  You can change it to `if (valuePair)` or just leave it as `if (valuePair !== undefined)`.  I think the 'truthy' compare argument for this case has been bolstered a bit by the recent regression, but I believe the API contract here is relatively well defined.

```
interface Map<K, V> {
    clear(): void;
    delete(key: K): boolean;
    forEach(callbackfn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any): void;
    get(key: K): V | undefined;
    has(key: K): boolean;
    set(key: K, value: V): this;
    readonly size: number;
}
```

WardenGnaw: Whoops, should have looked at the return values of get(). 🙈 

-resolved-

grdowns: I believe `if (valuePair)` is more idiomatic and as it does not change the meaning of the statement, I will substitute it in. Thanks

Extension/src/LanguageServer/client.ts
grdowns: That snuck in via autocomplete. Removed + noted for the future

Extension/src/LanguageServer/client.ts
grdowns: Guard has been brought back in + using `if (valuePair)` instead of `if (valuePair !== undefined)`

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +644,55 @@ class DefaultClient implements Client {
...
+            return false;
+        }
+
+        for (let e1 of r1) {
+            for (let e2 of r2) {
+                if (e1 !== e2) {
+                    return false;
+                }
+            }
+        }
```
WardenGnaw: return true

WardenGnaw (changes requested):

Extension/src/LanguageServer/client.ts
grdowns: :man_facepalming:

WardenGnaw: Did you also want a reference compare or actual equality?
https://github.com/Microsoft/vscode/blob/69f9a3b8080f5e90a47743dda1a6dc3f55972be4/src/vs/vscode.d.ts#L422

grdowns: Good catch! I was ignorant of the reference vs primitive comparison difference in this case. `if (!e1 == e2)` should be `if (!e1.isEqual(e2))` as it will compare the ints inside each range's positions.

WardenGnaw (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +644,57 @@ class DefaultClient implements Client {
...
+
+        // Apply the decorations to all *visible* text editors
+        let editors: vscode.TextEditor[] = vscode.window.visibleTextEditors.filter(e => e.document.uri.toString() === params.uri);
+        for (let e of editors) {
+            e.setDecorations(decoration, params.ranges);
+        }
+    }
+
+    // Helper method to compare two ranges for equality
+    private isRangesEqual(r1: vscode.Range[], r2: vscode.Range[]): boolean {
```
bobbrow: nit: comparing two things.  Use "are" instead of "is" for the naming. :)

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +644,57 @@ class DefaultClient implements Client {
...
+        if (r1.length !== r2.length) {
+            return false;
+        }
+
+        for (let e1 of r1) {
+            for (let e2 of r2) {
+                if (!e1.isEqual(e2)) {
```
bobbrow: This seems wrong.  It will compare r1[0] against r2[1] and fail.  If we are expecting the arrays to have the same ranges in the same order, this should be a single for-loop iterating from 0 to r1.length-1 (or r2.length-1 since they are equal at this point). Then it checks if r1[n] is equal to r2[n].

Extension/src/LanguageServer/client.ts
```diff
@@ -620,6 +644,57 @@ class DefaultClient implements Client {
...
+        let renderOptions: vscode.DecorationRenderOptions = {
+            light: { color: "rgba(175,175,175,1.0)" },
+            dark: { color: "rgba(155,155,155,1.0)" }
+        };
+        let decoration: vscode.TextEditorDecorationType = vscode.window.createTextEditorDecorationType(renderOptions);
+
+        // Recycle the active text decorations when we receive a new set of inactive regions
+        let valuePair: DecorationRangesPair = this.inactiveRegionsDecorations.get(params.uri);
+        if (valuePair) {
+            // The language server will send notifications regardless of whether the ranges have changed. Unfortunately we must check for changes ourselves
```
bobbrow: The "unfortunately" part of this comment is implying that there is a problem with the language server we are working around.  As we discussed earlier, we are making a conscious choice to do it this way and avoid a second copy of the cached results in the language server.  Perhaps this is unfortunate, but it is deliberate, and I don't think this comment helps us understand the code better or feel better about that decision, so my vote is to remove it. 😉

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
grdowns: Good point! I totally agree with that and it will be removed. :smile:

Extension/src/LanguageServer/client.ts
WardenGnaw: Are the arrays always going to have the Range in the same order?

If we are not guaranteedsame order, I think it should be e1 exists in e2, if it does not return false.

grdowns: D'oh. This is n squared as well as ineffective. Thanks

WardenGnaw:
https://github.com/Microsoft/vscode-cpptools/blob/master/Extension/test/unitTests/extension.test.ts is pretty empty. Never hurts to have more tests.

Extension/src/LanguageServer/client.ts
grdowns: It's a stylistic choice for a method returning a boolean. But I see your point and I'll concede. :laughing:

bobbrow (approved):

PR #1488, opened 2018-01-24
fix a bug in resolveVariables util function

bobbrow:
The change to triple equals had an unintended side effect for this utility function.  We should look for other potential regressions, but I need this now to get the language server unblocked.


WardenGnaw (approved):

WardenGnaw:
Possibility anything that was comparing to null/undefined.

sean-mcmanus:
This article recommends avoiding using null: https://medium.com/@basarat/null-vs-undefined-in-typescript-land-dc0c7a5f240a .

sean-mcmanus (approved):

WardenGnaw:
I think anywhere where we have === or !== to null or undefined we can just check to see if it is a Truthy or Falsey class.

PR #1493, opened 2018-01-26
tslint: Force semicolon in object literal

WardenGnaw:


sean-mcmanus (approved):

Extension/src/LanguageServer/ui.ts
```diff
@@ -165,7 +165,7 @@ export class UI {
             });
     }
 
-    public showWorkspaces(workspaceNames: { name: string, key: string }[]): Thenable<string> {
+    public showWorkspaces(workspaceNames: { name: string; key: string }[]): Thenable<string> {
```
bobbrow, line 168: i wonder why it doesn't force it for the second field

bobbrow (approved):

Extension/src/LanguageServer/ui.ts
WardenGnaw: Technically not required, similar to having a trailing comma at the end of the last element of the array.

I'm not sure if there is an existing core rule on this but I can create one.

bobbrow: nah, don't worry about it. I was just curious.

PR #1494, opened 2018-01-26
Seanmcm/telemetry fix

sean-mcmanus:


sean-mcmanus:
I'm still testing this...

Extension/src/LanguageServer/client.ts
```diff
@@ -134,7 +134,65 @@ function collectSettings(filter: (key: string, val: string, settings: vscode.Wor
         }
         if (filter(key, val, settings)) {
             previousCppSettings[key] = val;
-            result[key] = (key === "clang_format_path") ? "..." : String(previousCppSettings[key]);
+            switch (key.toLowerCase()) {
+                case "clang_format_path": {
```
bobbrow: I think we should just remove clang_format_path from telemetry

Extension/src/LanguageServer/client.ts
```diff
@@ -134,7 +134,65 @@ function collectSettings(filter: (key: string, val: string, settings: vscode.Wor
...
+                            case "chromium":
+                            case "mozilla":
+                            case "webkit":
+                            case "file":
+                            case "none": {
+                                result[key] = String(previousCppSettings[key]);
+                                break;
+                            }
+                            default: {
+                                if (val.startsWith("{") && val.endsWith("}")) {
```
bobbrow: is this distinction necessary? can we just send "..." for anything that doesn't match one of the enum values?

Extension/src/LanguageServer/client.ts
bobbrow: perhaps the above if-expression should be `if (filter(key, val, settings) && isTelemetryAllowed(key, val))` and then move all of the special cases to a new function `isTelemetryAllowed` and filter them there so that this function can remain simple.

Extension/src/LanguageServer/client.ts
```diff
@@ -134,7 +134,65 @@ function collectSettings(filter: (key: string, val: string, settings: vscode.Wor
         }
         if (filter(key, val, settings)) {
             previousCppSettings[key] = val;
-            result[key] = (key === "clang_format_path") ? "..." : String(previousCppSettings[key]);
+            switch (key.toLowerCase()) {
+                case "clang_format_path": {
+                    if (val) {
+                        switch (val.toLowerCase()) {
```
bobbrow: `val` is `any`, so I don't think you can safely call this without converting to string first.

Extension/src/LanguageServer/client.ts
```diff
@@ -134,7 +134,65 @@ function collectSettings(filter: (key: string, val: string, settings: vscode.Wor
...
+                        }
+                    } else {
+                        result[key] = "null";
+                    }
+                    break;
+                }
+                case "clang_format_style":
+                case "clang_format_fallbackstyle": {
+                    if (val) {
+                        switch (val.toLowerCase()) {
```
bobbrow: same comment as above.  `val` must be confirmed to be a string before this can be called on it.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -134,7 +134,65 @@ function collectSettings(filter: (key: string, val: string, settings: vscode.Wor
...
+                            case "visual studio":
+                            case "llvm":
+                            case "google":
+                            case "chromium":
+                            case "mozilla":
+                            case "webkit":
+                            case "file":
+                            case "none": {
+                                result[key] = String(previousCppSettings[key]);
```
bobbrow: Remember also that we have to rename the keys since we are making a change to what is logged for these settings.

sean-mcmanus: Oh, yeah, I forgot.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure

Extension/src/LanguageServer/client.ts
sean-mcmanus: What do you mean by "remove clang_format_path from telemetry"? Do you mean just use "..." like we were doing before?

bobbrow: I meant to just filter it out completely (never send that property)

sean-mcmanus: Okay, I stopped sending it.

Extension/src/LanguageServer/client.ts
sean-mcmanus: I added a 2 at the end.

bobbrow (approved):

PR #1549, opened 2018-02-13
add compilerPath, cStandard, and cppStandard settings, fix inactive region deletion bug

sean-mcmanus:
This isn't ready for code review yet...still testing stuff.


sean-mcmanus:
I need to update the msvc.json to get rid of the hard-coded version.

Extension/c_cpp_properties.schema.json
```diff
@@ -14,7 +14,11 @@
                         "description": "Platform name. Mac, Linux, or Win32 are the defaults unless a custom platform is added.",
                         "type": "string"
                     },
-                    "compileCommands":{
+                    "compilerPath": {
```
bobbrow, line 17: need to add code in client.ts to remove this setting from telemetry.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -97,7 +98,8 @@ export interface Configuration {
     browse?: Browse;
 }
 
-export interface DefaultPaths {
+export interface CompilerDefaults {
+    compilerPath: string;
```
bobbrow, line 105: It seems weird to send this back to the TypeScript code when the setting has already been sent down to the language server.  I presume you do this to resolve variables and such.  Can we intercept the didChangeConfiguration message and resolve the variables before they are sent down to the language server the first time?

bobbrow (changes requested):

Extension/c_cpp_properties.schema.json
sean-mcmanus: Ah, yeah.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: It's sent back so that when a user generates a new c_cpp_properties, the guessed one is shown. This gives more exposure to the setting and it's easier for users to change it to the correct compiler since the setting exists in the file already. I was thinking we'd eventually want to update the existing c_cpp_properties.json to have the default compilePath too, but that code doesn't exist yet. I haven't found a reason yet to differentiate between an empty compilePath and the guessed one, but I might discover one later...

Extension/c_cpp_properties.schema.json
sean-mcmanus: Wait, we don't send telemetry for this.

bobbrow: You're right.  the sliver of code I saw looked a lot like package.json, so I was in the wrong frame of mind.  Please ignore the comment.

sean-mcmanus:
I'm still fixing bugs with this -- it's not ready for a code review yet. I just pushed changes because I wanted to switch branches.

bobbrow (approved):

ThadHouse:
Just curious if this has been tested to work across the WSL boundary. I know a lot of people compile directly using WSL, and it seems like it should work, but just something that would definitely be helpful.

Extension/bin/msvc.json
```diff
@@ -28,14 +28,6 @@
       {
         "match": "^/dE--header_only_fallback",
         "replace": "--header_only_fallback"
-      },
-      {
-        "match": "^/lang_c_",
-        "replace": "--c\n--c11"
```
bobbrow, line 34: if this is being removed in this change, we will need to add the cStandard and cppStandard fields to c_cpp_properties.json.  I was thinking we were going to do this separately though.

bobbrow:
can you fix the lint errors so that this change can pass the Travis build?

```
Error: Failed to lint:
/home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/configurations.ts [231, 13]: if statements must be braced
/home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/configurations.ts [125, 13]: 'defaultCStandard' is declared but its value is never read.
/home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/configurations.ts [126, 13]: 'defaultCppStandard' is declared but its value is never read.
/home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/configurations.ts [192, 13]: misplaced opening brace.
```

sean-mcmanus:
Sure, I need to investigate why I don't see any linter errors...oh, my TSLint extension wasn't installed for some reason.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -186,13 +186,23 @@ export class CppProperties {
...
-            let config: Configuration = this.configurationJson.configurations[this.CurrentConfiguration];
-            if (!config.compilerPath || config.compilerPath.length === 0)
-            {
-                config.compilerPath = this.defaultCompilerPath; // Update the compilerPath with the default being used.
-                fs.writeFileSync(this.propertiesFile.fsPath, JSON.stringify(this.configurationJson, null, 4));
-            }
+        // Update the compilerPath, cStandard, and cppStandard with the default being used.
+        let doUpdate: boolean = false;
+        let config: Configuration = this.configurationJson.configurations[this.CurrentConfiguration];
+        if (this.defaultCompilerPath && this.defaultCompilerPath.length > 0 && (!config.compilerPath || config.compilerPath.length === 0)) {
```
bobbrow: The .length checks should be unnecessary.  If this.defaultCompilerPath evaluates to true it can't be the empty string.

sean-mcmanus:
@ThadHouse These settings that are dependent on querying the compiler won't work with WSL until we figure out how to run the clang/gcc compiler from within the WSL environment and pass the results back to Windows.

PR #1555, opened 2018-02-14
faster launch times for extension debugging

bobbrow:
Don't run `npm install` and the debugger copy script every time you launch the extension.


sean-mcmanus (approved):

Extension/src/Support/update.js
```diff
@@ -0,0 +1,16 @@
...
+'use strict';
+
+/**
+ * This file is used during local debugging of the extension and should not be referenced by any
+ * other source files.
+ */
+
+const fs = require("fs");
+if (!process.env.CPPTOOLS_DEV && fs.existsSync('./node_modules')) {
+    console.log("Skipping prepublish steps since they appear to have been executed already.");
```
pieandcakes, line 14: This is dangerous if you update the dependencies it will not know to run again.

sean-mcmanus: Can't we just rely on the user to manually run npm install if they update dependencies? We don't update dependencies very often. That's what I had to do previously.

bobbrow: On the language server side this is rarely an issue for us.  It shouldn't affect debugger scenarios since you always have CPPTOOLS_DEV set, right?

bobbrow: And I log a message to the Output window, so there shouldn't be any surprises here.

pieandcakes: ok that's fine then

WardenGnaw (approved):
LGTM

sean-mcmanus:
Is this going to be checked in?

PR #1595, opened 2018-02-21
Fixed cannot read property2 sometimes

urmysun:


bobbrow:
Help me understand the reasoning behind this change.  Putting the `"intelliSenseMode"` property at the top was intentional so that people don't miss it on accident.  Windows defaults to `"msvc-x64"` so it is extremely important for Windows users to know they need to use `"clang-x64"` mode for WSL.

urmysun:
Agree with you. I found that when I followed the documentation and added the configuration template to my cpp_properties.json file, the configuration still does actually not work. As a WSL users, I have to replace the whole   {"name": "Win32"...} with the template in the documentation. After replacing it, everything works fine.

sean-mcmanus:
I also don't understand this change. The parameter order shouldn't matter with json objects.

Issue #1574, opened 2018-02-16
Add a setting to disable inactive region code graying


bobbrow:
There are some cases where graying the code is undesirable for developers. A setting needs to be added to allow developers to disable the inactive region highlighting.

Consider also adding a lightbulb action to allow developers to toggle this without visiting the settings.json file.

See #1466 for a workaround if you need to disable this immediately.  We'll target March (possibly in the insiders release too).

bobbrow:
We just posted an insider release with the new setting to disable the inactive region colorization.  If you want to get an early preview, you can install the VSIX that matches your OS from the link below.

https://github.com/Microsoft/vscode-cpptools/releases/tag/v0.16.0-insiders

NiMiZ:
Can you also add option to define a list of defines that you want to be define?
In my project there are defines in the make file according to the make's arguments, so if it will be possible to give it a list or set of list of define it will be helpful

sean-mcmanus:
@NiMiZ When you use the "Edit Configurations" command, you should see the "defines" property in "c_cpp_properties.json" where you can add the list of defines. For defines with a value, use a = between the name and value. Let us know if you have a problem getting that to work.

NiMiZ:
I'm kinda new in VS Code and I don't know this "Edit Configurations" command bu tI opened the file and I ee the defines property there, but I don't understand how to add a define.
is ti like that:
                    "defines": {
                        "description": "A list of preprocessor definitions for the IntelliSense engine to use while parsing files. Optionally, use = to set a value, e.g. VERSION=1.",
                        "type": "array",
                        "items": {
                            "type": "string"
                            "value": [ "VERSION=1", "DEBUG=2"]
                        }
                    },


Can you give an example?

thanks

sean-mcmanus:
Use the View->Command Palette menu and type "Edit Configurations" after the >.
You can add defines like
``` 
"defines": [
                "_DEBUG=1",
                "UNICODE",
                "_UNICODE"
            ],
```

unimun:
I'm using -imacros option to implicitly include c macros to c source files. But cpptools do not appropriatly parse macros when I give header files to "defines" field of c_cpp_properties.json like:
``
            "defines": [
                "${workspaceRoot}/out/config.h",
                "_DEBUG",
                "UNICODE",
                "_UNICODE"
            ],
``
It enables syntax highlighting all conditional compile preprocessors including not defined symbols.
[...]

PR #1604, opened 2018-02-24
Add a setting to manage inactive region colorization

bobbrow:
It still needs a corresponding change from the language server process so that the lightbulb and toggle commands work, but this handles the TypeScript part.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -139,6 +139,12 @@
           "description": "Controls whether suspected compile errors detected by the IntelliSense engine will be reported back to the editor. Warnings about #includes that could not be located will always be reported to the editor. This setting is ignored by the Tag Parser engine.",
           "scope": "resource"
         },
+        "C_Cpp.dimInactiveRegions": {
+          "type": "boolean",
+          "default": true,
+          "description": "Controls whether inactive preprocessor blocks are colored differently than active code. This setting is ignored by the Tag Parser engine.",
+          "scope": "resource"
+        },
         "C_Cpp.formatting": {
           "type": "string",
           "enum": [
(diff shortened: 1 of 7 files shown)


bobbrow:
#1574

Extension/package.json
```diff
@@ -139,6 +139,12 @@
           "description": "Controls whether suspected compile errors detected by the IntelliSense engine will be reported back to the editor. Warnings about #includes that could not be located will always be reported to the editor. This setting is ignored by the Tag Parser engine.",
           "scope": "resource"
         },
+        "C_Cpp.dimInactiveRegions": {
+          "type": "boolean",
```
sean-mcmanus, line 143: Should we be forward looking and change this to an opacity setting, e.g. 100 would be equal to false? Then when VS Code enables real opacity this setting could be used to change the amount of opacity.

sean-mcmanus (approved):

Extension/package.json
bobbrow: I'm open to that, but has VSCode committed to giving us an opacity setting?

grdowns: We have opacity, we just don't have access to the opacity AND current color, OR just the opacity w/o color. They have not responded to my GitHub issue, but I could bump the issue until they bother to respond to me.

grdowns (approved):
As a note for the future, disabling inactive regions in this manner will cause colorization from the regular expressions to appear (e.g. coloration from #if 0).

Extension/package.json
sean-mcmanus: I assume they would "eventually" implement text opacity to match VS 2017, it seems like adding an opacity setting now would confuse users since it wouldn't work (unless we make it affect the gray color).

grdowns: @bobbrow @sean-mcmanus Just bumping this while looking through this PR to learn how to do my own setting. As of 3 days ago an opacity property was exposed. https://github.com/Microsoft/vscode/issues/43004
It was merged into master, though I'm not sure when it will ship. As soon as it is, however, we can expose our own opacity setting as a 0-100 or similar as Sean suggested.

PR #1617, opened 2018-02-27
copySourceDependencies needs to be run every time

WardenGnaw:
This fixes the nightly builds. It was missing cppdbg.ad7Engine.json.

### DIFF
--- a/Extension/src/Support/prepublish.js
+++ b/Extension/src/Support/prepublish.js
@@ -18,6 +18,7 @@ if (!process.env.CPPTOOLS_DEV && fs.existsSync('./node_modules')) {
     cp.execSync("npm install", {stdio:[0, 1, 2]});
     console.log(">> tsc -p ./");
     cp.execSync("tsc -p ./", {stdio:[0, 1, 2]});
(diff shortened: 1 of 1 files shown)


pieandcakes (approved):

bobbrow:
are your nightly builds running npm install before this prepublish step?  You should remove that step from your build definition rather than copy debugger dependencies on every launch.

pieandcakes:
yes and we want to make sure 100% this file is getting copied out so we're making sure it gets run. It is copying one JSON file.

bobbrow:
Can you please turn off the warning message in that script then?

bobbrow:
Is there a reason that one file doesn't already live in the Extension folder?  IOW, why does it need to be copied?

pieandcakes:
Because that folder doesn't exist until it is built. That's why we copy it. Andrew is going to streamline the overall process but this is to unbreak nightly tests.

bobbrow:
why doesn't it exist?  can't you just move the file from the Extension folder into Extension/debugAdapters/bin in the repository and skip needing to copy it ever again?

Issue #852, opened 2017-06-28
Need way to force-include files for intellisense


nwgh:
I came across this while trying to work around #661 (I'm on a Mac, building a native app). When building firefox, we force-include <objdir>/mozilla-config.h, which has a significant number of `#define`s (around 150 by my count right now). As it stands, I can copy/paste them from my mozilla-config.h, but this file is auto-generated, and subject to change (at least once per firefox release, every 6 weeks), so that's not the most useful or practical thing to do in the long run. Our command line for building with clang includes `-include /Users/nwgh/src/mozilla/gecko/obj-dbg/mozilla-config.h`. It would be nice to have this same effect in c_cpp_properties.json without having to copy/paste and double-check my work every once in a while.

I'll also note that adding the defines from that file does not actually fix #661 for me, but from reading that issue, it at least seems like the appropriate first step.

sean-mcmanus:
This feature exists in VS 2017 C++ IntelliSense, but it's in the part of the code we haven't ported yet, which we're working on now, but we don't have an ETA, other than it won't be "soon".

aenima1891:
This feature will be very usefull.
Its absence is one of the main reasons that does not make me migrate to VSCode.

sean-mcmanus:
@nwgh @aenima1891 A workaround is to modify the bin/msvc.json file in same folder the Microsoft.VSCode.CPP.Extension process is in by appending code to the "defaults" array like:
`      "-D_USE_DECLSPECS_FOR_SAL=1",
      "--preinclude",
      "F:\\forcedInclude.h"
    ],`
We might be able to add a setting to c_cpp_properties.json in the next release, since that file gets overwritten on updates.

nwgh:
@sean-mcmanus a setting would be great (since I use this for multiple checkouts, some of which may have different definitions in my force-include), but in the meantime, the workaround is awesome! Thanks!

aenima1891:
@sean-mcmanus  Thank you for the reply. I'll be waiting impatiently the c_cpp_properties.json setting, since every project has its own forced include files

alex-che:
By an unfortunate coincidence the force-include is not supported within the compile_commands.json also (#1252).

bobbrow:
We just posted an insider release with support for forced include files.  There is a new `"forcedInclude"` string array property in the **c_cpp_properties.json** configuration schema that you can use to set the forced includes.  If you want to get an early preview, you can install the VSIX that matches your OS.

[...]

PR #1630, opened 2018-03-02
add support for force includes in c_cpp_properties.json

bobbrow:
#852

### DIFF
--- a/Extension/c_cpp_properties.schema.json
+++ b/Extension/c_cpp_properties.schema.json
@@ -47,6 +47,13 @@
                         ],
                         "description": "If set, it overrides the default mode used by the IntelliSense engine. Windows defaults to msvc-x64 and Linux/Mac default to clang-x64."
                     },
+                    "forceIncludes": {
+                        "description": "A list of files that should be included before any include file in a translation unit",
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    },
                     "browse": {
                         "type": "object",
                         "properties": {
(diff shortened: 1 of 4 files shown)


sean-mcmanus (approved):

Extension/c_cpp_properties.schema.json
```diff
@@ -47,6 +47,13 @@
                         ],
                         "description": "If set, it overrides the default mode used by the IntelliSense engine. Windows defaults to msvc-x64 and Linux/Mac default to clang-x64."
                     },
+                    "forceIncludes": {
+                        "description": "A list of files that should be included before any include file in a translation unit",
```
sean-mcmanus: I think it's supposed to end in a period?

sean-mcmanus: I think it should be "forcedIncludes"? That's what VS calls it.

bobbrow: VS calls it "forcedInclude".  ok.

sean-mcmanus:
Actually VS uses “Forced Include Files”, so forcedIncludes would be more accurate. Without the “s” it makes it sound like only 1 can be specified.

sean-mcmanus (changes requested):
Actually VS uses “Forced Include Files”, so forcedIncludes would be more accurate. Without the “s” it makes it sound like only 1 can be specified.

bobbrow:
VS's CppProperties.json file (which we've modeled c_cpp_properties.json on) spells it `"forcedInclude"`.  I have similar qualms with plurality in the other property names, but if the aim is consistency, it needs to be `"forcedInclude"`

sean-mcmanus (approved):

Issue #1392, opened 2017-12-31
Add support for ${workspaceFolder} variable


azydevelopment:
- VSCode Version: Code 1.19.1 (0759f77bb8d86658bc935a10a64f6182c5a1eeba, 2017-12-19T09:46:23.884Z)
- OS Version: Windows_NT x64 10.0.16299
- Extensions: 

Extension|Author (truncated)|Version
---|---|---
githistory|don|0.2.3
cpptools|ms-|0.14.5
csharp|ms-|1.13.1
PowerShell|ms-|1.5.1

---

Repro project: [test.zip](https://github.com/Microsoft/vscode-cpptools/files/1595155/test.zip)

Steps to Reproduce:

1. Unzip attached repro test project zip
2. Open test\0 folder with VSCode
3. Try to ctrl+click the #include <1.h> file
4. Notice that it doesn't despite in the cpp properties json file there exists "${workspaceFolder}/../1" in the include path. Since test/0 and test/1 folders are peers, I'd imagine if I go up a directory then down into the test/1 directory, the 1.h file should be resolved.

Reproduces without extensions: Yes/No

sean-mcmanus:
We don't support the variable `${workspaceFolder}` only `${workspaceRoot}`, but we should probably add that so users don't get confused. Does it work after changing that? I'm pretty sure the ".." should work, but I can double check.

azydevelopment:
Trying right now. One minute.

azydevelopment:
Okay yes it does (but you might want to make sure I'm not doing something dumb still that makes it seem like it's working but isn't really :P). Thanks. It'll be nice if it matches the new recommendations from the multi-root feature additions.

Secondarily, regarding the multi-root feature additions; it would be nice to have an environment variable that refers to the multi-root workspace root. Meaning, I can add two folders from anywhere into my workspace but the environment variable can be used to make those folders look like they're peers removing the need for messy absolute or relative paths.

PS. I'll leave this open if you want to use this item to track the addition of the workspaceFolder variable and/or the above feature request.

sean-mcmanus:
Okay, I think I understand what you're describing, something like "${realWorkspaceRoot}/folder1", "${realWorkspaceRoot}/folder2". I wonder if VS Code itself plans to add that variable for tasks.json.

azydevelopment:
I certainly hope so :). I'd use it for include paths most of all.

Zooce:
[...]

PR #1641, opened 2018-03-06
replace workspaceRoot with workspaceFolder

bobbrow:
(actual support is on the server side)
#1392

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## Version 0.16.0: March 15, 2018
 * Enable autocomplete for local and global scopes. [#13](https://github.com/Microsoft/vscode-cpptools/issues/13)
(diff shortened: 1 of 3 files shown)


Extension/CHANGELOG.md
```diff
@@ -3,11 +3,12 @@
 ## Version 0.16.0: March 15, 2018
 * Enable autocomplete for local and global scopes. [#13](https://github.com/Microsoft/vscode-cpptools/issues/13)
 * Support additional multiline comment patterns. [#1100](https://github.com/Microsoft/vscode-cpptools/issues/1100), [#1539](https://github.com/Microsoft/vscode-cpptools/issues/1539)
-  * Add new setting: `C_Cpp.commentContinuationPatterns`.
+  * Adds new setting: `C_Cpp.commentContinuationPatterns`.
```
sean-mcmanus: "Adds" doesn't match the other usages of "Add"?

sean-mcmanus (approved):
It seems like we don't replace existing uses of ${workspaceRoot}, which seems okay to me since that should still work.

Extension/CHANGELOG.md
bobbrow: I made the lines more consistent

PR #1647, opened 2018-03-07
Check whether propertiesFile is null before accessing

grdowns:


bobbrow (approved):

sean-mcmanus (approved):

bobbrow:
Travis looks like it's stalled.  I'm going to force this in so we can build an insiders package.

PR #1649, opened 2018-03-08
Fix launch.json and activation stuck on cpptools.json downloading.

sean-mcmanus:
Fix breakpoints not being hittable.
Fix debugging being stuck on cpptools.json downloading.

### DIFF
--- a/Extension/.vscode/launch.json
+++ b/Extension/.vscode/launch.json
@@ -7,10 +7,14 @@
 			"type": "extensionHost",
 			"request": "launch",
 			"runtimeExecutable": "${execPath}",
-			"args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
+			"args": [
+				"--extensionDevelopmentPath=${workspaceFolder}"
+			],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outFiles": ["${workspaceRoot}/out/src"],
+			"outFiles": [
+				"${workspaceFolder}/out/**/*.js"
+			],
 			"preLaunchTask": "npm"
 		},
 		{
@@ -18,10 +22,15 @@
 			"type": "extensionHost",
 			"request": "launch",
 			"runtimeExecutable": "${execPath}",
-			"args": ["--extensionDevelopmentPath=${workspaceRoot}", "--extensionTestsPath=${workspaceRoot}/out/test" ],
+			"args": [
+				"--extensionDevelopmentPath=${workspaceFolder}",
+				"--extensionTestsPath=${workspaceFolder}/out/test"
+			],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outFiles": ["${workspaceRoot}/out/test"],
+			"outFiles": [
+				"${workspaceFolder}/out/test/**/*.js"
+			],
 			"preLaunchTask": "npm"
 		},
 		{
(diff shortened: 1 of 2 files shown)


Extension/.vscode/launch.json
```diff
@@ -7,21 +7,30 @@
 			"type": "extensionHost",
 			"request": "launch",
 			"runtimeExecutable": "${execPath}",
-			"args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
+			"args": [
+				"--extensionDevelopmentPath=${workspaceFolder}"
+			],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outFiles": ["${workspaceRoot}/out/src"],
```
sean-mcmanus, line 13: This was the bug. The other changes don't matter.

pieandcakes (approved):

Extension/.vscode/launch.json
pieandcakes: The outfiles is the bug?

pieandcakes: I think you have the wrong line highlighted.

Extension/src/main.ts
```diff
@@ -56,7 +56,7 @@ async function processRuntimeDependencies(): Promise<void> {
             }
         // The extension have been installed and activated before.
         } else {
-            await finalizeExtensionActivation();
+            finalizeExtensionActivation();
```
pieandcakes: I think you need to remove the waiting for the download, not the function. this probably isn't the right fix since you want the language server to start.

pieandcakes (changes requested):

Extension/src/main.ts
```diff
@@ -56,7 +56,7 @@ async function processRuntimeDependencies(): Promise<void> {
             }
         // The extension have been installed and activated before.
         } else {
-            await finalizeExtensionActivation();
```
sean-mcmanus: The await causes activation exiting to be blocked on the downloadCpptoolsJson promise resolving, which we don't want.

Extension/.vscode/launch.json
sean-mcmanus: Yeah, /out/src is the bug. The regression occurred after outDir got changed to outFiles 2 weeks ago. The strange thing is that it sometimes still works (I assume some VS Code randomness).

WardenGnaw (changes requested):
You want to remove the return on the download and change the return to be Promise.resolve(). 

The change should be here https://github.com/Microsoft/vscode-cpptools/blob/5e88dbfd7923cb1932965bc416c7e358d0feceb9/Extension/src/main.ts#L285

Extension/src/main.ts
sean-mcmanus: Activation of the language server is synchronous. So there's no need to wait. I'll double check...

pieandcakes: but that call is in your finalizeExtensionActivation.

sean-mcmanus: Okay, you're right...the other awaits in finalizeExtensionActivation are causing the activateLanguageServer to execute asynchronously...not sure what negative affect that has though. I think I know what to change...

pieandcakes (approved):
Please squash when committing. Thanks for making a fix!

WardenGnaw (approved):

PR #1650, opened 2018-03-08
Fix bug with compilerPath/cppVersion.

sean-mcmanus:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -187,25 +187,6 @@ export class CppProperties {
         this.defaultIncludes = compilerDefaults.includes;
         this.defaultFrameworks = compilerDefaults.frameworks;
 
-        // Update the compilerPath, cStandard, and cppStandard with the default being used.
-        let doUpdate: boolean = false;
-        let config: Configuration = this.configurationJson.configurations[this.CurrentConfiguration];
```
sean-mcmanus, line 192: For some reason, I didn't realize this was the hardcoded default config and not the user's actual config, which isn't parsed till later.

bobbrow (approved):

PR #1721, opened 2018-03-22
Update changelog.

sean-mcmanus:


WardenGnaw:
@sean-mcmanus We are making fix to MIEngine. We are now redirecting terminal stdout and stderr to engineLogging.

bobbrow (approved):

PR #1731, opened 2018-03-23
Dev/seanmcm/0 16 0 insiders2

sean-mcmanus:


sean-mcmanus:
I just merged in master and updated the version to 0.16.0-insiders2 in the package.json and changelog.md.

PR #1741, opened 2018-03-27
Don't set the compilerPath if the default is empty.

sean-mcmanus:


sean-mcmanus:
I'm still testing this change along with the language server side changes...

Extension/src/LanguageServer/configurations.ts
```diff
@@ -475,7 +477,7 @@ export class CppProperties {
 
             // Update the compilerPath, cStandard, and cppStandard with the default if they're missing.
             let config: Configuration = this.configurationJson.configurations[this.CurrentConfiguration];
-            if (config.compilerPath === undefined) {
+            if (config.compilerPath === undefined && this.defaultCompilerPath && !config.compileCommands) {
```
bobbrow: Why would we turn this off if compileCommands is set?  There are cases where we can't get the compiler info from compile_commands.json because the command begins with a custom script or something.  I believe we would fall back to the guessed compiler in that case and should propagate the chosen compiler back to c_cpp_properties.json

bobbrow (changes requested):

bobbrow (approved):

PR #1752, opened 2018-03-28
Adding snippet to fix pipeProgram in pipeTransport

WardenGnaw:
Now with VSCode 64 bit, WSL users need to use the correct pipeProgram in
the correct folder. We will attempt to help users with this.

@gregg-miskelly

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -5,8 +5,10 @@
 
 import * as os from 'os';
 import * as vscode from 'vscode';
+import * as process from 'process';
 import { IConfiguration, IConfigurationSnippet, DebuggerType, MIConfigurations, WindowsConfigurations, WSLConfigurations, PipeTransportConfigurations } from './configurations';
 import { parse } from 'jsonc-parser';
+import { getOutputChannelLogger, showOutputChannel } from './../logger';
 
 abstract class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
     private type: DebuggerType;
@@ -32,6 +34,26 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
         if (config.type === 'cppvsdbg' && os.platform() !== 'win32') {
             vscode.window.showErrorMessage("Debugger of type: 'cppvsdbg' is only available on Windows. Use type: 'cppdbg' on the current OS platform.");
             return undefined;
+        }
+
+        // Help WSL users with using the correct bash.exe
+        if (os.platform() === 'win32' && config.pipeTransport && config.pipeTransport.pipeProgram)
+        {
+            const pipeProgramStr: string = config.pipeTransport.pipeProgram;
+
+            if (pipeProgramStr.indexOf("sysnative") >= 0 && process.arch === 'x64') {
+                // User has sysnative but is running VSCode 64 bit. Should be using System32
+                config.pipeTransport.pipeProgram = pipeProgramStr.replace("sysnative", "System32");
+                getOutputChannelLogger().appendLine(`WARNING: 64-bit VSCode should use System32 for the directory for pipeProgram.`);
+                getOutputChannelLogger().appendLine(`pipeProgram has been modified to be: ${config.pipeTransport.pipeProgram}`);
+                showOutputChannel();
+            } else if (pipeProgramStr.indexOf("System32") >= 0 && process.arch === 'ia32') {
+                // User has System32 but is running VSCode 32 bit. Should be using sysnative
+                config.pipeTransport.pipeProgram = pipeProgramStr.replace("System32", "sysnative");
+                getOutputChannelLogger().appendLine(`WARNING: 32-bit VSCode should use sysnative for the directory for pipeProgram.`);
+                getOutputChannelLogger().appendLine(`pipeProgram has been modified to be: ${config.pipeTransport.pipeProgram}`);
+                showOutputChannel();
+            }
         }
 
         return config;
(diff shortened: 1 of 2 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,6 +36,25 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
             return undefined;
         }
 
+        // Help WSL users with using the correct bash.exe
+        if (os.platform() === 'win32' && config.pipeTransport && config.pipeTransport.pipeProgram) {
+            const pipeProgramStr: string = config.pipeTransport.pipeProgram;
+
+            if (pipeProgramStr.indexOf("sysnative") >= 0 && process.arch === 'x64') {
```
gregg-miskelly: Rather than using `indexOf` here, I would suggest a helper function to fix the following problems:

1. You need a case-insensitive comparison. Really case-insensitive is wrong here and you really need an ordinal case insensitive function, but you can only do some much in JavaScript. Use `toLowerCase` on the pipeProgramStr to fix this.
2. You need to make sure that the character before and after "sysnative" is a slash or backslash.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,6 +36,25 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
+                // User has sysnative but is running VSCode 64 bit. Should be using System32
+                config.pipeTransport.pipeProgram = pipeProgramStr.replace("sysnative", "System32");
```
gregg-miskelly: same comment about `replace`.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -36,29 +39,59 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
-                getOutputChannelLogger().appendLine(`WARNING: 64-bit VSCode should use System32 for the directory for pipeProgram.`);
-                getOutputChannelLogger().appendLine(`pipeProgram has been modified to be: ${config.pipeTransport.pipeProgram}`);
-                showOutputChannel();
-            } else if (pipeProgramStr.indexOf("System32") >= 0 && process.arch === 'ia32') {
-                // User has System32 but is running VSCode 32 bit. Should be using sysnative
-                config.pipeTransport.pipeProgram = pipeProgramStr.replace("System32", "sysnative");
-                getOutputChannelLogger().appendLine(`WARNING: 32-bit VSCode should use sysnative for the directory for pipeProgram.`);
-                getOutputChannelLogger().appendLine(`pipeProgram has been modified to be: ${config.pipeTransport.pipeProgram}`);
-                showOutputChannel();
+        const winDir = process.env.WINDIR? process.env.WINDIR.toLowerCase() : null;
```
gregg-miskelly: I think you want to check for `if (os.platform() === win32)` before you check for the `WINDIR` environment variable.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -36,29 +39,59 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
+        const winDirAltDirSep =  process.env.WINDIR ? process.env.WINDIR.replace('\\', '/').toLowerCase() : null;
+
+        // Help WSL users with using the correct pipeProgram if the one they selected does not exist.
+        if (os.platform() === 'win32' &&
+            config.pipeTransport &&
+            config.pipeTransport.pipeProgram &&
+            !fs.existsSync(config.pipeTransport.pipeProgram) &&
+            (config.pipeTransport.pipeProgram.toLowerCase().indexOf(winDir) >= 0 || config.pipeTransport.pipeProgram.toLowerCase().indexOf(winDirAltDirSep) >= 0)) {
```
gregg-miskelly: Should that be `indexOf(windir) === 0` instead of `>= 0` since `windir` is a fully qualified path.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -4,9 +4,13 @@
  * ------------------------------------------------------------------------------------------ */
 
 import * as os from 'os';
+import * as process from 'process';
+import * as fs from 'fs';
```
pieandcakes: NIT: sort?

gregg-miskelly (approved):
Otherwise LGTM

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,10 +38,55 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
             return undefined;
         }
 
+        const winDir: string = process.env.WINDIR ? process.env.WINDIR.toLowerCase() : null;
```
gregg-miskelly: I think you shouldn't check for `WINDIR` unless you are on Windows

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: Do we have to worry if users have something before %WINDIR%?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,10 +38,55 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
+        const winDirAltDirSep: string =  process.env.WINDIR ? process.env.WINDIR.replace('\\', '/').toLowerCase() : null;
+
+        // Help WSL users with using the correct pipeProgram if the one they selected does not exist.
+        if (os.platform() === 'win32' &&
+            config.pipeTransport &&
```
pieandcakes: We also probably want to check the instance where pipeProgram isn't rooted and pipeCwd is specified.

pieandcakes (approved):

PR #1804, opened 2018-04-09
Allow other VS Code extensions to provide custom intellisense configurations.

sherryyshi:
Create an API for registering configurations in the currently active language client.


sherryyshi:
Still a WIP at the moment. #Closed

Extension/tsconfig.json
```diff
@@ -2,6 +2,7 @@
 	"compilerOptions": {
 		"module": "commonjs",
 		"target": "es6",
+		"declaration": false,
```
WardenGnaw: The default value of `declaration` is false. Was there confusion that this needed to be included?

Extension/package.json
```diff
@@ -38,7 +38,26 @@
     "Linters"
```
WardenGnaw: The file is rewritten when the extension activates. Please only check in changes you have made to the package.json, and not the extension's rewritten content.

WardenGnaw (commented):
Thank you for your contribution. I will wait on a full review until you mention it is ready. 

I have just a few comments that can be addressed.

You can also ignore the 
continuous-integration/appveyor/pr failure. That was a leftover artifact from an investigation on testing on Windows.

sherryyshi:
@WardenGnaw thanks! I think this is ready to be reviewed. If you have questions regarding the background/purpose of this PR, feel free to contact me. #Closed

Extension/tsconfig.json
sherryyshi: Nope, I was experimenting and checked this in by accident :)

WardenGnaw:
@sherryyshi This seems to be mostly configurations from the Language Service. I will have @sean-mcmanus and @bobbrow to take a look at this PR. They may have questions about the purpose of these changes. 

Travis-CI is complaining about a couple errors. This it to make sure that the changes follow style guidelines, the project compiles, and it properly activates the extension for debugging. You can find more about the errors in the details link next to 
continuous-integration/travis-ci/pr.

sean-mcmanus:
Bob said he was going to code review this. #Closed

Extension/src/interfaces.ts
```diff
@@ -0,0 +1,42 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+export interface ICppTools {
```
bobbrow: >I [](start = 17, length = 1)

nit: we don't use Hungarian naming conventions in TypeScript.  Please remove the 'I' from the interface name. #Closed

bobbrow:
@sherryyshi, I sent you an email about this.  Let's discuss over email or we can set up another meeting if necessary. #Closed

Extension/src/LanguageServer/client.ts
```diff
@@ -482,6 +489,16 @@ class DefaultClient implements Client {
         this.trackedDocuments.add(document);
     }
 
+    public sendCustomConfiguration(document: vscode.TextDocument, config: SourceFileConfiguration): void {
+        // TODO: Implement this
```
sherryyshi: I would need some help on this.

Extension/cpptools.json
```diff
@@ -1,3 +1,3 @@
-{
-    "intelliSenseEngine_default_percentage": 100
+{
```
bobbrow: revert changes to this file (did you remove the last newline?) #Closed

Extension/src/api.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface CppToolsApi {
+    registerCustomConfigurationProvider(provider: CustomConfigurationProvider): void;
+}
+
+export interface SourceFileConfiguration {
+    includePaths: string[];    // This must also include the system include path (compiler defaults)
+    defines: string[];        // This must also include the compiler default defines (_MSC_VER, __cplusplus, etc)
+    intelliSenseMode: string; // Probably just ‘msvc-x64’ for Windows
+    forcedInclude?: string[]; // Any files that need to be included before the source file is parsed
+   standard: string;         // This is the C or C++ standard (language server will determine if the file is C or C++ based on this)
```
bobbrow: fix indentation (off by one) #Closed

Extension/src/api.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import * as vscode from 'vscode';
+
+export interface CppToolsApi {
+    registerCustomConfigurationProvider(provider: CustomConfigurationProvider): void;
+}
+
+export interface SourceFileConfiguration {
+    includePaths: string[];    // This must also include the system include path (compiler defaults)
+    defines: string[];        // This must also include the compiler default defines (_MSC_VER, __cplusplus, etc)
```
bobbrow: remove `_MSC_VER` (Windows-specific) #Closed

Extension/src/api.ts
```diff
@@ -0,0 +1,24 @@
...
+    intelliSenseMode: string; // Probably just ‘msvc-x64’ for Windows
```
bobbrow: Remove Windows-specific comment #Closed

Extension/src/cppTools.ts
```diff
@@ -0,0 +1,14 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import { CppToolsApi, CustomConfigurationProvider } from "./api";
```
bobbrow: nit: we always use single quote `'` for imports #Closed

Extension/src/LanguageServer/client.ts
```diff
@@ -262,6 +265,7 @@ class DefaultClient implements Client {
     private failureMessageShown = new PersistentState<boolean>("DefaultClient.failureMessageShown", false);
     private isSupported: boolean = true;
     private inactiveRegionsDecorations = new Map<string, DecorationRangesPair>();
+    private customConfigurationProvider: CustomConfigurationProvider;
```
bobbrow: I think you're going to need to keep an array of these as well as an index to the active one since multiple extensions can register providers.  The user will need to select which one is active via a `QuickPick` (see LanguageServer/ui.ts) when multiple providers are available. #Closed

Extension/src/LanguageServer/client.ts
bobbrow: Take a look at the notifications defined at the top of this file under the "Notifications to the server" comment.  You'll want to follow that pattern to create the JSON we'll want to send to the language server. It would probably be a `vscode.Uri` paired with a `SourceFileConfiguration`.

`ChangeFolderSettingsNotification` is probably the closest in function to what you will be doing.  Then do

```
this.notifyWhenReady(() => this.languageClient.sendNotification(CustomConfigurationNotification, params));
```

You should move this method down under the "notifications to the language server" comment below.  There are plenty of examples to emulate down there as well. #Closed

Extension/src/LanguageServer/client.ts
bobbrow: Thinking about multi-root workspaces though, I'm thinking maybe the providers will be in a global array (maybe owned by `ClientCollection`, or maybe just global to LanguageServer/extension.ts), but the Client will only store a reference to the active provider. #Closed

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,7 +24,13 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
         didOpen: (document, sendMessage) => {
             if (clients.checkOwnership(me, document)) {
                 me.TrackedDocuments.add(document);
-                sendMessage(document);
+
+                // If custom config provider exists, ask for the custom configuration and send a notification to the server
+                // before the textDocument/didOpen message so that the right information is used when the file is opened.
+                me.CustomConfigurationProvider.provideConfiguration(document.uri, (config: SourceFileConfiguration) => {
```
bobbrow: Make sure `me.CustomConfigurationProvider` is valid first. #Closed

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,7 +24,13 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
...
+                    me.sendCustomConfiguration(document, config);  // We have to add this method to client.ts
```
bobbrow: remove this comment (you already added the method) #Closed

Extension/src/api.ts
```diff
@@ -0,0 +1,24 @@
...
+    forcedInclude?: string[]; // Any files that need to be included before the source file is parsed
+   standard: string;         // This is the C or C++ standard (language server will determine if the file is C or C++ based on this)
+}
+ 
+export interface CustomConfigurationProvider {
+    name: string;
+    provideConfiguration(uri: vscode.Uri, callback: (config: SourceFileConfiguration) => any): void;
```
bobbrow: Don't make this a callback interface.  We need the answer immediately, so this method should just return it. #Closed

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,7 +24,13 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
...
+                    sendMessage(document);
```
bobbrow: This needs to come outside the callback.  This message must always be sent. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -104,6 +120,9 @@ function realActivation(): void {
 
     reportMacCrashes();
 
+    eventEmitter.emit(activationCompleteEventId);
```
bobbrow: Having this event may be unnecessary if we just keep track of the providers in this file and send them down to the Client when they are available.  I haven't thought this all the way through yet.  I might be able to be more prescriptive once you get the UI part in for selecting providers. #Closed

Extension/src/LanguageServer/client.ts
bobbrow: Then will we also need do to some work in configurations.ts to ensure the provider name shows up in the status bar UI instead of the c_cpp_properties.json config.

bobbrow (changes requested):

Extension/src/api.ts
sherryyshi: Hmm... Our server needs to do this asynchronously because it's communication with an out-of-proc process vis json rpc... #Closed

bobbrow: Making this async may result in vscode messages being sent out of order which will result in undefined behavior and confuse the language server.  If you can't synchronize here because you think that your out of proc server is going to take too long to service the request, then we'll need to rethink this. Regardless of what we do here, sendMessage needs to be called before this method returns so you might need to block anyway unless we decide to let the language server open the file with the wrong config until your extension comes back with the correct one.

We probably need some kind of timeout and/or a try/catch block in here to ensure the safety of the provider call (in case a bad extension hangs or crashes) #Closed

Extension/src/LanguageServer/client.ts
```diff
@@ -943,7 +948,8 @@ class NullClient implements Client {
     RootUri: vscode.Uri = vscode.Uri.file("/");
     Name: string = "(empty)";
     TrackedDocuments = new Set<vscode.TextDocument>();
-    CustomConfigurationProvider: CustomConfigurationProvider;
+    CustomConcigurationProviders: CustomConfigurationProvider[] = [];
```
bobbrow: typo #Closed

Extension/src/LanguageServer/client.ts
```diff
@@ -838,6 +845,14 @@ class DefaultClient implements Client {
         this.notifyWhenReady(() => this.languageClient.sendNotification(ChangeCompileCommandsNotification, params));
     }
 
+    public sendCustomConfiguration(document: vscode.TextDocument, config: SourceFileConfiguration): void {
+        let params: CustomConfigurationParams = {
+            uri: document.uri.toString(),
+            configuration: config
+        };
+        this.notifyWhenReady(() => this.languageClient.sendNotification(ChangeCompileCommandsNotification, params));
```
bobbrow: >ChangeCompileCommandsNotification [](start = 72, length = 33)

ChangeCompileCommandsNotification -> CustomConfigurationNotification #Closed

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -7,6 +7,29 @@
 import { Middleware } from 'vscode-languageclient';
 import { ClientCollection } from './clientCollection';
 import { Client } from './client';
+import { SourceFileConfiguration, CustomConfigurationProvider } from '../api';
+import { customConfigurationProviders } from './extension';
+
+function runThenableWithTimeout(promise: Thenable<any>, ms: number): Thenable<any> {
```
bobbrow: This method should block execution.  I don't think it should return a Thenable.  It should return the configuration or something 'falsy' if it times out. #Closed

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,6 +46,24 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
         didOpen: (document, sendMessage) => {
             if (clients.checkOwnership(me, document)) {
                 me.TrackedDocuments.add(document);
+
+                // Loop through registered providers until one is able to service the current document
+                customConfigurationProviders.some((provider: CustomConfigurationProvider): boolean => {
+                    let canProvide: boolean = provider.canProvideConfiguration(document.uri);
+                    if (canProvide) {
+                        runThenableWithTimeout(
```
bobbrow: This method should block execution.  You can update `canProvide` based on the result of this function.  When you successfully get a config, you can just fall out of the loop and use the `sendMessage(document);` call at the end of the method instead of including it here. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -28,6 +29,8 @@ let realActivationOccurred: boolean = false;
 let tempCommands: vscode.Disposable[] = [];
 let activatedPreviously: PersistentWorkspaceState<boolean>;
 
+export let customConfigurationProviders: CustomConfigurationProvider[] = [];
```
bobbrow: nit: I don't really like to export global vars for every file in the extension to use if we can avoid it. Maybe we should just move all of this (array and registration method) into protocolFilter.ts?  Or else, export a "provideConfiguration" function and pass a function with your logic from protocolFilter to it. #Closed

bobbrow (changes requested):
Sorry for the delay. I've been prepping changes for our insider release this week.  I think we're almost there.  Just a few comments remaining.

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -7,6 +7,29 @@
...
+    let timeout: Promise<any> = new Promise((resolve, reject) => {
+        timer = setTimeout(() => {
+            clearTimeout(timer);
+            reject("Timed out in " + ms + "ms.");
+        }, ms);
+    });
+
+    // Returns a race between our timeout and the passed in promise
+    return Promise.race([
+        promise,
```
bobbrow: One other thing about this race is that your extension won't be notified that we're going to ignore your result, but at the same time, the request isn't cancelled.  Should we attempt to cancel it? #Closed

Extension/src/LanguageServer/protocolFilter.ts
sherryyshi: Sorry if I'm asking a naïve question (still fairly new to JavaScript/TypeScript :) ) but how do I block on a Promise without making the current function return a Promise? It seems like if I await something, the whole stack just becomes asynchronous... Do you have any references or sample code that you can refer me to?

bobbrow: You're right. I'm equally new to JavaScript and assumed that such a construct existed.  Since it doesn't look like we can block, we're probably going to need to queue up a backlog of messages and send them later when the promise is resolved.  Ugh...  client.ts does something like this with the `notifyWhenReady` and `requestWhenReady` functions. #Closed

sherryyshi: The Middleware only lets us intercept "textDocument/*" messages... But those would be the only messages that we would care to block any ways right?

bobbrow: I'd need to think if there are any dependencies between LSP and cpptools messages.  I believe activeDocumentChanged and selectionChanged need to be sent in the correct order relative to the LSP messages to be safe.  I think we'll need to add a new method to `Client` that allows us to queue up the requests when a promise is pending.  I would expect it to be bad to infinitely chain promises off of these, so we might need to track the number of pending requests as they are serviced so we can clear the promise and go back to regular processing.  We might just need to make `requestWhenReady` and `notifyWhenReady` public and update them like this (not sure if this will work as written, but this is the general idea):

```
    public requestWhenReady<T>(request: () => Thenable<T>): Thenable<T> {
        if (this.pendingRequests == 0) {
            return request();
        } else if (this.isSupported && this.pendingTask) {  // this.onReadyPromise is replaced with this.pendingTask
            this.pendingRequests++;
            return this.pendingTask.then(() => {
                let result: T = request();
                this.pendingRequests--;
                if (this.pendingRequests == 0) {
                    this.pendingTask = undefined;
                }
                return result;
            });
        } else {
            return Promise.reject<T>("Unsupported client");
        }
    }
``` #Closed

Comment (sherryyshi) [reply Extension/src/LanguageServer/protocolFilter.ts]: Yeah that makes sense... Also, what about the other idea of making the server handle the delayed config notifications? Can we jam another parameter in the didOpen message to let the server know that there is a config on the way based on canProvide()? Maybe that's easier... Let me know if you want to meet up for a brainstorming session. #Closed

Comment (bobbrow) [reply Extension/src/LanguageServer/protocolFilter.ts]: didOpen is a LSP protocol message and we shouldn't tamper with that.
JavaScript is much more suited to do this delayed notifications stuff than the language server, so I would rather keep the logic in here nearby the code that has a need for it.

(i just noticed there is no logic in the code snippet above to set this.pendingTask, so it definitely won't work as written) #Closed

Comment (sherryyshi) [reply Extension/src/LanguageServer/protocolFilter.ts]: Okay... I will try out your method. Let me know if you think of other cpptools messages that need to be blocked. #Closed

**Comment by sherryyshi** [comment]
I think we also need another API for us to notify you that a config has changed... right now the config is only provided when asked upon file open.

**Comment by bobbrow** [comment]
Yes, I agree.  Are you tracking which files are opened so that you can send us just the configs for the affected files, or were you planning on notifying the extension about all files affected by the config change?  The latter seems more expensive to do, so perhaps we will also need an API to notify your extension when a file is closed, or an API that can provide you the list of open files since we already track that. #Closed

**Comment by sherryyshi** [comment]
As of now we are not tracking open files. I think sending the list of open files would make sense, it'll be like a batched version of provideConfiguration. So it'll be something like this:
1. provider sends didCustomConfigurationChange
2. cpptools sends batched provideConfigurations request, which returns a list of configurations/falsy values, depending on whether it can provide it
3. cpptools sends cpp LSP server sendConfiguration notification (can be batched too?) for the files with new configs
How does that sound?

File: Extension/src/LanguageServer/protocolFilter.ts (hunk 21 above)
Comment (sherryyshi) [reply Extension/src/LanguageServer/protocolFilter.ts]: Yeah good point. We can add a cancellation token or something.

File: Extension/src/LanguageServer/protocolFilter.ts (hunk 18 above)
Comment (sherryyshi) [reply Extension/src/LanguageServer/protocolFilter.ts]: Just uploaded the changes. I'm still testing it out, but thought I'd share what I have. I think all cpptools/ messages are already wrapped by notifyWhenReady or requestWhenReady, so they should be safe.

**Comment by bobbrow** [comment]
yes, that sounds good.  I'll update the PR with comments in api.ts. #Closed

File: Extension/src/api.ts [hunk 22]
```diff
@@ -0,0 +1,25 @@
...
+    intelliSenseMode: string; // "msvc-x64" or "clang-x64"
+    forcedInclude?: string[]; // Any files that need to be included before the source file is parsed
+    standard: string;         // The C or C++ standard
+}
+ 
+export interface CustomConfigurationProvider {
+    name: string;
+    canProvideConfiguration(uri: vscode.Uri): boolean;
+    provideConfiguration(uri: vscode.Uri): Thenable<SourceFileConfiguration>;
+}
```
bobbrow: Add method for cpptools to request updated custom configs for tracked files.  e.g. `provideConfigurations(uris: vscode.Uri[]): Thenable<SourceFileConfiguration[]>`

Consider removing `provideConfiguration` in favor of this newer general purpose API.  This would mean you need to update client.ts to send the batch instead of a single config at a time. #Closed

bobbrow: Also add `vscode.Event` property that cpptools can subscribe to and listen for configuration change events. e.g. `didChangeCustomConfiguration: vscode.EventEmitter<T>` (can T = void?)

If this doesn't work, the alternative is to add `didChangeCustomConfigurations(provider: CustomConfigurationProvider)` to the CppToolsApi interface.  Either way is ok. #Closed

bobbrow (changes requested):

Extension/src/api.ts
```diff
@@ -0,0 +1,31 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as vscode from 'vscode';
+
+export interface CppToolsApi {
+    registerCustomConfigurationProvider(provider: CustomConfigurationProvider): void;
```
saqadri: >registerCustomConfigurationProvider [](start = 4, length = 35)

Not blocking for this PR, but can you add comments in here (for each file/property/interface), especially for the API functions? Because this is the public contract for this extension, it should be well-documented

Extension/src/LanguageServer/client.ts
```diff
@@ -483,24 +491,53 @@ class DefaultClient implements Client {
     }
 
     /*************************************************************************************
-     * wait until the language client is ready for use before attempting to send messages
+     * wait until the all pendingTasks are complete (e.g. language client is ready for use)
+     * before attempting to send messages
      *************************************************************************************/
+    public runBlockingTask(task: Thenable<any>): Thenable<any> {
```
saqadri, line 443: >any [](start = 42, length = 3)

Should this function be generic (i.e. accept type T), so this can be Thenable<T>?

saqadri: Same for requestWhenReady below

---
In reply to: [185363898](https://github.com/Microsoft/vscode-cpptools/pull/1804#discussion_r185363898) [](ancestors = 185363898)

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -7,30 +7,60 @@
 import { Middleware } from 'vscode-languageclient';
 import { ClientCollection } from './clientCollection';
 import { Client } from './client';
+import { SourceFileConfiguration, CustomConfigurationProvider } from '../api';
+import { provideCustomConfiguration } from './extension';
+
+function runThenableWithTimeout(promise: Thenable<any>, ms: number): Thenable<any> {
```
saqadri: >runThenableWithTimeout [](start = 9, length = 22)

btw general comment: might be nice to have a utils.ts for the extension with common functions such as this being co-located. 

Not blocking for this PR

saqadri (approved):
:shipit:

Extension/package.json
```diff
@@ -38,7 +38,26 @@
...
+    "onCommand:C_Cpp.ToggleErrorSquiggles",
+    "onCommand:C_Cpp.ToggleIncludeFallback",
+    "onCommand:C_Cpp.ToggleDimInactiveRegions",
+    "onCommand:C_Cpp.ShowReleaseNotes",
+    "onCommand:C_Cpp.ResetDatabase",
+    "onCommand:C_Cpp.PauseParsing",
+    "onCommand:C_Cpp.ResumeParsing",
+    "onCommand:C_Cpp.ShowParsingCommands",
+    "onCommand:C_Cpp.TakeSurvey",
+    "onDebug"
```
sherryyshi: I keep checking this file in by accident... :(

Extension/src/LanguageServer/protocolFilter.ts
sherryyshi: This should be removed. There is another copy of this in extension.ts which is what is actually being used.

Extension/src/LanguageServer/extension.ts
```diff
@@ -60,6 +62,68 @@ export function activate(activationEventOccurred: boolean): void {
...
+    let timeout: Promise<any> = new Promise((resolve, reject) => {
+        timer = setTimeout(() => {
+            clearTimeout(timer);
+            reject("Timed out in " + ms + "ms.");
+        }, ms);
+    });
+
+    // Returns a race between our timeout and the passed in promise
+    return client.runBlockingTask(Promise.race([thenable(), timeout]).then((result: any) => {
+        clearTimeout(timer);
```
sherryyshi: This line is unnecessary since the timer is cleared in the timeout promise. Will remove in the next push.

Extension/src/LanguageServer/client.ts
```diff
@@ -104,6 +105,7 @@ const ChangeFolderSettingsNotification: NotificationType<FolderSettingsParams, v
 const ChangeCompileCommandsNotification: NotificationType<FileChangedParams, void> = new NotificationType<FileChangedParams, void>('cpptools/didChangeCompileCommands');
 const ChangeSelectedSettingNotification: NotificationType<FolderSelectedSettingParams, void> = new NotificationType<FolderSelectedSettingParams, void>('cpptools/didChangeSelectedSetting');
 const IntervalTimerNotification: NotificationType<void, void> = new NotificationType<void, void>('cpptools/onIntervalTimer');
+const CustomConfigurationNotification: NotificationType<SourceFileConfigurationItem[], void> = new NotificationType<SourceFileConfigurationItem[], void>('cpptools/sendCustomConfiguration');
```
bobbrow: Can you keep this as an interface instead of an array of items?  It will make our life easier on the language server side. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -60,6 +63,72 @@ export function activate(activationEventOccurred: boolean): void {
...
+        return Promise.reject("No providers found for " + document.uri);
+    }, 1000, client, tokenSource).then((configs: SourceFileConfigurationItem[]) => {
+        if (configs !== null && configs.length > 0) {
+            client.sendCustomConfigurations(configs);
+        }
+    });
+}
+
+export function onDidCustomConfigurationChange(customConfigurationProvider: CustomConfigurationProvider): void {
+    // TODO: How to handle multiple workspaces? Only works with the current workspace.
```
bobbrow: multiple workspaces are handled by ClientCollection.  You probably just need to forward this message to ClientCollection and have it iterate on its clients. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -60,6 +63,72 @@ export function activate(activationEventOccurred: boolean): void {
...
+    // Should this be called when the workspace changes?
+    if (vscode.workspace.textDocuments === undefined || vscode.workspace.textDocuments.length === 0) {
+        return;
+    }
+
+    let documents: vscode.Uri[] = [];
+    for (let i: number = 0; i < vscode.workspace.textDocuments.length; ++i) {
```
bobbrow: You're not going to want to do it this way.  This doesn't have any notion of which workspace (read: `Client`) owns which files.  Each `Client` tracks the files it owns, so you want to have each client request updates for its tracked documents and then send the updated configs to its language server. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -60,6 +63,72 @@ export function activate(activationEventOccurred: boolean): void {
...
+        if (document.languageId === "cpp" || document.languageId === "c") {
+            documents.push(document.uri);
+        }
+    }
+
+    let tokenSource: CancellationTokenSource = new CancellationTokenSource();
+
+    // TODO: Can we rely on clients.ActiveClient to be the client for the current workspace?
+    runBlockingThenableWithTimeout(() => customConfigurationProvider.provideConfigurations(documents, tokenSource.token), 1000, clients.ActiveClient, tokenSource)
+    .then((configs: SourceFileConfigurationItem[]) => {
```
bobbrow: i think you want to `.then` the `runBlockingThenableWithTimeout` call rather than the `provideConfigurations` call so that the language server notification doesn't count toward the timeout. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -60,6 +63,72 @@ export function activate(activationEventOccurred: boolean): void {
...
+}
+
+function runBlockingThenableWithTimeout(thenable: () => Thenable<any>, ms: number, client: Client, tokenSource?: CancellationTokenSource): Thenable<any> {
+    let timer: NodeJS.Timer;
+    // Create a promise that rejects in <ms> milliseconds
+    let timeout: Promise<any> = new Promise((resolve, reject) => {
+        timer = setTimeout(() => {
+            clearTimeout(timer);
+            if (tokenSource) {
+                tokenSource.cancel();
```
bobbrow: I see you are marking the operation as cancelled, but I don't see any code that actually does anything about it.  Are there future changes coming for this? #Closed

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
sherryyshi: The token is passed to our extension, and then passed to our "configuration server", which will react to it. Is there something to be done here as well?

Extension/src/LanguageServer/extension.ts
sherryyshi: Hmm... I think that's what it's doing now

---
In reply to: [185637994](https://github.com/Microsoft/vscode-cpptools/pull/1804#discussion_r185637994) [](ancestors = 185637994)

bobbrow: Ah! I got confused parsing the long line in my head.  Can you reformat?

```
runBlockingThenableWithTimeout(() => {
    return customConfigurationProvider.provideConfigurations(documents, tokenSource.token);
}, 1000, clients.ActiveClient, tokenSource)
``` #Closed

File: Extension/src/LanguageServer/extension.ts (hunk 32 above)
Comment (bobbrow) [reply Extension/src/LanguageServer/extension.ts]: Not sure how I missed that.  I see it now. #Closed

**Comment by ThadHouse** [comment]
Looking through this, I've noticed that the system headers and system defines need to be included in the provided item. However, the code that grabbed those when passed in a compilerPath seems to be in the low level language service rather then accessable in typescript. Is there any way to get that added to the API? There are some uses for this where the build system doesn't inject the system included directly, but instead just depends on the compiler to have them.

**Comment by bobbrow** [comment]
@ThadHouse, that's a reasonable request.

File: Extension/src/api.ts [hunk 33]
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * The model representing the custom IntelliSense configurations for a source file.
+ */
+export interface SourceFileConfiguration {
+    /**
+     * This must also include the system include path (compiler defaults)
+     */
+    includePaths: string[];
```
bobbrow: this should be renamed `includePath` to match `c_cpp_properties.json`

Extension/src/api.ts
```diff
@@ -0,0 +1,110 @@
...
+     * The C or C++ standard
+     */
+    standard: string;
+
+    /**
+     * Any files that need to be included before the source file is parsed
+     */
+    forcedInclude?: string[];
+
+    /**
```
sherryyshi: @ThadHouse @bobbrow does this address the request?

ThadHouse: Does the low level language service cache values for the same compiler path? 

This looks great, but I would worry about performance if it checked the compiler for every file, even with the same path.

bobbrow: 👍

bobbrow: @ThadHouse, yes the language service will cache values to avoid invoking the compiler repeatedly. I am working on this change right now.

ThadHouse: One more question. The normal docs say skip the headers and the compilerPath is msvc mode is set. Is that still the case with this API?

bobbrow: Which docs? Clang for Windows also requires msvc mode, so you should specify the compilerPath in all cases where you want the extension to add system includes and defines for you.  We don't specifically query cl.exe in msvc mode yet since the mechanism isn't readily available (we read the registry instead), but we will in the future.

ThadHouse: https://github.com/Microsoft/vscode-cpptools/blob/master/Documentation/LanguageServer/FAQ.md

Specifically this one, where if you want to use MSVC it says don't provide a compilerPath, and only add a compilerPath if using clang for windows.

bobbrow (approved):

PR #1829, opened 2018-04-13
Adding snippet to fix pipeProgram in pipeTransport

WardenGnaw:
Note: The other PR #1752 got into a bad state so I created a new one.

Now with VSCode 64 bit, WSL users need to use the correct pipeProgram in
the correct folder. We will attempt to help users with this.

@gregg-miskelly can you take a look at this one?

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -6,7 +6,6 @@
 import * as debugUtils from './utils';
 import * as os from 'os';
 import * as path from 'path';
-import * as process from 'process';
 import * as vscode from 'vscode';
 
 import { IConfiguration, IConfigurationSnippet, DebuggerType, MIConfigurations, WindowsConfigurations, WSLConfigurations, PipeTransportConfigurations } from './configurations';
@@ -38,9 +37,8 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
             return undefined;
         }
 
-        // Modify WSL config on 64-bit VSCode for OpenDebugAD7
+        // Modify WSL config for OpenDebugAD7
         if (os.platform() === 'win32' &&
-            process.arch === 'x64' &&
             config.pipeTransport &&
             config.pipeTransport.pipeProgram) {
             let replacedPipeProgram: string = null;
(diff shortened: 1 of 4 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,6 +39,30 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
             return undefined;
         }
 
+        // Modify WSL config on 64-bit VSCode for OpenDebugAD7
+        if (os.platform() === 'win32' &&
+            process.arch === 'x64' &&
```
gregg-miskelly: Why does `process.arch` mater for this code?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -34,6 +39,30 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
+            let replacedPipeProgram: string = null;
+            const pipeProgramStr: string = config.pipeTransport.pipeProgram.toLowerCase().trim();
+
+            // OpenDebugAD7 is a 32-bit process. Make sure the WSL pipe transport is using the correct program.
+            replacedPipeProgram = debugUtils.ArchitectureReplacer.checkAndReplaceWSLPipeProgram(pipeProgramStr, debugUtils.ArchType.ia32);
+
+            // If pipeProgram does not get replaced and there is a pipeCwd, concatenate with pipeProgramStr and attempt to replace.
+            if (!replacedPipeProgram && config.pipeTransport.pipeCwd) {
+                const pipeCwdStr: string = config.pipeTransport.pipeCwd.toLowerCase().trim();
+                const newPipeProgramStr: string = path.join(pipeCwdStr, pipeProgramStr);
```
gregg-miskelly: What does `path.join` do if you give it to fully-qualified paths? Do you need to check that `pipeProgramStr` is relative first? Or does it just ignore `pipeCwdStr` if `pipeProgramStr` is already fully qualified?

gregg-miskelly (approved):
Otherwise LGTM

Extension/src/Debugger/configurationProvider.ts
gregg-miskelly: I just checked, and the results are a bit odd. I might suggest first checking if `pipeProgramStr` is relative.

```
> path.join("c:\\", "d:\\jam\\full-path.txt")
'c:\\d:\\jam\\full-path.txt'
```

WardenGnaw:
```
path.join(process.env.windir, 'temp')
>'C:\\windows\\temp'
```
I believe path.join is just ```leftParam + path.sep + rightParam```

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: I had a thought this only mattered for 64bit. I guess it would matter for 32 bit and they put system32 in the pipeProgram.

pieandcakes (approved):

PR #1841, opened 2018-04-16
Update MinGW.md

sean-mcmanus:
Bob feel free to update the docs better. I just wanted to make sure users set their compilerPath instead of deleting it after following the docs.


bobbrow:
This is fine.  I had the same thought that we need to update this documentation just now and then I saw your pull request.  I will make some additional changes today.

bobbrow (approved):

PR #1844, opened 2018-04-16
Updates to MinGW documentation

bobbrow:
Updates to MinGW documentation

### DIFF
--- a/Documentation/LanguageServer/MinGW.md
+++ b/Documentation/LanguageServer/MinGW.md
@@ -10,7 +10,7 @@ Starting with version 0.16.1, if you set `compilerPath` and `intelliSenseMode`, 
         {
             "name": "Win32",
             "intelliSenseMode": "clang-x64",
-            "compilerPath": "C:/MinGW/bin/gcc.exe"
+            "compilerPath": "C:/MinGW/bin/gcc.exe",
             "includePath": [
                 "${workspaceRoot}",
             ],
@@ -28,7 +28,7 @@ Starting with version 0.16.1, if you set `compilerPath` and `intelliSenseMode`, 
                 "databaseFilename": ""
             }
         }
-    ]
+    ],
     "version": 3
 }
 ```
--- a/Documentation/LanguageServer/MinGW.md
+++ b/Documentation/LanguageServer/MinGW.md
@@ -2,13 +2,13 @@ For developers using MinGW on Windows, we recommend you start with the following
 
 ## With extension version 0.16.1 and higher:
 
-Starting with version 0.16.1, if you set `compilerPath` and `intelliSenseMode`, you no longer need to copy the system include path or defines to `includePath` and `defines` to enable IntelliSense to work properly. However, `browse.path` still needs to be updated manually to add the system include paths to enable code browsing for system headers. For example:
+Starting with version 0.16.1, if you set the `compilerPath` property and change `intelliSenseMode` to `clang-x64`, you no longer need to copy the system include path or defines to `includePath` and `defines` to enable IntelliSense to work properly. However, `browse.path` still needs to be updated manually to add the system include paths to enable code browsing for system headers. For example:
 
 ```json
 {
     "configurations": [
         {
-            "name": "Win32",
+            "name": "MinGW",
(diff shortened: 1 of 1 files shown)


Documentation/LanguageServer/MinGW.md
```diff
@@ -1,6 +1,43 @@
 For developers using MinGW on Windows, we recommend you start with the following **c_cpp_properties.json** template.  Select "C/Cpp: Edit Configurations" from the command palette to create this file if you haven't already.
 
-Note that you may have to change the MinGW version number to match what you have installed. Eg. `C:/MinGW/lib/gcc/mingw32/5.3.0/` instead of `C:/MinGW/lib/gcc/mingw32/6.3.0/`.
+## With extension version 0.16.1 and higher:
+
+Starting with version 0.16.1, if you set `compilerPath` and `intelliSenseMode`, you no longer need to copy the system include path or defines to `includePath` and `defines` to enable IntelliSense to work properly. However, `browse.path` still needs to be updated manually to add the system include paths to enable code browsing for system headers. For example:
```
sean-mcmanus: Why are you mentioning intelliSenseMode? It doesn't actually need to be set if the OS/compiler match.

Documentation/LanguageServer/MinGW.md
```diff
@@ -1,6 +1,43 @@
...
+
+```json
+{
+    "configurations": [
+        {
```
sean-mcmanus, line 10: Can you add the cppStandard, cStandard? It's recommended to set those correctly too.

Documentation/LanguageServer/MinGW.md
```diff
@@ -1,6 +1,43 @@
...
+                    "C:/MinGW/lib/gcc/mingw32/6.3.0/include-fixed",
+                    "C:/MinGW/include/*"
+                    "${workspaceRoot}",
+                ],
+                "limitSymbolsToIncludedHeaders": true,
+                "databaseFilename": ""
+            }
+        }
+    ],
+    "version": 3
```
sean-mcmanus, line 32: Does this help doc apply to MinGW64? compilerPath works with mingw64, not sure about the other stuff. Can you add a comment about MinGW64?

sean-mcmanus (changes requested):

Documentation/LanguageServer/MinGW.md
bobbrow: The extension always sets this, and it doesn't look like the query compiler message returns a value back for this.  Do we override it if it's wrong?

Documentation/LanguageServer/MinGW.md
bobbrow: There are lots of MinGW distributions and didn't feel it necessary to mention them all, but I can add a comment that basically says to set the compilerPath to whatever compiler you want.

Documentation/LanguageServer/MinGW.md
bobbrow: We have reasonable defaults for those (and I think the extension will automatically add them for you once you set the compilerPath) so I didn't think it was necessary to talk about it.  The less information we have to convey to make things work, the better, IMO.

Documentation/LanguageServer/MinGW.md
sean-mcmanus: Oh, you're trying to say the user needs to set the intelliSenseMode to match the compilerPath compiler if there's a mismatch (e.g. msvc-x64 with a gcc/clang compiler).

Documentation/LanguageServer/MinGW.md
sean-mcmanus: Ok

sean-mcmanus (approved):

Documentation/LanguageServer/MinGW.md
bobbrow: yes.  I'm just trying to call out that property since some users weren't changing it to `clang-x64` and it defaults to `msvc-x64` on Windows.

bobbrow: I might be able to word this better

Issue #1338, opened 2017-12-12
Configuration improvements


csholmq:
When creating a new `c_cpp_properties.json` the default system includes comes first meaning that any project modules with the same name is not included. (E.g Io.h gets matched in `C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/ucrt` before the project specific)

Similarly, adding local paths gets appended to the end of the `includePath` array.

![image](https://user-images.githubusercontent.com/1297882/33876794-cae0be38-df26-11e7-8a42-0ad56fa09eb4.png)

```
{
    "name": "Win32",
    "includePath": [
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/um",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/ucrt",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/shared",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/winrt",
        "${workspaceRoot}",
        "${workspaceRoot}/A_XMC/out/obj"
    ],
    "defines": [
        "_DEBUG",
        "UNICODE"
    ],
    "intelliSenseMode": "msvc-x64",
    "browse": {
        "path": [
            "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/um",
            "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/ucrt",
            "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/shared",
            "C:/Program Files (x86)/Windows Kits/10/Include/10.0.14393.0/winrt",
            "${workspaceRoot}"
        ],
        "limitSymbolsToIncludedHeaders": true,
        "databaseFilename": ""
    }
}
```

sean-mcmanus:
Yeah, changing the ordering makes sense to me.

bobbrow:
The trouble is going to be that the default ordering will be temporary if developers use the light bulb to add include paths to this property (since they are added at the end and not in the middle).  We really need to split out the system includes from the project includes.  I hate to add yet another 'include path' property but there may be an opportunity to do this if we move c_cpp_properties.json to settings.json in the near future.  @csholmq, do you mind if we morph this issue into a design spec for how include paths could be handled in the post-c_cpp_properties.json world?  I would like to hear the community's input on the subject as well.

csholmq:
[...]

PR #1848, opened 2018-04-17
Add VS Code settings for configuration properties

bobbrow:
This adds the new settings for c_cpp_properties.json to implement the spec in #1338 

(I haven't been able to write any tests for this because the vscode.WorkspaceConfiguration API doesn't persist my attempts to update settings.  I have reached out to the VS Code team for help.)


Extension/src/LanguageServer/client.ts
```diff
@@ -177,6 +177,9 @@ function collectSettingsForTelemetry(filter: (key: string, val: string, settings
                     break;
                 }
                 default: {
+                    if (key.startsWith("default.")) {
+                        continue;   // Don't log c_cpp_properties.json defaults since they may contain PII.
```
bobbrow, line 181: Open Issue: Do we want to log a generic event to track feature usage though?

Extension/package.json
```diff
@@ -229,6 +229,118 @@
...
+          "description": "The value to use in a configuration if \"browse.databaseFilename\" is not specified, or the values to insert if \"${default}\" is present in \"browse.databaseFilename\".",
+          "scope": "resource"
+        },
+        "C_Cpp.default.browse.limitSymbolsToIncludedHeaders": {
+          "type": "boolean",
+          "default": true,
+          "description": "The value to use in a configuration if \"browse.limitSymbolsToIncludedHeaders\" is not specified, or the values to insert if \"${default}\" is present in \"browse.limitSymbolsToIncludedHeaders\".",
+          "scope": "resource"
+        },
+        "C_Cpp.default.systemIncludePath": {
```
sean-mcmanus, line 369: Why is this needed again? I keep forgetting. Oh, is this for cl.exe/Windows paths?

Extension/src/LanguageServer/client.ts
sean-mcmanus: I don't know...up to your or Rong.

sean-mcmanus (approved):

sean-mcmanus (approved):

sean-mcmanus:
Should we publish newer binaries so people running the extension from the typescript aren't broken? Maybe we can just wait till people complain...

sean-mcmanus:
Oh -- can you add this back to activationEvents: `"workspaceContains:**/c_cpp_properties.json"`, adding `"workspaceContains:**/compile_commands.json"` also seems good (assuming that doesn't exist in non-C/C++ projects)...`"workspaceContains:**/cpp.hint"` would also be valid but lower pri...we don't want `"workspaceContains:**/*.cpp"` due to the scenario where a primarily non-C/C++ project happens to have some subfolder with C/C++.

PR #1854, opened 2018-04-19
Add preferredPathSeparator setting

sean-mcmanus:


sean-mcmanus:
The change seems to work, without any typescript changes.

bobbrow (approved):
We usually send the settings the language server needs down in the createLanguageClient method in client.ts.  Perhaps this is unnecessary and can be removed in a future update since we get a didChangeConfiguration event shortly after that and we could delay initialization until then.  Until we change this though, it might be best to follow the current pattern.  I don't feel strongly about it though.

sean-mcmanus:
Sure, I followed the current pattern -- also updated the package.json version and fwlinks . There npm added some update package and the package-lock.json changed...not sure exactly why.

Extension/package.json
```diff
@@ -1295,6 +1295,7 @@
     "minimatch": "~3.0.4",
     "mkdirp": "~0.5.1",
     "tmp": "~0.0.33",
+    "update": "^0.7.4",
```
WardenGnaw: What is this package used for?

This is the one adding a lot to the package-lock.json.
It goes from 182652 bytes to 414669 bytes.

sean-mcmanus:
@WardenGnaw Do you understand the package.json update package change and the changes to package-lock.json?

WardenGnaw (changes requested):
If we don't use the update package or have an updatefile.js, please remove the update package.

WardenGnaw (approved):

sean-mcmanus:
The fwlinks all currently point to the same 0.16.1 binaries, but the targets will be updated when they're ready.

PR #1872, opened 2018-04-24
wrong name for enum

bobbrow:


Extension/package.json
```diff
@@ -205,7 +205,7 @@
           "type": "string",
           "enum": [
             "Forward Slash",
-            "Back Slash"
+            "Backslash"
```
sean-mcmanus, line 208: What?

sean-mcmanus: Okay, looks like "Backslash" is a special word.

bobbrow: nah, just a regular word 😉

PR #1886, opened 2018-04-26
Initial implementation for supporting vcpkg

greazer:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -225,6 +229,35 @@ export class CppProperties {
...
+                               }
+                            }
+                        });
+                        this.vcpkgPathReady = true;
+                        this.handleConfigurationChange();
+                    })
+                };
+            }
+        }
+        catch(error) {}
```
bobbrow: catch should set this.vcpkgPathReady = true and call this.handleConfigurationChange() so that we don't get stuck forever waiting for data that's never coming. #Closed

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -197,15 +202,14 @@ export class CppProperties {
...
+                configuration.includePath = ["${workspaceFolder}"].concat(this.vcpkgIncludes);
+            }
+            if (!settings.defaultBrowsePath) {
+                configuration.browse.path = ["${workspaceFolder}"].concat(this.vcpkgIncludes);
             }
             if (!settings.defaultDefines) {
                 configuration.defines = (process.platform === 'win32') ? ["_DEBUG", "UNICODE", "_UNICODE"] : [];
             }
-            if (!settings.defaultBrowsePath) {
-                // We don't add system includes to the browse.path anymore. The language server has this information.
```
bobbrow, line 206: can you add this comment back to the block where you update the browse.path? #Closed

Extension/src/LanguageServer/configurations.ts
greazer: Ah yeah. But a finally would be better for this, no? We need to call these two operations no matter what.  #Closed

sean-mcmanus (approved):

bobbrow (approved):

Issue #1873, opened 2018-04-24
Extension fails silently to install if extensions-home-folder is missing execution rights on Windows


ankostis:
The extension (in VSCode-1.22.2 ) cannot work under Windows if folder `%USERPROFILE%\.vscode\extensions` has not *execution* rights.
This problem can be solved using `--extensionHomePath` option.

The real problem is that there is no log record explaining this installation problem when it happens!

Below are the relevant logs,  after having waited for at least 5':

### after *install*:
```
[2018-04-25 18:59:45.056] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:45.219] [sharedprocess] [info] Scanned system extensions: 69
[2018-04-25 18:59:47.152] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:47.154] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:47.191] [sharedprocess] [info] Scanned system extensions: 69
[2018-04-25 18:59:51.990] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:58.192] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:58.195] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:58.195] [sharedprocess] [info] Scanned user extensions: 0
[2018-04-25 18:59:58.280] [sharedprocess] [info] Scanned system extensions: 69
[2018-04-25 18:59:58.283] [sharedprocess] [info] Scanned system extensions: 69
[2018-04-25 19:00:15.607] [sharedprocess] [info] Installing extension: cpptools
[2018-04-25 19:00:16.257] [sharedprocess] [info] Downloaded extension: cpptools
[2018-04-25 19:00:21.298] [sharedprocess] [info] Extracted extension to C:\Users\username\.vscode\extensions\.ms-vscode.cpptools-0.16.1: ms-vscode.cpptools-0.16.1
[2018-04-25 19:00:21.300] [sharedprocess] [info] Renamed to C:\Users\username\.vscode\extensions\ms-vscode.cpptools-0.16.1
[2018-04-25 19:00:21.300] [sharedprocess] [info] Installation completed. ms-vscode.cpptools-0.16.1
[2018-04-25 19:00:21.398] [sharedprocess] [info] Extensions installed successfully: ms-vscode.cpptools
[...]

PR #1893, opened 2018-04-27
shouldn't be attempting to re-register vscode's commands

bobbrow:
Noticed in customer logs for issue #1873


sean-mcmanus (approved):
This used to be required to get the menu commands to work, but this no longer appears to be the case.

PR #1896, opened 2018-04-27
Recursive includePath changes.

sean-mcmanus:


sean-mcmanus:
I added a small change to make the config and icons visible with c_cpp_properties.json/settings.json active.

Extension/src/LanguageServer/ui.ts
```diff
@@ -110,12 +110,15 @@ export class UI {
 
     public activeDocumentChanged(): void {
         let activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
-        let show: boolean = (activeEditor && (activeEditor.document.languageId === "cpp" || activeEditor.document.languageId === "c"));
+        let isCpp: boolean = (activeEditor && ((activeEditor.document.languageId === "cpp" || activeEditor.document.languageId === "c")));
```
bobbrow: Why add the extra parentheses?

bobbrow (approved):

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Oops, that's stale code -- also I just noticed a missing parenthesis in the isSettingsJson so I'll fix that.

PR #1910, opened 2018-05-01
Provide user hint through IntelliSense fallback msg that vcpkg is installed and is being used.

greazer:
In addition, I needed to change the directory that our package uses to determine whether vcpkg is installed and has any installed libraries.


sean-mcmanus (approved):

Extension/package.json
```diff
@@ -1306,7 +1306,7 @@
     "vscode-debugadapter": "~1.24.0",
     "vscode-debugprotocol": "~1.24.0",
     "vscode-extension-telemetry": "~0.0.11",
-    "vscode-languageclient": "~3.4.5",
+    "vscode-languageclient": "3.5.1",
```
greazer, line 1309: This was necessary to allow builds on Mac (and probably linux) to work.

Extension/src/LanguageServer/client.ts
```diff
@@ -686,7 +686,10 @@ class DefaultClient implements Client {
             if (showIntelliSenseFallbackMessage.Value) {
                 let learnMorePanel: string = "Learn More";
                 let dontShowAgain: string = "Don't Show Again";
-                vscode.window.showInformationMessage("Configure includePath for better IntelliSense results.", learnMorePanel, dontShowAgain).then((value) => {
+                let fallbackMsg: string = this.configuration.VcpkgInstalled ?   
+                    "Update your IntelliSense settings or use vcpkg to install libraries to help find missing headers." :
```
bobbrow: double check if there is any capitalization for `vcpkg` and then you're good to go.

bobbrow (approved):

PR #1916, opened 2018-05-02
Move settings update into updateToVersion4 function

grdowns:
Cleanup code by moving resolution of non-existent entries into an updateToVersion function, which is then triggered if configVersion is less than the new version, 4.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -10,7 +10,7 @@ import * as vscode from 'vscode';
 import * as util from '../common';
 import { PersistentFolderState } from './persistentState';
 import { CppSettings } from './settings';
-const configVersion: number = 3;
+const configVersion: number = 4;
 
 // No properties are set in the config since we want to apply vscode settings first (if applicable).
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -606,6 +586,30 @@ export class CppProperties {
...
 
+    private updateToVersion4(): void {
+        this.configurationJson.version = 4;
+        // Update intelliSenseMode, compilerPath, cStandard, and cppStandard with the defaults if they're missing.
+        // If VS Code settings exist for these properties, don't add them to c_cpp_properties.json
+        let settings: CppSettings = new CppSettings(this.rootUri);
+        for (let i: number = 0; i < this.configurationJson.configurations.length; i++) {
+            let config: Configuration = this.configurationJson.configurations[i];
+
+            if (config.intelliSenseMode === undefined && !settings.defaultIntelliSenseMode) {
```
bobbrow, line 602: if you move intelliSenseMode here, then the language server needs to be updated too since this property is not marked as optional.

bobbrow (approved):

Issue #1857, opened 2018-04-20
Request Feature: Support customize environment variable in c_cpp_properites.json


453jerry:
When a project has a lot of include path and all include path is absolute path, then maintain c_cpp_properites.json is very uncomfortable. 
for example:   
``` JSON
"name": "Win32",
            "includePath": [
                "C:/Program Files (x86)/GNU Tools ARM Embedded/7 2017-q4-major/arm-none-eabi/include",
                "C:/Program Files (x86)/GNU Tools ARM Embedded/7 2017-q4-major/arm-none-eabi/include/c++/7.2.1",
                "C:/Program Files (x86)/GNU Tools ARM Embedded/7 2017-q4-major/arm-none-eabi/include/c++/7.2.1/arm-none-eabi",
                "C:/Program Files (x86)/GNU Tools ARM Embedded/7 2017-q4-major/lib/gcc/arm-none-eabi/7.2.1/include",
                "C:/Program Files (x86)/GNU Tools ARM Embedded/7 2017-q4-major/lib/gcc/arm-none-eabi/7.2.1/include-fixed",
                "${workspaceFolder}",
                "C:/nRF5_SDK_15.0.0_a53641a/components",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/experimental_memobj",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/experimental_section_vars",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/balloc",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/experimental_log",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/experimental_log/src",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/delay",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/bsp",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/strerror",
                "C:/nRF5_SDK_15.0.0_a53641a/components/libraries/util",
                "C:/nRF5_SDK_15.0.0_a53641a/components/drivers_nrf/nrf_soc_nosd",
                "C:/nRF5_SDK_15.0.0_a53641a/components/softdevice/mbr/nrf52832/headers",
                "C:/nRF5_SDK_15.0.0_a53641a/components/boards",
                "C:/nRF5_SDK_15.0.0_a53641a/components/toolchain/cmsis/include",
[...]

Issue #368, opened 2016-11-24
can we set includePath in c_cpp_properties.json globally?


abc100m:
I would like to set includePath  globally, not set it in every folder/project.
can we make this config file available in  settings.json?

```json
{
    "configurations": [
        {
            "name": "Mac",
            "includePath": ["/usr/include"],
            "browse" : {
                "limitSymbolsToIncludedHeaders" : true,
                "databaseFilename" : ""
            }
        },
        {
            "name": "Linux",
            "includePath": ["/usr/include"],
            "browse" : {
                "limitSymbolsToIncludedHeaders" : true,
                "databaseFilename" : ""
            }
        },
        {
            "name": "Win32",
            "includePath": ["c:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/include"],
            "browse" : {
                "limitSymbolsToIncludedHeaders" : true,
                "databaseFilename" : ""
            }
        }
    ]
}
```

BillDenton:
This would also be very useful for me as whole my projects/folders use the same settings.

sean-mcmanus:
I agree. We'll consider adding that later.

BillDenton:
I got around the problem by create a common config, with symbolic links to it in each project/folders .vscode directory. However, it would still be better if the settings where in settings.json. Like a lot of other extensions shouldn't the settings be in the standard settings.json. That way they could be set globally or locally as required. It would also remove one settings file.

Gruntfuggly:
Surely the simplest way to address this would be to look for the c_cpp_properties.json file in the user settings folder if it isn't found for the workspace?

luni3359:
Right now I just copy-paste it into every project, it's a bit annoying and it wouldn't be a bad idea to have this.

hb9tst:
Same issue for me, sometimes I need to update the settings for particular projects, but in most cases they are the same for all of them.

Alex031544:
Hi, same issue here. It would be very comfortable if the search paths can be defined globally. Search paths depending on the OS set up and can be different. So it is not nice to have this inside the  git managed project folders.

It would be really nice if you can find a way to bring this configuration file into */home/user/.vscode/**

Thanks

seanchann:
same issue for me. it should be work together with user custom library.

horvay:
What I do right now is create a VisualCode folder somewhere, and put my task, settings, cpp config in there.

Then I create a whatever.code-workspace file that looks something like this:
```json
{
[...]

PR #1921, opened 2018-05-03
add support for custom variables

bobbrow:


Extension/c_cpp_properties.schema.json
```diff
@@ -107,6 +107,14 @@
                 }
             }
         },
+        "variables" : {
```
bobbrow: I debated between "environment" and "variables", but I'm open to suggestions on what to call this.  Visual Studio's CppProperties.json has an array called "environments" (yes, plural) that lets you do fancy namespacing, but i don't think we want that.

bobbrow:
#1857, #368

sean-mcmanus (approved):

Extension/c_cpp_properties.schema.json
bobbrow: launch.json and tasks.json use "env", so I'm thinking maybe we'll go with that to be consistent.

PR #1937, opened 2018-05-04
Adding OptionsSchema generation

WardenGnaw:
This is to prevent repetitive fields from being copy and pasted
improperly.

There are some parts of OptionsSchema.json that needs to be fixed. E.g. logging for cppvsdbg.

### DIFF
--- a/Extension/tools/GenerateOptionsSchema.ts
+++ b/Extension/tools/GenerateOptionsSchema.ts
@@ -96,7 +96,7 @@ function ReplaceReferences(definitions: any, objects: any) {
     return objects;
 }
 
-function mergeReferences(baseDefinitions: any, additionalDefinitions: any) : void {
+function MergeReferences(baseDefinitions: any, additionalDefinitions: any) : void {
     for (let key in additionalDefinitions) {
         if (baseDefinitions[key]) {
             throw `Error: '${key}' defined in multiple schema files.`;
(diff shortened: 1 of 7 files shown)


Extension/tools/GenerateOptionsSchema.ts
```diff
@@ -0,0 +1,133 @@
...
+        // Recursively replace references if the array has objects in items.
+        if (objects[key].hasOwnProperty('type') && objects[key].type === "array" && objects[key].items != null && objects[key].items.hasOwnProperty('$ref')) {
+            objects[key].items = RefReplace(definitions, objects[key].items);
+        }
+    }
+
+    return objects;
+}
+
+function mergeReferences(baseDefinitions: any, additionalDefinitions: any) : void {
```
bobbrow: inconsistent capitalization of functions.  We generally follow the JavaScript conventions of camelCasing for functions in the extension, but I know there are some places where PascalCasing has been used.  Maybe we need a tslint rule for that.

bobbrow (approved):

Extension/tools/GenerateOptionsSchema.ts
WardenGnaw: There not one in the core ruleset. https://github.com/palantir/tslint/issues/2277

There is a package that does it: https://www.npmjs.com/package/vrsource-tslint-rules#ext-variable-name

I'll add this as its own PR since it might affect a lot of things. And since its a dev-dependency, we dont need to update third party licenses text?

sean-mcmanus (approved):

PR #1948, opened 2018-05-07
update release notes and don't enable recursive includes by default in new folders

bobbrow:


sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -206,7 +206,7 @@ export class CppProperties {
 
             if (!settings.defaultIncludePath) {
                 // We don't add system includes to the includePath anymore. The language server has this information.
-                configuration.includePath = ["${workspaceFolder}/**"].concat(this.vcpkgIncludes);
+                configuration.includePath = ["${workspaceFolder}"].concat(this.vcpkgIncludes);
```
WardenGnaw, line 209: nit: `this.vcpkgIncludes.unshift("${workspaceFolder}")`

WardenGnaw (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
bobbrow: My understanding is that unshift modifies the source array and returns the length instead of a new array.  This is not what we want here.

WardenGnaw: Whoops, I thought we were adding to this.vspkgIncludes. Nevermind!

PR #1983, opened 2018-05-11
Fix the file watcher

bobbrow:
Something must have changed in VS Code with the file watcher API because we stopped getting file created/deleted events.  This change fixes it.


Extension/src/LanguageServer/client.ts
```diff
@@ -450,7 +450,7 @@ class DefaultClient implements Client {
         if (this.rootFolder) {
             // WARNING: The default limit on Linux is 8k, so for big directories, this can cause file watching to fail.
             this.rootPathFileWatcher = vscode.workspace.createFileSystemWatcher(
-                path.join(this.RootPath, "*"),
+                "**/*",
```
bobbrow, line 453: note that we don't get file delete notifications for every file in a deleted folder, so we'll need to update the language server to take care of that specially.

sean-mcmanus (approved):

PR #2059, opened 2018-05-30
Disable recursive `includePath` for ${workspaceFolder} in default c_cpp_properties.json

yunlingz:
# Disable recursive `includePath` for ${workspaceFolder} in default c_cpp_properties.json

## Origin c_cpp_properties.json generated

```
"includePath": [
	"${workspaceFolder}/**"
],
```

## Now

```
"includePath": [
	"${workspaceFolder}"
],
```

## Why

On macOS 10.13.4, VS Code Version 1.23.1. **C/C++ Extension 0.17.3**

Maybe you can test the Extension based on this environment.

If we use the origin auto-generated c_cpp_properties.json which suggest `${workspaceFolder}/**` in includePath, this will cause the automatic completion to fail, and sometimes it will lead to high CPU usage.

qq20180530-123411 2x

After I change `${workspaceFolder}/**` to `${workspaceFolder}` in includePath, Automatic completion started working properly.

screen shot 2018-05-30 at 12 35 48 pm

I searched the internet and I found this bug not only happened to me. I just happened to find that disable recursive `includePath` for ${workspaceFolder} in default c_cpp_properties.json can solve this problem. (But I think this may be a bug about recursive `includePath`.)

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -207,7 +207,7 @@ export class CppProperties {
 
             if (!settings.defaultIncludePath) {
                 // We don't add system includes to the includePath anymore. The language server has this information.
-                configuration.includePath = ["${workspaceFolder}/**"].concat(this.vcpkgIncludes);
+                configuration.includePath = ["${workspaceFolder}"].concat(this.vcpkgIncludes);
             }
             if (!settings.defaultBrowsePath) {
                 // We don't add system includes to the includePath anymore. The language server has this information.


sean-mcmanus:
Can you give more repro info on the case where automatic completion fails? If you add a new header you may need to save first before the header will be picked up for completions.

yunlingz:
@sean-mcmanus  The environment description has been updated in the previous paragraph, maybe you can test the Extension based on this environment (macOS 10.13.4, VS Code Version 1.23.1. C/C++ Extension 0.17.3).

sean-mcmanus:
We've fixed the 100% CPU "livelock" and improved the perf by ~4 times with recursive includes for 0.17.4-insiders so we don't think disabling it by default will be necessary, but we might change our minds. We're also investigating additional performance improvements for later.

I think you can also set the C_Cpp.default.includePath setting to "${workspaceFolder}" as another workaround.

yunlingz:
Thanks for your great work, I will try the insiders branch.

bobbrow:
@Hyzeta we plan to publish a VSIX with the fix this week.  Cloning the insiders branch and building yourself won't actually pick up the fix since it's in the server code which is not open source.

sean-mcmanus:
@Hyzeta Our perf fixes aren't ready yet, so we decided to disable recursive includes by default for 0.17.4.

sean-mcmanus (approved):

PR #2061, opened 2018-05-30
Grdowns/inactive region opacity

grdowns:
Add opacity setting for inactive preprocessor blocks to replace text greying


Extension/src/LanguageServer/client.ts
```diff
@@ -705,12 +705,11 @@ class DefaultClient implements Client {
...
-        let renderOptions: vscode.DecorationRenderOptions = {
-            light: { color: "rgba(175,175,175,1.0)" },
-            dark: { color: "rgba(155,155,155,1.0)" },
-            rangeBehavior: vscode.DecorationRangeBehavior.ClosedOpen
-        };
-        let decoration: vscode.TextEditorDecorationType = vscode.window.createTextEditorDecorationType(renderOptions);
+        let settings: CppSettings = new CppSettings(this.RootUri);
+        
+        let decoration: vscode.TextEditorDecorationType = vscode.window.createTextEditorDecorationType({
+            opacity: String(settings.inactiveRegionOpacity)
```
WardenGnaw: Do we need any error validation here?

WardenGnaw (approved):

Extension/package.json
```diff
@@ -146,6 +146,12 @@
           "description": "Controls whether inactive preprocessor blocks are colored differently than active code. This setting is ignored by the Tag Parser engine.",
           "scope": "resource"
         },
+        "C_Cpp.inactiveRegionOpacity": {
+          "type:": "number",
+          "default": 0.5,
```
sean-mcmanus: Can this be 0.55 to match Visual Studio's default opacity?

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, should be a number between 0 and 100...although maybe 0 is too low if the text becomes invisible.

Extension/package.json
```diff
@@ -146,6 +146,12 @@
...
+          "Description": "Controls the opacity of inactive preprocessor blocks. Scales between 0.0 and 1.0. This setting is overridden by dimInactiveRegions.",
```
bobbrow: I think it would be more clear if we said "This setting only applies when inactive region dimming is enabled."

Extension/package.json
bobbrow: Can you add `"minimum": 0` and `"maximum": 1` to the schema so that VS Code will squiggle values outside the valid range?

Extension/src/LanguageServer/client.ts
bobbrow: you shouldn't need to make this a string since you already enforced the inactiveRegionOpacity setting to be a string in settings.ts.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
grdowns: ```opacity``` is a ```S```tring, not a ```s```tring, unfortunately. Is it okay to do the conversion here?

Extension/package.json
grdowns: Added!

Extension/src/LanguageServer/client.ts
grdowns: We don't do it elsewhere in the codebase. I believe the line of thought is ```CppSettings``` should always be constructible, as ```this.RootUri``` should always be valid. @bobbrow?

bobbrow: I think Andrew is referring to the inactiveRegionOpacity setting

Extension/src/LanguageServer/client.ts
```diff
@@ -705,12 +705,11 @@ class DefaultClient implements Client {
     }
 
     private updateInactiveRegions(params: InactiveRegionParams): void {
-        let renderOptions: vscode.DecorationRenderOptions = {
-            light: { color: "rgba(175,175,175,1.0)" },
-            dark: { color: "rgba(155,155,155,1.0)" },
-            rangeBehavior: vscode.DecorationRangeBehavior.ClosedOpen
```
bobbrow: do we need to keep the ClosedOpen rangeBehavior?

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, we need that.

grdowns: Re-added!

sean-mcmanus (approved):

sean-mcmanus:
@bobbrow Can you unblock this now?

bobbrow (approved):

olr666:
Is it possible to restore the greyed text and leave users with the option to choose either grey inactive regions OR dimmed but coloured inactive regions? Some people do prefer greyed sections to dimmed-but-still-coloured.

sean-mcmanus:
@olr666 It's possible, but I don't think we were planning to implement that unless we got more user feedback. We could accept a pull request for some bool grayInactiveRegions setting that defaulted to false...maybe the inactiveRegionOpacity could be used to change the gray color as well?

bobbrow:
I think adding an `inactiveRegionColor` setting would be better.  When not set we leave the colors alone. When set, we override the color.

PR #2126, opened 2018-06-13
Updated the link for creating issues.

jameshamm:
The current link returns a 301 (permanent redirect) so we may as well just point straight to the site.


WardenGnaw (approved):
Thank you for your contribution to the extension.

PR #2135, opened 2018-06-15
make browse.path optional

bobbrow:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -403,7 +391,23 @@ export class CppProperties {
...
-            configuration.browse.path = this.updateConfiguration(configuration.browse.path, settings.defaultBrowsePath);
+
+            if (!configuration.browse.path) {
+                if (settings.defaultBrowsePath) {
+                    configuration.browse.path = settings.defaultBrowsePath;
+                } else if (configuration.includePath) {
+                    // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
+                    configuration.browse.path = configuration.includePath.slice(0);
+                    if (-1 === configuration.includePath.findIndex((value: string, index: number) => {
+                        return !!value.match(/^\$\{(workspaceRoot|workspaceFolder)\}(\\|\\\*\*|\/|\/\*\*)?$/g);
```
sean-mcmanus, line 402: Why !!?

sean-mcmanus: Oh, value.match return null if it's not found, so !null == true, and !true is false...maybe using != null would be less confusing?

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
bobbrow: i didn't want to check for empty array vs null vs undefined

PR #2136, opened 2018-06-15
Bobbrow/recursive includes

bobbrow:
I'm still waiting for the updated json file to upload, but this code can be reviewed now

### DIFF
--- a/Extension/cpptools.json
+++ b/Extension/cpptools.json
@@ -1,3 +1,5 @@
 {
-    "intelliSenseEngine_default_percentage": 100
+    "intelliSenseEngine_default_percentage": 100,
+    "defaultIntelliSenseEngine": 100,
+    "recursiveIncludes": 100
 }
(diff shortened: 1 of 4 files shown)


Extension/cpptools.json
```diff
@@ -1,3 +1,5 @@
 {
-    "intelliSenseEngine_default_percentage": 100
+    "intelliSenseEngine_default_percentage": 100,
```
bobbrow, line 2: keeping the old one around unless we want to create a new fwlink url

Extension/src/abTesting.ts
```diff
@@ -9,88 +9,132 @@ import * as https from 'https';
...
-                let redirectUrl: string | string[];
-                if (typeof response.headers.location === "string") {
-                    redirectUrl = response.headers.location;
-                } else {
-                    redirectUrl = response.headers.location[0];
+    private async updateSettingsAsync(): Promise<void> {
+        const cpptoolsJsonFile: string = util.getExtensionFilePath(localConfigFile);
+    
+        try {
+            const exists: boolean = await util.checkFileExists(cpptoolsJsonFile);
```
bobbrow, line 66: not sure we need to check the file exists since we're in a try/catch already...

sean-mcmanus (approved):

PR #2137, opened 2018-06-15
Grdowns/detect compile commands

grdowns:
Respond to a `compileCommandsPaths` message by prompting the user to select a `compile_commands.json` which will then populate `compileCommands` in `c_cpp_properties.json` for the active configuration.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -92,6 +92,10 @@ interface CustomConfigurationParams {
     configurationItems: SourceFileConfigurationItem[];
 }
 
+interface CompileCommandsPaths {
+    paths: string[];
+}
+
 // Requests
 const NavigationListRequest: RequestType<TextDocumentIdentifier, string, void, void> = new RequestType<TextDocumentIdentifier, string, void, void>('cpptools/requestNavigationList');
 const GoToDeclarationRequest: RequestType<void, void, void, void> = new RequestType<void, void, void, void>('cpptools/goToDeclaration');
@@ -122,6 +126,7 @@ const ReportStatusNotification: NotificationType<ReportStatusNotificationBody, v
 const DebugProtocolNotification: NotificationType<OutputNotificationBody, void> = new NotificationType<OutputNotificationBody, void>('cpptools/debugProtocol');
 const DebugLogNotification:  NotificationType<OutputNotificationBody, void> = new NotificationType<OutputNotificationBody, void>('cpptools/debugLog');
 const InactiveRegionNotification:  NotificationType<InactiveRegionParams, void> = new NotificationType<InactiveRegionParams, void>('cpptools/inactiveRegions');
+const CompileCommandsPathsNotification:  NotificationType<CompileCommandsPaths, void> = new NotificationType<CompileCommandsPaths, void>('cpptools/compileCommandsPaths');
 
 let failureMessageShown: boolean = false;
 
@@ -521,6 +526,7 @@ class DefaultClient implements Client {
         this.languageClient.onNotification(ReportStatusNotification, (e) => this.updateStatus(e));
         this.languageClient.onNotification(ReportTagParseStatusNotification, (e) => this.updateTagParseStatus(e));
         this.languageClient.onNotification(InactiveRegionNotification, (e) => this.updateInactiveRegions(e));
+        this.languageClient.onNotification(CompileCommandsPathsNotification, (e) => this.promptCompileCommands(e));
         this.setupOutputHandlers();
     }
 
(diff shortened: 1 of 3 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,43 @@ class DefaultClient implements Client {
...
     }
 
+    private promptCompileCommands(params: CompileCommandsPaths) : void {
+        if (this.configuration.Configurations[this.configuration.CurrentConfiguration].compileCommands !== undefined) {
+            return;
+        }
+
+        let showCompileCommandsSelection: PersistentState<boolean> = new PersistentState<boolean>("CPP.showPromptCompileCommands", true);
+        if (!showCompileCommandsSelection.Value) {
+           return;
```
bobbrow: >    [](start = 8, length = 3)

nit: looks like indentation is off by one space here

grdowns: Fixed!

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,43 @@ class DefaultClient implements Client {
...
+        }
+
+        let message: string = "Detected compile_commands.json. Would you like to select a compile_commands.json for this configuration?";
```
bobbrow: Do we want to separate the simple case (only 1 found) from the "we need more information" case so that the user only has to click once?

If there is only one `compile_commands.json` found, we could just say "Would you like to use `[path]` to configure IntelliSense?"

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,43 @@ class DefaultClient implements Client {
...
+        let yes: string = "Yes";
+        let later: string = "Later";
```
bobbrow: "Later" feels like an affirmative choice.  I think you meant "Ask Me Later".  In the interest of keeping labels short, "Not Now" is another alternative.

grdowns: screenshot
I did it to match VSCode's update message. I do agree that the alternatives you offer sound better.

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,43 @@ class DefaultClient implements Client {
...
+        let dontShowAgain: string = "Don't Show Again";
```
bobbrow: I think "Don't Ask Again" matches the context of this popup a little better

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,43 @@ class DefaultClient implements Client {
         }
     }
 
+    private promptCompileCommands(params: CompileCommandsPaths) : void {
+        if (this.configuration.Configurations[this.configuration.CurrentConfiguration].compileCommands !== undefined) {
+            return;
+        }
+
+        let showCompileCommandsSelection: PersistentState<boolean> = new PersistentState<boolean>("CPP.showPromptCompileCommands", true);
```
bobbrow: PersistentState is global.  We might only want to disable it for the current workspace.  You can use `PersistentFolderState` for that.

Extension/src/LanguageServer/client.ts
bobbrow: Thanks for the context. They are not asking a question, so "Later" is not an answer.

bobbrow (changes requested):
We need to settle the UI presentation

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,47 @@ class DefaultClient implements Client {
...
+        if (this.configuration.Configurations[this.configuration.CurrentConfiguration].compileCommands !== undefined) {
+            return;
+        }
+
+        let showCompileCommandsSelection: PersistentState<boolean> = new PersistentState<boolean>("CPP.showCompileCommandsSelection", true);
+        if (!showCompileCommandsSelection.Value) {
+            return;
+        }
+
+        let message: string = "Would you like to use compile_command.json files to auto-configure IntelliSense?";
```
WardenGnaw: nit: these `let` should be `const`

Extension/src/LanguageServer/client.ts
```diff
@@ -761,27 +761,33 @@ class DefaultClient implements Client {
             return;
         }
 
-        let message: string = "Detected compile_commands.json. Would you like to select a compile_commands.json for this configuration?";
+        let message: string = "Would you like to use compile_command.json files to auto-configure IntelliSense?";
```
bobbrow: [sp] `compile_commands.json`

Extension/src/LanguageServer/ui.ts
```diff
@@ -168,6 +168,25 @@ export class UI {
...
 
+    public showCompileCommands(paths: string[]): Thenable<number> {
+        let options: vscode.QuickPickOptions = {};
+        options.placeHolder = "Select a compile_commands.json...";
+
+        let items: IndexableQuickPickItem[] = [];
+        for (let i: number = 0; i < paths.length; i++) {
+            items.push({label: paths[i], description: "", index: i});
+        }
+        items.push({ label: "Edit Configurations...", description: "", index: paths.length });
```
bobbrow: I didn't see this in the spec.  You should ask Rong if she wants this or not.  My vote is that it's unnecessary.

Extension/src/LanguageServer/ui.ts
```diff
@@ -168,6 +168,25 @@ export class UI {
...
+
+        return vscode.window.showQuickPick(items, options)
+        .then(selection => {
```
bobbrow: nit: please indent the `then` block to match coding style elsewhere in the file.

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,47 @@ class DefaultClient implements Client {
...
+        let dontAskAgain: string = "Don't Ask Again";
+        vscode.window.showInformationMessage(message, yes, notNow, dontAskAgain).then((value) => {
+            switch (value) {
+                case yes:
+                    if (params.paths.length > 1) {
+                        ui.showCompileCommands(params.paths).then((index) => {
+                            if (index < 0) {
+                                return;
+                            }
+                            if (index >= params.paths.length) {
```
bobbrow: if we keep this, please add a comment that this is because "edit configurations" is added to the list.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
grdowns: Removed after speaking with Rong :smile:

Extension/src/LanguageServer/client.ts
```diff
@@ -745,6 +751,46 @@ class DefaultClient implements Client {
...
+            return;
+        }
+
+        let showCompileCommandsSelection: PersistentState<boolean> = new PersistentState<boolean>("CPP.showCompileCommandsSelection", true);
+        if (!showCompileCommandsSelection.Value) {
+            return;
+        }
+
+        let compileCommandStr: string = params.paths.length > 1 ? "a compile_commands.json file" : params.paths[0];
+        let folderStr: string = vscode.workspace.workspaceFolders.length > 1 ? "the " + this.Name : "this";
```
bobbrow: >vscode.workspace.workspaceFolders [](start = 32, length = 33)

should always check that vscode.workspace.workspaceFolders is defined before asking for .length

bobbrow (approved):
:shipit:

PR #2143, opened 2018-06-19
Update changelog, fwlinks, and fix an issue with abTesting.

sean-mcmanus:


sean-mcmanus:
Let me know what changelog changes you guys want. Some of the fixes we made were for bugs not in 0.17.4 so I excluded those.

bobbrow (approved):

pieandcakes:
@wardengnaw can you add the debugger fixes we have including the `--` and the main breakpoint fix?

sean-mcmanus:
@bobbrow Can you add/provide a changelog line about the configuration providers or can you add that in your pull request? Also, that pull request handles the package.json issue differently...your method seems fine, but there will be a conflict.

bobbrow:
I don't think the config provider update is going to make it into 0.17.5.  I'll publish an insiders soon after the release with it in.

PR #2157, opened 2018-06-20
Change persistent state to folder

grdowns:
Don't Ask Again only applies to the current folder

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -756,7 +756,7 @@ class DefaultClient implements Client {
             return;
         }
 
-        let showCompileCommandsSelection: PersistentFolderState<boolean> = new PersistentFolderState<boolean>("CPP.showCompileCommandsSelection", true, this.Name);
+        let showCompileCommandsSelection: PersistentFolderState<boolean> = new PersistentFolderState<boolean>("CPP.showCompileCommandsSelection", true, this.RootPath);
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -756,7 +756,7 @@ class DefaultClient implements Client {
             return;
         }
 
-        let showCompileCommandsSelection: PersistentState<boolean> = new PersistentState<boolean>("CPP.showCompileCommandsSelection", true);
+        let showCompileCommandsSelection: PersistentFolderState<boolean> = new PersistentFolderState<boolean>("CPP.showCompileCommandsSelection", true, this.Name);
```
sean-mcmanus: Isn't this.Name just "FolderName"? Shouldn't this be workspaceFolder.fsPath ? I think it's supposed to be the full path (see the other usage of PersistentFolderState in configuration.ts); otherwise, all folders with the same name would share the same setting.

bobbrow: I believe the 3rd parameter is supposed to be an absolute path.  e.g. `this.RootPath`

bobbrow (changes requested):

bobbrow (approved):

sean-mcmanus (approved):

sean-mcmanus:
I think we're skipping this for now....

PR #2166, opened 2018-06-21
Fix ${config:a.b} https://github.com/Microsoft/vscode-cpptools/issues…

### LINKED ISSUE #2165: Fails to resolve config variables containing colons
Type: LanguageService

- OS and Version:  Mac OS X 10.11.6
- VS Code Version: 1.24.1
- C/C++ Extension Version: 0.17.4

I have some configuration variables that share the same prefix and use '.' as a delimiter.  This looks to be standard practice for vscode settings.  e.g.
ml.sdk
ml.username
ml.foobar

I'm trying to include some sdk-relative include paths for IntelliSense like this in the workspace folder's settings.json:
    "ml.sdk": "/path/to/sdk",
    "C_Cpp.default.includePath": [
                "${workspaceFolder}/**",
                "${config:ml.sdk}/foo",
                "${config:ml.sdk}/bar",
                ],

IntelliSense is unable to locate any of the headers in /path/to/sdk/foo or /path/to/sdk/bar.

If I change the '.' to '_' instead, e.g. ml_sdk, IntelliSense is able to locate the headers.

I have an extension that reads and writes these settings, e.g.
vscode.workspace.getConfiguration('ml').get('sdk');

so I'm not sure I can do that without having a '.' as a settings section-identifier.

sean-mcmanus:
Fixes https://github.com/Microsoft/vscode-cpptools/issues/2165 .

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -175,7 +175,15 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
             }
             case "config": {
                 let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
-                newValue = (config) ? config.get(name).toString() : undefined;
+                try {
+                    // Look up setting directly.
+                    newValue = (config) ? config.get(name).toString() : undefined;
+                } catch (error) {
+                    // If that fails, fallback to the previous behavior.
+                    let keys: string[] = name.split('.');
+                    keys.forEach((key: string) => { config = (config) ? config.get(key) : config; });
+                    newValue = (config) ? config.toString() : undefined;
+                }
                 break;
             }
             case "workspaceFolder": {
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -175,14 +175,8 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
             }
             case "config": {
                 let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
-                try {
-                    // Look up setting directly.
-                    newValue = (config) ? config.get(name).toString() : undefined;
-                } catch (error) {
-                    // If that fails, fallback to the previous behavior.
-                    let keys: string[] = name.split('.');
-                    keys.forEach((key: string) => { config = (config) ? config.get(key) : config; });
-                    newValue = (config) ? config.toString() : undefined;
+                if (config) {
+                    newValue = config.get<string>(name);
                 }
                 break;
             }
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
Can we get this into 0.17.5?

Extension/src/common.ts
```diff
@@ -175,9 +175,7 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
             }
             case "config": {
                 let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
-                let keys: string[] = name.split('.');
-                keys.forEach((key: string) => { config = (config) ? config.get(key) : config; });
-                newValue = (config) ? config.toString() : undefined;
+                newValue = (config) ? config.get(name).toString() : undefined;
```
greazer: I'm not sure about this change. Basically you're removing a line of code that looks like it's trying to change config for a reason. However, based on my limited knowledge of this, I'm not sure what the point would be. The problem is that the code is so specific, that it seems questionable to change it so dramatically.

bobbrow (approved):

Extension/src/common.ts
```diff
@@ -175,7 +175,15 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
             }
             case "config": {
                 let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
-                newValue = (config) ? config.get(name).toString() : undefined;
+                try {
```
greazer: This seems backwards to me. The previous case first did the iteration over the key array and threw an exception (the bug we're trying to fix). So, we should try that first and if we throw, then we fall back to the direct setting.

greazer (changes requested):

Extension/src/common.ts
sean-mcmanus: It doesn't make sense to try the code that we expect to fail and have no repro which actually works. I think the VS Code API changed. I just added that try/catch in case there's some reason the single "get" might fail that we're not aware of. I'd prefer to avoid throwing an exception in the common/expected code path.

greazer: But what this means is that if there IS any behavior that we haven't been able to discover, then that behavior will break. If we do the opposite, that won't occur. It is a good point that we shouldn't regularly throw in a common code-path, so is there a way to do the same thing without relying on an exception?

sean-mcmanus: The old code no longer matches the VS Code APIs used so it doesn't make sense to fall back to it. Either it never worked or the API/types changed.

greazer (approved):

PR #2189, opened 2018-06-26
start sending installation failure telemetry again

bobbrow:
This regressed in 0.15.0.

### DIFF
--- a/Extension/src/packageManager.ts
+++ b/Extension/src/packageManager.ts
@@ -183,11 +183,10 @@ export class PackageManager {
             } catch (error) {
                 retryCount += 1;
                 lastError = error;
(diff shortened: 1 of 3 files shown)


Extension/src/packageManager.ts
```diff
@@ -183,11 +183,10 @@ export class PackageManager {
             } catch (error) {
                 retryCount += 1;
                 lastError = error;
-                if (retryCount >= MAX_RETRIES) {
+                if (retryCount > MAX_RETRIES) {
```
bobbrow, line 186: we used to do 5 retries, but currently are only doing 4 because of retryCount + 1 before this check

sean-mcmanus (approved):

PR #2213, opened 2018-06-30
Update changelog.

sean-mcmanus:


sean-mcmanus:
0.17.6 got delayed till Monday due to publishing issues, so we added some more fixes.

sean-mcmanus:
The debugger change log entries have 2 duplicates...we should remove the old ones.

bobbrow (approved):

Issue #2214, opened 2018-06-30
Edit Configuration fails to create c_cpp_properties.json after it's deleted with VS Code Insiders 1.25.0


sean-mcmanus:
Looks like they added a feature that causes the deleted file to be viewed (showing "deleted from disk") which prevents the new c_cpp_properties.json from being generated.

sean-mcmanus:
I filed bug https://github.com/Microsoft/vscode/issues/53368 .

PR #2235, opened 2018-07-06
Fix bug c_cpp_properties.json creation after being deleted.

sean-mcmanus:


sean-mcmanus:
This is for https://github.com/Microsoft/vscode-cpptools/issues/2214 , which repros with VS Code Insiders (not sure why it only repros on certain machines).

sean-mcmanus:
There's a problem -- "#" in the path fails, working on a fix...

bobbrow (approved):

PR #2241, opened 2018-07-07
Seanmcm/windows sdk version

sean-mcmanus:
I'm still reviewing/testing this some more, so you could wait a while if you want...

### DIFF
--- a/Extension/c_cpp_properties.schema.json
+++ b/Extension/c_cpp_properties.schema.json
@@ -58,9 +58,10 @@
                             "type": "string"
                         }
                     },
-                    "windowsSDKVersion": {
+                    "windowsSdkVersion": {
                         "description": "Version of the Windows SDK include path to use on Windows, e.g. '10.0.17134.0'.",
-                        "type": "string"
+                        "type": "string",
+                        "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8.1$"
                     },
                     "defines": {
                         "description": "A list of preprocessor definitions for the IntelliSense engine to use while parsing files. Optionally, use = to set a value, e.g. VERSION=1.",
--- a/Extension/c_cpp_properties.schema.json
+++ b/Extension/c_cpp_properties.schema.json
@@ -61,7 +61,7 @@
                     "windowsSdkVersion": {
                         "description": "Version of the Windows SDK include path to use on Windows, e.g. '10.0.17134.0'.",
                         "type": "string",
-                        "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8.1$"
+                        "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8\\.1$"
                     },
                     "defines": {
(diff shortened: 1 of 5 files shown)


Extension/package.json
```diff
@@ -286,6 +286,15 @@
           "description": "The value to use in a configuration if \"macFrameworkPath\" is not specified, or the values to insert if \"${default}\" is present in \"macFrameworkPath\".",
           "scope": "resource"
         },
+        "C_Cpp.default.windowsSDKVersion": {
```
bobbrow: `SDK` should be `Sdk` per convention

bobbrow (changes requested):

Extension/c_cpp_properties.schema.json
```diff
@@ -58,6 +58,10 @@
                             "type": "string"
                         }
                     },
+                    "windowsSDKVersion": {
+                        "description": "Version of the Windows SDK include path to use on Windows, e.g. '10.0.17134.0'.",
+                        "type": "string"
```
bobbrow: should probably add `patternProperties` ([reference](https://spacetelescope.github.io/understanding-json-schema/reference/object.html)) to this so we can get squiggles for invalid versions.  (see the `env` property).

The regexp is this, I think: `^\d{2}\.\d{1}\.\d{5}\.\d{1}$`

Extension/package.json
sean-mcmanus: Ah, okay -- that's easier to read :)

Extension/c_cpp_properties.schema.json
sean-mcmanus: Ok.

Extension/c_cpp_properties.schema.json
```diff
@@ -58,6 +58,11 @@
                             "type": "string"
                         }
                     },
+                    "windowsSdkVersion": {
+                        "description": "Version of the Windows SDK include path to use on Windows, e.g. '10.0.17134.0'.",
+                        "type": "string",
+                        "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8.1$"
```
bobbrow: The 8.1 part should have a backslash before the dot

Extension/package.json
```diff
@@ -286,6 +286,15 @@
...
           "scope": "resource"
         },
+        "C_Cpp.default.windowsSdkVersion": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": null,
+          "description": "Version of the Windows SDK include path to use on Windows, e.g. '10.0.17134.0'.",
+          "scope": "resource"
```
bobbrow: Should the pattern go in here too?

bobbrow (approved):

Extension/c_cpp_properties.schema.json
sean-mcmanus: Ooops...

Extension/package.json
sean-mcmanus: Hmm...it doesn't seem to be required...I could try adding it...oh, this is the default one...I didn't test that yet...

bobbrow (approved):

PR #2261, opened 2018-07-12
Update cpp.hint to use the injected defaults.

sean-mcmanus:


bobbrow (changes requested):
I'm not familiar enough with the changes.  is it possible to still add definitions to this file?

sean-mcmanus:
Yeah, you can add new definitions to cpp.hint file after the pragma -- the pragma just uses the defaults that are built into our binary.

bobbrow (approved):

PR #2262, opened 2018-07-12
Remove main snippet and replace with more friendly imain and vmain.

greazer:


bobbrow (approved):

sean-mcmanus:
@greazer Is this going to be checked in?

Issue #2212, opened 2018-06-29
Allow users to set the inactiveRegion color


bobbrow:
From PR #2061

Add an `inactiveRegionColor` setting that allows users to set the region color.  It should work alongside the opacity setting.  If not set, we leave the color alone.  If set, we change the color of the inactive regions.  VSCode accepts ThemeColors for the decoration, so those could potentially be allowed too.

bobbrow:
Found a duplicate issue: #1620.  Need to close both when this goes in.

PR #2308, opened 2018-07-22
adds settings for inactive region font & bg color

john-patterson:
This adds the inactiveRegionFontColor and inactiveRegionBackgroundColor
settings. If a preprocessor block is found to be inactive, it will be
colored by these fields if they are set. They accept hexadecimal font
strings or valid theme colors as their values.These fields also
interactive with the dimInactiveRegions setting.

This was requested in issue
https://github.com/Microsoft/vscode-cpptools/issues/2212.

I tried launching the tests, but they crashed for me with error `Error: Cannot find module 'C:\Users\John\Projects\vscode-cpptools\Extension\out\test'` but I don't think this is related to my change.

I also feel like `dimInactiveRegions` should now be changes to `colorizeInactiveRegions` as this verb change captures both changes in opacity and color, but I wanted to get input before I made a user-facing alteration.

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -1,5 +1,8 @@
 # C/C++ for Visual Studio Code Change Log
 
+## Version 0.17.8: August 16, 2018
+* Add `inactiveRegionForegroundColor` and `inactiveRegionBackgroundColor` to settings. [#2212](https://github.com/Microsoft/vscode-cpptools/issues/2212)
+
 ## Version 0.17.7: July 22, 2018
 * Fix `Go to Definition` for code scoped with an aliased namespace. [#387](https://github.com/Microsoft/vscode-cpptools/issues/387)
 * Fix incorrect IntelliSense errors with template template-arguments. [#1014](https://github.com/Microsoft/vscode-cpptools/issues/1014)
@@ -24,7 +27,6 @@
 * Add a setting to silence configuration provider warnings. [#2292](https://github.com/Microsoft/vscode-cpptools/issues/2292)
 * Fix for debugging async Visual C++ causing debugger to hang.
 * Fix `main` snippet.
-* Add `inactiveRegionFontColor` and `inactiveRegionBackgroundColor` to settings. [#2212](https://github.com/Microsoft/vscode-cpptools/issues/2212)
 
 ## Version 0.17.6: July 2, 2018
 * Fix the database icon getting stuck with recursive includes. [#2104](https://github.com/Microsoft/vscode-cpptools/issues/2104)
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -1,6 +1,6 @@
 # C/C++ for Visual Studio Code Change Log
 
-## Version 0.17.7: July 19, 2018
+## Version 0.17.7: July 22, 2018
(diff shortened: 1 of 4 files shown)


Extension/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # C/C++ for Visual Studio Code Change Log
 
-## Version 0.17.7: July 19, 2018
+## Version 0.17.7: July 22, 2018
```
sean-mcmanus: 0.17.7 already shipped -- can you add a 0.17.8, with some date such as August 16?

Extension/package.json
```diff
@@ -154,6 +154,18 @@
           "minimum": 0.1,
           "maximum": 1
         },
+        "C_Cpp.inactiveRegionFontColor": {
+          "type": "string",
+          "default": null,
+          "description": "Controls the font coloring of inactive preprocessor blocks. Input is in the form a hexidecimal color code or a valid Theme Color. If not set, this defaults to the syntax coloring scheme of the editor. This setting only applies when inactive region dimming is enabled.",
```
sean-mcmanus: Should be "hexadecimal".

sean-mcmanus (changes requested):

sean-mcmanus:
I think the tests are run as part of the GitHub Travis CI so you don't have to run it locally.

Extension/package.json
```diff
@@ -154,6 +154,18 @@
           "minimum": 0.1,
           "maximum": 1
         },
+        "C_Cpp.inactiveRegionFontColor": {
+          "type": "string",
```
sean-mcmanus: The type should accept "null" as valid, since that's the default, e.g.
[
            "string",
            "null"
],

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -154,6 +154,18 @@
           "minimum": 0.1,
           "maximum": 1
         },
+        "C_Cpp.inactiveRegionFontColor": {
```
sean-mcmanus: Should this be "inactiveRegionForegroundColor"? VS Code seems to use the foreground instead of font term: https://code.visualstudio.com/docs/getstarted/theme-color-reference .

bobbrow:
We do plan to rename a bunch of settings in the future.  I'll add this one to the list.  I'd rather do it all at once.

Extension/package.json
```diff
@@ -154,6 +154,21 @@
...
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": null,
+          "description": "Controls the font coloring of inactive preprocessor blocks. Input is in the form a hexadecimal color code or a valid Theme Color. If not set, this defaults to the syntax coloring scheme of the editor. This setting only applies when inactive region dimming is enabled.",
+          "scope": "resource"
+        },
+        "C_Cpp.inactiveRegionBackgroundColor": {
+          "type": "string",
```
bobbrow: this property should be string | null as well.

bobbrow (changes requested):

bobbrow (approved):

sean-mcmanus (approved):
Thanks a lot 👍

yachnyymaxim:
Is this an intended behavior: if you select a piece of text within an inactive code block your selection is not visible? 
Here is the example below, it has "text2" string selected on the line 4:
image

PR #2315, opened 2018-07-23
Add gcc-x64 intelliSenseMode.

sean-mcmanus:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -300,15 +300,17 @@ export class CppProperties {
...
+            return "gcc-x64";
+        } else if (name === "Mac") {
             return "clang-x64";
         } else if (name === "Win32") {
             return "msvc-x64";
         } else if (process.platform === 'win32') {
             // Custom configs default to the OS's preference.
             return "msvc-x64";
         } else {
-            return "clang-x64";
```
bobbrow: we should probably do a process.platform check for 'darwin' and pick clang for that.  'gcc' remains the catch-all at the end.

bobbrow (changes requested):

bobbrow (approved):

Issue #2057, opened 2018-05-29
Multi-pass environment variable resolution


bobbrow:
@Gouthamm4G
> Copied from #957 

The initialized variables don't work inside `env` when it is referenced right next to the definition. For example,
```
{
    "env": {
        "envRoot": "apps/tool/buildenv",
        "arm6.include": "${envRoot}/arm6/include",
    },
    "configurations": [
        {
            "name": "Custom",
            "includePath": [
                "${arm6.include}"
            ],
            "browse": {
                "limitSymbolsToIncludedHeaders": false,
                "path": [
                     "${arm6.include}"
                ]
            },
            …
        }
    ],
    "version": 4
}
```
This doesn't seem to work. That would be really useful

john-patterson:
I'll take a crack at this.

sean-mcmanus:
@john-patterson Cool, thanks a lot 👍

PR #2322, opened 2018-07-24
Add multi-pass environment variable resolution

john-patterson:
Per issue [#2057](https://github.com/Microsoft/vscode-cpptools/issues/2057).

This allows a user to nest environment tags, or define tags based on the definition of other environment variables. If the result of a resolution still contains a `${...}` tag, we will continue to scan the input until we reach a steady state where the variable expansion results in the same input.

This should allow the following forms which were previously not possible in the environment definition:

* Dependent variable definition
```
"env": {
    "envRoot": "apps/tool/builldenv",
    "arm6.include": "${envRoot}/arm6/include"
},
...
"some_config": "${arm6.include}
```

* Nested variable definition of any degree
```
"env": {
    "root": "apps/tool",
    "armVersion": "${config:armVersion}",
    "arm5.include": "${root}/legacy/arm/include",
    "arm6.include": "${root}/arm/include",
    "armTestPath": "${arm${armVersion}.include}/test"
}
```

* Additionally, this fixes an open bug that was not reported. If a variable contained "env", "config", or "workspaceFolder" it would not be parsed correctly. If you used a variable ${envRoot} it would match "envR" as the type and "oot" as the variable. This has been resolved.

### DIFF
--- a/Extension/.vscode/launch.json
+++ b/Extension/.vscode/launch.json
@@ -24,7 +24,7 @@
 			"runtimeExecutable": "${execPath}",
 			"args": [
 				"--extensionDevelopmentPath=${workspaceFolder}",
-				"--extensionTestsPath=${workspaceFolder}/out/test"
+				"--extensionTestsPath=${workspaceFolder}/out/test/unitTests"
 			],
 			"stopOnEntry": false,
 			"sourceMaps": true,
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -169,6 +169,104 @@ export function showReleaseNotes(): void {
     vscode.commands.executeCommand('vscode.previewHtml', vscode.Uri.file(getExtensionFilePath("ReleaseNotes.html")), vscode.ViewColumn.One, "C/C++ Extension Release Notes");
 }
 
+export function resolveInnerVariable(originalInput: string, wholeVariable: string, innerVariable: string, additionalEnvironment: {[key: string]: string | string[]}): string {
+    let varType: string;
+    let name: string = innerVariable.replace(/(env|config|workspaceFolder)[\.|:](.*)/, (ignored: string, type: string, rest: string) => {
+        varType = type;
+        return rest;
+    });
+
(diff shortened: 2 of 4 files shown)


Extension/.vscode/launch.json
```diff
@@ -24,7 +24,7 @@
 			"runtimeExecutable": "${execPath}",
 			"args": [
 				"--extensionDevelopmentPath=${workspaceFolder}",
-				"--extensionTestsPath=${workspaceFolder}/out/test"
+				"--extensionTestsPath=${workspaceFolder}/out/test/unitTests"
```
sean-mcmanus, line 27: Oh, that's interesting...I guess we don't Launch Tests very much :)

sean-mcmanus: Looks good so far. I'm not done reviewing yet...

john-patterson:
@bobbrow Raised a good point in that this a lot of added complexity for nesting variable, and the case that this is a useful feature for customers is a nebulous prospect at best. With that in might, I pushed a change which simplifies this feature at the expense of nested variables.

Extension/src/common.ts
```diff
@@ -178,54 +178,59 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
...
-                if (typeof v === "string") {
-                    newValue = v;
-                } else if (input === match && v instanceof Array) {
-                    newValue = v.join(";");
-                }
-                if (!newValue) {
-                    newValue = process.env[name];
-                }
-                break;
+    let regexp: () => RegExp = () => /\$\{((env|config|workspaceFolder)(\.|:))?(.*?)\}/g;
```
sean-mcmanus, line 181: Good catch 👍

sean-mcmanus (approved):
Seems good to me. Bob will probably want to finish reviewing as well.

sean-mcmanus:
Do you know what's up with the Travis CI failure?

bobbrow (approved):

john-patterson:
These tests should pass in the future. I added the fix for Travis CI.

PR #2332, opened 2018-07-26
Allow telemetry to detect failed downloads

WardenGnaw:


bobbrow (approved):
This should work.  The "false" data we have appears to only be logged when an exception is thrown during the installation process.  Since the download code doesn't throw, we weren't getting the telemetry in the place I expected.  Thanks for fixing this.

PR #2340, opened 2018-07-30
Using sh instead of bash

### LINKED ISSUE #569: getRemoteOSAndProcesses uses bash
Not all Linuxes have `bash`, but practically all of them have `sh`. From what I can see, there are no features from `bash` that are not in `sh` being used. Would it be possible to change the `attachToProcess.js` from

```javascript
        const command = `bash -c 'uname && if [ $(uname) == "Linux" ] ; then ${nativeAttach_1.PsProcessParser.psLinuxCommand} ; elif [ $(uname) == "Darwin" ] ; ` +
            `then ${nativeAttach_1.PsProcessParser.psDarwinCommand}; fi'`;
```

to

```javascript
        const command = `sh -c 'uname && if [ $(uname) == "Linux" ] ; then ${nativeAttach_1.PsProcessParser.psLinuxCommand} ; elif [ $(uname) == "Darwin" ] ; ` +
            `then ${nativeAttach_1.PsProcessParser.psDarwinCommand}; fi'`;
```

This comes up more often when using the `pipeTransport` to debug on other operating systems.

andyneff:
Fixes #569

### DIFF
--- a/Extension/src/Debugger/attachToProcess.ts
+++ b/Extension/src/Debugger/attachToProcess.ts
@@ -120,7 +120,7 @@ export class RemoteAttachPicker {
 
     private getRemoteOSAndProcesses(pipeCmd: string): Promise<AttachItem[]> {
         // Commands to get OS and processes
-        const command: string = `bash -c 'uname && if [ $(uname) == "Linux" ] ; then ${PsProcessParser.psLinuxCommand} ; elif [ $(uname) == "Darwin" ] ; ` +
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):
Seems good to me, but the debugger guys should approve as well.

WardenGnaw (changes requested):
Blocking for testing.

WardenGnaw (approved):

timedroid:
This broke debugging for me: #2473

PR #2343, opened 2018-07-30
Update FAQs.md to include debugging

pieandcakes:
@bobbrow @sean-mcmanus @ronglums @WardenGnaw @grdowns 

Let me know if you have feedback.

@leslierichardson95

### DIFF
--- a/Documentation/FAQs.md
+++ b/Documentation/FAQs.md
@@ -3,7 +3,8 @@
 ## Table of Contents
 * Setup: [Debugging Setup](#debugging-setup)
 * Debugger: [Why is debugging not working?](#why-is-debugging-not-working)
-* Build: [How to enable debug symbols?](#how-to-enable-debug-symbols)
+* Build: [How to enable debug symbols](#how-to-enable-debug-symbols)
+* Logging: [How to enable logging](#how-to-enable-logging)
 
 ## Debugging Setup
 The debugger needs to be configured to know which executable and debugger to use:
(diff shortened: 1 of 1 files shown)


Documentation/FAQs.md
```diff
@@ -43,3 +44,39 @@ When in doubt, please check your compiler's documentation for the options necess
 
 * #### cl.exe
   Symbols are located in the `*.pdb` file.
+
+## How to enable logging
+
+Enabling logging will show communication information between VS Code and our extension and between our extension and the debugger.
```
leslierichardson95, line 50: This sentence seems a little vague.  Maybe be more specific on what you mean when you say "our extension" and the "debugger" just in case people don't know what's being referred to?

pieandcakes: Leslie and I talked offline. This is documentation in our repo so "this extension" should be clear but i'll change it to say "cpptools extension"

Documentation/FAQs.md
```diff
@@ -43,3 +44,39 @@ When in doubt, please check your compiler's documentation for the options necess
...
+
+### Logging for `MI` debuggers
+
+The logging block with its defaults is as follows:
```
bobbrow, line 54: I think it might be better to show code snippets with the values you want people to copy.  If you redirect people here, they are likely to want to do that rather than see what the defaults are.

sean-mcmanus (approved):

Documentation/FAQs.md
pieandcakes: There are different combinations depending on what the user wants to see.

bobbrow: Sure, but the topic is called "how to enable logging" and I suspect not everyone is going to read deeply.  You don't have to change it.  It's just my 2 cents.

I don't imagine most people will know what an "MI" debugger is either.

sean-mcmanus:
What's the status of this?

Issue #2344, opened 2018-08-01
C_Cpp.clang_format_path won't accept environment variable


benlindsay:
Type: LanguageService

**Describe the bug**
- OS and Version: Mac OSX 10.12.6
- VS Code Version: 1.25.1
- C/C++ Extension Version: 0.17.7
- A clear and concise description of what the bug is.

I have my clang-format in a custom path, specifically $HOME/.local/bin/clang-format. I have this on two different computers, both of which I use to edit some code via sshfs. I want to be able to specify something like this in my user settings:

```
{
    "C_Cpp.clang_format_path": "${env:HOME}/.local/bin/clang-format",
}
```

so that I get the same formatting functionality on both computers. Formatting works fine if I put in the absolute path with no variables, but all my attempts to use a variable have disabled the formatting functionality. I've tried `"${env:HOME}/.local/bin/clang-format"`,  `"${env.HOME}/.local/bin/clang-format"`, `"~/.local/bin/clang-format"`, and `"$HOME/.local/bin/clang-format"`, all with no success. After any of these attempts if I replace the path with the absolute path it immediately works again.

Is there a way to use a variable here, or some other way to fix my problem?

Thanks in advance!

bobbrow:
There is no workaround right now (besides maybe editing the extension JavaScript yourself), but this is easy enough to fix.  We should have it for the next update.

sean-mcmanus:
The fix I added will pick up the env in the clang_format_path when the extension starts, but any changes to the clang_format_path that has an env won't be picked up until a Reload Window is done...which seems okay to me (it'll fall back to the previous clang_format_path before the settings change).

benlindsay:
@sean-mcmanus Awesome, thanks so much! This will be a great feature to have :)

smhc:
Would this also fix the ability to use '${workspaceFolder}' in compilerPath?

I would like to use a 'dummy/proxy' compiler to return the include/defines to enable me to work cross-platform and need access to workspaceFolder in the path.

sean-mcmanus:
@smhc Yes -- ${workspaceFolder} should work as well with our next release.

sean-mcmanus:
0.18.0 should have the fix (it doesn't need a Reload Window either, as mentioned previously).

benlindsay:
@sean-mcmanus works great! Thanks!

PR #2380, opened 2018-08-09
Enable environment variables with clang_format_path.

sean-mcmanus:
For https://github.com/Microsoft/vscode-cpptools/issues/2344 .

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -71,7 +71,7 @@
             "null"
           ],
           "default": null,
-          "description": "The full path of the clang-format executable.",
+          "description": "The full path of the clang-format executable. Changing this setting requires a Reload Window if it contains environment variables.",
           "scope": "resource"
         },
         "C_Cpp.clang_format_style": {
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -71,7 +71,7 @@
             "null"
           ],
           "default": null,
-          "description": "The full path of the clang-format executable. Changing this setting requires a Reload Window if it contains environment variables.",
+          "description": "The full path of the clang-format executable.",
           "scope": "resource"
         },
         "C_Cpp.clang_format_style": {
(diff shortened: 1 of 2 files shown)


Extension/package.json
```diff
@@ -71,7 +71,7 @@
             "null"
           ],
           "default": null,
-          "description": "The full path of the clang-format executable.",
+          "description": "The full path of the clang-format executable. Changing this setting requires a Reload Window if it contains environment variables.",
```
bobbrow: did it not work to intercept the didChangeConfigurations message in protocallFilter.ts?

bobbrow (changes requested):

Extension/package.json
sean-mcmanus: I didn't investigate that. I can.

sean-mcmanus: It doesn't appear possible to "replace" a setting in that filter, but I can change it to ignore that particular setting change and send our own message with the correct setting.

bobbrow: Yes, I think that would be better.  Thanks!

Extension/src/LanguageServer/client.ts
```diff
@@ -402,6 +403,10 @@ class DefaultClient implements Client {
             if (changedSettings["commentContinuationPatterns"]) {
                 updateLanguageConfigurations();
             }
+            if (changedSettings["clang_format_path"]) {
+                let settings: CppSettings = new CppSettings(this.RootUri);
+                this.languageClient.sendNotification(UpdateClangFormatPathNotification, util.resolveVariables(settings.clangFormatPath, null));
```
bobbrow, line 408: Can you just use `changedSettings["clang_format_path"]` without creating a new `CppSettings` object?

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: No, the value is "...".

bobbrow: 😆 ok

bobbrow (approved):

Issue #2240, opened 2018-07-06
“Configure your IntelliSense settings to help find missing headers.” should not show if there are CustomConfigurationProvider extensions


sherryyshi:
Type: LanguageService
**----- Input information below -----**

This message directs users to edit their c_cpp_properties.json file, which might conflict with CustomConfigurationProvider behaviors.

sean-mcmanus:
Sounds like this will be fixed via https://github.com/Microsoft/vscode-cpptools/issues/2350 .

bobbrow:
Yes a combination of #2350 and #2324 (fixed) will address this.

sean-mcmanus:
@bobbrow This is fixed now, right?

bobbrow:
Yes.  The providers will need to update to the v2 API to get the fixed behavior.

MSeutin:
I have a mac and I have this erro on my vscode.  the thing is I am a newbie so I do not understand API, or #2350.  can someone give me simple steps to fix this?

PR #2385, opened 2018-08-10
Stop showing the fallback message for custom config providers.

sean-mcmanus:
For https://github.com/Microsoft/vscode-cpptools/issues/2240

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -809,7 +809,7 @@ class DefaultClient implements Client {
             util.setProgress(util.getProgressParseRootSuccess());
         } else if (message.endsWith("No Squiggles")) {
             util.setIntelliSenseProgress(util.getProgressIntelliSenseNoSquiggles());
-        } else if (message.endsWith("IntelliSense Fallback")) {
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -809,7 +809,7 @@ class DefaultClient implements Client {
             util.setProgress(util.getProgressParseRootSuccess());
         } else if (message.endsWith("No Squiggles")) {
             util.setIntelliSenseProgress(util.getProgressIntelliSenseNoSquiggles());
-        } else if (message.endsWith("IntelliSense Fallback")) {
+        } else if (!this.configurationProvider && message.endsWith("IntelliSense Fallback")) {
```
bobbrow, line 812: It's not quite this simple.  If the provider is absent or misbehaving, cpptools will be using c_cpp_properties.json to configure IntelliSense, so this message might still be needed.

The issue that @sherryyshi and her team actually want to solve is the first launch scenario (e.g. before the config provider is even added to c_cpp_properties.json), and this won't solve that.

bobbrow (changes requested):

Issue #785, opened 2017-05-30
Add option to turn off language server features


bobbrow:
Some users of the extension may only wish to use the debugger features and don't want to be bothered with any error notifications from the language server.

moving request out of closed issue #755

vscpp:
@bobbrow, actually, I really want to disable debugger feature of cpp extension. VS code should be a code editor rather than debugger. We have more professional IDE or debug tool. I wonder that why a code editor adding a debugger function. Till now, the intelliSense is not full functionality.

bobbrow:
The debugger feature doesn't get in the way of the editor for you, does it?  There's really nothing to disable if you don't try to use it.  The language server features, on the other hand, will consume CPU and memory resources by simply opening a file.

pieandcakes:
@vscpp Can you explain what you mean by:
> I really want to disable debugger feature of cpp extension.

What specifically do you want disabled? Like @bobbrow mentioned, unless you configure for debugging, the debug extension doesn't do any work at all.

vscpp:
@bobbrow, @pieandcakes, I using VS code for C coding, searching and reading, and I debug my code by other tools. It means vs code is a source editor like Source Insight. I hope CPP extension have more function like Source Insight.
But currently, seems MSFT take more effort on Debugger feature, not basic code editor improving. If you use VS code for C coding that you will find there is a big gap with Source Insight.
e.g. for a data struct, if you type . or ->  , there is no correct struct member display.

bobbrow:
@vscpp, semantic auto-complete is something I am working on currently (tracked by issue #13).  We are a small team and appreciate your patience as we try to simultaneously address the incoming issues reported here.

zajo:
@bobbrow Thank you for creating this issue -- in the meantime, what are my options for preventing intellisense from spawning 10 threads and using 100% of the CPU power for 15 minutes every time I open a C++ project in vscode (the end result for me being a whole lots of bogus red squiggles) without disabling the debugger as well?

bobbrow:
Until this issue is addressed the way to consume the least amount of power is to use the following settings:

### In settings.json:
```
"C_Cpp.intelliSenseEngine": "Tag Parser"
```

### In c_cpp_properties.json
```
        {
            "name": "whatever",
            "includePath": [],
            "defines": [],
            "browse": {
                "limitSymbolsToIncludedHeaders": true,
                "databaseFilename": "",
                "path": []
            },
            "compilerPath": ""
        }
```
[...]

PR #2392, opened 2018-08-13
Add a disableLanguageService setting.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/785 .

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -104,7 +104,8 @@
           "type": "string",
           "enum": [
             "Default",
-            "Tag Parser"
+            "Tag Parser",
+            "Disabled"
           ],
           "default": "Default",
           "description": "Controls the IntelliSense provider. \"Tag Parser\" provides \"fuzzy\" results that are not context-aware. \"Default\" provides context-aware results and is in preview mode - member list, hover tooltips, and error squiggles are currently implemented. Features not yet implemented in the new default engine will use the tag parser engine instead.",
@@ -454,12 +455,6 @@
           },
           "default": null,
           "description": "The value to use for the system include path. If set, it overrides the system include path acquired via \"compilerPath\" and \"compileCommands\" settings.",
-          "scope": "resource"
-        },
-        "C_Cpp.disableLanguageService": {
-          "type": "boolean",
-          "default": false,
-          "description": "If set, all the C/C++ language service features are disabled (i.e. only the debugger is enabled). Requires a Reload Window to take effect.",
           "scope": "resource"
         }
       }
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -455,6 +455,12 @@
           "default": null,
           "description": "The value to use for the system include path. If set, it overrides the system include path acquired via \"compilerPath\" and \"compileCommands\" settings.",
           "scope": "resource"
+        },
+        "C_Cpp.disableLanguageService": {
+          "type": "boolean",
+          "default": false,
+          "description": "If set, all the C/C++ language service features are disabled (i.e. only the debugger is enabled). Requires a Reload Window to take effect.",
+          "scope": "resource"
         }
       }
     },
--- a/Extension/src/commands.ts
+++ b/Extension/src/commands.ts
@@ -11,7 +11,7 @@ class TemporaryCommandRegistrar {
     // Used to save/re-execute commands used before the extension has activated (e.g. delayed by dependency downloading).
     private delayedCommandsToExecute: Set<string>;
     private tempCommands: vscode.Disposable[]; // Need to save this to unregister/dispose the temporary commands.
-    private isDisabled: boolean;
+    private isLanguageServerDisabled: boolean;
 
     private commandsToRegister: string[] = [
         "C_Cpp.ConfigurationEdit",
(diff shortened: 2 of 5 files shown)


Extension/package.json
```diff
@@ -455,6 +455,12 @@
           "default": null,
           "description": "The value to use for the system include path. If set, it overrides the system include path acquired via \"compilerPath\" and \"compileCommands\" settings.",
           "scope": "resource"
+        },
+        "C_Cpp.disableLanguageService": {
```
bobbrow: I was planning to expose this differently, via `C_Cpp.intelliSenseEngine` (until we refactor settings and have `intelliSense.enabled`).

Extension/src/main.ts
```diff
@@ -279,6 +284,11 @@ async function postInstall(info: PlatformInformation): Promise<void> {
 }
 
 async function finalizeExtensionActivation(): Promise<void> {
+    if (vscode.workspace.getConfiguration("C_Cpp", null).get<boolean>("disableLanguageService")) {
```
bobbrow: It seems like this feature should be a little more involved than just this.  What happens to all the commands that are registered in package.json?  Are we able to remove the context menu items?

bobbrow (changes requested):

Extension/package.json
sean-mcmanus: You mean a "Disabled" setting for intelliSenseEngine? I could do that.

Extension/src/main.ts
sean-mcmanus: Yeah, all the package.json and context menu stuff remains activated, but they don't do anything. I was thinking that didn't matter, and/or wait till people complain about that. I don't think there's a way to remove the package.json stuff without modifying the file and doing a Reload Window and/or splitting the extension into multiple ones and disabling the separate sub-extension.

Extension/src/main.ts
```diff
@@ -279,6 +284,12 @@ async function postInstall(info: PlatformInformation): Promise<void> {
 }
 
 async function finalizeExtensionActivation(): Promise<void> {
+    if (vscode.workspace.getConfiguration("C_Cpp", null).get<string>("intelliSenseEngine") === "Disabled") {
```
bobbrow, line 290: There doesn't seem to be a way to re-enable the language server after it's been disabled.  It also appears that disabling the language server after it has already been loaded will have no effect.

Extension/src/commands.ts
```diff
@@ -11,6 +11,7 @@ class TemporaryCommandRegistrar {
     // Used to save/re-execute commands used before the extension has activated (e.g. delayed by dependency downloading).
     private delayedCommandsToExecute: Set<string>;
     private tempCommands: vscode.Disposable[]; // Need to save this to unregister/dispose the temporary commands.
+    private isDisabled: boolean;
```
bobbrow: Since this is just about disabling the language server, I think this variable name should reflect that.

Extension/src/commands.ts
```diff
@@ -44,10 +45,18 @@ class TemporaryCommandRegistrar {
 
     public registerTempCommand(command: string): void {
         this.tempCommands.push(vscode.commands.registerCommand(command, () => {
+            if (this.isDisabled) {
+                vscode.window.showInformationMessage("The C/C++ command is disabled due to C_Cpp.intelliSenseEngine set to Disabled.");
```
bobbrow: "This command is disabled when \"C_Cpp.intelliSenseEngine\" is set to \"Disabled\""

bobbrow (changes requested):

Extension/src/main.ts
sean-mcmanus: Yeah, that was "by design" :)...my assumption was that users who wanted to use this setting wouldn't mind having to Reload Window for it to work.

Extension/src/commands.ts
sean-mcmanus: Sure.

Extension/src/main.ts
sean-mcmanus: I added the reload prompt.

Extension/src/main.ts
```diff
@@ -279,6 +287,25 @@ async function postInstall(info: PlatformInformation): Promise<void> {
...
+        languageServiceDisabled = true;
+        getTemporaryCommandRegistrarInstance().disableLanguageServer();
+        disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
+            if (!reloadMessageShown && vscode.workspace.getConfiguration("C_Cpp", null).get<string>("intelliSenseEngine") !== "Disabled") {
+                reloadMessageShown = true;
+                util.promptForReloadWindowDueToSettingsChange();
+            }
+        }));
+        Telemetry.logLanguageServerEvent("intelliSenseEngine disabled");
```
bobbrow: we don't have any events with spaces in them that I am aware of and the normal telemetry should cover this already.

Extension/src/main.ts
```diff
@@ -279,6 +287,25 @@ async function postInstall(info: PlatformInformation): Promise<void> {
...
+        return;
+    } else {
+        disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
```
bobbrow: since the "true" case does an early return, you don't need to put this part in an "else" block.

bobbrow (changes requested):

Extension/src/main.ts
sean-mcmanus: Ah, yeah.

bobbrow (approved):

PR #2412, opened 2018-08-16
Update package-lock and fix new linter errors.

sean-mcmanus:


sean-mcmanus:
Does the package-lock look correct to you or did I do something incorrect? It appears to have fixed the url-parse vulnerability. I'll add a task to update our gulp dev dependencies.

bobbrow:
That doesn't look right.  I see no edits to package.json.  I remember your clone having trouble with this...

sean-mcmanus:
I deleted my node_modules and tried again, but I get the same results -- I think it's because it's a level 2 dependency, so it doesn't require a modification of our package.json dependencies.

bobbrow (approved):

PR #2455, opened 2018-08-27
Implement v2 API

bobbrow:
Mostly complete implementation of v2 API.
* need to settle on behavior of extension between `register` and `ready` events.


sean-mcmanus (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/customProviders.ts
```diff
@@ -1,185 +1,208 @@
...
+    private getId(provider: string|CustomConfigurationProvider): string {
+        if (typeof provider === "string") {
+            return provider;
+        } else if (provider.extensionId) {
+            return provider.extensionId;
+        } else if (provider.name) {
+            return provider.name;
+        } else {
+            console.error(`invalid provider: ${provider}`);
+            return "";
```
grdowns: Would it be considered better TypeScript practice to `return null;` instead of `return "";`?

grdowns: Noticing that this is a part of the original code, although GitHub is showing entire file diffs. Is there a way to more precisely isolate these diffs?

sean-mcmanus: @grdowns There's an option to "Hide whitespace changes" which fixes the diff problems in this case.

bobbrow: I'm not sure what happened.  I think I had the vscode setting to trim extra whitespace enabled at some point.  I wonder if it changed my line endings...

grdowns: @sean-mcmanus Yep, that does the trick. Since this is the case then yeah @bobbrow it sounds like the line endings were converted

bobbrow: Yeah, it looks like the line endings were \n for those files and now they are \r\n.  I had thought that git would normalize everything to \n on the server, but we might need to set that attribute manually in the .gitattributes file.  But we should do that in a separate change.  I'll see if I can revert these files back to \n to reduce the diff noise.

bobbrow: It turned out that the line endings thing happened as a result of me installing a private fork of git to support the Windows codebase.  I have since fixed the settings problem so this doesn't happen again.

bobbrow:
I'm going to close this for now.  I want to finish up the language server side and will iterate on the TypeScript stuff as a result.

PR #2465, opened 2018-08-29
Add updateChannel setting

grdowns:
Preliminary PR for updateChannel setting. When turned to "Insiders", the extension will periodically check for new releases. If one is found, the new build of the extension will automatically be installed


Extension/package.json
```diff
@@ -456,6 +456,16 @@
...
           "scope": "resource"
+        },
+        "C_Cpp.updateChannel": {
+          "type": "string",
+          "enum": [
+            "Default",
+            "Insiders"
+          ],
+          "default": "Default",
+          "description": "TODO",
```
sean-mcmanus: Where you going to fill this out now?

Extension/src/LanguageServer/extension.ts
```diff
@@ -25,9 +31,11 @@ let ui: UI;
 let disposables: vscode.Disposable[] = [];
 let languageConfigurations: vscode.Disposable[] = [];
 let intervalTimer: NodeJS.Timer;
+let insiderUpdateTimer: NodeJS.Timer;
 let realActivationOccurred: boolean = false;
 let tempCommands: vscode.Disposable[] = [];
 let activatedPreviously: PersistentWorkspaceState<boolean>;
+const insiderUpdateTimerInterval: number = 1000 * 15;
```
sean-mcmanus: Why do you need a special timer? I thought we were going to piggyback on the cpptools.json -- i.e. we would just check for a new version whenever we check that (i.e. on startup or every 30 minutes or whatever).

WardenGnaw:
@grdowns There seems to be a couple of todo's, are those going to be fixed later?

grdowns:
Yeah, this review is more preliminary so others can offer feedback. It also allows @bobbrow to get a better idea of my status. I'll be finished with it once the TODO's are gone :)

Extension/package.json
```diff
@@ -456,6 +456,16 @@
...
           "description": "The value to use for the system include path. If set, it overrides the system include path acquired via \"compilerPath\" and \"compileCommands\" settings.",
           "scope": "resource"
+        },
+        "C_Cpp.updateChannel": {
+          "type": "string",
+          "enum": [
+            "Default",
+            "Insiders"
+          ],
+          "default": "Default",
```
sean-mcmanus, line 466: Can you add code to set the default to "Insiders" if they're using VS Code Insiders or an insiders vsix? There is related code in finalizeExtensionActivation().

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
grdowns: We should only check for an update when the user has the `updateChannel` setting set to `Insiders`; I suppose we could instead query what the setting is turned to once the `intervalTimer` goes off. In fact, that might even be the correct thing to do. What do you think?

Extension/package.json
grdowns: Absolutely! @bobbrow and I briefly discussed that and it should definitely be a part of the change.

Extension/src/LanguageServer/extension.ts
bobbrow: I think piggybacking on the abTesting update timer is fine, but the setInterval call will need to be factored out of the ABTestSettings class.  Whichever way you go, 15 seconds is too frequent to check (though I think you've only set it this way for testing purposes).

Extension/src/LanguageServer/extension.ts
```diff
@@ -141,7 +150,22 @@ export function updateLanguageConfigurations(): void {
  *********************************************/
 
 function onDidChangeSettings(): void {
+    // Check whether the user changed updateChannel to "Insiders", else return
+    let tracker: SettingsTracker = clients.ActiveClient.SettingsTracker;
+    let newUpdateChannel: string = tracker.getChangedSettings()["updateChannel"];
```
bobbrow: Unfortunately, you can't call `getChangedSettings` here or you will break the client's handling of changed settings on the line under this one (they will be erased on the next call to `getChangedSettings` since no settings will have changed since the last call).

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
     clients.forEach(client => client.onDidChangeVisibleTextEditors(editors));
 }
 
+// TODO move this fn to a common area -- Replace abTesting's downloadCpptoolsJsonAsync with this fn
+function downloadFileToDestination(urlString, destinationPath: string, headers?: OutgoingHttpHeaders): Promise<void> {
```
bobbrow: `urlString` -> `url: string`

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+        });
+        request.on('error', (error) => { reject(); });
+        request.end();
+    });
+}
+
+// TODO factor this out?
+async function parseJsonAtPath(path: string): Promise<any> {
+    try {
+        const exists: boolean = await util.checkFileExists(path);
```
bobbrow: do you need to check if the file exists since this is wrapped by a try/catch already?

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+        if (exists) {
+            const fileContent: string = await util.readFileText(path);
+            return JSON.parse(fileContent);
+        }
+    } catch (error) {
+        console.log(error);
+    }
```
bobbrow: I think we should reject the promise here

Extension/src/LanguageServer/extension.ts
grdowns: Yes, 15 seconds is much too often, even for testing! I managed to hit GitHub's 60 requests per hour (for unauthorized accesses) just by leaving debug on for a few minutes. Woops.
At any rate, I think it might be useful to keep this timer independent so as to allow for a configuration point on Insiders updating specifically

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+        let version: string = util.packageJson["version"];
+        const dashOffset: number = version.indexOf("-");
+        if (dashOffset !== -1) {
+            if (version.substr(dashOffset + 1) !== "insiders") {
+                return null;
+            }
+            version = version.substr(0, dashOffset); // Strip out the suffix
+        }
+        return version;
+    }();
```
bobbrow: We tend not to use a lot of IIFE's in our extension. It's my personal preference to avoid using them. That's not saying you can't, but for something as complex as version management, it might be better to write up a class and have named functions doing the stuff you need.  It makes the code easier to maintain without requiring lots of comments.

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+    if (!version) {
+        return;
+    }
+
+    // Download and parse the JSON release list from GitHub to get the latest build
+    const releaseJsonPath: string = util.getExtensionFilePath("releases.json");
+    await downloadFileToDestination("https://api.github.com/repos/Microsoft/vscode-cpptools/releases",
+                                    releaseJsonPath, { "User-Agent": "vscode-cpptools" });
+    const latestBuild: any = await parseJsonAtPath(releaseJsonPath)[0]; // [0] to get latest build
```
bobbrow: there should be a type check before indexing into what we assume will be an array.

bobbrow: We should also add an interface for the expected GitHub json so we can access the members in a cleaner format. We'll also want to validate the json since we shouldn't blindly trust 3rd-party input.

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+            version = version.substr(0, dashOffset); // Strip out the suffix
+        }
+        return version;
+    }();
+    if (!version) {
+        return;
+    }
+
+    // Download and parse the JSON release list from GitHub to get the latest build
+    const releaseJsonPath: string = util.getExtensionFilePath("releases.json");
```
bobbrow: Are you still planning to download stuff into a tmp folder?  VSCode detects changes in the extension directory and pops up a reload prompt.  We don't want that to happen while we check for updates.

Extension/src/LanguageServer/extension.ts
```diff
@@ -194,6 +218,141 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
...
+    await downloadFileToDestination("https://api.github.com/repos/Microsoft/vscode-cpptools/releases",
+                                    releaseJsonPath, { "User-Agent": "vscode-cpptools" });
+    const latestBuild: any = await parseJsonAtPath(releaseJsonPath)[0]; // [0] to get latest build
+    if (!latestBuild) {
+        return;
+    }
+
+    // Check whether the user actually needs to update
+    if (version >= latestBuild["name"]) { // latestBuild["name"] excludes version suffix
```
bobbrow: I thought the latest build could still have the "-insiders" in it, no?

bobbrow: This comparison is dangerous anyway since "2" > "10".  A "version" class that parses the triplet and compares the pieces numerically is safer.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: There's a "prerelease" key:bool pair in the json which determines whether it is an insiders build. It looks as though the "name" key value pair is the numbered portion only.
https://api.github.com/repos/Microsoft/vscode-cpptools/releases

Good catch on the comparison. Since the comparison is more complex, this strengthens the argument that a `version` class is the way to go

Extension/src/LanguageServer/extension.ts
grdowns: Can do!

Extension/src/LanguageServer/extension.ts
grdowns: Yeah. I fumbled around with import statements and finally got to using the `tmp` library. I have a TODO at the bottom of the function regarding cleaning up the temp files

Extension/src/LanguageServer/extension.ts
grdowns: Good point. I did copy most of this function from updateSettingsAsync in `abTesting.ts`. Perhaps it's not needed there either, though I should limit the scope of changes to this feature only. What do you think?

Extension/src/LanguageServer/extension.ts
grdowns: Thanks

Extension/src/LanguageServer/extension.ts
grdowns: Interesting, though it does explain why I had to do it this way, as trying to access it via the Client yielded no results. I'll look into a non-modifying alternative

Extension/src/LanguageServer/extension.ts
bobbrow: It seems small enough of a change that it won't overly complicate this PR.  I would say go ahead and factor this out to common and update abtesting.ts so we don't leave debt for this case.

Extension/src/LanguageServer/extension.ts
bobbrow: I see `"name": "0.17.8-insiders"` in the result a few entries down.  If an insiders release is the most recent release, then the first one in the list will have `-insiders` in the name.  Am I misunderstanding?

Extension/src/LanguageServer/extension.ts
grdowns: Sounds good!

Extension/src/LanguageServer/extension.ts
grdowns: No, that's my bad. I had misunderstood the `prerelease` meaning. I'll fix this up; Thanks :smile:

Extension/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "cpptools",
   "displayName": "C/C++",
   "description": "C/C++ IntelliSense, debugging, and code browsing.",
-  "version": "0.18.2-master",
+  "version": "0.18.0-master",
```
bobbrow: This file appears to have been merged incorrectly.  The version number and fwlinks look wrong.  They should all be reverted

Extension/package.json
```diff
@@ -126,8 +104,7 @@
           "type": "string",
           "enum": [
             "Default",
-            "Tag Parser",
-            "Disabled"
+            "Tag Parser"
```
bobbrow: revert this.

Extension/src/abTesting.ts
```diff
@@ -83,45 +83,11 @@ export class ABTestSettings {
         }
     }
 
-    // NOTE: Code is copied from DownloadPackage in packageManager.ts, but with ~75% fewer lines.
-    private downloadCpptoolsJsonAsync(urlString): Promise<void> {
```
bobbrow, line 87: With this method removed, I suspect that there are some imports that can be removed as well.  Please check for the grayed out ones.  (TSLint might complain about these too)

Extension/src/common.ts
```diff
@@ -557,3 +559,40 @@ export function promptForReloadWindowDueToSettingsChange(): void {
...
+            }
+            if (response.statusCode !== 200) { // If request is not successful
+                return reject();
+            }
+            // Write file using downloaded data
+            let downloadedBytes = 0; // tslint:disable-line
+            let createdFile: fs.WriteStream = fs.createWriteStream(destinationPath);
+            response.on('data', (data) => { downloadedBytes += data.length; });
+            response.on('end', () => { createdFile.close(); });
+            createdFile.on('close', () => { resolve(); this.updateSettingsAsync(); });
```
bobbrow: `this.updateSettingsAsync` is part of the abtesting class.  It should be removed from here and invoked in the `.then` after the call to download the file.

Extension/src/LanguageServer/extension.ts
```diff
@@ -17,6 +17,12 @@ import { CppSettings } from './settings';
 import { PersistentWorkspaceState } from './persistentState';
 import { getLanguageConfig } from './languageConfig';
 import { getCustomConfigProviders } from './customProviders';
+import { SettingsTracker, getTracker } from './settingsTracker';
```
bobbrow: remove any unneeded imports

Extension/src/LanguageServer/extension.ts
```diff
@@ -17,6 +17,12 @@ import { CppSettings } from './settings';
...
+import { PlatformInformation } from '../platform';
+import { exec, execSync } from 'child_process';
+import * as tmp from 'tmp';
+import { isArray, isString } from 'util';
```
bobbrow: `util` is already imported

Extension/src/LanguageServer/extension.ts
```diff
@@ -141,7 +150,21 @@ export function updateLanguageConfigurations(): void {
  *********************************************/
 
 function onDidChangeSettings(): void {
+    const changedActiveClientSettings: { [key: string] : string } = clients.ActiveClient.onDidChangeSettings();
     clients.forEach(client => client.onDidChangeSettings());
+
+    // Implicitly check whether user changed "updateChannel" setting to "Insiders"
```
bobbrow: this comment seems incorrect and probably can be removed

Extension/src/LanguageServer/extension.ts
```diff
@@ -141,7 +150,21 @@ export function updateLanguageConfigurations(): void {
...
+    const newUpdateChannel: string = changedActiveClientSettings['updateChannel'];
+    if (newUpdateChannel) {
+        if (newUpdateChannel === 'Default') {
+            // clearInterval(insiderUpdateTimer);
+        } else
```
bobbrow: `else if` goes on one line

Extension/src/LanguageServer/extension.ts
```diff
@@ -125,6 +133,7 @@ function realActivation(): void {
 
     reportMacCrashes();
 
+    // insiderUpdateTimer = setInterval(checkAndApplyUpdate, insiderUpdateTimerInterval);
```
bobbrow: is this ready to enable?

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+        const exists: boolean = await util.checkFileExists(path);
+        if (exists) {
+            const fileContent: string = await util.readFileText(path);
+            return JSON.parse(fileContent);
+        }
+    } catch (error) {
+    }
+}
+
+interface ParsedVersion {
```
bobbrow: This should be an 'export' class in its own typescript file.  Rename it `PackageVersion` or `CppToolsVersion`

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+    major: number;
+    minor: number;
+    patch: number;
+    suffix?: string;
+}
+
+function getParsedVersion(versionStr: string): ParsedVersion | undefined {
```
bobbrow: This should be the constructor of the version class

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+    }(tokens);
+
+    if (!parsedVersion.major || !parsedVersion.minor || !parsedVersion.patch) {
+        telemetry.logLanguageServerEvent('versionParsingFailure', { 'versionString': versionStr });
+    }
+
+    return parsedVersion;
+}
+
+function parsedVersionGreater(v1: ParsedVersion, v2: ParsedVersion): boolean {
```
bobbrow: this should be another member of the version class.  Rename it to `isGreaterThan` and remove 1 of the parameters.

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+            const unInstallCommand: string = vsCodeScriptPath + ' --uninstall-extension ms-vscode.cpptools';
+            await execSync(unInstallCommand);
+        }
+        await execSync(installCommand);
+    } catch (error) {
+        telemetry.logLanguageServerEvent("installCommandFailure");
+    }
+}
+
+async function downloadUrlForPlatform(build: Build): Promise<string | undefined> {
```
bobbrow: our linting rules don't require `| undefined`.  This is implicit

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+    if (diff = v2.patch - v1.patch) {
+        return diff > 0;
+    } else
+    if (!v2.suffix && v1.suffix === 'insiders') {
+        return true;
+    }
+    return false;
+}
+
+interface Asset {
```
bobbrow: I think all of the stuff related to the GitHub API should move to its own file.  Export a single function that gets you a VSIX download url.  Something like `export function getTargetBuild(updateChannel: string): string`.  If an empty string or undefined is returned, you do nothing.

All of the github api stuff, version checking, etc, happens as a result of calling this export.

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+function isAsset(input: any): input is Asset {
+    return input && input.name && typeof(input.name) === "string" && input.browser_download_url && input.browser_download_url && typeof(input.browser_download_url) === "string"; 
+}
+
+function isBuild(input: any): input is Build {
+    return input && input.name && typeof(input.name) === "string" && isArrayOfAssets(input.assets) && input.assets.length >= 4;
+}
+
+function isArrayOfAssets(input: any): input is Asset[] {
+    if (!(input instanceof Array)) {
```
bobbrow: nit: a more typescripty way of doing this would be:
```
return (input instanceof Array) && input.every(item => isAsset(item));
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+        if (!isAsset(item)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+function isReleaseJson(input: any): input is Build[] {
+    if (!input || !(input instanceof Array) || input.length === 0) {
```
bobbrow:
```
return input && (input instanceof Array) && input.length > 0 && input.every(item => isBuild(item));
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+async function installVsix(vsixLocation: string, updateChannel: string): Promise<void> {
+    // Get the path to the VSCode command -- replace logic later when VSCode allows calling of
+    // workbench.extensions.action.installVSIX from TypeScript w/o instead popping up a file dialog
+    const platformInfo: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+    const vsCodeScriptPath: string | undefined = await async function(platformInfo): Promise<string | undefined> {
+        if (platformInfo.platform === 'win32') {
+            const vsCodeProcessPath: string = path.dirname(process.execPath);
+            return path.join(vsCodeProcessPath, 'bin', 'code.cmd');
+        } else {
+            const stdout: Buffer = await execSync("which code");
```
bobbrow: This doesn't seem to account for if the user is using the insiders build.  I know the insiders version on Windows has a different name than the stable version and I would expect the mac/linux versions to have different names too.

bobbrow: Also, I wouldn't expect `await execSync` to be doing what you expect.  Is there an async version of `exec` you can use instead?

bobbrow: (there are additional `await execSync` lines that will likely need to be updated too)

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+                }
+            }
+        }
+    }(platformInfo);
+    if (!vsixName) {
+        return;
+    }
+
+    // Get the URL to download the VSIX, using vsixName as a key
+    const downloadUrl: string | undefined = build.assets.find((asset) => {
```
bobbrow: when you have a single-line lambda that returns a result, you should use this syntax:
```
build.assets.find(asset => asset.name === vsixName);
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+        return asset.name === vsixName;
+    }).browserDownloadUrl;
+
+    return downloadUrl;
+}
+
+// Determines whether there exists a build that should be installed; returns the build if there is
+function getTargetBuild(releaseJson: Build[], updateChannel: string): Build | undefined {
+    // Get predicates to determine the build to install, if any
+    let needsUpdatePred: any;
```
bobbrow: 'Pred' is a kind of Hungarian which we should avoid.  I would recommend:
```
let needsUpdate: (v1: ParsedVersion, v2: ParsedVersion) => boolean;
let useBuild: (build: Build) => boolean;
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+    } else
+    if (updateChannel === 'Default') {
+        needsUpdatePred = function(v1: ParsedVersion, v2: ParsedVersion): boolean { return parsedVersionGreater(v2, v1); };
+        buildPred = function(build: Build): boolean { return build.name.indexOf('-') === -1; };
+    } else {
+        return;
+    }
+
+    // Get the build to install
+    const targetBuild: Build = releaseJson.find((build) => {
```
bobbrow:
```
find(build => buildPred(build));
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +222,253 @@ function onInterval(): void {
...
+
+    return targetBuild;
+}
+
+async function getReleaseJson(): Promise<Build[] | undefined> {
+    const releaseJsonFile: any = tmp.fileSync();
+
+    // Download json from GitHub
+    await util.downloadFileToDestination('https://api.github.com/repos/Microsoft/vscode-cpptools/releases',
+        releaseJsonFile.name, { 'User-Agent': 'vscode-cpptools' }).catch(() => {
```
bobbrow: nit: make the url a const var so that you can fit all the parameters on one line.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: How should we deal with whether the constructor fails? (If the incoming string is malformed)

Extension/src/LanguageServer/extension.ts
grdowns: Yeah, `exec` itself is async. Can't remember why I chose `await execSync`. Doesn't make sense either in hindsight

Extension/src/LanguageServer/extension.ts
bobbrow: Adding an `isValid` field would be one way to deal with it.  Another way would be to make the constructor take in the 4 data fields and have a helper function (static member?) do the parsing and return either a valid version or undefined.

Extension/src/LanguageServer/extension.ts
grdowns: I think we should leave it disabled until we're ready to approve the PR. The rate limiting has hampered my testing previously. The alternative is to use [authentication](https://developer.github.com/v3/#rate-limiting) which means adding some code I'd have to take out later anyway

Extension/src/abTesting.ts
bobbrow: Here is the TSLint report:

```
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/abTesting.ts[7, 13]: 'url' is declared but its value is never read.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/abTesting.ts[8, 13]: 'https' is declared but its value is never read.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/abTesting.ts[9, 1]: All imports are unused.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/abTesting.ts[10, 13]: 'vscode' is declared but its value is never read.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/abTesting.ts[11, 13]: 'fs' is declared but its value is never read.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/extension.ts[20, 1]: All imports are unused.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/extension.ts[22, 10]: 'exec' is declared but its value is never read.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/extension.ts[24, 1]: All imports are unused.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/extension.ts[25, 1]: All imports are unused.
ERROR: /home/travis/build/Microsoft/vscode-cpptools/Extension/src/LanguageServer/extension.ts[38, 7]: 'insiderUpdateTimerInterval' is declared but its value is never read.
```

Extension/src/LanguageServer/extension.ts
bobbrow: don't waste any time on authentication

Extension/src/ghAPI.ts
```diff
@@ -0,0 +1,152 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
```
bobbrow: We should avoid abbreviations in the file name.  Please rename this file `githubApi.ts` or similar.

Extension/src/ghAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+function isBuild(input: any): input is Build {
+    return input && input.name && typeof(input.name) === "string" && isArrayOfAssets(input.assets) && input.assets.length >= 4;
+}
+
+function isArrayOfAssets(input: any): input is Asset[] {
+    return input instanceof Array && input.every(item => isAsset(item));
+}
+
+function isReleaseJson(input: any): input is Build[] {
+    return input && input instanceof Array && input.length !== 0 && input.every(item => isBuild(item));
```
bobbrow: as we discussed, maybe don't validate "every" item, just the first few.  (or maybe validate until you find both a prerelease and stable release?)

Extension/src/ghAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+async function downloadUrlForPlatform(build: Build): Promise<string> {
```
bobbrow: nit: casing for `URL` -> `Url`

Extension/src/ghAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+    // Get the URL to download the VSIX, using vsixName as a key
+    const downloadUrl: string = build.assets.find((asset) => {
+        return asset.name === vsixName;
+    }).browserDownloadUrl;
+
+    return downloadUrl;
+}
+
+export async function getTargetBuildURL(updateChannel: string): Promise<string> {
+    const releaseJson: Build[] = await getReleaseJson();
```
bobbrow: It's typed as a Build[], so I'd prefer removing "json" from the name of the variable.  Consider calling it `releases` or `builds` (plural because it's an array) instead.

Extension/src/ghAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+    return targetBuild;
+}
+
+async function getReleaseJson(): Promise<Build[]> {
+    const releaseJsonFile: any = tmp.fileSync();
+
+    // Download json from GitHub
+    const releaseUrl: string = 'https://api.github.com/repos/Microsoft/vscode-cpptools/releases';
+    await util.downloadFileToDestination(releaseUrl, releaseJsonFile.name, { 'User-Agent': 'vscode-cpptools' }).catch(() => {
+            // TODO check internet connection before attempting? That way we can mitigate false positive telemetry
```
bobbrow: It's hard to determine what to check.  They could have Internet, but be behind a proxy and they haven't configured it correctly.  It's good for us to see those errors to we can attempt to mitigate them.  You can remove this comment

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (tokens.length < 3) {
+            this.isValid = false;
+            telemetry.logLanguageServerEvent('versionParsingFailure', { 'versionString': versionStr });
+            return;
+        }
+
+        this.major = parseInt(tokens[0]);
+        this.minor = parseInt(tokens[1]);
+        this.patch = parseInt(tokens[2]);
+        this.suffix = tokens[3];
```
bobbrow: do you need to check the length of the array before grabbing the 4th element?

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+            telemetry.logLanguageServerEvent('versionParsingFailure', { 'versionString': versionStr });
+            return;
+        }
+
+        this.isValid = true;
+    }
+
+    public isGreaterThan(other: PackageVersion): boolean {
+        // PackageVersions cannot be compared if either have a suffix that is not 'insiders'
+        if ((this.suffix && this.suffix !== 'insiders') || (other.suffix && other.suffix !== 'insiders')) {
```
bobbrow: You might want to do `.startsWith('insiders')`.
If we have more than one insiders release the Nth one will be called `-insidersN`

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+            return false;
+        }
+
+        let diff: number = other.major - this.major;
```
bobbrow: Based on the comparisons below, this seems backwards.  We're checking if `this > other`, so if you return true for a positive `diff` you should do `this - other`.

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (diff) {
+            return diff > 0;
+        } else if (diff = other.minor - this.minor) {
+            return diff > 0;
+        } else if (diff = other.patch - this.patch) {
+            return diff > 0;
+        } else if (!other.suffix && this.suffix === 'insiders') {
```
bobbrow: This also seems backwards.  0.18.0 > 0.18.0-insiders.
You also need to take into consideration insiders3 > insiders2 > insiders

Extension/src/packageVersion.ts
bobbrow: You may also want to return false if either PackageVersion is invalid.

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +218,56 @@ function onInterval(): void {
...
+            execSync(uninstallCommand);
+        }
+        execSync(installCommand);
+    } catch (error) {
+        telemetry.logLanguageServerEvent("installCommandFailure");
+    }
+}
+
+async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
+    // TODO query whether user has extensions.autoUpdate turned off?
```
bobbrow: Nah.  If they opt in to the insiders channel, it is assumed that they want us to auto-update.  You can remove this comment.

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +218,56 @@ function onInterval(): void {
...
+    const downloadUrl: string = await getTargetBuildURL(updateChannel);
+    if (!downloadUrl) {
+        return;
+    }
+
+    // Download the target version and install it
+    const vsixFile: any = tmp.fileSync({ postfix: '.vsix' });
```
bobbrow: is there an async version?

bobbrow (changes requested):

Extension/src/packageVersion.ts
grdowns: `suffix` is marked optional, which according to [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html) means it is also an `undefined` type. There is no crash after assignment when `tokens[3]` evaluates to `undefined`, so my assumption is it's well formed. What do you think?

Extension/src/packageVersion.ts
grdowns: Good catches. Thanks!

Extension/src/LanguageServer/extension.ts
```diff
@@ -221,51 +221,55 @@ function onInterval(): void {
...
-    const vsCodeScriptPath: string = function(platformInfo): string {
-        if (platformInfo.platform === 'win32') {
-            const vsCodeProcessPath: string = path.dirname(process.execPath);
-            const scriptName: string = path.basename(process.execPath) + '.cmd';
-            return path.join(vsCodeProcessPath, 'bin', scriptName);
-        } else {
-            const vsCodeProcessPath: string = path.basename(process.execPath);
-            const stdout: Buffer = execSync("which " + vsCodeProcessPath);
-            return stdout.toString().trim();
+    return PlatformInformation.GetPlatformInformation().then((platformInfo) => {
```
bobbrow: I'm not sure why you removed `await` here.

grdowns: Good point! I'll add that back; it was more clear the other way

bobbrow:
I'm going through the telemetry changes now and will make some name change requests for the next review.

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,70 @@
...
+    patch: number;
+    suffix?: string;
+    suffixVersion: number;
+    isValid: boolean;
+
+    constructor(versionStr: string) {
+        let tokens: string[] = versionStr.split(new RegExp('[-\\.]', 'g')); // Match against dots and dashes
+        if (tokens.length < 3) {
+            this.isValid = false;
+            telemetry.logLanguageServerEvent('versionParsingFailure', { 'versionString': versionStr });
```
bobbrow: rename event -> `'invalidPackageVersion'`
rename data -> `{ 'value': versionStr }`

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        this.major = parseInt(tokens[0]);
+        this.minor = parseInt(tokens[1]);
+        this.patch = parseInt(tokens[2]);
+        this.suffix = tokens[3];
+        this.suffixVersion = tokens.length >= 5 ? parseInt(tokens[4]) : 0;
+
+        if (this.major === undefined || this.minor === undefined || this.patch === undefined) {
+            this.isValid = false;
+            telemetry.logLanguageServerEvent('versionParsingFailure', { 'versionString': versionStr });
```
bobbrow: rename event -> `'invalidPackageVersion'`
rename data -> `{ 'value': versionStr }`

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export class PackageVersion {
+    major: number;
+    minor: number;
+    patch: number;
+    suffix?: string;
+    suffixVersion: number;
+    isValid: boolean;
+
+    constructor(versionStr: string) {
```
bobbrow: nit: `Str` should not be in the name of the variable.  You can just call it `version`.

Extension/src/packageVersion.ts
bobbrow: it seems to work, so you can leave it.

Extension/package.json
```diff
@@ -456,6 +456,16 @@
           "default": null,
           "description": "The value to use for the system include path. If set, it overrides the system include path acquired via \"compilerPath\" and \"compileCommands\" settings.",
           "scope": "resource"
+        },
+        "C_Cpp.updateChannel": {
```
bobbrow, line 460: All new settings should get a property added to `CppSettings` in `settings.ts`

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +218,60 @@ function onInterval(): void {
...
+                const scriptName: string = path.basename(process.execPath) + '.cmd';
+                return path.join(vsCodeProcessPath, 'bin', scriptName);
+            } else {
+                const vsCodeProcessPath: string = path.basename(process.execPath);
+                const stdout: Buffer = execSync("which " + vsCodeProcessPath);
+                return stdout.toString().trim();
+            }
+        }(platformInfo);
+        if (!vsCodeScriptPath) {
+            telemetry.logLanguageServerEvent('findVsCodeScriptFailure');
```
bobbrow: Rather than have a bunch of different telemetry event names, I think there should be one event with a result status payload.  We can communicate the error type via the rejected promise.  So for this case we should remove the telemetry call and reject the promise with a `new Error("findVSCodeScript")` object. In the final catch or error handler, we take the string from the error object and log it with the `'installVsix'` event using `error` or `success` as the payload.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+        try {
+            if (updateChannel === 'Default') {
+                // Uninstall the current version, as the version to install is a previous version
+                const uninstallCommand: string = vsCodeScriptPath + ' --uninstall-extension ms-vscode.cpptools';
+                execSync(uninstallCommand);
+            }
+            execSync(installCommand);
+            return Promise.resolve();
+        } catch (error) {
+            return Promise.reject(new Error('Install command failed'));
```
bobbrow: To be consistent with other errors, can you format it like: `"Failed to install vsix"`?
I also want to add the `err` info to the message, but it might have PII in it, so it could be off limits.  We can come back to this later if we see a lot of failure telemetry.

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
+    return getTargetBuildURL(updateChannel).then((downloadUrl) => {
+        if (!downloadUrl) {
+            return Promise.resolve();
+        }
+
+        // Create a temporary file, download the vsix to it, then install the vsix
+        tmp.file({postfix: '.vsix'}, async (err, vsixPath, fd, cleanupCallback) => {
+            if (err) {
+                cleanupCallback();
```
bobbrow: Based on what I read on the `tmp` documentation, it looks like we don't need to manually call the cleanup callback since we don't pass in the option to keep the file.
https://www.npmjs.com/package/tmp

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+                return Promise.reject(new Error('Failed to create vsix file'));
```
bobbrow: I don't think returning a promise here actually does anything.  You need to wrap the call to `tmp.file` in the constructor of a new promise and call the `reject` function here instead.  Examples of dealing with callback APIs are in `common.ts`  (e.g. `readFileText`).

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+            }
+            await util.downloadFileToDestination(downloadUrl, vsixPath).catch(() => {
+                cleanupCallback();
+                return Promise.reject(new Error('Failed to download vsix'));
+            });
+            await installVsix(vsixPath, updateChannel).then(() => {
+                cleanupCallback();
+                telemetry.logLanguageServerEvent('installVsix', { 'Success': '' } );
```
bobbrow: Don't forget to set the value of the payload to "true"

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+            }, error => { cleanupCallback(); return Promise.reject(error); });
+        });
+    }, (error: Error) => { telemetry.logLanguageServerEvent('installVsix', { 'Error': error.message }); });
```
bobbrow: You need to also add 'Success': 'false' to the payload

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
...
+            }
+            execSync(installCommand);
+            return Promise.resolve();
+        } catch (error) {
+            return Promise.reject(new Error('Install command failed'));
+        }
+    });
+}
+
+async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
```
bobbrow: If this function succeeds in installing a VSIX, we need to disable the interval timer so that we don't try to re-download and reinstall the VSIX multiple times.  In the best case, the user will reload the window for the update to take effect, but in reality, users may not reload the window for some time (especially since we don't notify them of the update right now)

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+    browser_download_url: string;
+}
+
+interface Build {
+    name: string;
+    assets: Asset[];
+}
+
+function isAsset(input: any): input is Asset {
+    const ok: boolean = input && input.name && typeof(input.name) === "string" &&
```
bobbrow: nit: I imagine the `ok` vars were added for debugging.  We probably don't need to keep them long term.

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+            default: {
+                switch (platformInfo.architecture) {
+                    case 'x86': return 'cpptools-linux32.vsix';
+                    case 'x86_64': return 'cpptools-linux.vsix';
+                }
+            }
+        }
+    }(info);
+    if (!vsixName) {
+        const distro: string = info.distribution ? (':' + info.distribution.name + '-' + info.distribution.version) : '';
```
bobbrow: For custom builds of linux, I think some of this info could have PII in it that we don't want to log.  Check with @sean-mcmanus which fields are safe.

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+    const info: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+    const vsixName: string = vsixNameForPlatform(info);
+
+    // Get the URL to download the VSIX, using vsixName as a key
+    const downloadUrl: string = build.assets.find((asset) => {
+        return asset.name === vsixName;
+    }).browser_download_url;
+
+    if (!downloadUrl) {
+        return Promise.reject(new Error('Failed to find VSIX: ' + vsixName + ' in build: ' + build.name));
```
bobbrow: in an `async` function it's better if this is a `throw` rather than a `reject`.

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }(info);
+    if (!vsixName) {
+        const distro: string = info.distribution ? (':' + info.distribution.name + '-' + info.distribution.version) : '';
+        throw (new Error('Failed to match VSIX name for: ' + info.platform + distro + ':' + info.architecture));
+    }
+    return vsixName;
+}
+
+async function downloadUrlForPlatform(build: Build): Promise<string> {
```
bobbrow: consider renaming this to `getVsixDownloadUrl`

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+        return asset.name === vsixName;
+    }).browser_download_url;
+
+    if (!downloadUrl) {
+        return Promise.reject(new Error('Failed to find VSIX: ' + vsixName + ' in build: ' + build.name));
+    }
+    return downloadUrl;
+}
+
+export async function getTargetBuildURL(updateChannel: string): Promise<string> {
```
bobbrow: nit: `URL` -> `Url`

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+    return getReleaseJson()
+        .then(builds => getTargetBuild(builds, updateChannel))
+        .then(build => { return build ? downloadUrlForPlatform(build) : Promise.resolve(undefined); } );
+}
+
+// Determines whether there exists a build that should be installed; returns the build if there is
+function getTargetBuild(builds: Build[], updateChannel: string): Build | undefined {
+    // Get predicates to determine the build to install, if any
+    let needsUpdate: (v1: PackageVersion, v2: PackageVersion) => boolean;
```
bobbrow: Self-documentation would improve if the parameters were renamed: `v1` -> `installed` and `v2` -> `latest` (or similar)

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    return targetBuild;
+}
+
+async function getReleaseJson(): Promise<Build[]> {
+    return new Promise<Build[]>((resolve, reject) => {
+        // Create temp file to hold json
+        tmp.file(async (err, releaseJsonPath, fd, cleanupCallback) => {
+            if (err) {
+                cleanupCallback();
```
bobbrow: see comment above. `cleanupCallback` should not be needed.

Extension/src/githubAPI.ts
```diff
@@ -0,0 +1,152 @@
...
+                // Read json
+                .then(() => { return util.readFileText(releaseJsonPath); })
+                .catch(() => { cleanupCallback(); reject(new Error('Failed to read release json file')); })
+                // Parse json
+                .then(fileContent => { cleanupCallback(); return Promise.resolve(fileContent); })
+                .then(JSON.parse)
+                .catch(() => { reject(new Error('Failed to parse release json')); })
+                // Type check and return
+                .then(releaseJson => {
+                    isArrayOfBuilds ? resolve(releaseJson) : reject('Release json is not Build[]');
```
bobbrow: looks like a potential remnant on the syntax we discussed on Friday.  Make sure you pass in `releaseJson` to `isArrayOfBuilds` here.

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,58 @@
...
+        let tokens: string[] = version.split(new RegExp('[-\\.]', 'g')); // Match against dots and dashes
+        if (tokens.length < 3) {
+            throw new Error('Failed to parse version string: ' + version);
+        }
+
+        this.major = parseInt(tokens[0]);
+        this.minor = parseInt(tokens[1]);
+        this.patch = parseInt(tokens[2]);
+        this.suffix = tokens[3];
+        this.suffixVersion = tokens.length >= 5 ? parseInt(tokens[4]) : 0;
```
bobbrow: I don't think this works with our naming conventions.  There is no `tokens[4]` because `insiders[N?]` doesn't split `insiders` from `N`.  (and there is no `insiders1`).  You need to remove `"insiders"` from tokens[3] and then parse the resulting integer, if present.

Extension/src/packageVersion.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        let diff: number = this.major - other.major;
+        if (diff) {
+            return diff > 0;
+        } else if (diff = this.minor - other.minor) {
+            return diff > 0;
+        } else if (diff = this.patch - other.patch) {
+            return diff > 0;
+        } else {
```
bobbrow: simplifying from here down:

```
} else if (this.suffix) {
    return (other.suffix && this.suffixVersion > other.suffixVersion);
} else {
    return other.suffix;
}
```

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -199,6 +223,64 @@ function onInterval(): void {
     clients.ActiveClient.onInterval();
 }
 
+async function installVsix(vsixLocation: string, updateChannel: string): Promise<void> {
+    // Get the path to the VSCode command -- replace logic later when VSCode allows calling of
+    // workbench.extensions.action.installVSIX from TypeScript w/o instead popping up a file dialog
+    return PlatformInformation.GetPlatformInformation().then((platformInfo) => {
+        const vsCodeScriptPath: string = function(platformInfo): string {
```
WardenGnaw, line 235: Why is this an anonymous function? Just assign vsCodeScriptPath instead of returning.

WardenGnaw:
[nit] Any reason why you are using `.then()` `.catch()` instead of `await` `try/catch`?

Extension/src/LanguageServer/extension.ts
grdowns: `vsCodeScriptPath` needs to be checked before being returned, as it should never be undefined. I wrapped it in an IIFE to make it const, though admittedly it's a bit overkill as it was refactored into its own function

grdowns:
Refactoring and PR feedback is all responded to. Upgrade + downgrade have been tested on Linux since the PR feedback. I will need to run the scenarios for both Mac + Win

grdowns:
@WardenGnaw I chose promises largely to learn how async programming works in TS. I found that for this await, when I used `await` I missed a lot of exceptions. It seems that without a strong foundation in using promises, some of the bugs/problems posed by using `await` don't easily present themselves. There's a really good article on it [here](https://medium.com/@bluepnume/learn-about-promises-before-you-start-using-async-await-eb148164a9c8) if you're interested in reading more.

Extension/src/githubAPI.ts
```diff
@@ -64,44 +71,37 @@ function vsixNameForPlatform(info: PlatformInformation): string {
...
-        return asset.name === vsixName;
-    }).browser_download_url;
-
-    if (!downloadUrl) {
-        return Promise.reject(new Error('Failed to find VSIX: ' + vsixName + ' in build: ' + build.name));
-    }
-    return downloadUrl;
-}
+export async function getTargetBuildUrl(updateChannel: string): Promise<string> {
+    const boundGetTargetBuild: (builds: Build[]) => Build = getTargetBuild.bind(undefined, updateChannel);
```
bobbrow: Please don't use `bind`

Extension/src/LanguageServer/extension.ts
```diff
@@ -250,35 +250,41 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
-
-        // Create a temporary file, download the vsix to it, then install the vsix
-        tmp.file({postfix: '.vsix'}, async (err, vsixPath, fd, cleanupCallback) => {
-            if (err) {
-                cleanupCallback();
-                return Promise.reject(new Error('Failed to create vsix file'));
+    const p: Promise<void> = new Promise<void>((resolve, reject) => {
+        getTargetBuildUrl(updateChannel).then((downloadUrl) => {
+            if (!downloadUrl) {
+                return resolve();
```
bobbrow: I don't think it is technically correct to `return resolve();`  `resolve()` is a function that returns nothing (void).  You might be confusing it with `return Promise.resolve();` which creates a new, resolved promise and returns it.

I think it is more correct to do:

```
    resolve();
    return;
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -250,35 +250,41 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+
+            // Create a temporary file, download the vsix to it, then install the vsix
+            tmp.file({postfix: '.vsix'}, (err, vsixPath, fd, cleanupCallback) => {
+                if (err) {
+                    return reject(new Error('Failed to create vsix file'));
+                }
+    
+                const downloadVsix: any = util.downloadFileToDestination.bind(undefined, downloadUrl, vsixPath);
+                const boundInstallVsix: any = installVsix.bind(undefined, vsixPath, updateChannel);
+                const logSuccess: any = () => { telemetry.logLanguageServerEvent('installVsix', { 'Success': 'true' } ); };
```
bobbrow: Please don't use `bind`.  I thought we talked about not using `.then(func)` syntax in our code (which seems to be the underlying reason for this).  Please use `var => func(var)` syntax for the continuation clauses.  It makes the code easier to read and maintain.

Extension/src/LanguageServer/extension.ts
```diff
@@ -250,35 +250,41 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
             });
-            await installVsix(vsixPath, updateChannel).then(() => {
-                cleanupCallback();
-                telemetry.logLanguageServerEvent('installVsix', { 'Success': '' } );
-            }, error => { cleanupCallback(); return Promise.reject(error); });
         });
-    }, (error: Error) => { telemetry.logLanguageServerEvent('installVsix', { 'Error': error.message }); });
+    });
+    return p.catch((error: Error) => {
+        telemetry.logLanguageServerEvent('installVsix', { 'Error': error.message, 'Success': 'false' });
```
bobbrow: The payload property names should be all lower-case: `error`, `success`

Extension/src/githubAPI.ts
```diff
@@ -116,37 +116,36 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build | undefin
...
-                .catch(() => { cleanupCallback(); reject(new Error('Failed to read release json file')); })
-                // Parse json
-                .then(fileContent => { cleanupCallback(); return Promise.resolve(fileContent); })
-                .then(JSON.parse)
-                .catch(() => { reject(new Error('Failed to parse release json')); })
-                // Type check and return
-                .then(releaseJson => {
-                    isArrayOfBuilds ? resolve(releaseJson) : reject('Release json is not Build[]');
-                });
+            const downloadReleaseJson: any = util.downloadFileToDestination.bind(undefined, releaseUrl, releaseJsonPath, header);
```
bobbrow: Please don't use `bind`.

bobbrow (changes requested):

Extension/src/githubAPI.ts
```diff
@@ -119,33 +164,34 @@ function getTargetBuild(updateChannel: string, builds: Build[]): Build {
...
 
-            return downloadReleaseJson()
-                .catch(rejectDownload)
-                .then(readJson, rejectRead)
-                .then(JSON.parse, rejectParse)
-                .then(typeCheck);
+            const releaseUrl: string = 'https://api.github.com/repos/Microsoft/vscode-cpptools/releases';
+            const header: OutgoingHttpHeaders = { 'User-Agent': 'vscode-cpptools' };
+            return util.downloadFileToDestination(releaseUrl, releaseJsonPath, header)
+                .then(() => util.readFileText(releaseJsonPath), () => { return rejectDownload(); })
```
bobbrow: this doesn't work the way you want it to.  All of the `.then` branches are going to be executed even if one of the promises is rejected.

Extension/src/githubAPI.ts
```diff
@@ -119,33 +164,34 @@ function getTargetBuild(updateChannel: string, builds: Build[]): Build {
...
-            const header: OutgoingHttpHeaders = { 'User-Agent': 'vscode-cpptools' };
-            const downloadReleaseJson: any = util.downloadFileToDestination.bind(undefined, releaseUrl, releaseJsonPath, header);
-            const rejectDownload: any = () => { return reject(new Error('Failed to download release json')); };
-
-            const readJson: any = util.readFileText.bind(undefined, releaseJsonPath);
-            const rejectRead: any = () => { return reject(new Error('Failed to read release json file')); };
-
-            const rejectParse: any = () => { return reject(new Error('Failed to parse release json')); };
+            // Helper functions to handle promise rejection
+            const rejectDownload: any = () => { return reject(new Error('Failed to download release JSON')); };
```
bobbrow: Avoid using `any` when possible.  This is masking a bug in the `.then` chain below.

bobbrow (changes requested):

bobbrow (approved):
Go ahead and commit, but we still need to test Insiders on Windows and make sure that works (might need another PR).

grdowns:
Linux has its testing complete. Win + Mac will require minor changes. To make the diffs easier, I'm going to squash + merge this PR, then start up a second PR for Win + Mac stuff, and anything else that arises out of testing.

Issue #2491, opened 2018-09-06
Variable expansion does not work for default setttings


g-arjones:
Type: General
- OS and Version: Ubuntu 16.04
- VS Code Version: 1.27.1
- C/C++ Extension Version: 0.18.1
- Other extensions you installed (and if the issue persists after disabling them): none
- Variables expansion do not work for default settings

1. Open user settings.json
2. Add ""C_Cpp.default.compileCommands": "${workspaceRoot}/build/compile_commands.json"

**Expected behavior**
${workspaceRoot} should be expanded to the absolute path of the folder

sean-mcmanus:
Hmm...it's working for me. It's possible something else is causing your compile_commands.json to not be used. In the error message we send back, we show "${workspaceFolder}" intentionally. If the path doesn't exist, it will be silently ignored (it might be better if we showed some warning message). @bobbrow Are you able to repro any issue?

g-arjones:
Is ${workspaceFolderBasename} supposed to be expanded as well?

bobbrow:
> Is ${workspaceFolderBasename} supposed to be expanded as well?

We don't currently support that particular variable.

g-arjones:
Is there any way to get the path where the .code-workspace file is?

bobbrow:
I don't believe VS Code has a variable for that.  https://code.visualstudio.com/docs/editor/variables-reference

g-arjones:
Any plans to add support for ${workspaceFolderBasename} ? Maybe I can open a pull request for that if you point me in the right direction...

bobbrow:
You could search for usages of `util.resolveVariables` and add `"workspaceFolderBasename": <value>` to the  `additionalEnvironment` passed in.
* For `configurations.ts` you should be able to get that value by extracting it from `this.rootUri`.
* For `client.ts` you should be able to get it from `this.Name`.

g-arjones:
Great! I will give it a try... Thanks.

g-arjones:
Done: #2495

PR #2495, opened 2018-09-07
Add support for 'workspaceFolderBasename' expansion (#2491)

g-arjones:
@bobbrow Could you take a look? Please, let me know if there are any issues with this implementation... Thanks!

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -240,6 +240,10 @@ class DefaultClient implements Client {
         return this.trackedDocuments;
     }
 
+    private get AdditionalEnvironment(): { [key: string]: string | string[] } {
+        return { workspaceFolderBasename: this.Name };
+    }
+
     private getName(workspaceFolder?: vscode.WorkspaceFolder): string {
         return workspaceFolder ? workspaceFolder.name : "untitled";
     }
@@ -337,7 +341,7 @@ class DefaultClient implements Client {
             },
             workspaceFolder: workspaceFolder,
             initializationOptions: {
-                clang_format_path: util.resolveVariables(settings.clangFormatPath, null),
+                clang_format_path: util.resolveVariables(settings.clangFormatPath, this.AdditionalEnvironment),
                 clang_format_style: settings.clangFormatStyle,
                 clang_format_fallbackStyle: settings.clangFormatFallbackStyle,
                 clang_format_sortIncludes: settings.clangFormatSortIncludes,
@@ -405,7 +409,7 @@ class DefaultClient implements Client {
             }
             if (changedSettings["clang_format_path"]) {
                 let settings: CppSettings = new CppSettings(this.RootUri);
-                this.languageClient.sendNotification(UpdateClangFormatPathNotification, util.resolveVariables(settings.clangFormatPath, null));
+                this.languageClient.sendNotification(UpdateClangFormatPathNotification, util.resolveVariables(settings.clangFormatPath, this.AdditionalEnvironment));
             }
             this.configuration.onDidChangeSettings();
             telemetry.logLanguageServerEvent("CppSettingsChange", changedSettings, null);
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -611,7 +611,11 @@ export class CppProperties {
                 delete this.configurationJson.env['workspaceRoot'];
                 delete this.configurationJson.env['workspaceFolder'];
                 delete this.configurationJson.env['default'];
+            } else {
+                this.configurationJson.env = {};
             }
+
+            this.configurationJson.env["workspaceFolderBasename"] = path.basename(this.rootUri.fsPath);
 
             // Warning: There is a chance that this is incorrect in the event that the c_cpp_properties.json file was created before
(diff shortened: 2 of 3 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -393,11 +393,18 @@ export class CppProperties {
         return result;
     }
 
+    private get additionalEnvironment(): { [key: string]: string | string[] } {
+        let env: { [key: string]: string | string[] } = this.configurationJson.env;
+        env["workspaceFolderBasename"] = path.basename(this.rootUri.fsPath);
```
bobbrow: I think it would be better to just add "workspaceFolderBasename" to configurationJson.env right after it is parsed (it could be `undefined` here).  In `parsePropertiesFile()` we remove a couple of disallowed variables because we want to resolve them elsewhere.  At that point we should ensure that `.env` is an object (create one if it's undefined), and then add this key.

We should also update `c_cpp_properties.schema.json`'s `patternProperties` to disallow `workspaceFolderBasename` as a property of `env` so that green squiggles will appear in `c_cpp_properties.json` if anyone tries to set it manually.

Extension/src/LanguageServer/client.ts
```diff
@@ -310,6 +310,13 @@ class DefaultClient implements Client {
         }
     }
 
+    private get additionalEnvironment(): { [key: string]: string | string[] } {
+        let env: { [key: string]: string | string[] } = {};
+        env["workspaceFolderBasename"] = this.Name;
```
bobbrow: It might save a few lines (and type definitions) if you just do:

```
return { workspaceFolderBasename: this.Name };
```

Extension/src/LanguageServer/client.ts
```diff
@@ -310,6 +310,13 @@ class DefaultClient implements Client {
         }
     }
 
+    private get additionalEnvironment(): { [key: string]: string | string[] } {
```
bobbrow: minor style issue - we name property getters/setters with a leading capital letter (e.g. AdditionalEnvironment).

bobbrow: Or you could just make this a readonly variable on the class and leave the casing the same.  Either way.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
g-arjones: >  think it would be better to just add "workspaceFolderBasename" to configurationJson.env right after it is parsed 

I considered that first but then I thought that it sounded wrong, semantically. I mean, one might need a way to get the actual environment that was set by the user in the configuration json. Thoughts?

> We should also update c_cpp_properties.schema.json's patternProperties to disallow workspaceFolderBasename as a property of env so that green squiggles will appear in c_cpp_properties.json if anyone tries to set it manually.

Definitely! I'm on it.

Extension/src/LanguageServer/client.ts
g-arjones: 👍

Extension/src/LanguageServer/client.ts
g-arjones: Kept as a property and capitalized the leading letter.

g-arjones:
Ok, I have amended the commit:

1. Renamed `additionalEnvironment` to `ExtendedEnvironment` to make the meaning clearer and follow the style guideline (configurations.ts);
2. Explicitly removed `workspaceFolderBasename` from the parsed configuration json in `parsePropertiesFile()`;
3. Updated c_cpp_properties.schema.json to disallow `workspaceFolderBasename` in `env`;
4. Fixed the style issues in client.ts.

I can still do as you suggested and just add `workspaceFolderBasename` key to the env object in `parsePropertiesFile()` if it makes more sense to you. Your call, just let me know...

Could you please have a look again?

Extension/src/LanguageServer/configurations.ts
bobbrow: > I considered that first but then I thought that it sounded wrong, semantically. I mean, one might need a way to get the actual environment that was set by the user in the configuration json. Thoughts?

I don't think TypeScript/JavaScript does a deep copy on assignment so you're modifying the `.env` here anyway (if I'm reading [this](https://docstore.mik.ua/orelly/web/jscript/ch09_03.html) right).  I think it's fine to do it this way since we disallow user override per the schema and the results are correct.  And if we do it in `parsePropertiesFile()` we only have to do it once every time `c_cpp_properties.json` changes instead of every time the getter is accessed.

bobbrow (changes requested):
I responded to your question in the comment thread

Extension/src/LanguageServer/configurations.ts
g-arjones: > so you're modifying the .env here anyway

Yeah, you are right. 

> And if we do it in parsePropertiesFile() we only have to do it once every time

True, let's go with that... One possible caveat: will `parsePropertiesFile()` be called again if the folder is renamed?

g-arjones:
Updated again! Third time is the charm... :-)

bobbrow (approved):

bobbrow:
Looks good.  Thanks for your help on this!

PR #2511, opened 2018-09-13
Add the active selection end.

sean-mcmanus:
This requires the engine changes or it'll crash (hopefully, no externals will be using master?).
Don't check this in until the engine part is done.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -63,7 +63,7 @@
     ],
     "configuration": {
       "type": "object",
-      "title": "C/C++ Configuration",
+      "title": "C/C++",
       "properties": {
         "C_Cpp.clang_format_path": {
           "type": [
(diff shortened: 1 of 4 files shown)


sean-mcmanus:
This change isn't ready yet...it's not sending the expected JSON. VS Code is sending the JSON as 2 unnamed objects with line/character (i.e. 2 positions) -- I think I can just parse the json as 2 positions instead of a Range (which doesn't work because there are no "start/end" objects being sent in the JSON).

Extension/package.json
```diff
@@ -63,7 +63,7 @@
     ],
     "configuration": {
       "type": "object",
-      "title": "C/C++ Configuration",
```
bobbrow, line 66: what's this about?

bobbrow (approved):

Extension/package.json
sean-mcmanus: It matches the other extensions in the new VS Code settings view. I can show you.

bobbrow: ok

Issue #2237, opened 2018-07-06
Add browse.path API for CustomConfigurationProviders


sherryyshi:
Type: LanguageService
**----- Input information below -----**

Currently there is no API for passing browse.path information from a "CustomConfigurationProvider". The only way to pass it in is to populate the c_cpp_properties.json file. Ideally there should be an API to allow this to be passed in memory.

chkp-roic:
Hi @bobbrow , do you know if and when this feature is planned?
I need this feature for a project and I might implement it myself or search some sort of a workaround to this.

bobbrow:
As we speak, I am working on version 2 of the API which includes this feature.  Expect an update this week with support in cpptools rolling out in September.

chkp-roic:
Thanks for the update!

bobbrow:
support for this has been committed (#2531)

sean-mcmanus:
The v2 API support has shipped with 0.19.0.

sherryyshi:
thanks!

PR #2531, opened 2018-09-18
Implement provider API v2

bobbrow:
#2237

### DIFF
--- a/Extension/src/cppTools.ts
+++ b/Extension/src/cppTools.ts
@@ -7,38 +7,98 @@
 import { CustomConfigurationProvider, Version } from 'vscode-cpptools';
 import { CppToolsTestApi, CppToolsTestHook } from 'vscode-cpptools/out/testApi';
 import { CustomConfigurationProvider1, getCustomConfigProviders, CustomConfigurationProviderCollection } from './LanguageServer/customProviders';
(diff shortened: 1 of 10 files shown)


Extension/src/cppTools.ts
```diff
@@ -7,46 +7,106 @@
...
         this.version = version;
     }
 
-    registerCustomConfigurationProvider(provider: CustomConfigurationProvider): void {
+    private addNotifyReadyTimer(provider: CustomConfigurationProvider1): void {
+        if (this.version >= Version.v2) {
+            const timeout: number = 30;
+            let timer: NodeJS.Timer = setTimeout(() => {
+                console.warn(`registered provider ${provider.extensionId} did not call 'notifyReady' within ${timeout} seconds`);
+            }, timeout * 1000);
```
sean-mcmanus, line 33: The capitalization at the start and the period at the end is not consistent with the console.warn above. Not sure if it matters.

sean-mcmanus (approved):

Extension/src/cppTools.ts
bobbrow: There were two sentences in the first one.  That's why it was doing sentence casing.  I split it into two separate messages instead.

PR #2540, opened 2018-09-21
Fixing queuing logic that stalls LS on Mac

bobbrow:
* adding additional logging for custom configuration providers
* remove an assertion causing the LS to fail to activate when invalid configuration providers attempt to register
* fixing a queuing bug that causes an indefinite hang on Mac
* removing an unhandled promise rejection

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -667,13 +667,14 @@ class DefaultClient implements Client {
                 }
             };
             
-            this.pendingRequests++;
-            console.assert(this.pendingRequests > 0);
-            if (this.pendingRequests === 1) {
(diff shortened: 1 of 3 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1147,6 +1170,12 @@ class DefaultClient implements Client {
             return Promise.resolve();
         }
 
+        let settings: CppSettings = new CppSettings(this.RootUri);
+        let out: logger.Logger = logger.getOutputChannelLogger();
+        if (settings.loggingLevel == "Debug") {
```
sean-mcmanus: tslint wants this to be === (and another occurrence on line 1128).

PR #2541, opened 2018-09-21
Update channel feature pt 2

grdowns:
Find Mac code script; Commenting; Move code into try catches correctly filter errors; Add notification on installation; Correctly grandfather in insiders users; Fix downloads not downloading fully

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -572,8 +572,12 @@ export async function renamePromise(oldName: string, newName: string): Promise<v
 }
 
 export function promptForReloadWindowDueToSettingsChange(): void {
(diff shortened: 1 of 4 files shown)


Extension/src/common.ts
```diff
@@ -176,19 +176,19 @@ export function isUri(input: any): input is vscode.Uri {
 }
 
 export function isString(input: any): input is string {
-    return typeof(input) === "string";
+    return input && typeof(input) === "string";
```
bobbrow: revert the changes in these helper functions and in `resolveVariables`. this is a result of not having the latest changes from master when you copied over your branch.

bobbrow (changes requested):

bobbrow (approved):

PR #2543, opened 2018-09-22
Update Channel pt 3

grdowns:
Instead of uninstalling then installing the extension, install and override when prompted.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -258,22 +258,24 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
         }
 
         // Install the VSIX
-        const installCommand: string = vsCodeScriptPath + ' --install-extension ' + vsixLocation;
-        try {
-            let process: ChildProcess = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation]);
-            // If downgrading, the VS Code CLI will prompt whether the user is sure they would like to downgrade
-            // Respond to this by writing 0 to stdin (the option to override and install the VSIX package)
-            let sentOverride: boolean = false;
-            process.stdout.on('data', (data: Buffer) => {
-                if (!sentOverride) {
-                    process.stdin.write('0\n');
-                    sentOverride = true;
-                }
-            });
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -249,27 +248,42 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
-            return Promise.resolve();
-        } catch (error) {
-            return Promise.reject(new Error('Failed to install VSIX'));
-        }
+
+            // Timeout the process if no response is sent back. Ensures this Promise resolves/rejects
+            const timer: NodeJS.Timer = setTimeout(() => {
+                process.kill();
+                reject(new Error('Failed to receive response from VS Code script process for installation within 1s.'));
+            }, 1000);
```
bobbrow: I think you should allow a few more seconds to this timer.  There's no real rush to kill the process.  How about 10-15 seconds?

Extension/src/LanguageServer/extension.ts
```diff
@@ -249,27 +248,42 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
-                return stdout.toString().trim();
+                try {
+                    const stdout: Buffer = execSync('which ' + vsCodeBinName);
+                    return stdout.toString().trim();
+                } catch (error) {
+                    return undefined;
+                }
             }
         }(platformInfo);
-        if (!vsCodeScriptPath) {
```
bobbrow: Why is this removed?  There is a case above that could result in vsCodeScriptPath being undefined.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: Sure!

Extension/src/LanguageServer/extension.ts
grdowns: I'll add it back in

bobbrow (approved):

PR #2559, opened 2018-09-25
Add ab testing for Go to Def (fix some ab testing bugs)

sean-mcmanus:


Extension/src/LanguageServer/client.ts
```diff
@@ -400,7 +402,8 @@ class DefaultClient implements Client {
                 default: {
                     systemIncludePath: settings.defaultSystemIncludePath
                 },
-                vcpkg_root: util.getVcpkgRoot()
+                vcpkg_root: util.getVcpkgRoot(),
+                goto_def_intelliSense: abTestSettings.UseGoToDefIntelliSense
```
bobbrow: please rename to `gotoDefIntelliSense` or similar.

Extension/src/LanguageServer/persistentState.ts
```diff
@@ -20,7 +20,11 @@ class PersistentStateBase<T> {
     }
 
     public get Value(): T {
-         return this.state.get<T>(this.key, this.defaultvalue);
+        return this.state.get<T>(this.key, this.defaultvalue);
+    }
+
+    public get DefaultValue(): T {
```
bobbrow: nit: Can you move this under the setter for `Value` to keep the two properties together?

Extension/src/LanguageServer/client.ts
bobbrow: Doing this in init works for now, but I'd imagine that at some point in the future we'll want a separate `abTestingSettingsChanged` message or something similar in the future.  That will allow us to make the settings read async again too.

bobbrow (approved):

Extension/src/LanguageServer/persistentState.ts
sean-mcmanus: Ok

PR #2570, opened 2018-09-27
Return null config to VS Code

pieandcakes:
return null to trigger VS Code to open a configuration file
https://github.com/Microsoft/vscode/issues/54213

@isidorn


WardenGnaw (approved):

isidorn:
@pieandcakes awesome, thanks for jumping on this so quickly!

PR #2576, opened 2018-09-28
Grdowns/update channel4

grdowns:
Remove redundant telemetry on failure
Check GitHub API rate limit prior to attempting to download release 
Update error message for failing to parse release JSON into type Build[]

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -617,3 +617,36 @@ export function downloadFileToDestination(urlStr: string, destinationPath: strin
         request.end();
     });
 }
+
+export function downloadFileToStr(urlStr: string, headers?: OutgoingHttpHeaders): Promise<any> {
+    return new Promise<string>((resolve, reject) => {
+        let parsedUrl: url.Url = url.parse(urlStr);
+        let request: ClientRequest = https.request({
+            host: parsedUrl.host,
+            path: parsedUrl.path,
+            agent: getHttpsProxyAgent(),
+            rejectUnauthorized: vscode.workspace.getConfiguration().get('http.proxyStrictSSL', true),
+            headers: headers
+        }, (response) => {
+            if (response.statusCode === 301 || response.statusCode === 302) { // If redirected
+                // Download from new location
+                let redirectUrl: string;
+                if (typeof response.headers.location === 'string') {
+                    redirectUrl = response.headers.location;
+                } else {
+                    redirectUrl = response.headers.location[0];
+                }
+                return resolve(downloadFileToStr(redirectUrl, headers));
+            }
+            if (response.statusCode !== 200) { // If request is not successful
+                return reject();
+            }
+            let downloadedData: string = '';
+            response.on('data', (data) => { downloadedData += data; });
+            response.on('error', (error) => { reject(); });
+            response.on('end', () => { resolve(downloadedData); });
+        });
+        request.on('error', (error) => { reject(); });
+        request.end();
+    });
+}
--- a/Extension/src/githubAPI.ts
+++ b/Extension/src/githubAPI.ts
@@ -222,45 +222,29 @@ async function rateLimitExceeded(): Promise<boolean> {
  * @return Information about the released builds of the C/C++ extension.
  */
 async function getReleaseJson(): Promise<Build[]> {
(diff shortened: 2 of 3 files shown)


Extension/src/githubAPI.ts
```diff
@@ -177,6 +177,54 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build {
...
+interface RateLimit {
+    rate: Rate;
+}
+
+function isRate(input: any): input is Rate {
+    return input && input.remaining && typeof(input.remaining) === 'number';
+}
+
+function isRateLimit(input: any): input is RateLimit {
+    return input && input.rate && isRate(input.rate);
```
bobbrow: nit: checking `input.rate` here should be unnecessary (`isRate` will do it)

Extension/src/githubAPI.ts
```diff
@@ -177,6 +177,54 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build {
...
+interface Rate {
+    remaining: number;
+}
+
+interface RateLimit {
+    rate: Rate;
+}
+
+function isRate(input: any): input is Rate {
+    return input && input.remaining && typeof(input.remaining) === 'number';
```
bobbrow: I think Sean added a `util.isNumber` function.

Extension/src/githubAPI.ts
```diff
@@ -177,6 +177,54 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build {
...
+        } catch (error) {
+            throw new Error('Failed to parse rate limit JSON');
+        }
+
+        if (isRateLimit(rateLimit)) {
+            return Promise.resolve(rateLimit);
+        } else {
+            throw new Error('Rate limit JSON is not of type RateLimit');
+        }
+    } catch (error) {
```
bobbrow, line 203: Since this is an `async` function using `await`, you shouldn't need to catch/return reject explicitly here.  Just let the exception be thrown out of this method.  (e.g. remove the outer try/catch)

Extension/src/githubAPI.ts
```diff
@@ -177,6 +177,54 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build {
...
+        return Promise.reject(error);
+    }
+}
+
+async function rateLimitExceeded(): Promise<boolean> {
+    const rateLimit: RateLimit = await getRateLimit();
+    if (rateLimit.rate.remaining <= 0) {
```
bobbrow: this is an `async` function using `await`.  You should be able to just return `rateLimit.rate.remaining` here.

bobbrow (changes requested):

Extension/src/githubAPI.ts
```diff
@@ -177,6 +177,46 @@ function getTargetBuild(builds: Build[], updateChannel: string): Build {
...
+function isRate(input: any): input is Rate {
+    return input && input.remaining && util.isNumber(input.remaining);
+}
+
+function isRateLimit(input: any): input is RateLimit {
+    return input && isRate(input.rate);
+}
+
+async function getRateLimit(): Promise<RateLimit> {
+        const header: OutgoingHttpHeaders = { 'User-Agent': 'vscode-cpptools' };
```
bobbrow: it looks like this is double-indented

Extension/src/common.ts
```diff
@@ -620,4 +620,37 @@ export function downloadFileToDestination(urlStr: string, destinationPath: strin
         request.on('error', (error) => { reject(); });
         request.end();
     });
+}
+
+export function downloadFileToStr(urlStr: string, headers?: OutgoingHttpHeaders): Promise<any> {
```
bobbrow, line 625: nifty.  Should we update the query for the release JSON to use this new method too?

bobbrow (changes requested):

Extension/src/common.ts
grdowns: Totally. I figured it was best to ask first, then I ended up forgetting :laughing:

bobbrow (approved):

bobbrow:
TSlint: `tmp` is no longer needed in the githubApi.ts import list

bobbrow (approved):

PR #2592, opened 2018-10-02
don't use falsy comparisons when resolving default settings.

shankaryoga21:


bobbrow:
This looks like my branch.  Did you test this for me?

sean-mcmanus (approved):

PR #2611, opened 2018-10-05
check before filling issues

Alokpro1:


bobbrow:
I believe this page is about contributing to the code.  This change might be better placed in the issue templates.

CONTRIBUTING.md
```diff
@@ -3,7 +3,8 @@
 ## Contribution Steps
 
 * [Build and debug the extension](Documentation/Building%20the%20Extension.md).
-* File an [issue](https://github.com/Microsoft/vscode-cpptools/issues) and a [pull request](https://github.com/Microsoft/vscode-cpptools/pulls) with the change and we will review it.
+* Before filling a new issue **check if the issue already exists in it.**
```
sean-mcmanus: Should be spelled "filing".

sean-mcmanus: I think "in it" should be removed. I don't understand what "it" is referring to.

sean-mcmanus:
Bob is referring to https://github.com/Microsoft/vscode-cpptools/tree/master/.github/ISSUE_TEMPLATE , which gets shown to users when the file a new issue.

CONTRIBUTING.md
Alokpro1: okay. Sorry for the mistake

sean-mcmanus:
@Alokpro1 The language service issue template already has ``, so I don't think we need to add this to the contributing.md.

Issue #2613, opened 2018-10-05
2 uncaught errors


areee:
- Extension Name: cpptools
- Extension Version: 0.19.0
- OS Version: Darwin x64 18.0.0
- VSCode version: 1.28.0-insider

 We have written the needed data into your clipboard. Please paste:
```json 
{
	"activationTimes": {
		"startup": false,
		"codeLoadingTime": 6582,
		"activateCallTime": 120,
		"activateResolvedTime": 479,
		"activationEvent": "onLanguage:cpp"
	},
	"runtimeErrors": [
		{
			"name": "TypeError",
			"message": "Cannot read property 'fsPath' of null"
		},
		{
			"name": "TypeError",
			"message": "Cannot read property 'fsPath' of null"
		}
	]
}
```

sean-mcmanus:
Thanks for reporting this. Yeah, I'm seeing this error when opening a file without a workspace folder. You could use Open Folder instead of open file to avoid this error. Looks like a regression from https://github.com/Microsoft/vscode-cpptools/pull/2552 .

sean-mcmanus:
Fixed with https://github.com/Microsoft/vscode-cpptools/releases/download/0.19.1-insiders/cpptools-osx.vsix .

rsthn:
I can confirm it happens on Windows as well, please add a fix to it.

    Extension Name: cpptools
    Extension Version: 0.19.0
    OS Version: Microsoft Windows [Version 10.0.17134.228]
    VSCode version: 1.28.1

PR #2614, opened 2018-10-05
Fix error when rootRui is null.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/2613 .


grdowns (approved):
I had always thought it was strange that strings are default initialized to "undefined". I suppose it's an optimization of some kind for type comparison, but man does it lead to some painful bugs.

bobbrow (approved):

Issue #2536, opened 2018-09-21
IntelliSense-based Go to Definition falls back to the previous implementation if the definition isn't in the same TU


sean-mcmanus:
This is applies to 0.19.0 or later.

sean-mcmanus:
This is fixed for the "main" scenarios, but there are subclasses that still don't give correct results, such as
1. Overloaded methods return all overloads instead of the one that exactly matches the input params (https://github.com/Microsoft/vscode-cpptools/issues/2677 ).
2. I suspect there are probably other buggy scenarios as well (e.g. templates). Let us know if you encounter any...

sean-mcmanus:
Fixed in https://github.com/Microsoft/vscode-cpptools/releases/tag/0.19.1-insiders2 .

PR #2615, opened 2018-10-06
Update the changlog for 0.19.1-insiders.

sean-mcmanus:


sean-mcmanus:
We're waiting on another debugger fix. @pieandcakes Can you add that to the changelog when it's ready?

Let me know if anyone thinks anything else should be fixed for 0.19.1-insiders; otherwise, I'm continuing work on https://github.com/Microsoft/vscode-cpptools/issues/2536 .

bobbrow (approved):

PR #2637, opened 2018-10-09
Grdowns/update channel fix

grdowns:
Removed surrounding quotes on calls to spawn on Windows and Mac
Add try/catch around spawn to override any possible exceptions
Default vsixname to cpptools-linux32.vsix
Move logFailure inline
Remove potential PII from error message to be on the safe side

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -262,8 +262,13 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
 
         // Install the VSIX
         return new Promise<void>((resolve, reject) => {
-            let process: ChildProcess = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation], {shell: true});
-            if (process.pid === undefined) {
+            let process: ChildProcess;
+            try {
+                process = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation], {shell: true});
+                if (process.pid === undefined) {
+                    throw new Error;
+                }
+            } catch (error) {
                 reject(new Error('Failed to launch VS Code script process for installation'));
                 return;
             }
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -344,7 +344,7 @@ async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
         });
     });
     await p.catch((error: Error) => {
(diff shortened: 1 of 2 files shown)


sean-mcmanus (approved):

Extension/src/githubAPI.ts
```diff
@@ -100,8 +100,8 @@ function vsixNameForPlatform(info: PlatformInformation): string {
             case 'darwin': return 'cpptools-osx.vsix';
             default: {
                 switch (platformInfo.architecture) {
-                    case 'x86': return 'cpptools-linux32.vsix';
                     case 'x86_64': return 'cpptools-linux.vsix';
+                    default: return 'cpptools-linux32.vsix';
```
bobbrow: Should we allow auto-updating when we are not 100% certain of the OS architecture?  I think I would rather fail here - just to be safe.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -262,8 +262,13 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
 
         // Install the VSIX
         return new Promise<void>((resolve, reject) => {
-            let process: ChildProcess = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation]);
-            if (process.pid === undefined) {
+            let process: ChildProcess;
+            try {
+                process = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation], {shell: true});
```
bobbrow: Does turning on `shell` result in a terminal/cmd window opening up?  I would think that the real problem here is simply that the `vsCodeScriptPath` has quotes in it and we just need to remove them.

bobbrow: I just tested this on Windows and removing the quotes from `vsCodeScriptPath` fixes the spawn issue.  It should also fix it for macOS.

Extension/src/githubAPI.ts
grdowns: Okay. If that's the case then I should add cases for `x86` and `i386`, yeah? Are there any other architecture types that we support?

grdowns: Sean let me know that the supported 32-bit architectures are listed in the `package.json`. I'll add the cases `x86`, `i386`, as well as `i686`.

bobbrow (approved):

PR #2649, opened 2018-10-11
Update changelog for 0.19.1-insiders2

sean-mcmanus:


Extension/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # C/C++ for Visual Studio Code Change Log
 
-## Version 0.19.1-insiders2: October 11th, 2018
+## Version 0.19.1-insiders2: October 12th, 2018
 * Add IntegratedTerminal support for Linux and Windows. [#35](https://github.com/microsoft/vscode-cpptools/issues/35)
 * Unify Visual Studio Code debug protocol parsing by using a shared library with Visual Studio.
+* Fix `-isystem` without a space after getting ignored in `compile_comamands.json`. [#2629](https://github.com/Microsoft/vscode-cpptools/issues/2629)
+* Fix Insiders update channel installation failure (due to `spawn`) on Windows/Mac. [#2636](https://github.com/Microsoft/vscode-cpptools/issues/2636)
```
bobbrow: not due to spawn, due to quotes in the path, but this is an implementation detail, and we don't need to say why it was broken so I would just remove it.

Extension/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # C/C++ for Visual Studio Code Change Log
 
-## Version 0.19.1-insiders2: October 11th, 2018
+## Version 0.19.1-insiders2: October 12th, 2018
```
bobbrow: remove the `th` from this entry and the other insiders entry.

bobbrow (changes requested):

bobbrow (approved):

PR #2686, opened 2018-10-17
Ensure that package version suffix + version suffix version are set

grdowns:
Ensure the package suffix is set if it's in the version string, and also default the package suffix version suffix (ugh) to 0.

### DIFF
--- a/Extension/src/packageVersion.ts
+++ b/Extension/src/packageVersion.ts
@@ -28,11 +28,11 @@ export class PackageVersion {
                 this.suffixVersion = parseInt(tokens[3].substring(firstDigitOffset));
             } else {
                 this.suffix = tokens[3];
-                this.suffixVersion = 0;
+                this.suffixVersion = 1;
             }
         } else {
-            this.suffix = tokens[3];
-            this.suffixVersion = 0;
+            this.suffix = undefined;
+            this.suffixVersion = 1;
         }
 
         if (this.major === undefined || this.minor === undefined || this.patch === undefined) {
--- a/Extension/src/packageVersion.ts
+++ b/Extension/src/packageVersion.ts
@@ -26,7 +26,13 @@ export class PackageVersion {
             if (firstDigitOffset !== -1) {
                 this.suffix = tokens[3].substring(0, firstDigitOffset);
                 this.suffixVersion = parseInt(tokens[3].substring(firstDigitOffset));
+            } else {
+                this.suffix = tokens[3];
+                this.suffixVersion = 0;
             }
+        } else {
+            this.suffix = tokens[3];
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

pieandcakes:
@bobbrow I think this warrants a bigger discussion on whether or not we want to continue with this strange naming for insiders. could we go with 0.19.1-insiders.2 ? I feel like that might be easier that a split on `.` will yield 4 and your last one is that new version

sean-mcmanus:
I don't see how 0.19.1-insiders.2 is any better or less strange than 0.19.1-insiders2. Adding the "." seems more strange to me, but I don't know of other examples to compare to. Using a `0.19.1.<build number>` where build number just increments for every "insider" release is what we do for the binaries we ship, since they don't allow arbitrary text, but that would make our public release have a version such as 0.19.1.3 instead of just 0.19.1.

Extension/src/packageVersion.ts
```diff
@@ -26,7 +26,13 @@ export class PackageVersion {
             if (firstDigitOffset !== -1) {
                 this.suffix = tokens[3].substring(0, firstDigitOffset);
                 this.suffixVersion = parseInt(tokens[3].substring(firstDigitOffset));
+            } else {
+                this.suffix = tokens[3];
+                this.suffixVersion = 0;
```
bobbrow: suffixVersion should be 1, right?

Extension/src/packageVersion.ts
```diff
@@ -26,7 +26,13 @@ export class PackageVersion {
...
             }
+        } else {
+            this.suffix = tokens[3];
```
bobbrow: should suffix be `undefined`?  (I suppose tokens[3] is already undefined in this block, but it seems cleaner to me to just say `undefined` here)

bobbrow (changes requested):

bobbrow:
@pieandcakes we know that vscode will do the right thing for `#.#.#-S` versions.  I suppose we could test out whether adding another dot messes this up or not, but I'd rather stick with the current formatting which I think is fine.

Extension/src/packageVersion.ts
grdowns: Yeah, definitely. Whether it's 1 or 0 will not affect version comparison (as we only apply suffices with a number of 2 or higher), but it's best to be technically correct in case we need to depend on that logic in the future.

Extension/src/packageVersion.ts
grdowns: Agreed

bobbrow (approved):

sean-mcmanus:
Is this ready to check in?

Issue #2656, opened 2018-10-12
Request to divide into two separate: IntelliSense and Debugging


dmyger:
Is it possible split current extension in two separate: one for language service another for debugging?

As for me, there is a small issue: no matter how hard I tried, I couldn't make C++ IntelliSense works as I want for my projects, so I just don't use all features of the IntelliSense. I have to set 'disable' settings wherever it's possible, to avoid interfering with cquery, which satisfies me completely. But actually it's possible only if disable extension at all.

On other hand, your extension has very good features for debugging and I prefer use it. So I need to enable back the extension before debugging. This case is not very convenient to use.

As for me, it seems that would be more useful to have two extensions with clearly separated functionality. What do you think about it?

pieandcakes:
@bobbrow Do we have a way to disable Intellisense for @dmyger  ?

pieandcakes:
@dmyger We are trying to provide a complete experience, especially for new users. If you are having issues with Intellisense and would like to report issues, they can look into your scenario and see if they can resolve the problems you are running into.

sean-mcmanus:
We already have a "Disabled" setting for "C_Cpp.intelliSenseEngine". Does that work for you?

dmyger:
I have following disables:
```json
"C_Cpp.intelliSenseEngine": "Disabled",
"C_Cpp.autocomplete": "Disabled",
"C_Cpp.formatting": "Disabled",
"C_Cpp.errorSquiggles": "Disabled",
"C_Cpp.intelliSenseEngineFallback": "Disabled",
```
And when I click mouse on function with Ctrl pressed (`Go to Definition`) I get two offers one from cquery another one from "disabled" IntelliSense, both are in the same file and the same line. But, more worst situation when I try do it with any variable I get a lot of definitions wherever it find variables with same name, instead of completely accurate navigation to the desired definition of a single variable as it works when the IntelliSense extension disabled at all.

bobbrow:
Changing the `intelliSenseEngine` setting requires you to reload the VS Code window to take effect.  Did you reload the window and the language service is still running?

dmyger:
Of course, I'm reloading a window and doing full restart of application, too.
But I don't said that _"language service is still running"_, I said that some of its features are keep working.

sean-mcmanus:
[...]

PR #2692, opened 2018-10-18
Don't allow configuration providers to activate the extension when IntelliSense is disabled

bobbrow:
#2656

### DIFF
--- a/Extension/src/LanguageServer/customProviders.ts
+++ b/Extension/src/LanguageServer/customProviders.ts
@@ -4,8 +4,9 @@
  * ------------------------------------------------------------------------------------------ */
 'use strict';
 
-import { CustomConfigurationProvider, Version, SourceFileConfigurationItem, WorkspaceBrowseConfiguration } from 'vscode-cpptools';
 import * as vscode from 'vscode';
+import { CustomConfigurationProvider, Version, SourceFileConfigurationItem, WorkspaceBrowseConfiguration } from 'vscode-cpptools';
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/customProviders.ts
```diff
@@ -135,6 +136,11 @@ export class CustomConfigurationProviderCollection {
     }
 
     public add(provider: CustomConfigurationProvider, version: Version): boolean {
+        if (new CppSettings().intelliSenseEngine === "Disabled") {
```
grdowns, line 139: Should the user of this function already have determined whether IntelliSense is disabled before preparing their `CustomConfigurationProvider`? If that is the case, perhaps an exception would be more apt

bobbrow: Even if they should, it doesn't mean they will, so we need to protect against it.

bobbrow: I think I had a different context in mind when responding.  The sole caller of this function is the `registerCustomConfigurationProvider` method which is invoked by the provider.  I don't think we should expect the providers to check our settings before registering, but I do think that we should protect against adding providers to the collection and since this is the method that does that, this is where the check should be made (rather than requiring every potential caller to know that they have to make this check)

pieandcakes (approved):

grdowns (approved):

sean-mcmanus (approved):

PR #2693, opened 2018-10-18
Don't update pre-releases

grdowns:
Do not try to update if the current version is greater than the latest release or incomparable

### DIFF
--- a/Extension/src/githubAPI.ts
+++ b/Extension/src/githubAPI.ts
@@ -131,6 +131,10 @@ export interface BuildInfo {
 export async function getTargetBuildInfo(updateChannel: string): Promise<BuildInfo> {
     return getReleaseJson()
         .then(builds => {
+            if (!builds || builds.length === 0) {
+                return undefined;
+            }
+
             // If the user version is greater than or incomparable to the latest available verion then there is no need to update
             // Allows testing pre-releases without accidentally downgrading to the latest version
             const userVersion: PackageVersion = new PackageVersion(util.packageJson.version);
@@ -159,6 +163,7 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
 /**
  * Determines whether there exists a Build in the given Build[] that should be installed.
  * @param builds The GitHub release list parsed as an array of Builds.
+ * @param userVersion The verion of the extension that the user is running.
  * @param updateChannel The user's updateChannel setting.
  * @return The Build if the user should update to it, otherwise undefined.
  */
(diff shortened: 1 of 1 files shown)


Extension/src/githubAPI.ts
```diff
@@ -130,7 +130,17 @@ export interface BuildInfo {
  */
 export async function getTargetBuildInfo(updateChannel: string): Promise<BuildInfo> {
     return getReleaseJson()
-        .then(builds => getTargetBuild(builds, updateChannel))
+        .then(builds => {
+            // If the user version is greater than or incomparable to the latest available verion then there is no need to update
+            // Allows testing pre-releases without accidentally downgrading to the latest version
+            const userVersion: PackageVersion = new PackageVersion(util.packageJson.version);
+            const latestVersion: PackageVersion = new PackageVersion(builds[0].name);
+            if (userVersion.isGreaterThan(latestVersion) || (userVersion.suffix && userVersion.suffix !== 'insiders')) {
```
pieandcakes, line 142: might want to do a starts with 'insiders' or contains 'insiders' since we have the insiders2 problem right now.

pieandcakes (dismissed):

Extension/src/githubAPI.ts
bobbrow: for insiders2, the 2 will be removed and `.suffix` will just have `"insiders"` and `.suffixVersion` will have 2

bobbrow, line 142: You don't need to check `userVersion.suffix` before checking `userVersion.suffix !== 'insiders'`

Extension/src/githubAPI.ts
```diff
@@ -130,7 +130,17 @@ export interface BuildInfo {
  */
 export async function getTargetBuildInfo(updateChannel: string): Promise<BuildInfo> {
     return getReleaseJson()
-        .then(builds => getTargetBuild(builds, updateChannel))
+        .then(builds => {
+            // If the user version is greater than or incomparable to the latest available verion then there is no need to update
+            // Allows testing pre-releases without accidentally downgrading to the latest version
+            const userVersion: PackageVersion = new PackageVersion(util.packageJson.version);
+            const latestVersion: PackageVersion = new PackageVersion(builds[0].name);
```
bobbrow, line 141: is it safe to just pass in `build[0]` here?

bobbrow (changes requested):

Extension/src/githubAPI.ts
grdowns: I believe that would change the behavior to make it such that users on a release version would never be able to update. Is that right?

grdowns: I can test this

grdowns: Yeah, that's the case. I'll leave this in

bobbrow: Sorry, my original comment was bogus.

bobbrow (approved):

PR #2772, opened 2018-11-06
Add avoidWindowsConsoleRedirection and fix externalConsole comment

pieandcakes:


Extension/package.json
```diff
@@ -801,7 +801,12 @@
               },
               "externalConsole": {
                 "type": "boolean",
-                "description": "If true, a console is launched for the debuggee. If false, no console is launched. Note this option is ignored in some cases for technical reasons.",
+                "description": "If true, a console is launched for the debuggee. If false, on Linux and Windows, it will appear in the Integrated Console.",
+                "default": false
+              },
+              "avoidWindowsConsoleRedirection": {
```
WardenGnaw, line 807: You will want to add this to the Extension\tools\OptionsSchema.json and run `npm run generateOptionsSchema`

WardenGnaw (changes requested):
🙅‍♂️

WardenGnaw (approved):

Extension/package-lock.json
```diff
@@ -447,7 +447,7 @@
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
-      "integrity": "sha512-L3hKV5R/p5o81R7O02IGnwpDmkp6E982XhtbuwSe3O4qOtMMMtodicASA1Cny2U+aCXcNpml+m4dPsvsJ3jatg==",
+      "integrity": "sha1-NgSLv/TntH4TZkQxbJlmnqWukfE=",
```
pieandcakes, line 450: https://github.com/npm/npm/issues/16938#issuecomment-339863980

Extension/package.json
bobbrow: I see this PR is already committed, but I wanted to comment on the naming of the setting.  The default behavior is a double negative: "don't avoid console redirection."  I think default behaviors should not use double negatives.  A preferred naming would be along the lines of `"useWindowsConsoleRedirection": true` so if people wanted to "disable" the feature they set it to false, instead of true.  Just a minor semantic thing, but I think it makes everything more clear for the user.

Extension/package-lock.json
bobbrow: What does this mean?  Sean and I have some bad npm state?

pieandcakes: No. there is an issue in npm that generates sha1 and sha512, but it keeps swapping it out. Andrew told me to update my npm and clean out my npm cache and repo and then reinit it and it fixed it. I was adding the "fix" comment from npm so I don't lose the history of what I found.

Extension/package.json
pieandcakes: @bobbrow its on by default and not shown in the configuration. The only time we want users to override it is for one off cases where we should have caught ourselves and not redirected. This isn't something widely used but is offered as a workaround.

bobbrow: The default value is `false`, so it appears this is "off" by default.  Am I missing something?

Do we own the naming of this setting, or does VS Code?  Regardless of whether it's widely used or not, using double negatives is still undesirable.

pieandcakes: Sorry, I guess I wasn't clear. This is to turn off MIEngine adding console redirection on Windows. It is a flag we set and it is "enabled" by default. I wanted a feature where the user could disable it. I guess "avoid" wasn't the right word and "disable" might be better. 

the default is "allow Windows console redirection".

bobbrow: "disable" is the same as "avoid".  Both result in a double negative.  "allowWindowsConsoleRedirection" is a better name and you would set the default value to `true`

pieandcakes: At the end of the day I wanted someone to be able to flip a switch "on" to enable the "disable" feature. I don't want people to see that it is on and have to add a setting in their launch.json and set it to false to turn it off. If I were to go with your option, I would feel compelled to add the setting to the default launch.json and turn it "on" by default because if I put my end user hat on, if a feature is on, I want to know its on and turn it off. My wording is to show that this "disable" feature is off and if you want to disable it, then you can turn it on.

At the end of the day we're going to disagree on this one.

Issue #2822, opened 2018-11-14
powershell.exe -Command [System.Security.Principal.WindowsIdentity]::GetCurrent().Name


BoskiDialer:
Not sure if it is "C/C++ 0.20.1" issue or "vs code" issue, but this issue only happens when i have "c/c++ 0.20.1" enabled

When using VS Code (ver 1.29.0) with C/C++ (ver 0.20.1), during working with project (changing files, saving files), number of powershell (i am not using it! only editing files) processes increases about 1 per 1~5 minutes (depends on user usage). Number of this processes is not bounded and after couple of hours RAM in my PC depleted (around 100 powershell processes, each using 30MB of RAM)

```
C:\Users\BoskiDialer>ver
Microsoft Windows [Wersja 6.1.7601]
```
-> Windows 7 Pro

```
C:\Users\BoskiDialer>code --status

Version:          Code 1.29.0 (5f24c93878bd4bc645a4a17c620e2487b11005f9, 2018-11
-12T07:47:15.448Z)
OS Version:       Windows_NT x64 6.1.7601
CPUs:             Intel(R) Core(TM) i7-2640M CPU @ 2.80GHz (4 x 2793)
Memory (System):  7.92GB (3.48GB free)
VM:               0%
Screen Reader:    no
Process Argv:
GPU Status:       2d_canvas:                    enabled
                  checker_imaging:              disabled_off
                  flash_3d:                     enabled
                  flash_stage3d:                enabled
                  flash_stage3d_baseline:       enabled
                  gpu_compositing:              enabled
                  multiple_raster_threads:      enabled_on
                  native_gpu_memory_buffers:    disabled_software
                  rasterization:                unavailable_software
                  video_decode:                 enabled
                  video_encode:                 enabled
                  webgl:                        enabled
                  webgl2:                       enabled
CPU %   Mem MB     PID  Process
    0       87    3744  code main
    0       85    1788     shared-process
    0      145    2180     gpu-process
    0      186    7472     window ([REDACTED] - [REDACTED] - Visual Studio Cod
e)
    0       63    3024       extensionHost
    0       56    1564         C:/windows/system32/windowspowershell/v1.0/powers
hell.exe -Command [System.Security.Principal.WindowsIdentity]::GetCurrent().Name

    0       57    5396         C:/windows/system32/windowspowershell/v1.0/powers
hell.exe -Command [System.Security.Principal.WindowsIdentity]::GetCurrent().Name

[...]

PR #2828, opened 2018-11-15
Update telemetry API.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/2822 .


sean-mcmanus:
I think the 1k limit bug will be okay for our next Insiders due to the additional filtering of removing the line numbers and the __1::. If not, we can revert it back for the following release. This will at least get the people hitting this off of 0.20.0.

pieandcakes (approved):

bobbrow (approved):

PR #2871, opened 2018-11-27
Seanmcm/fix insiders debugging

sean-mcmanus:


Extension/.vscode/tasks.json
```diff
@@ -5,7 +5,6 @@
     "tasks": [
         {
             "label": "TypeScript Compile",
-            "identifier": "compile",
```
sean-mcmanus, line 8: This was causing debugging to not work with VS Code Insiders.

WardenGnaw (approved):

PR #2872, opened 2018-11-28
bump up the timeout for installing the insiders vsix

bobbrow:


sean-mcmanus (approved):
Yeah, this is the top failure.

bobbrow:
Yeah, that's why I did it.

PR #2931, opened 2018-12-15
Attempt to fix download failure.

sean-mcmanus:
This explains why we have 1000 users with "Failed to download VSIX" results.

The current users of VS Code Insiders won't be able to auto-upgrade to 0.21.0-insiders until they get this fix in the non-Insiders release.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -327,11 +327,36 @@ async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
                 // then the .catch call will return a resolved promise
                 // Thusly, the .catch call must also throw, as a return would simply return an unused promise
                 // instead of returning early from this function scope
+                let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
+                let originalProxySupport: string = config.inspect<string>('http.proxySupport').globalValue;
+                while (true) { // Might need to try again with a different http.proxySupport setting.
+                    try {
+                        await util.downloadFileToDestination(buildInfo.downloadUrl, vsixPath)
+                            .catch(() => { throw new Error('Failed to download VSIX package'); });
+                    } catch (error) {
+                        // Try again with the proxySupport to "off".
+                        let currentProxySupport = config.inspect<string>('http.proxySupport').globalValue;
+                        if (currentProxySupport !== originalProxySupport) {
+                            config.update('http.proxySupport', originalProxySupport, true); // Reset the http.proxySupport.
+                            reject(error); // Changing the proxySupport didn't help.
+                            return;
+                        }
+                        if (config.get('http.proxySupport') !== "off" && originalProxySupport !== "off") {
+                            config.update('http.proxySupport', "off", true);
+                            continue;
+                        }
+                        reject(error);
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
This change is not working as expected. The 2nd config.update isn't working. Also, when I Launch Extension the http.request is no longer failing for me.

sean-mcmanus:
There's not much point in pushing this change for Insiders, because the affected users won't be able to auto-install the fix. The VS Code people are saying they think it's a bug with their stuff, but they're not able to repro it. When we release our Insiders, we should be able to see how many download success versus failures we get in telemetry.

Extension/src/LanguageServer/extension.ts
```diff
@@ -327,11 +327,35 @@ async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
                 // then the .catch call will return a resolved promise
                 // Thusly, the .catch call must also throw, as a return would simply return an unused promise
                 // instead of returning early from this function scope
+                let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
+                let originalProxySupport: string = config.inspect<string>('http.proxySupport').globalValue;
+                while (true) { // Might need to try again with a different http.proxySupport setting.
```
sean-mcmanus, line 332: I don't believe this can infinite loop...I could add some extra checks in case though.

sean-mcmanus:
This change shouldn't be noticeable until a 0.21.0-insiders3 or 0.22.0-insiders.

Extension/src/LanguageServer/extension.ts
```diff
@@ -327,11 +327,35 @@ async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
...
+                    try {
+                        await util.downloadFileToDestination(buildInfo.downloadUrl, vsixPath)
+                            .catch(() => { throw new Error('Failed to download VSIX package'); });
```
WardenGnaw: Why are you catching the promise and throwing a new error just to catch it in the try catch block?
You can remove the `.catch(() => { throw new Error('') });` and create the error object in the catch block on line 336.

Extension/src/LanguageServer/extension.ts
```diff
@@ -327,11 +327,35 @@ async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
...
+                        telemetry.logLanguageServerEvent('installVsix', { 'error': "Success with http.proxySupport off", 'success': 'true' });
+                    }
+                    break;
+                }
                 try {
-                    await util.downloadFileToDestination(buildInfo.downloadUrl, vsixPath)
-                        .catch(() => { throw new Error('Failed to download VSIX package'); });
                     await installVsix(vsixPath, updateChannel)
-                        .catch((error: Error) => { throw error; });
+                            .catch((error: Error) => { throw error; });
```
WardenGnaw: Also not needed.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Okay, I was just following the pattern used by the existing code. I'll try changing it.

WardenGnaw (approved):

Issue #2943, opened 2018-12-19
The default VS Code wordBasedSuggestions setting causes incorrect completions


sean-mcmanus:
With the default VS Code settings (i.e. editor.wordBasedSuggestions set to "true"), type:

int some_unrelated_variable_with_std_in_the_middle;
namespace std`<enter>`

Bug: "std" gets annoyingly auto-completed to "some_unrelated_variable_with_std_in_the_middle". We complained to VS Code a year ago (https://github.com/Microsoft/vscode/issues/31075), but they didn't provide a solution.

sean-mcmanus:
Fixed with our latest [0.21.0-insiders3](https://github.com/Microsoft/vscode-cpptools/releases). Setting `C_Cpp.updateChannel` to `"Insiders"` is another way to get the update.

PR #2944, opened 2018-12-19
Disable wordBasedSuggestions by default.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/2943 .

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1367,10 +1367,10 @@
     },
     "configurationDefaults": {
       "[cpp]": {
-        "editor.autoIndent": false
+        "editor.wordBasedSuggestions": false
       },
       "[c]": {
-        "editor.autoIndent": false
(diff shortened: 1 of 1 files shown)


Extension/package.json
```diff
@@ -1367,10 +1367,10 @@
     },
     "configurationDefaults": {
       "[cpp]": {
-        "editor.autoIndent": false
```
sean-mcmanus, line 1370: This hasn't been need over a year (VS Code fixed a bug with this).

WardenGnaw (approved):

Extension/package.json
```diff
@@ -1367,10 +1367,10 @@
...
+        "editor.wordBasedSuggestions": false
```
sean-mcmanus, line 1370: Users who really want this can opt-in.

PR #3065, opened 2019-01-22
Fix CLI mismatch for downgrading

grdowns:
The September update for VS Code added a CLI change that breaks downgrading. Check user's VS Code/-Insiders version then use the appropriate install command

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -252,7 +252,7 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
                 return path.join(vsCodeExeDir, 'bin', cmdFile);
             } else if (platformInfo.platform === 'darwin') {
                 return path.join(process.execPath, '..', '..', '..', '..', '..',
-                'Resources', 'app', 'bin', 'code');
+                    'Resources', 'app', 'bin', 'code');
             } else {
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -266,7 +267,25 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+        if (userVersion.isGreaterThan(breakingVersion, 'insider')) {
+            return new Promise<void>((resolve, reject) => {
+                let process: ChildProcess;
+                try {
+                    process = spawn(vsCodeScriptPath, ['--install-extension', vsixLocation, '--force']);
+                    if (process.pid === undefined) {
+                        throw new Error();
+                    }
+                } catch (error) {
+                    reject(new Error('Failed to launch VS Code script process for installation'));
```
bobbrow, line 282: Let's update the telemetry message to include the version number so we can differentiate

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: Talked offline. This telemetry only reports on whether the child process was spawned, not whether the supplied arguments were correct and the command succeeded. Therefore there is no need to split the code as it depends only on whether the executable's path was found

bobbrow (approved):

Issue #3100, opened 2019-01-26
Show a Warning message if the C/C++ Clang Command Adapter extension is enabled and C_Cpp.intelliSenseEngine is not Disabled


sean-mcmanus:
30 or so users have reported error squiggles coming from the C/C++ Clang Command Adapter -- hopefully, showing this warning will help avoid the confusion?

sean-mcmanus:
We might re-consider this later...

PR #3101, opened 2019-01-26
Seanmcm/0 22 0 changelog1

sean-mcmanus:
Show a warning for the C/C++ Clang Command Adapter: fix for https://github.com/Microsoft/vscode-cpptools/issues/3100 .

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -109,10 +109,10 @@ function realActivation(): void {
     if (new CppSettings().intelliSenseEngine === "Disabled") {
         throw new Error("Do not activate the extension when IntelliSense is disabled.");
     } else {
-        let clangCommandAdapterEnabled: boolean = vscode.extensions.all.some((extension: vscode.Extension<any>, index: number, array: vscode.Extension<any>[]): boolean => {
+        let clangCommandAdapterActive: boolean = vscode.extensions.all.some((extension: vscode.Extension<any>, index: number, array: vscode.Extension<any>[]): boolean => {
             return extension.isActive && extension.id === "mitaki28.vscode-clang";
         });
-        if (clangCommandAdapterEnabled) {
+        if (clangCommandAdapterActive) {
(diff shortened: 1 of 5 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -108,6 +108,13 @@ function onActivationEvent(): void {
 function realActivation(): void {
     if (new CppSettings().intelliSenseEngine === "Disabled") {
         throw new Error("Do not activate the extension when IntelliSense is disabled.");
+    } else {
+        let clangCommandAdapterActive: boolean = vscode.extensions.all.some((extension: vscode.Extension<any>, index: number, array: vscode.Extension<any>[]): boolean => {
+            return extension.isActive && extension.id === "mitaki28.vscode-clang";
+        });
+        if (clangCommandAdapterActive) {
+            vscode.window.showWarningMessage("Disable the C/C++ Clang Command Adapter extension or set C_Cpp.intelliSenseEngine to Disabled.");
```
bobbrow: I don't think we should call out any one extension.  Let's gather telemetry on how big a problem this is first.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -105,9 +105,19 @@ function onActivationEvent(): void {
     activatedPreviously.Value = true;
 }
 
+const conflictingExtensionsStr: string = "CPP." + util.packageJson.version + ".ConflictingExtensions";
+
 function realActivation(): void {
     if (new CppSettings().intelliSenseEngine === "Disabled") {
         throw new Error("Do not activate the extension when IntelliSense is disabled.");
+    } else if (util.extensionContext.globalState.get<boolean>(conflictingExtensionsStr, true)) {
```
bobbrow: Can you use the `PersistentState<T>` class instead of raw calls?

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Sure.

bobbrow (approved):

PR #3118, opened 2019-01-30
Grdowns/build tasks

grdowns:
Generate build tasks for the user which will compile the active file and output a binary in the same directory
The build tasks are generated based off of compilers found on the user's machine, as well as their compilerPath setting


Extension/src/LanguageServer/extension.ts
```diff
@@ -82,6 +99,113 @@ export function activate(activationEventOccurred: boolean): void {
...
+    }
+
+    if (!compilerPaths) {
+        telemetry.logLanguageServerEvent('noCompilerFound'); // Don't prompt a message yet until we can make a data-based decision
+        // Display a message prompting the user to install compilers if none were found
+        // const dontShowAgain: string = "Don't Show Again";
+        // const learnMore: string = "Learn More";
+        // const message: string = "No C/C++ compiler found on the system. Please install a C/C++ compiler to use the C/Cpp: build active file tasks.";
+
+        // let showNoCompilerFoundMessage: PersistentState<boolean> = new PersistentState<boolean>("CPP.showNoCompilerFoundMessage", true);
```
pieandcakes, line 174: is this commented out for a reason? or is that what that first line of the comment is explaining? If so, i'd put it on a new line so its less confusing

Extension/src/LanguageServer/extension.ts
```diff
@@ -82,6 +99,113 @@ export function activate(activationEventOccurred: boolean): void {
...
+        exeName = '${fileBasenameNoExtension}.exe';
+    } else {
+        exeName = '${fileBasenameNoExtension}';
+    }
+
+    // Generate tasks
+    let result: vscode.Task[] = [];
+    compilerPaths.forEach(compilerPath => {
+        const taskName: string = path.basename(compilerPath) + " build active file";
+        const args: string[] = ['-g', '${fileDirname}/${fileBasename}', '-o', '${fileDirname}/' + exeName];
```
pieandcakes: To avoid accidental collisions, should the output exes be put in a \bin folder inside the directory?

pieandcakes: Also, to be easier on the user, could we do `<command> -g *.<whatever their file extension choice is such as cpp> -o <filename>.<exe>` ?

grdowns: As in attempt to compile all files within the current working directory? And how would they select the file extension of choice through the quickpick menu? I'm not sure I understand

grdowns: Not a bad idea. Another thought is putting it into the `.vscode` folder. @bobbrow?

pieandcakes: Based on the file extension of the currently opened file? Or chain all of them? I don't imagine beginning users who would use this to use .c .cpp and .cxx all together in the same project.

pieandcakes: I would be against `.vscode` as that seems to be where configuration and extension data is going and not anything to do with the customer's coding experience. For me it would be between `/bin` or `/out`.

Extension/src/LanguageServer/extension.ts
```diff
@@ -83,6 +100,117 @@ export function activate(activationEventOccurred: boolean): void {
...
+            type: 'shell',
+            label: taskName,
+            command: compilerPath,
+            args: args,
+        };
+
+        const command: vscode.ShellExecution = new vscode.ShellExecution(compilerPath, args, { cwd: path.dirname(compilerPath) });
+        const target: vscode.WorkspaceFolder = vscode.workspace.getWorkspaceFolder(clients.ActiveClient.RootUri);
+        let task: vscode.Task = new vscode.Task(kind, target, taskName, 'C/Cpp', command, '$gcc');
+        task.definition = kind; // The constructor for vscode.Task will eat the definition. Reset it by reassigning
```
sean-mcmanus: I'd prefer more usage of periods at the end of comments...not sure if we're enforcing that.

Extension/src/LanguageServer/extension.ts
```diff
@@ -83,6 +100,119 @@ export function activate(activationEventOccurred: boolean): void {
...
+ * Generate tasks to build the current file based on the user's detected compilers, the user's compilerPath setting, and the current file's extension.
+ */
+async function getBuildTasks(): Promise<vscode.Task[]> {
+    const activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
+    if (!activeEditor) {
+        return [];
+    }
+
+    // Don't offer tasks for header files.
+    const activeFileExt: string = path.extname(activeEditor.document.fileName);
```
sean-mcmanus: This code is case-sensitive, right? It seems like it needs to handle both cases then. For example, .H should count as a header file too, and .C is supposed to be a C++ file.

Also, extensionless files should count as header files, if that's possible.

Extension/src/LanguageServer/extension.ts
```diff
@@ -83,6 +100,119 @@ export function activate(activationEventOccurred: boolean): void {
...
+
+    // Don't offer tasks for header files.
+    const activeFileExt: string = path.extname(activeEditor.document.fileName).toLowerCase();
+    const isHeader: boolean = [".hpp", ".hh", ".hxx", ".h", ""].some(ext => activeFileExt === ext);
+    if (isHeader) {
+        return [];
+    }
+
+    // Don't offer tasks if the active file's extension is not a recognized C/C++ extension.
+    const activeFileIsCpp: boolean = [".cpp", ".cc", ".cxx", ".mm", ".ino", ".inl"].some(ext => activeFileExt === ext);
```
bobbrow: if this is case-sensitive, consider dealing with `.C` (which is c++, by convention).  Perhaps we should offer both in this case.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -46,6 +46,7 @@ export interface ConfigurationJson {
 export interface Configuration {
     name: string;
     compilerPath?: string;
+    compilerInfo?: CompilerInfo[];
```
bobbrow: I usually use plurals in the variable names for arrays/lists.  e.g. consider renaming to `'compilers` or `knownCompilers` (`compilerInfos` sounds weird)

Extension/src/LanguageServer/extension.ts
```diff
@@ -128,7 +128,7 @@ async function getBuildTasks(): Promise<vscode.Task[]> {
         activeFileLanguage = 'cpp';
     } else if (activeFileExt === '.c') {
         activeFileLanguage = 'c';
-    } else if (activeFileExt !== '.C') { // Ambiguous file extension
+    } else if (activeFileExt !== '.C') { // Ambiguous file extension.
```
bobbrow: ambiguous means show tasks for both c and c++

Extension/src/LanguageServer/extension.ts
```diff
@@ -83,6 +100,128 @@ export function activate(activationEventOccurred: boolean): void {
...
+        }
+
+        compilerPaths = [...map.values()];
+    } else if (userCompilerPath) {
+        compilerPaths = [userCompilerPath];
+    }
+
+    if (!compilerPaths) {
+        // Don't prompt a message yet until we can make a data-based decision.
+        telemetry.logLanguageServerEvent('noCompilerFound');
```
bobbrow, line 168: can we call it "buildTaskNoCompiler"?

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: I updated the comment to reflect that's the case

bobbrow (approved):

Issue #3121, opened 2019-01-31
Unable to set non-recursive includePath in workspace root


mill1000:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10 (Windows_NT x64 10.0.17134)
- VS Code Version: 1.30.2
- C/C++ Extension Version: 0.21.0
- Other extensions you installed (and if the issue persists after disabling them): None

I am unable to set a non-recursive includePath to the workspace root. I wish to add specific folders to the includePath, and exclude others. However since they are all subfolders of the workspace root, they are automatically included. It appears the extension always adds the workspace root as a recursive includePath.

Example c_cpp_properties.json
```
{
  "configurations": [
      {   
          "name": "MinGW",
          "intelliSenseMode": "gcc-x64",
          "compilerPath": "C:/msys32/opt/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc.exe",
          "includePath": [
              "${workspaceFolder}/Sub/Include/**",
              "${workspaceFolder}/Main/**",
              "${workspaceFolder}/*"
          ],
          "defines": [
            "__GNUC__=5"
          ],
          "cStandard": "c11",
          "cppStandard": "c++14"
      }
  ],
  "version": 4
}
```

The C/C++ log shows the root folder is included recursively.
```
Code browsing service initialized
Populate include completion cache.
Discovering files...
  Processing folder (recursive): C:/MSYS32/OPT/XTENSA-ESP32-ELF/LIB/GCC/XTENSA-ESP32-ELF/5.2.0/INCLUDE/
  Processing folder (recursive): C:/MSYS32/OPT/XTENSA-ESP32-ELF/LIB/GCC/XTENSA-ESP32-ELF/5.2.0/INCLUDE-FIXED/
  Processing folder (recursive): C:/MSYS32/OPT/XTENSA-ESP32-ELF/XTENSA-ESP32-ELF/INCLUDE/
  Processing folder (recursive): C:/MSYS32/OPT/XTENSA-ESP32-ELF/XTENSA-ESP32-ELF/SYSROOT/USR/INCLUDE/
  Processing folder (recursive): C:/PATH/TO/ROOT/
  Discovering files: 1060 file(s) processed
  0 file(s) removed from database
Done discovering files.
Parsing remaining files...
  Parsing: 3 files(s) processed
Done parsing remaining files.
```
**Expected behavior**
[...]

PR #3122, opened 2019-01-31
Fix regex to allow non-recursive includePath

mill1000:
Allow non-recursive ${workspaceRoot} or ${workspaceFoldeR} in includePath when constructing the browse.path from the includePath. Solves #3121.

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -1,5 +1,8 @@
 # C/C++ for Visual Studio Code Change Log
 
+## Pull Request
+* Allow the workspace root to be a non-recursive includePath. [#3121](https://github.com/Microsoft/vscode-cpptools/issues/3121)
+
 ## Version 0.22.0-insiders February 5, 2019
 * Fix signature help active parameter selection when parameter names are missing or subsets of each other. [#2952](https://github.com/Microsoft/vscode-cpptools/issues/2952)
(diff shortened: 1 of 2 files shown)


Extension/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # C/C++ for Visual Studio Code Change Log
 
+## Pull Request
+* Allow the workspace root to be a non-recursive includePath. [#3121](https://github.com/Microsoft/vscode-cpptools/issues/3121)
```
sean-mcmanus: Can you move this under the 0.22.0-insiders section below?

Can you re-word it to say something like ```"Allow `*` in `includePath` to apply to `browse.path`"```? Since non-recursive includPath were already supported, they just didn't auto-copy to browse.path.

Also, you can also add your "credit" line, such as
```* Tucker Kern (@mill1000) [PR #3122](https://github.com/Microsoft/vscode-cpptools/pull/3122)```

Or we can make these changelog changes ourselves later on (which is we've done in the past with other submissions).

sean-mcmanus (changes requested):
Seems good to me.

Extension/CHANGELOG.md
mill1000: Yeah not a problem

sean-mcmanus (approved):

PR #3146, opened 2019-02-08
Update extend version.

sean-mcmanus:


WardenGnaw (approved):

pieandcakes:
It looks like you are adding the module `extend`.

sean-mcmanus:
@pieandcakes It's used by one of our dependencies. Do you know another way to make our dependencies use a higher version?

Issue #3176, opened 2019-02-14
c_cpp_properties.json is blank after Edit Configuration if it's invoked too soon


sean-mcmanus:
1. Open a new folder (without a c_cpp_properties.json) and then do "Edit Configuration". I'm not sure what the timing is exactly, but I've hit this at least 2 times under "normal usage" of the extension.

Bug: Depending on your timing you may hit a command not found (https://github.com/Microsoft/vscode-cpptools/issues/3160) or you get a blank config:

![image](https://user-images.githubusercontent.com/19859882/52810769-3070fb00-3048-11e9-9c7a-1a64b8132495.png)

Issue #3160, opened 2019-02-13
command 'C_Cpp.GoToDeclaration' not found


rivergold:
Type: General

**Describe the bug**
- OS and Version: Ubuntu 18.04
- VS Code Version: 1.30.2
- C/C++ Extension Version: 0.21.0
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is.
    When I use `ctrl` + `F12` to go to a declaration, occur error command 'C_Cpp.GoToDeclaration' not found

I have tried other lower C/C++ Extension Version, it can run correctly. Is there something wrong in version 0.21.0?

sean-mcmanus:
Our process is probably failing to start. I assume other C_Cpp commands also fail? We haven't gotten other reports of 0.21.0 failing yet. Your machine might be lacking some runtime shared library our Microsoft.VSCode.CPP.Extension.linux process requires. Are you sure you have 0.21.0 and not 0.22.0-insiders? Does 0.22.0-insiders fail too? When you do Help->Toggle Developer Tools does it show any errors in the console? Can you launch ~/.vscode/extesnsions/ms-vscode-cpptools-0.21.0/bin/Microsoft.VSCode.CPP.Extension.linux ?

Are you missing any of these shared libraries?
```ldd Microsoft.VSCode.CPP.Extension.linux
	linux-vdso.so.1 =>  (0x00007ffee7b73000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2ca3136000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2ca2f18000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2ca2b4f000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2ca294b000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f2ca2743000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2ca252b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2ca343c000)
```

rivergold:
There are some reasons caused the program failed to start. After I uninstall the extension, restart the computer and reinstall the extension, it can work……

sean-mcmanus:
[...]

PR #3178, opened 2019-02-15
Fix command not found and empty c_cpp_properties.json if activation is too slow.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3160 and https://github.com/Microsoft/vscode-cpptools/issues/3176.


Extension/src/LanguageServer/client.ts
```diff
@@ -321,6 +321,10 @@ class DefaultClient implements Client {
                     // The event handlers must be set before this happens.
                     languageClient.sendRequest(QueryCompilerDefaultsRequest, {}).then((compilerDefaults: configs.CompilerDefaults) => {
                         this.configuration.CompilerDefaults = compilerDefaults;
+                        
+                        // Only register the real commands after the extension has finished initializing,
+                        // e.g. prevents empty c_cpp_properties.json from generation.
+                        registerCommands();
```
sean-mcmanus, line 327: I could delay only a subset of commands, but that didn't seem necessary to bother with.

sean-mcmanus (commented):
I could delay only a subset of commands, but that didn't seem necessary to bother with.

bobbrow (approved):

PR #3182, opened 2019-02-15
Add Pipe Transport Configurations to MacOS

WardenGnaw:
We didn't provide Pipe Transport Configurations for MacOS


sean-mcmanus:
Are you going to check this in?

sean-mcmanus (approved):

PR #3184, opened 2019-02-16
Add settings for auto pch

michelleangela:
Add these settings for auto pch:

1. Cache path
2. Cache size


Extension/package.json
```diff
@@ -308,6 +308,19 @@
...
+          "default": "${workspaceFolder}/.vscode/ipch/",
+          "description": "Defines the folder path for cached precompiled headers. The default path is \"${workspaceFolder}/.vscode/ipch/\"",
+          "scope": "resource"
+        },
+        "C_Cpp.intelliSense.cacheSize": {
+          "type": "number",
+          "default": 5120,
+          "description": "Maximum size of the per-workspace hard drive space in megabytes for cached precompiled headers; the actual usage may fluctuate around this value. The default size is 5120 MB. Precompiled header caching is disabled when the size is 0.",
+          "scope": "resource",
+          "minimum": 0
```
grdowns, line 322: Should we also set a maximum value? Even if we constrain the size after reading the setting's value, it would be helpful to set a maximum to give the user feedback

bobbrow: Visual Studio does not have a maximum for this value

bobbrow: Well, technically it does. The max value of uint32.

Extension/package.json
```diff
@@ -308,6 +308,19 @@
           "description": "Determines whether pop up notifications will be shown when a configuration provider extension is unable to provide a configuration for a source file.",
           "scope": "resource"
         },
+        "C_Cpp.intelliSense.cachePath": {
```
sean-mcmanus: This naming style seems inconsistent with our existing settings like intelliSenseEngine, intelliSenseMode, and intelliSenseEngineFallback.

sean-mcmanus (dismissed):

Extension/package.json
bobbrow: Yeah, we were trying to be forward-looking, but I agree this might be premature.
We should rename these `intelliSenseCachePath` and `intelliSenseCacheSize`.

Extension/src/LanguageServer/client.ts
```diff
@@ -410,6 +410,10 @@ class DefaultClient implements Client {
                 default: {
                     systemIncludePath: settings.defaultSystemIncludePath
                 },
+                intelliSense: {
```
bobbrow: This will also need to change to `intelliSenseCachePath` and `intelliSenseCacheSize`

bobbrow (changes requested):

Extension/package.json
michelleangela: Won't add a maximum value for now.

bobbrow (approved):

LinAGKar:
Does intelliSenseCachePath need to be unique per workspace, or can it be set to a global location? Really, on Linux it should point to somewhere in XDG_CACHE_HOME by default.

bobbrow:
We were emulating Visual Studio which sets this per project, but you can certainly set `C_Cpp.intelliSenseCachePath` to a global location that all projects share.

renkun-ken:
It looks like `C_Cpp.intelliSenseCachePath` currently does not translate environment variables like `${env:HOME}` so that I can put these files somewhere into my home folder. I'm not sure if I'm missing something?

sean-mcmanus:
@renkun-ken It's a bug -- we have a fix at https://github.com/Microsoft/vscode-cpptools/pull/3368 for our next release.

renkun-ken:
Thanks, @sean-mcmanus!

LinAGKar:
`~/.cache/vscode/` does not seem to work either. Also, the setting says that there is a fallback if an invalid path is given, but there is no indication whether the currently set path is considered valid.

bobbrow:
@LinAGKar, it's the same bug as `${env:HOME}` and will be fixed in the next release.

PR #3191, opened 2019-02-19
Set up CI with Azure Pipelines

WardenGnaw:
Added azure-pipelines.yml and build templates to allow using Azure Pipelines as part of our PR CI check. 

Fixed the missing error reporting when integration tests failed. 
Debug integration tests now test with a mock debugger. 
Seperated languageServer and debugger integration tests. 
Factory provider now takes a context in order for us to find the extension path without using utils.


Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -9,32 +9,51 @@ import * as path from 'path';
...
 
+    constructor(context: vscode.ExtensionContext) {
+        super(context);
+    }
+
     createDebugAdapterDescriptor(session: vscode.DebugSession, executable: vscode.DebugAdapterExecutable | undefined): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
         return util.isExtensionReady().then(ready => {
             if (ready) {
-                let command: string = path.join(util.extensionContext.extensionPath, './debugAdapters/OpenDebugAD7');
+                let command: string = path.join(this.context.extensionPath, './debugAdapters/OpenDebugAD7');
```
pieandcakes, line 35: Is the `context` new?

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -9,32 +9,51 @@ import * as path from 'path';
...
-// Note: util.extensionContext.extensionPath is needed for the commands because VsCode does not support relative paths for adapterExecutableComand
+// NOTE: This file is not automatically tested.
 
-export class CppdbgDebugAdapterDescriptorFactory implements vscode.DebugAdapterDescriptorFactory {
+abstract class DebugAdapterDescriptorFactoryWithContext implements vscode.DebugAdapterDescriptorFactory {
+    protected readonly context: vscode.ExtensionContext;
+
+    // This is important for the Mock Debugger since it can not use src/common
+    constructor(context: vscode.ExtensionContext) {
+        this.context = context;
```
pieandcakes, line 19: Could this be stale at any point in that it might change after activation but since we are caching it, we might have a stale one?

Extension/test/integrationTests/MockDebugger/debugAdapterDescriptorFactory.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as vscode from "vscode";
+import * as path from 'path';
+
+// import * as util from '../src/common' <- DO NOT USE. Also do not use anything with relative paths, it will break during hijacking in test/debug.integration.test.ts
+
+abstract class DebugAdapterDescriptorFactoryWithContext implements vscode.DebugAdapterDescriptorFactory {
```
pieandcakes, line 11: Why not abstract this to another file so its not duplicated?

Extension/test/integrationTests/MockDebugger/debugAdapterDescriptorFactory.ts
```diff
@@ -0,0 +1,47 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as vscode from "vscode";
+import * as path from 'path';
+
+// import * as util from '../src/common' <- DO NOT USE. Also do not use anything with relative paths, it will break during hijacking in test/debug.integration.test.ts
```
pieandcakes: Change wording to not use term `hijack`

Extension/test/integrationTests/MockDebugger/debugAdapterDescriptorFactory.ts
```diff
@@ -0,0 +1,47 @@
...
+    protected readonly context: vscode.ExtensionContext;
+
+    constructor(context: vscode.ExtensionContext) {
+        this.context = context;
+    }
+
+    abstract createDebugAdapterDescriptor(session: vscode.DebugSession, executable: vscode.DebugAdapterExecutable | undefined): vscode.ProviderResult<vscode.DebugAdapterDescriptor>;
+}
+
+export class CppdbgDebugAdapterDescriptorFactory extends DebugAdapterDescriptorFactoryWithContext {
```
pieandcakes, line 21: Does this support interfacing so that we can create an interface of `ITestDebugAdapter` and then we can make sure in the production case that we don't have that interface but we do in the test case?

Extension/test/integrationTests/debug/debug.integration.test.ts
```diff
@@ -7,14 +7,19 @@ import * as vscode from 'vscode';
 import * as assert from 'assert';
  
 suite(`Debug Integration Test: `, function(): void {
```
pieandcakes: NIT: Rename the file to just be integration.test.ts

azure-pipelines.yml
```diff
@@ -0,0 +1,17 @@
...
+  parameters:
+    name: macOS
+    pool:
+      vmImage: 'macOS-10.13'
+
+- template: Extension/jobs/build.yml
+  parameters:
+    name: Linux
+    pool:
+      vmImage: 'ubuntu-16.04'
```
pieandcakes, line 15: Do we want to check Arch or CentOS too or something that is different (aka doesn't support bashisms very well)?

pieandcakes (approved):

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
WardenGnaw: Its the same context that is set in common.ts (util.extensionContext)

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
WardenGnaw: It is the same one that VS Code passes in when we activate, if its stale then that's how VSCode handles it.

Extension/test/integrationTests/MockDebugger/debugAdapterDescriptorFactory.ts
WardenGnaw: The .js file is copied during the pipeline, trying to remove any relative imports as possible.

azure-pipelines.yml
WardenGnaw: Only if you want to maintain the machines for Arch/CentOS

Issue #2799, opened 2018-11-09
No warning/log is shown if a non-existent compilerPath is used


sean-mcmanus:
I accidentally used a `\` instead of `\\` on Windows, causing the compilerPath to not be used -- the scheme only gives warnings for certain `\` characters, e.g. using `\f` gives no error, but `\c` does. No warning message or logging is shown explaining why it is choosing a different compilerPath.

PR #3196, opened 2019-02-21
Seanmcm/cpp properties squiggles

sean-mcmanus:


sean-mcmanus:
This doesn't work for compilerPath -- it seems like that is the most important case to handle next: https://github.com/Microsoft/vscode-cpptools/issues/2799 .

Extension/src/LanguageServer/configurations.ts
```diff
@@ -117,32 +118,7 @@ export class CppProperties {
         let rootPath: string = rootUri ? rootUri.fsPath : "";
         this.currentConfigurationIndex = new PersistentFolderState<number>("CppProperties.currentConfigurationIndex", -1, rootPath);
         this.configFolder = path.join(rootPath, ".vscode");
-
-        let configFilePath: string = path.join(this.configFolder, "c_cpp_properties.json");
```
sean-mcmanus, line 121: I moved this because it was causing the JSON to be parsed 2 extra times unnecessarily.

sean-mcmanus:
I forgot to add telemetry.

sean-mcmanus:
Working on fixing a bug when one path is a substring of another.

sean-mcmanus:
This change doesn't handle relative paths -- I don't think we can check this in until that is fixed, because otherwise it would incorrectly squiggle any cases of relative paths.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -699,13 +699,13 @@ export class CppProperties {
...
-                        for (let path of pathArray) {
-                            paths.add(path);
+                        for (let curPath of pathArray) {
+                            paths.add('"' + curPath + '"');
                         }
                     }
                 }
                 if (this.CurrentConfiguration.compileCommands) {
-                    paths.add(this.CurrentConfiguration.compileCommands);
+                    paths.add('"' + this.CurrentConfiguration.compileCommands + '"');
```
grdowns: This will work. If you feel like it you can also use a format string which is a little bit less of a hassle to modify. https://basarat.gitbooks.io/typescript/docs/template-strings.html

sean-mcmanus: Yeah, I like that style better -- I forgot TypeScript had that.

sean-mcmanus:
I think I need to handle/disable WSL path squiggles that we translate in the server.

Extension/src/LanguageServer/configurations.ts
bobbrow: Can you double check to make sure that a bogus config doesn't get sent to the extension while we wait for the compiler defaults to arrive? I seem to recall adding this to the constructor for a reason.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -562,7 +569,10 @@ export class CppProperties {
                             });
                             settings.update("default.configurationProvider", undefined); // delete the setting
                         }
+                        let savedKnownCompilers: KnownCompiler[] = this.configurationJson.configurations[0].knownCompilers;
+                        this.configurationJson.configurations[0].knownCompilers = undefined;
```
bobbrow: Is this the same as
```ts
delete this.configurationJson.configurations[0].knownCompilers
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -583,12 +593,15 @@ export class CppProperties {
...
+        if (this.propertiesFile === undefined) {
+            return; // Occurs when propertiesFile hasn't been checked yet.
+        }
         this.configFileWatcherFallbackTime = new Date();
         if (this.propertiesFile) {
             this.parsePropertiesFile();
             // parsePropertiesFile can fail, but it won't overwrite an existing configurationJson in the event of failure.
             // this.configurationJson should only be undefined here if we have never successfully parsed the propertiesFile.
-            if (this.configurationJson !== undefined) {
+            if (this.configurationJson !== null) {
```
bobbrow: I think it was a bug that we tested `this.configurationJson` against `undefined`.  This should just be a 'falsy' check.
```ts
if (this.configurationJson)
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -597,7 +610,7 @@ export class CppProperties {
             }
         }
 
-        if (this.configurationJson === undefined) {
+        if (this.configurationJson === null) {
```
bobbrow: This should also be a 'falsy' check.
```ts
if (!this.configurationJson)
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -675,15 +692,170 @@ export class CppProperties {
...
+            this.handleSquiggles();
         } catch (err) {
-            vscode.window.showErrorMessage('Failed to parse "' + this.propertiesFile.fsPath + '": ' + err.message);
+            vscode.window.showErrorMessage(`Failed to parse "${this.propertiesFile.fsPath}": ${err.message}`);
             throw err;
         }
     }
 
+    private handleSquiggles(): void {
+        vscode.workspace.openTextDocument(this.propertiesFile).then((document: vscode.TextDocument) => {
```
bobbrow, line 712: For future-proofing, this function should check if `this.propertiesFile` is valid before trying to open it.  The places you call `handleSquiggles` today are safe, but may not be if we ever support multiple configs without a properties file.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -675,15 +692,170 @@ export class CppProperties {
...
+                    } else if (this.rootfs && this.rootfs.length > 0) {
+                        resolvedPath = this.rootfs + resolvedPath.substr(1);
+                        resolvedPath = resolvedPath.replace(/\//g, path.sep);
+                        // TODO: Handle WSL symlinks.
+                    }
+                }
+
+                let pathExists: boolean = true;
+                if (!fs.existsSync(resolvedPath)) {
+                    // Check again for a relative path.
```
bobbrow, line 802: I thought we didn't support relative paths at all

Extension/src/LanguageServer/configurations.ts
```diff
@@ -675,15 +692,170 @@ export class CppProperties {
...
+                        resolvedPath = relativePath;
+                    }
+                }
+
+                // Iterate through the text and apply squiggles.
+                for (let curOffset: number = curText.indexOf(curPath); curOffset !== -1; curOffset = curText.indexOf(curPath, curOffset + curPath.length)) {
+                    let message: string;
+                    if (!pathExists) {
+                        message = `Cannot find "${resolvedPath}".`;
+                        newSquiggleMetrics.PathNonExistent++;
```
bobbrow, line 816: I thought we were going to do metrics on which properties had squiggles instead of what the errors were.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -117,32 +122,7 @@ export class CppProperties {
...
-        this.configFileWatcher.onDidDelete(() => {
-            this.propertiesFile = null;
-            this.resetToDefaultSettings(true);
-            this.handleConfigurationChange();
-        });
-
-        this.configFileWatcher.onDidChange(() => {
-            this.handleConfigurationChange();
-        });
+        this.diagnosticCollection = vscode.languages.createDiagnosticCollection(rootPath);
```
bobbrow, line 125: is this object supposed to be disposed at some point?

bobbrow: and is it supposed to be for the c_cpp_properties.json file instead of the workspace folder?

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I believe this change requires my other pull request to be 100% correct during the initial load: https://github.com/Microsoft/vscode-cpptools/pull/3178 . Was there something else you wanted me to double check in particular? I tested compile_commands and configurationProvider scenarios. The handleConfigurationChange() method checks for undefined now (i.e. before it's set), which can occur with the async buildVcpkgIncludePath -- that might be what you're referring to?

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: For this usage, it leads to identical behavior (see https://stackoverflow.com/questions/14967535/delete-a-x-vs-a-x-undefined ), but I can change it to use delete though, since is a more "complete" deletion than "= undefined".

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Sure, I can change that, but I don't think it's ever undefined.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: What do you mean? What didn't support relative paths? The language server supports relative paths. Do you see a problem with the squiggle support for this?

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: That's not what I thought :). Why does it matter which properties have the squiggles? That doesn't appear to be important info to me.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Not sure, I'll check...

Extension/src/LanguageServer/configurations.ts
bobbrow: This prevents us from attaching different meanings to null and undefined (which is undesirable)

Extension/src/LanguageServer/configurations.ts
bobbrow: Either way seems fine. We just need to make sure the feature is working.  You don't have to change this.

Extension/src/LanguageServer/configurations.ts
bobbrow: If it's disposable, we just need to add it to the disposables array

Extension/src/LanguageServer/configurations.ts
bobbrow: I approved that PR

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: This seems correct to me. It doesn't need to be disposed. Using rootPath as the "name" should be equivalent to the rootPath + .vscode/c_cpp_properties.json and that last part is unnecessary, right?

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I double checked and tested it and it seems good to me.

Extension/src/LanguageServer/configurations.ts
bobbrow: I just looked at the API.  The string you pass in is just a name for the collection, not a uri or filepath.  And it does have a dispose method on it, so you should call `dispose` on it in the `dispose` method at the bottom of this file.  It doesn't look like it inherits `vscode.Disposable` so you can't add it to the disposables array.

Extension/src/LanguageServer/configurations.ts
bobbrow: `includePath`, for example doesn't support relative paths.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Okay, you're right. I tested it and it seemed to work, but it was actually using the file-relative path and not the workspaceFolder relative path.

sean-mcmanus: Wait -- I'm seeing relative include paths work sometimes. Might be some bug.

sean-mcmanus: Relative paths seem to all work for me for all the properties. It's better to not show squiggles if there arer working cases.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I added a dispose call.

bobbrow (approved):

PR #3215, opened 2019-02-25
Grdowns/launch configs

grdowns:
- Implement debug configurations that will build the active C/C++ file and launch a debugger
- Add command to run the configuration
- Modify build tasks to resolve the compiler via PATH instead of hard-coding path


grdowns:
TODO:
- Get working scenario for workspaceFolder open but active file outside of workspaceFolder
- Test Mac debugger name; should be `lldb` for Mac and `lldb-mi` I believe

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -7,11 +7,73 @@ import * as debugUtils from './utils';
...
+            if (label.indexOf("(gdb) Launch") !== -1) {
+                label = "Default Configuration";
+            }
+            return {label: label, configuration: config};
+        });
+
+        return vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"}).then(async selection => {
+            // Wrap in new Promise to make sure task kicks off before VS Code switches the active document to launch.json
+            return new Promise<vscode.DebugConfiguration[]>(async resolve => {
+                if (selection.label.indexOf("Build and Debug active file") !== -1 && selection.configuration.preLaunchTask) {
```
bobbrow: need to handle undefined `selection` (e.g. user cancelled operation)

Extension/src/Debugger/extension.ts
```diff
@@ -28,7 +30,67 @@ export function initialize(context: vscode.ExtensionContext): void {
...
+        interface MenuItem extends vscode.QuickPickItem {
+            configuration: vscode.DebugConfiguration;
+        }
+
+        const items: MenuItem[] = configs.map<MenuItem>(config => {
+            return {label: config.name, configuration: config};
+        });
+
+        vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"}).then(async selection => {
+            if (selection.configuration.preLaunchTask) {
```
bobbrow, line 70: need to handle undefined `selection` (e.g. user cancelled operation)

Extension/src/Debugger/extension.ts
```diff
@@ -28,7 +30,67 @@ export function initialize(context: vscode.ExtensionContext): void {
...
+
+        interface MenuItem extends vscode.QuickPickItem {
+            configuration: vscode.DebugConfiguration;
+        }
+
+        const items: MenuItem[] = configs.map<MenuItem>(config => {
+            return {label: config.name, configuration: config};
+        });
+
+        vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"}).then(async selection => {
```
bobbrow: if `items` is empty, we should not show the quick pick.  (On Windows, we might want to show a message that msvc is not supported yet)

Extension/src/common.ts
```diff
@@ -36,6 +37,61 @@ export function getRawPackageJson(): any {
...
+    assert(selectedTask);
+
+    let definition: vscode.TaskDefinition = selectedTask.definition as vscode.TaskDefinition;
+    if (definition && definition.compilerPath) {
+         // TODO: add desired properties to empty object, don't delete.
+        delete definition.compilerPath;
+     }
+    rawTasksJson.tasks.push(selectedTask.definition);
+    // TODO: It's dangerous to overwrite this file. We could be wiping out comments.
+    await writeFileText(getTasksJsonPath(), JSON.stringify(rawTasksJson, null, 2));
```
bobbrow, line 91: Is there a way we can use the vscode API to write to this file?  I would like to avoid losing existing comments in the file and it might be a bit of work to parse them out ourselves.  Perhaps the jsonc parser can help us out, but I'm unfamiliar with the API.

Extension/src/common.ts
```diff
@@ -36,6 +37,61 @@ export function getRawPackageJson(): any {
...
+    }
+    return new Promise<any>((resolve, reject) => {
+        fs.exists(path, exists => {
+            if (!exists) {
+                return resolve({});
+            }
+            const fileContents: Buffer = fs.readFileSync(path);
+            let rawTasks: any = {};
+            try {
+                rawTasks = JSON.parse(fileContents.toString());
```
bobbrow, line 53: This will likely fail.  We should probably use the jsonc parser instead.

bobbrow (changes requested):

Extension/src/Debugger/extension.ts
```diff
@@ -28,7 +30,67 @@ export function initialize(context: vscode.ExtensionContext): void {
...
+        if (!folder) {
+            // Not enabled because we do not react to single-file mode correctly yet.
+            // We get an ENOENT when the user's c_cpp_properties.json is attempted to be parsed.
+            // The DefaultClient will also have its configuration accessed, but since it doesn't exist it errors out.
+            vscode.window.showErrorMessage('This command is not yet available for single-file mode.');
+            return Promise.resolve();
+        }
+
+        if (!util.fileIsCOrCppSource(textEditor.document.uri.fsPath)) {
+            vscode.window.showErrorMessage('The command is disabled because the active file is not a C or C++ source file.');
```
bobbrow: "the command is disabled" -> "Cannot build and debug"

sean-mcmanus:
I'm making these code review changes and not Griffin right?

bobbrow:
> I'm making these code review changes and not Griffin right?

yes, Sean

Extension/src/common.ts
grdowns: I'm not sure! One possible solution is to invoke the "Configure Tasks" command, however it will pop up a quickpick menu for the user to select, and as far as I know the command itself doesn't take any arguments (gross). Otherwise the VS Code API does not expose any means as far as my research goes.

Another idea is to look at how VS Code does it to derive some possible solutions.

sean-mcmanus:
Do we want the debug configuration show "clang Build and Debug active file"? The casing seems odd to me. I changed the casing of the command to be "title case" like the other commands, e.g. "Build and Debut Active File".

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -23,8 +85,59 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
+            newConfig.name = compilerName + " Build and Debug active file";
+            newConfig.preLaunchTask = task.name;
+            newConfig.externalConsole = false;
+            const exeName: string = path.join("${fileDirname}", "${fileBasenameNoExtension}");
+            newConfig.program = platform === "win32" ? path.join(exeName, ".exe") : exeName;
+
+            let debuggerName: string;
+            if (compilerName.startsWith("clang")) {
+                newConfig.MIMode = "lldb";
+                const suffix: string = compilerName.substr(compilerName.indexOf("-"));
```
sean-mcmanus: I don't understand what this suffix is supposed to be. It's turning "clang" into "g" and then lldb into lldbg.

bobbrow: I think it's supposed to be the `6.0` after the `-` in `clang-6.0`.  Problem is that the user could just be using `clang`.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -7,11 +7,73 @@ import * as debugUtils from './utils';
...
+        this.underlyingProvider = provider;
+    }
+
+    async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        const configs: vscode.DebugConfiguration[] = await this.underlyingProvider.provideDebugConfigurations(folder, token);
+        
+        const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+        if (!editor || !util.fileIsCOrCppSource(editor.document.fileName) || configs.length <= 1) {
+            const defaultConfig: vscode.DebugConfiguration = configs.find(config => { return config.name === "(gdb) Launch" || config.name === "(lldb) Launch"; });
+            assert(defaultConfig);
```
bobbrow: I don't think we want to use `assert` in the extension.  If I recall correctly, `assert` will terminate the extension process if the condition fails.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -23,8 +85,59 @@ abstract class CppConfigurationProvider implements vscode.DebugConfigurationProv
...
-        return this.provider.getInitialConfigurations(this.type);
+    async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        const buildTasks: vscode.Task[] = await getBuildTasks();
+        if (!buildTasks.length) {
+            return Promise.resolve(this.provider.getInitialConfigurations(this.type));
+        }
+        const defaultConfig: vscode.DebugConfiguration = this.provider.getInitialConfigurations(this.type).find(config => {
+            return config.name === "(gdb) Launch" || config.name === "(lldb) Launch";
+        });
+        assert(defaultConfig, "Could not find default debug configuration.");
```
bobbrow: don't use `assert`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -7,11 +7,73 @@ import * as debugUtils from './utils';
...
+    public constructor(provider: CppConfigurationProvider) {
+        this.underlyingProvider = provider;
+    }
+
+    async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        const configs: vscode.DebugConfiguration[] = await this.underlyingProvider.provideDebugConfigurations(folder, token);
+        
+        const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+        if (!editor || !util.fileIsCOrCppSource(editor.document.fileName) || configs.length <= 1) {
+            const defaultConfig: vscode.DebugConfiguration = configs.find(config => { return config.name === "(gdb) Launch" || config.name === "(lldb) Launch"; });
```
bobbrow: There are a lot of hardcoded "(gdb) Launch" and "(lldb) Launch" strings in this file. Can we declare a const for these somewhere and use those instead? There is a lot of potential for misspellings and stuff getting out of sync if these labels ever change.

sean-mcmanus: Sure.

grdowns: A string enumeration would do the trick. I had some thoughts about that.

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: Do you know how to repro this scenario? If not, I can investigate later...

sean-mcmanus:
Yeah, it's creating multiple configs after every debug launch, so I need to fix that...

sean-mcmanus:
I'm seeing C++ debug configs appearing for .c files...

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: I handled it by using the "Default Configuration"...I could cause it to fail with an error too.

Extension/src/Debugger/extension.ts
sean-mcmanus: I show "No compiler found" in the quick pick.

sean-mcmanus:
I disabled the debug and launch.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -93,8 +93,8 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
 	 * Returns a list of initial debug configurations based on contextual information, e.g. package.json or folder.
 	 */
     async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
-        const buildTasks: vscode.Task[] = await getBuildTasks();
-        if (!buildTasks.length) {
+        let buildTasks: vscode.Task[]; // = await getBuildTasks(); // TODO
+        if (!buildTasks || buildTasks.length === 0) {
```
grdowns, line 97: Btw, `buildTasks` is undefined as it is never initialized. Technically the `buildTasks.length === 0` check isn't necessary

sean-mcmanus: Yeah, I know -- I just kept that code in there for later...I could have commented it out as well.

Extension/src/common.ts
```diff
@@ -81,8 +81,18 @@ export async function ensureBuildTaskExists(taskName: string): Promise<void> {
     if (definition && definition.compilerPath) {
          // TODO: add desired properties to empty object, don't delete.
         delete definition.compilerPath;
-     }
-    rawTasksJson.tasks.push(selectedTask.definition);
+    }
+    
+    let isDuplicateTask: boolean = false;
+    for (let task of rawTasksJson.tasks) {
```
grdowns: You should rewrite this as
```
if (!rawTasksJson.tasks.find(task => { return task.label === selectedTask.definition.label }) {
    rawTasksJson.tasks.push(selectTask.definition);
}

Comment (sean-mcmanus) [reply Extension/src/common.ts]: Ok

File: Extension/src/LanguageServer/configurations.ts [hunk 13]
```diff
@@ -832,7 +842,7 @@ export class CppProperties {
                         }
                     }
                     let diagnostic: vscode.Diagnostic = new vscode.Diagnostic(
-                        new vscode.Range(document.positionAt(curTextStartOffset + curOffset + 1), document.positionAt(curTextStartOffset + curOffset + curPath.length - 1)),
+                        new vscode.Range(document.positionAt(curTextStartOffset + curOffset + 1), document.positionAt(curTextStartOffset + curOffset + (curPath.endsWith("\"") ? curPath.length - 1 : curPath.length))),
```
grdowns: Btw, you can embed quotes in a string in JavaScript, as the begin quote type will match to the end quote type. I.e. you can write: `curPath.endsWith('"')`

sean-mcmanus: Yeah, I used '"' a few lines above. I can change it to that this case.

sean-mcmanus (approved):

bobbrow (approved):
:shipit:

Issue #1524, opened 2018-02-06
Allow other extensions to query properties of the current configuration


thejcannon:
I'm writing an extension that plugs in our companies build processes on top of vscode-cpptools. Stuff like generating c_cpp_properties.json, launch.json, etc...

Ideally, the user would have their workspace set up so that they've selected a configuration for intellisense, and when they hit F5, the appropriate launch.json configuration is selected for their selected c_cpp_properties.json configuration.

The easiest way I can think of would be to allow other extensions to query vscode-cpptools for the current configuration, and then clients can use that to get the name or any other property.

I'm still getting my hands dirty with extensions, so let me know if this already exists, or there's a better way to accomplish this.

Thanks!

sean-mcmanus:
VS code allows extensions to export functions that other extensions can call. So we could export some GetCurrentConfiguration method. Another potential option would be to store the current configuration name in a setting, and then the setting could be referenced using ${config.<setting>} in launch.json. We currently just save the current configuration as a number/index, which we send to the processes when the number changes. You should be able to implement this yourself with our open source typescript and we could accept a pull request.

thejcannon:
I think exposing the current configuration (as the full JSON) would probably be more beneficial to clients.

As far as exposing a function goes, I found a way of [executing another extension's commands](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_commands), is that what you're referring to? If so I think it would be easy to add a new command (that isn't attached to anything in the UI) that returns the current configuration.

sean-mcmanus:
I don't know what you mean by "exposing the current configuration (as the full JSON)". We currently use a c_cpp_properties.json that other extensions can read/write (the same way our extension does using JSON apis), but the currently selected configuration is not available anywhere. We plan to migrate away from that json file and just use the normal settings.json later though, although c_cpp_properties.json might still be supported for backwards compatibility unless do a conversion, in which case the currently selected configuration might just be a setting as well.

I don't think we want to add a "command", which are listed in the command palette for users to execute. I'm referring to returning a public API from our activate method: https://code.visualstudio.com/docs/extensionAPI/vscode-api#_extensions:
`Extension writers can provide APIs to other extensions by returning their API public surface from the activate-call.
export function activate(context: vscode.ExtensionContext) {
    let api = {
        sum(a, b) {
            return a + b;
        },
        mul(a, b) {
            return a * b;
        }
    };
    // 'export' public api-surface
    return api;
}
[...]

PR #3219, opened 2019-02-26
Enable ${command:cpptools.activeConfigName} in tasks #1524

bobbrow:
Allow for tasks to query the active configuration as requested in issue #1524


Extension/src/LanguageServer/extension.ts
```diff
@@ -584,6 +584,7 @@ export function registerCommands(): void {
     disposables.push(vscode.commands.registerCommand('C_Cpp.ResumeParsing', onResumeParsing));
     disposables.push(vscode.commands.registerCommand('C_Cpp.ShowParsingCommands', onShowParsingCommands));
     disposables.push(vscode.commands.registerCommand('C_Cpp.TakeSurvey', onTakeSurvey));
+    disposables.push(vscode.commands.registerCommand('cpptools.activeConfig', onGetActiveConfig));
```
sean-mcmanus: Should this be activeConfigName to be more explicit and/or if we want to expose the actual config or an index later?

sean-mcmanus (approved):

PR #3221, opened 2019-02-26
Suggest user switching to insiders channel

### LINKED ISSUE #3089: If C_Cpp.updateChannel is Default and an insiders is available, show a prompt
Showing the prompt should increase usage of the Insiders channel. The prompt should have a Don't Show Again option.

lh123:
If C_Cpp.updateChannel is Default and an insiders is available, show a prompt.
fix #3089


Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,38 @@ function realActivation(): void {
     updateLanguageConfigurations();
 
     reportMacCrashes();
-
+    
     const settings: CppSettings = new CppSettings(clients.ActiveClient.RootUri);
-    if (settings.updateChannel === 'Insiders') {
+    let suggestInsider: PersistentState<boolean> = new PersistentState<boolean>("CPP.suggestInsider", true);
```
sean-mcmanus: "suggestInsiders" would be better.

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,38 @@ function realActivation(): void {
...
+    
+    if (suggestInsider.Value && settings.updateChannel === 'Default') {
+        getTargetBuildInfo("Insiders").then(buildInfo => {
+            if (!buildInfo) {
+                return;
+            }
+            const message: string = `The insider version (${buildInfo.name}) is available, whether to switch to the insider channel?`;
```
sean-mcmanus: The Insiders version...switch to the Insiders channel?.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Also, ", whether to" doesn't seem like the correct wording. Can you remove that, i.e. "is available. Switch to...".

sean-mcmanus:
Cool -- thanks a lot for this pull request. I had forgotten about this issue.

Extension/src/LanguageServer/extension.ts
lh123: Yes, you are right, I have made change.

Extension/src/LanguageServer/extension.ts
lh123: Yes, you are right, I made a change.

sean-mcmanus: I don't see the change to the `The Insiders version...` string.

sean-mcmanus:
I think @bobbrow might have had some feedback he's planning to post later. We'd like to get this in for our 0.22.0 release next week, but I don't think it will get into 0.22.0-insiders2 this week (it wouldn't be usable in an Insiders release anyway).

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,43 @@ function realActivation(): void {
...
     const settings: CppSettings = new CppSettings(clients.ActiveClient.RootUri);
-    if (settings.updateChannel === 'Insiders') {
+    let suggestInsiders: PersistentState<boolean> = new PersistentState<boolean>("CPP.suggestInsiders", true);
+    
+    if (suggestInsiders.Value && settings.updateChannel === 'Default') {
+        getTargetBuildInfo("Insiders").then(buildInfo => {
+            if (!buildInfo) {
+                return;
+            }
+            const message: string = `The insider version (${buildInfo.name}) is available, switch to the insider channel?`;
```
bobbrow: `Insiders version: ${buildInfo.name} is available. Would you like to switch to the Insiders channel and install this update?`

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,43 @@ function realActivation(): void {
...
+            const yes: string = "Yes";
+            const askLater: string = "Ask Me Later";
+            const dontShowAgain: string = "Don't Show Again";
+            vscode.window.showInformationMessage(message, yes, askLater, dontShowAgain).then(selection => {
+                switch (selection) {
+                    case yes:
+                        // Avoid double calling applyUpdate
+                        shouldCheckAndApplyUpdate = false;
+                        vscode.workspace.getConfiguration("C_Cpp").update("updateChannel", "Insiders", true);
+                        return applyUpdate(buildInfo, 'Insiders');
```
bobbrow: I'm unsure as to why you split this up.  Can you just update the configuration value and let `onDidChangeSettings` call `checkAndApplyUpdate`?  Your comment suggests that this is what would happen automatically.  If you can just let the settings change do it's thing, you wouldn't need to split out `applyUpdate` from `checkAndApplyUpdate` nor would you need the `shouldCheckAndApplyUpdate` variable.

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,43 @@ function realActivation(): void {
...
+                        suggestInsiders.Value = false;
+                        break;
+                    case askLater:
+                        break;
+                    default:
+                        break;
+                }
+            });
+        }).catch(error => {
+            // Handle .then following getTargetBuildInfo rejection
```
bobbrow: Rejection shouldn't go through `.catch`.  Hmm... I think I see a bug in `getTargetBuildInfo` that makes it throw in some cases instead of returning a rejected promise.  I think this should use `await getTargetBuildInfo`.  We'll need to clean up the code in general to follow the async/await model instead of a combination of Promises and async/await.

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,43 @@ function realActivation(): void {
...
+            if (error.message.indexOf('/') !== -1 || error.message.indexOf('\\') !== -1) {
+                error.message = "Potential PII hidden";
+            }
+            telemetry.logLanguageServerEvent('installVsix', { 'error': error.message, 'success': 'false' });
```
bobbrow: We need to think about a different telemetry event here or simply removing this.  This telemetry event name is used for the path that actually installs the VSIX.  I'll discuss with @sean-mcmanus.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
lh123: If I use `onDidChangeSettings` to call `checkAndApplyUpdate`, `getTargetBuildInfo` will be called twice. So I use `applyUpdate` to consume `buildInfo` to avoid calling `checkAndApplyUpdate` again.

Extension/src/LanguageServer/extension.ts
lh123: The `getTargetBuildInfo` is a `async` function that throws an error in it, I think catch can catch this exception. But we need to add `return` before `vscode.window.showInformationMessage(message,yes,askLater,dontShowAgain)` to enable catch to handle exceptions thrown by `applyUpdate`.

lh123: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Return_value
> A Promise which will be resolved with the value returned by the async function, **or rejected with an uncaught exception thrown from within the async function**.

bobbrow: You are right, you can catch this if the throwing function is marked `async`.  I tested some cases in the TypeScript playground and this should work.  I would prefer that you use `try { await } catch { }` instead of `.then` and `.catch`, but I won't make you convert this function to async.

Extension/src/LanguageServer/extension.ts
```diff
@@ -477,83 +515,79 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+ * Query package.json and the GitHub API to determine whether the user should update, if so then install the update.
+ * The update can be an upgrade or downgrade depending on the the updateChannel setting.
+ * @param updateChannel The user's updateChannel setting.
+ */
+async function checkAndApplyUpdate(updateChannel: string): Promise<void> {
+    let buildInfo: BuildInfo;
+    try {
+        buildInfo = await getTargetBuildInfo(updateChannel);
+        if (!buildInfo) {
+            return undefined;
```
bobbrow: we don't usually return 'undefined' for functions marked 'void'.

Extension/src/LanguageServer/extension.ts
bobbrow: Got it. The comment made me think it was for a different reason.  I'm still trying to think of a different way to do this since I feel the code would flow more naturally if the settings change logic managed the update.  Perhaps we can cache the results of `getTargetBuildInfo` and have the update code clear the stored value when an actual update completes (or fails).

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Sounds good to me except "insiders channel" should be "Insiders channel".

Extension/src/LanguageServer/extension.ts
sean-mcmanus: What's wrong with this telemetry? Isn't it install related?

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Isn't it because it's a `Promise<void>`?

bobbrow: It's an `async` function, so you don't return a `Promise` explicitly. You return the wrapped type: `void` in this case.

Extension/src/LanguageServer/extension.ts
bobbrow: I updated the original comment

Extension/src/LanguageServer/extension.ts
bobbrow: Yes, but this is not for people on the Insiders channel. This is for **everyone** (at least once).  I don't want to merge telemetry from `getTargetBuildInfo` errors for people on the Default channel with people on the Insiders channel.

I think there's two options here:
1. Insert a `.catch` between `getTargetBuildInfo` and `.then` so we can filter out those errors (maybe fire a different event there)
2. Move this catch to `applyUpdate(...).catch`

Extension/src/LanguageServer/extension.ts
lh123: Yes. I think we should change `return undefined` to `return`.

Extension/src/LanguageServer/extension.ts
lh123: What should we do with the exception thrown by `getTargetBuildInfo`, ignore it directly?

sean-mcmanus: Can you send the same telemetry but with EventName "suggestInsiders" instead of "installvsix"?

lh123:
I have refactored the code using `async` and `await`.

Extension/src/LanguageServer/extension.ts
lh123: Ok, I have cached `buildInfo` and refactored the code.

Extension/src/LanguageServer/extension.ts
```diff
@@ -477,85 +481,125 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+    // If we have buildinfo cache, we should use it.
+    let buildInfo: BuildInfo | null = buildInfoCache;
+    // clear buildinfo cache.
+    buildInfoCache = null;
+    
+    if (!buildInfo) {
+        try {
+            buildInfo = await getTargetBuildInfo(updateChannel);
+        } catch (error) {
+            telemetry.logLanguageServerEvent('suggestInsiders', { 'error': error.message, 'success': 'false' });
```
bobbrow: This event should still be 'installVsix'

Extension/src/LanguageServer/extension.ts
```diff
@@ -477,85 +481,125 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+            suggestInsiders.Value = false;
+            break;
+        case askLater:
+            break;
+        default:
+            break;
+    }
+}
+
+async function applyUpdate(buildInfo: BuildInfo, updateChannel: string): Promise<void> {
```
bobbrow: If this function is meant to be `async`, it should await the new Promise instead of returning it.  I believe that's what it did before.

Extension/src/LanguageServer/extension.ts
```diff
@@ -293,9 +294,12 @@ function realActivation(): void {
     updateLanguageConfigurations();
 
     reportMacCrashes();
-
+    
     const settings: CppSettings = new CppSettings(clients.ActiveClient.RootUri);
-    if (settings.updateChannel === 'Insiders') {
+    
+    if (settings.updateChannel === 'Default') {
+        suggestInsidersChannel();
```
bobbrow, line 316: Thanks for refactoring this.  I believe this is much easier to follow. 👍 👍

Extension/src/LanguageServer/extension.ts
```diff
@@ -477,85 +481,125 @@ async function installVsix(vsixLocation: string, updateChannel: string): Promise
...
+    const yes: string = "Yes";
+    const askLater: string = "Ask Me Later";
+    const dontShowAgain: string = "Don't Show Again";
+    let selection: string = await vscode.window.showInformationMessage(message, yes, askLater, dontShowAgain);
+    switch (selection) {
+        case yes:
+            // Cache buidinfo.
+            buildInfoCache = buildInfo;
+            // It will call onDidChangeSettings.
+            vscode.workspace.getConfiguration("C_Cpp").update("updateChannel", "Insiders", true);
```
bobbrow: I believe the preferred value for the 3rd parameter of `update` is `ConfigurationTarget.Global`.  It means the same as `true` but makes it easier to remember the behavior just by reading the code.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
lh123:
```suggestion
function applyUpdate(buildInfo: BuildInfo, updateChannel: string): Promise<void> {
```
Yes, but I think the best way is to delete `async`.

bobbrow (approved):

sean-mcmanus (approved):

sean-mcmanus:
@lh123 We'll add your credit to the changelog when we write it.

sean-mcmanus:
I think this change has some minor "issues" -- I'll fix them in https://github.com/Microsoft/vscode-cpptools/pull/3251 .

PR #3237, opened 2019-03-01
Fix "cl.exe" in the compilerPath squiggle.

sean-mcmanus:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -753,8 +753,9 @@ export class CppProperties {
                 paths.add(`"${this.CurrentConfiguration.compileCommands}"`);
             }
 
-            let compilerPathAndArgs: util.CompilerPathAndArgs;
-            if (this.CurrentConfiguration.compilerPath) {
+            if (this.CurrentConfiguration.compilerPath &&
+                    !(os.platform() === 'win32' && this.CurrentConfiguration.compilerPath.startsWith("cl.exe"))) { // Skip this special case.
```
bobbrow: They could have an old path (e.g. Preview VS) and the extension will still find a cl.exe.  This should be `endsWith`. We talked about doing this check as you go through the arguments.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I fixed "cl.exe" and stuff with missing ".exe", but "cl" without an .exe...not sure if that matters.

bobbrow (approved):

PR #3244, opened 2019-03-02
Seanmcm/debug fixes

sean-mcmanus:
Still need to more testing (on Mac...)

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -54,24 +54,20 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
             return menuItem;
         });
 
-        return vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"}).then(async selection => {
-            // Wrap in new Promise to make sure task kicks off before VS Code switches the active document to launch.json
-            return new Promise<vscode.DebugConfiguration[]>(async (resolve, reject) => {
-                if (!selection) {
-                    return reject([]); // User canceled it.
-                }
-                if (selection.label.indexOf(buildAndDebugActiveFileStr()) !== -1 && selection.configuration.preLaunchTask) {
-                    try {
-                        await util.ensureBuildTaskExists(selection.configuration.preLaunchTask);
-                        await vscode.debug.startDebugging(folder, selection.configuration);
-                        Telemetry.logDebuggerEvent("buildAndDebug", { "success": "true" });
-                    } catch (e) {
-                        Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
-                    }
-                }
-                return resolve([selection.configuration]);
-            });
-        });
+        const selection: MenuItem = await vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"});
+        if (!selection) {
+            throw new Error(); // User canceled it.
+        }
+        if (selection.label.indexOf(buildAndDebugActiveFileStr()) !== -1 && selection.configuration.preLaunchTask) {
+            try {
+                await util.ensureBuildTaskExists(selection.configuration.preLaunchTask);
+                await vscode.debug.startDebugging(folder, selection.configuration);
+                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "true" });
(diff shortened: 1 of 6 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -127,7 +126,7 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
                 const suffix: string = suffixIndex === -1 ? "" : compilerName.substr(suffixIndex);
                 debuggerName = (platform === "darwin" ? "lldb" : "lldb-mi") + suffix;
             } else {
-                debuggerName = "gdb";
+                debuggerName = platform === "win32" ? "gdb.exe" : "gdb";
```
bobbrow: Do we need to do something similar for lldb on Windows?

Extension/src/Debugger/extension.ts
```diff
@@ -35,7 +35,7 @@ export function initialize(context: vscode.ExtensionContext): void {
         disposables.push(vscode.debug.registerDebugConfigurationProvider('cppvsdbg', new CppVsDbgConfigurationProvider(configurationProvider)));
     }
     const provider: CppDbgConfigurationProvider = new CppDbgConfigurationProvider(configurationProvider);
-    disposables.push(vscode.debug.registerDebugConfigurationProvider('cppdbg', provider)); //new QuickPickConfigurationProvider(provider))); // TODO
+    disposables.push(vscode.debug.registerDebugConfigurationProvider('cppdbg', new QuickPickConfigurationProvider(provider))); // TODO
```
bobbrow: remove "TODO" comment?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -93,8 +92,8 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
 	 * Returns a list of initial debug configurations based on contextual information, e.g. package.json or folder.
 	 */
     async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
-        let buildTasks: vscode.Task[]; // = await getBuildTasks(); // TODO
-        if (!buildTasks || buildTasks.length === 0) {
+        let buildTasks: vscode.Task[] = await getBuildTasks(); // TODO
```
bobbrow: You should try/catch this.  If the promise is rejected, you should get an exception here, not an empty array.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -57,9 +56,9 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
 
         return vscode.window.showQuickPick(items, {placeHolder: "Select a configuration"}).then(async selection => {
             // Wrap in new Promise to make sure task kicks off before VS Code switches the active document to launch.json
-            return new Promise<vscode.DebugConfiguration[]>(async resolve => {
+            return new Promise<vscode.DebugConfiguration[]>(async (resolve, reject) => {
                 if (!selection) {
-                    return resolve([defaultConfig]); // User canceled it. Choose the default config?
+                    return reject([]); // User canceled it.
```
bobbrow: If you want to return an empty array, the way to do it is with `resolve([])`, not `reject([])`.  reject goes through the error callback (or throws an exception if you're using `await`)

bobbrow (changes requested):

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: Griffin was saying that was intentional. I'll discuss it with you tomorrow.

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: The promise is never rejected, right?

bobbrow (approved):

Extension/src/Debugger/configurationProvider.ts
bobbrow: There is an async anti-pattern in here.  We talked about it and replaced the promises with async/await

Issue #3206, opened 2019-02-23
[windows insiders]the extension always notify me to update to new version 0.22.0


JoeyRxy:
the notification is:

> The C/C++ Extension has been updated to version 0.22.0-insiders. Please reload the window for the changes to take effect.

so I **reload again and again**.But it still show up the notification!

sean-mcmanus:
What version of VS Code do you have? This might be a bug if your version of VS Code is too old.

JoeyRxy:
the newest except the insider version.
now I solve the problem by disabling the extension's notification.
but I now have a secondary problem:the gdb debugger doesn't work on win10
after I configure those .JSON files while it works on Linux(Ubuntu).
when I click the dubug button,I can see the output from task.json in the
terminalm.but then the whole thing just stop.


Sean McManus <notifications@github.com> 于 2019年2月26日周二 上午3:38写道：

aparks5:
happening to me on insiders 1.31.1

"The C/C++ Extension has been updated to version 0.22.0-insiders2. Please reload the window for the changes to take effect."

^ never goes away 

version:

Version: 1.31.1
Commit: 1b8e8302e405050205e69b59abb3559592bb9e60
Date: 2019-02-12T02:19:29.629Z
Electron: 3.1.2
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Linux x64 4.4.0-142-generic

bobbrow:
I'm seeing this too.  It seems that VSCode is not uninstalling the previous version, and it keeps loading the older version of the extension.  Removing the older version of the extension manually fixed the problem for me.  We'll investigate why this is suddenly broken.

sean-mcmanus:
1. We weren't waiting for the install to finish before showing the reload prompt.
2. We weren't checking the exit code before showing the reload prompt.
3. We weren't logging any error if the install failed (we just sent telemetry).

JoeyRxy:
OK,thank you all!
unfortunately I have reset my computer for other reason so that the problem disappear

sean-mcmanus:
We made fixes for this with 0.22.0-insiders3 that should prevent the incorrect extra reload/installation prompts.

PR #3251, opened 2019-03-05
Fix install vsix.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3206 .
Also, fixed some minor issues with https://github.com/Microsoft/vscode-cpptools/pull/3221 .

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -40,7 +40,6 @@ let activatedPreviously: PersistentWorkspaceState<boolean>;
 const insiderUpdateTimerInterval: number = 1000 * 60 * 60;
 let buildInfoCache: BuildInfo | null = null;
 const taskSourceStr: string = "C/C++";
-const noBuildInfoStr: string = 'Failed to get build info';
 const cppInstallVsixStr: string = 'C/C++ install vsix: ';
 let taskProvider: vscode.Disposable;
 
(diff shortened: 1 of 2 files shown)


lh123 (commented):
I think `async function installVsix(vsixLocation: string, updateChannel: string): Promise<void> `
should be `function installVsix(vsixLocation: string, updateChannel: string): Promise<void> `

sean-mcmanus:
> I think async function installVsix(vsixLocation: string, updateChannel: string): Promise<void> 
> should be function installVsix(vsixLocation: string, updateChannel: string): Promise<void>

Yeah, I think whoever added the async might have been confused by the linter which shows "This maybe be converted to an async function" when the async is removed -- the Quick Fix code action for that converts the .then to an await.

Extension/src/LanguageServer/extension.ts
```diff
@@ -40,7 +40,7 @@ let activatedPreviously: PersistentWorkspaceState<boolean>;
 const insiderUpdateTimerInterval: number = 1000 * 60 * 60;
 let buildInfoCache: BuildInfo | null = null;
 const taskSourceStr: string = "C/C++";
-
+const cppInstallVsixStr: string = 'C/C++ install vsix: ';
```
bobbrow: I was thinking at some point we'd want to prefix our log messages with `cpptools:`.  Perhaps this would be a good place to start.  It could be `cpptools: Install VSIX: ` if you want to have a separate category here.

bobbrow (approved):

PR #3264, opened 2019-03-08
Seanmcm/add cl debug support

sean-mcmanus:


sean-mcmanus:
This works in the basic case. Still investigating if there are any special cases.

Extension/src/Debugger/extension.ts
```diff
@@ -67,6 +67,12 @@ export function initialize(context: vscode.ExtensionContext): void {
             if (!selection) {
                 return; // User canceled it.
             }
+            if (selection.label.startsWith("cl.exe")) {
+                if (!process.env.DevEnvDir || process.env.DevEnvDir.length === 0) {
+                    vscode.window.showErrorMessage('cl.exe build and debug is only usable when Code is run from the Developer Command Prompt for VS.');
```
bobbrow: Should it be "VS Code" (the product) or "code" (the executable) instead of "Code"?

Extension/src/LanguageServer/extension.ts
```diff
@@ -219,19 +219,22 @@ export async function getBuildTasks(): Promise<vscode.Task[]> {
         // Handle compiler args in compilerPath.
         let compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(compilerPath);
         compilerPath = compilerPathAndArgs.compilerPath;
-        const taskName: string = path.basename(compilerPath) + " build active file";
-        let args: string[] = ['-g', '${file}', '-o', path.join('${fileDirname}', '${fileBasenameNoExtension}' + (isWindows ? '.exe' : ''))];
+        const filePath: string = path.join('${fileDirname}', '${fileBasenameNoExtension}');
+        const compilerPathBase: string = path.basename(compilerPath);
+        const taskName: string = compilerPathBase + " build active file";
+        const isCl: boolean = taskName.startsWith("cl.exe");
+        let args: string[] = isCl ? [ '${file}', '/Zi' ] : ['-g', '${file}', '-o', filePath + (isWindows ? '.exe' : '')];
```
bobbrow: Do we want to add `/EHsc` and `/Fe:` (like I put in the [pending documentation](https://github.com/mikeblome/vscode-docs/blob/mb-cpp-updates/docs/cpp/cpp-config-msvc.md#create-a-build-task))?

bobbrow (changes requested):

Extension/src/Debugger/extension.ts
sean-mcmanus: Sure -- at one point the branding was "Code", but I don't know if they're still following that.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I can add /EHsc, but I don't think /Fe is needed, right?

sean-mcmanus: Oh, the docs use FE to make the .exe have a different name from the source...should I add anyway, even it even though it's not needed in this case?

sean-mcmanus:
Build and Debug fails if it's launched before the configuration is set -- looking into a fix now.

sean-mcmanus:
Oh -- F5 doesn't work with cl.exe. If this doesn't get in for 0.22.0, the fix for Build and Debug "too early" should get ported to another pull request.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -132,10 +132,10 @@ export class CppProperties {
     public get ConfigurationsChanged(): vscode.Event<Configuration[]> { return this.configurationsChanged.event; }
     public get SelectionChanged(): vscode.Event<number> { return this.selectionChanged.event; }
     public get CompileCommandsChanged(): vscode.Event<string> { return this.compileCommandsChanged.event; }
-    public get Configurations(): Configuration[] { return this.configurationJson.configurations; }
+    public get Configurations(): Configuration[] { return this.configurationJson ? this.configurationJson.configurations : null; }
```
bobbrow, line 135: Does this end up causing a null reference somewhere else? (as opposed to a undefined reference here)

bobbrow (approved):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: No.

sean-mcmanus:
F5 debugging works now, but it is buggy if the (Windows) launch is selected with a gdb debugger or the (gdb) launch is selected with a cl.exe debugger, as in we should be removing the available compilers from the list if they don't match the 1st selection...not sure if that is a ship blocking issue, since it could be considered "user error".

bobbrow (approved):

bobbrow (approved):

PR #3293, opened 2019-03-14
Disable the IntelliSense cache on Mac.

sean-mcmanus:


Extension/src/LanguageServer/client.ts
```diff
@@ -383,11 +383,13 @@ class DefaultClient implements Client {
         
         let intelliSenseCacheDisabled: boolean = false;
         if (os.platform() === "darwin") {
-            const releaseParts: string[] = os.release().split(".");
-            if (releaseParts.length >= 1) {
-                // AutoPCH doesn't work for older Mac OS's.
-                intelliSenseCacheDisabled = parseInt(releaseParts[0]) < 17;
-            }
+            intelliSenseCacheDisabled = true;
+            // TODO: Re-enable this after it works on Mac.
```
bobbrow: It "works", but it isn't fast.  Maybe update the comment to say that we'll re-enable this after we work around performance issues with mmap.

bobbrow (approved):

Issue #3300, opened 2019-03-15
Clarification on *.vsix offline installs


CBtiger:
you have many posts clarifiying the problem  that the standard *.vsix package causes errors at the first start because the cpptools extensions wants to download extensions. I think I read all your issues on that and I also found the Release folders where you provide the different version per OS.

What I do not understand is that nobody is asking why the main target version seems to be missing there. You provide
 * Linux32
 * Linux  (I take its Linux64)
 * MacOS
 * Win32

What about Win64? The majority of downloaders should run a Win X64 platform. I ran into the described problem with the standard *.vsix download. Now I am confused. Would I compromise my Win x64 system my installing the win32 version? I did not see a '*.vsix' uninstall...

You would help people a lot when addind a small section in your ReadMe.md about "Offline Install". Because its surprising to users that your extension is the only one that does **not** work for an offline case. Technically totally feasible that dependencies vary with the traget system, why not just telling people in the readme that everyone who downloaded the "standard" *.vsix file should consider to take one (or more) of those <link to releases>. If you could add a sentence what Win-64bit users should select than you would have made my day!

bobbrow:
Thanks for the feedback.  Adding a section to the readme on this is a good suggestion.  Our extension should be outputting an error message with instructions when it tries to download dependencies in the "offline" scenario, but it's possible that people might not see them.

Win32 binaries run on Win64, so all Windows users can install the same package.  But I think renaming these would be good to reduce the potential for confusion.

CBtiger:
You are welcome, glad if you find it useful.
I must say I am very impressed by your Extension and even more by your support. 

Your instructions included indeed. Just ran it again. It says to download the release vsix files. So that is great and very helpful. After reading the message I only hung at the x64bit naming issue. 

The very best solution imho is leave every file name as it is and add a section in your Readme.md. Its not too long to become unreadable. And its fine to put this section "Offline use" to the very end for such a non-standard use case. But I would really appreciate to understand the issue before hand and get the links just there. To any coder it should be easy to understand that 

>"..the tools are OS dependent and thus can not be resolved in the single vsix download. Thus initiual setup must be online or for pure offline use the correct files should be used <link>. Win32 is for x64 and x32."

(your wording will be better) That would be an outrageous documentation, indeed!

PR #3307, opened 2019-03-18
readme feedback

bobbrow:
#3300 
@CBtiger


CBtiger (approved):
Great. Even when people get stuck using the *.vsix from the VS Code main page, they will directly get the correct download here. Those who are not aware of the *.vsix download from the VS Code main page will not be confused and get the correct file right away. 
I think its a very good solution.

WardenGnaw (approved):

Issue #3309, opened 2019-03-18
${env} isn't processed in C_Cpp.default.includePath


ijash:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Linux mint 19.01
- VS Code Version: 1.32.3
- C/C++ Extension Version:0.21
- Other extension: arduino  0.2.25

I can't get the  `C_Cpp.default.IncludePath` get to work. please see details below.

**To Reproduce**

1. i have followed the instruction in https://github.com/Microsoft/vscode-cpptools/blob/master/Documentation/LanguageServer/Customizing%20Default%20Settings.md
2. my `settings.json` is 
```json
{    "C_Cpp.default.includePath": [
        "${env.HOME}/.arduino"
        ]
}
```
my `c_cpp_properties.json` is:
```json
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "/opt/arduino/arduino-1.8.8/tools/**",
                "/opt/arduino/arduino-1.8.8/hardware/arduino/avr/**",
                "${default}"
            ],
            "forcedInclude": [
                "/opt/arduino/arduino-1.8.8/hardware/arduino/avr/cores/arduino/Arduino.h"
            ],
            "intelliSenseMode": "gcc-x64",
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "c11",
            "cppStandard": "c++17"
        }
    ],
    "version": 4
}
```
mostly auto generated except `${default}`

3. i write in `main.ino` which use C++ language as follows:
```c
#include <MFRC522.h>


```
and got a problem `Include file not found in browse.path.`

4. BUT, when i move the `"${env.HOME}/.arduino"` to `c_cpp_properties.json` inside `configurations`->`includePath` array, the problem disappears.

**Expected behavior**

I expect something like global configuration or default configuration in `settings.json` so i don't have to re-write configuration each new project to `c_cpp_properties.json`. and my expectation is the same as issue https://github.com/Microsoft/vscode-cpptools/issues/368#issue-191469357
 
**Screenshots**

this is the current problem:
![image](https://user-images.githubusercontent.com/30673286/54560174-4cbfd900-49f4-11e9-875f-d0637267f147.png)

this is the behavior after i edit the `c_cpp_properties.json`:
![image](https://user-images.githubusercontent.com/30673286/54560843-d3c18100-49f5-11e9-8de7-947d25cac0a7.png)

**Additional context**

-none

[...]

PR #3310, opened 2019-03-18
Fix variable resolving not working with ${default}

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3309 .


sean-mcmanus:
@bobbrow You wrote this code 11 months ago -- was there a reason you didn't resolve the "${default}" before the other variables? I don't think we need to handle "${default}" being the result of an environment variable resolution, right?

bobbrow (approved):

bobbrow:
I think I might have done it so that I was only ever working with paths that don't have `;` in them.  If we want to prioritize resolving `${default}` over `${env:}`, I think that's fine.

PR #3312, opened 2019-03-18
Update changelog.

sean-mcmanus:


sean-mcmanus:
FYI, I removed the changes/fixes that only applied to the Insiders builds, i.e. stuff not affecting 0.21.0 users.

bobbrow (approved):

PR #3313, opened 2019-03-19
Fix squiggles with WSL compilerPath with args.

sean-mcmanus:


sean-mcmanus:
This change removes the `""` requirement around matching, which could lead to squiggles appearing at incorrect spots -- however, that seems okay, because the correct spot will have a squiggle as well and it isn't expected to occur too often. E.g. if "foo" was a relative includePath, then all instances of "foo" would squiggle, even "/usr/bin/foo". Maybe I could add the `""` requirement back, but it would break the `;` scenario.

UPDATE: I think I'll remove the `""` change.

bobbrow (approved):

PR #3328, opened 2019-03-20
Build tasks.json deletion. Configure Task.

sean-mcmanus:


Extension/src/common.ts
```diff
@@ -47,11 +49,13 @@ export function getRawTasksJson(): Promise<any> {
...
             }
-            const fileContents: Buffer = fs.readFileSync(path);
+            let fileContents: string = fs.readFileSync(path).toString();
+            fileContents = fileContents.replace(/^\s*\/\/.*$/gm, ""); // Remove start of line // comments.
             let rawTasks: any = {};
             try {
-                rawTasks = JSON.parse(fileContents.toString()); 
+                rawTasks = JSON.parse(fileContents); 
             } catch (error) {
+                return reject(failedToParseTasksJson);
```
bobbrow: when rejecting, it should be a `new Error(failedToParseTasksJson)`

Extension/src/Debugger/extension.ts
```diff
@@ -91,12 +91,15 @@ export function initialize(context: vscode.ExtensionContext): void {
                         await util.ensureBuildTaskExists(selection.configuration.preLaunchTask);
                         Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
                     } catch (e) {
+                        if (e === util.failedToParseTasksJson) {
```
bobbrow: if you make the e an `Error` instead of a raw string, then this needs to be updated.

bobbrow (approved):

PR #3359, opened 2019-03-25
Update for 0.22.2.

sean-mcmanus:


Extension/package.json
```diff
@@ -1431,7 +1431,7 @@
     "tslint-no-unused-expression-chai": "0.0.3",
     "typescript": "^2.9.2",
     "vrsource-tslint-rules": "^5.8.3",
-    "vscode": "^1.1.30"
+    "vscode": "^1.1.33"
```
bobbrow, line 1434: Why are we incrementing this?

bobbrow (changes requested):

Extension/package.json
sean-mcmanus: It's a bug fix update from running npm install.

bobbrow: As long as it doesn't update our min version dependency, then this is fine.

bobbrow (approved):

Issue #3367, opened 2019-03-26
C_Cpp.intelliSenseCachePath doesn't resolve tilde or env for home directory


rewrking:
**Type: LanguageService**

**Describe the bug**
- OS and Version: All (tested on Win/Mac)
- VS Code Version: 1.32.3
- C/C++ Extension Version: 0.22.1
- Other extensions you installed (and if the issue persists after disabling them): None
- A clear and concise description of what the bug is.
"C_Cpp.intelliSenseCachePath" setting does not respect the tilde ~ as the home directory path, nor does it take ${env:HOME}, ${env.HOME}, ${HOME}... ${env:HOME} I think is the correct syntax, but wanted to cover my bases.

**To Reproduce**
1. Open a terminal (use git-bash on windows), go to ~/.vscode and type "mkdir cpptools-cache", or create the folder through explorer/finder
1. Go into settings.json, add "C_Cpp.intelliSenseCachePath" : "~/.vscode/cpptools-cache" 
or "C_Cpp.intelliSenseCachePath" : "${env:HOME}/.vscode/cpptools-cache"
3. Reload the window with a project open
4. Delete the ipch folder currently in the project directory
5. Open some cpp files
6. Notice the ipch folder generates in the project directory's .vscode folder again.

**Expected behavior**
The extension should resolve the home directory and place its ipch stuff there instead of the project directory's .vscode folder. Instead, it does nothing, and cache files continue to be created in the project's .vscode folder.

**Additional context**
This is a pretty annoying. Personally, I'd prefer that the extension didn't clutter up my project folders with cache files and instead default to some other folder I don't know of or care about (Roaming for instance), instead of having to configure yet another setting to get them out of the project directory.

sean-mcmanus:
Environment variables in the intelliSenseCachePath should expand with the latest Insiders: https://github.com/Microsoft/vscode-cpptools/releases/tag/0.22.2-insiders .

rewrking:
Works great. Thanks!

PR #3368, opened 2019-03-27
Enable environment variables with intelliSenseCachePath.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3367 . Requires a language server change too.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -476,7 +476,8 @@ class DefaultClient implements Client {
             if (changedSettings["clang_format_path"]) {
                 let settings: CppSettings = new CppSettings(this.RootUri);
                 this.languageClient.sendNotification(UpdateClangFormatPathNotification, util.resolveVariables(settings.clangFormatPath, this.AdditionalEnvironment));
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -475,6 +476,9 @@ class DefaultClient implements Client {
             if (changedSettings["clang_format_path"]) {
                 let settings: CppSettings = new CppSettings(this.RootUri);
                 this.languageClient.sendNotification(UpdateClangFormatPathNotification, util.resolveVariables(settings.clangFormatPath, this.AdditionalEnvironment));
+            } else if (changedSettings["intelliSenseCachePath"]) {
```
bobbrow: I don't think it should be `else if `. You can change multiple things at once, right?

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, good catch.

sean-mcmanus:
@bobbrow Can you approve this?

bobbrow (approved):

PR #3377, opened 2019-03-28
Fix for null ref race condition when using custom config providers.

sean-mcmanus:


Extension/src/LanguageServer/client.ts
```diff
@@ -312,23 +312,13 @@ class DefaultClient implements Client {
 
             // requests/notifications are deferred until this.languageClient is set.
             this.queueBlockingTask(() => languageClient.onReady().then(
-                () => {
+                async () => {
```
bobbrow: We shouldn't mix `async` with `then`. Can the whole function be switched to `async`?

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Umm...this isn't really mixing async and then, because the lambda is async, right? I could try to make the main function async still though.

bobbrow: You're using async inside the then.

sean-mcmanus: Yeah, that's just declaring the lambda as async -- the async lambda itself just uses await and not any "then". Are you saying that it still counts as "mixing"?

sean-mcmanus: @bobbrow What do you want me to change?

bobbrow: The only thing this change is trying to do is change the order of operations, right?  If that's the case you don't need to convert the lambda to async/await.

sean-mcmanus: I just changed the await to return the promise. We could switch everything to await later.

bobbrow (approved):

PR #3393, opened 2019-04-02
Add use of queryTranslationUnitSource to determine if custom configuration can be skipped

Colengms:
Add call in provideCustomConfiguration to new QueryTranslationUnitSourceRequest.  If a system header or there is already a TU loaded, skip custom configuration.


Extension/src/LanguageServer/client.ts
```diff
@@ -604,6 +616,16 @@ class DefaultClient implements Client {
     }
 
     public async provideCustomConfiguration(document: vscode.TextDocument): Promise<void> {
+
+        let params: QueryTranslationUnitSourceParams = {
+            uri: document.uri.fsPath
```
bobbrow: I believe `fsPath` will get the converted path (e.g. c:\source\1.cpp) instead of the URI (e.g. file:///c:/source/1.cpp).
I believe `path` is what you want here.

bobbrow: (though we also talked about renaming `uri` to `path`, so let's just pick the one that matches)

bobbrow:
>                     this.sendCustomConfigurations(configs, true);

in the header case, we should send the configs for both header and source files.

---
Refers to: Extension/src/LanguageServer/client.ts:666 in e5a3c39. [](commit_id = e5a3c39789f601f5514cbb25169d47f407ee4019, deletion_comment = False)

Extension/src/LanguageServer/client.ts
```diff
@@ -113,10 +113,22 @@ interface CompileCommandsPaths {
     paths: string[];
 }
 
+interface QueryTranslationUnitSourceParams {
+    uri: string;
+}
+
+interface QueryTranslationUnitSourceResult {
+    uri: string;
+    is_header: boolean;
```
bobbrow: consider using an enum to describe the cases we want to handle.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: Elsewhere in the file, it looks like toString() is used.  Switched to toString().  Now passing uri's to/from (as strings).

Extension/src/LanguageServer/client.ts
```diff
@@ -604,6 +632,16 @@ class DefaultClient implements Client {
     }
 
     public async provideCustomConfiguration(document: vscode.TextDocument): Promise<void> {
+
```
bobbrow: nit: remove empty newline here

bobbrow (approved):
We should wait until the language server change is ready before we commit this, but this change is basically ready to go.

PR #3404, opened 2019-04-03
Ensure clearCustomConfigurations() is called when changing the configurationProvider

Colengms:
Clears custom configurations when the configurationProvider is changed.


sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1206,6 +1206,9 @@ class DefaultClient implements Client {
         }).then(() => {
             let newProvider: string = this.configuration.CurrentConfigurationProvider;
             if (this.configurationProvider !== newProvider) {
+                if (this.configurationProvider !== undefined) {
```
sean-mcmanus: Should this be `if (this.configurationProvider)`?

PR #3406, opened 2019-04-04
Add call to notifyWhenReady, to prevent crash

Colengms:
Addresses an issue with a prior change, which added use of languageClient when it may not yet be available.

### DIFF
--- a/Extension/src/LanguageServer/protocolFilter.ts
+++ b/Extension/src/LanguageServer/protocolFilter.ts
@@ -23,12 +23,10 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
         didOpen: (document, sendMessage) => {
             if (clients.checkOwnership(me, document)) {
                 me.TrackedDocuments.add(document);
-                me.notifyWhenReady(() => {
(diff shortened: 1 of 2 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,10 +23,12 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
         didOpen: (document, sendMessage) => {
             if (clients.checkOwnership(me, document)) {
                 me.TrackedDocuments.add(document);
-                me.provideCustomConfiguration(document).then(() => {
-                    sendMessage(document);
-                }, () => {
-                    sendMessage(document);
+                me.notifyWhenReady(() => {
```
bobbrow: I think you should await `requestWhenReady` on the QTUS request inside `provideCustomConfiguration` in client.ts (line 643) instead of using `notifyWhenReady` here.

bobbrow (changes requested):

sean-mcmanus:
I verified the fix.

Issue #3088, opened 2019-01-24
Reprompt for the configuration provider if c_cpp_properties.json is deleted


sean-mcmanus:
Suggested in https://github.com/Microsoft/vscode-cpptools/issues/2843#issuecomment-457136598 .

Maybe we can add other code to detect/prompt users for a missing configurationProvider setting as well.

Didn't make 0.23.0-insiders.

bobbrow:
This was addressed along with #2346 in 0.27.0.

sean-mcmanus:
@bobbrow Can you clarify? That issue seems unrelated. I don't recall seeing a fix for this. I had PR with the fix, but then I never got finished with my testing of it.

sean-mcmanus:
Ok, I see the fix now.

PR #3407, opened 2019-04-04
Reprompt on c_cpp_properties.json deletion.

sean-mcmanus:
Fix for #3088.


sean-mcmanus:
I'm not done testing this -- there may be some bugs, but now I'm not sure how I reproed them yet.

sean-mcmanus:
c_cpp_properties.json gets created from the prompt if it doesn't exist, but for the configurationProvider prompt it adds it to the default.configurationProvider -- is that intentional? The difference in behavior confused me.

sean-mcmanus:
I can't full test this until the bug with c_cpp_properties.json deletion causing creation to fail is fixed.

sean-mcmanus:
This pull request is just needs further testing -- the last time I tried to test it, another bug was interfering with my testing, but that got fixed and I never got around to re-test.

ThadHouse:
This won't pop up if settings.json is used to set defaults, right? We use the provider API and have noticed random bugs that happen when the API is used and a user accidentally adds c_cpp_properties.json without understanding. We always recommend our users delete that file if they see it, and I don't want this to be come a feedback loop.

sean-mcmanus:
> This won't pop up if settings.json is used to set defaults, right? We use the provider API and have noticed random bugs that happen when the API is used and a user accidentally adds c_cpp_properties.json without understanding. We always recommend our users delete that file if they see it, and I don't want this to be come a feedback loop.

I'm not sure, but when I have time to revisit this, I'll double check that the config prompts don't appear if the configuration provider (or compileCommands) is already set via the settings.json.

bobbrow:
This was addressed with PR #4987.

Issue #3213, opened 2019-02-25
Extension causes high cpu load


germinolegrand:
- Issue Type: `Performance`
- Extension Name: `cpptools`
- Extension Version: `0.21.0`
- OS Version: `Linux x64 4.9.0-8-amd64`
- VSCode version: `1.31.1`

[ms-vscode.cpptools-unresponsive.cpuprofile.txt](https://github.com/Microsoft/vscode-cpptools/files/2899824/ms-vscode.cpptools-unresponsive.cpuprofile.txt)

Find more details here: https://github.com/Microsoft/vscode/wiki/Explain:-extension-causes-high-cpu-load

WardenGnaw:
From the .cpuprofile, it is taking a long time to load a module. VS Code's solution is https://github.com/Microsoft/vscode/wiki/Explain:-extension-causes-high-cpu-load#code-loading

sean-mcmanus:
@WardenGnaw So we should be using WebPack? Do we know how severe of an issue this is, i.e. how much of a benefit would be gained?

WardenGnaw:
@sean-mcmanus We would have to do profiling on WebPacked code vs non-WebPacked code and see what the improvement is.

germinolegrand:
Same notification while loading the same code folder (which is quite huge, tens of thousands of files, that might be of importance ?).

[ms-vscode.cpptools-unresponsive.cpuprofile.txt](https://github.com/Microsoft/vscode-cpptools/files/2911569/ms-vscode.cpptools-unresponsive.cpuprofile.txt)

WardenGnaw:
Time is still mostly spent doing module loads. Profiler is saying it is taking 2500 ms.

philkuz:
I noticed in the messages that @sean-mcmanus is leaving on duplicate issues, there's apparently a fix to this issue in the upcoming release. Is there a branch where this is currently located? 

Thanks!

WardenGnaw:
@philkuz The branch would be https://github.com/Microsoft/vscode-cpptools/pull/3411

sean-mcmanus:
Should be fixed with 0.23.0-insiders2: https://github.com/Microsoft/vscode-cpptools/releases/tag/0.23.0-insiders2

larshb:
Version 1.40.2

![image](https://user-images.githubusercontent.com/8710787/70141907-524f7300-1698-11ea-9012-98d5e3e1b047.png)

[...]

PR #3411, opened 2019-04-04
Bobbrow/webpack

sean-mcmanus:
Debugging doesn't work. Something wrong with the source mapping I guess.


Extension/src/LanguageServer/client.ts
```diff
@@ -336,6 +339,7 @@ class DefaultClient implements Client {
                     // Listen for messages from the language server.
                     this.registerNotifications();
                     this.registerFileWatcher();
+                    console.log('registered stuff');
```
bobbrow: probably don't want to keep these logs. :)

sean-mcmanus: Oh, I thought those logs were intentional. Although "registered stuff" seemed odd.

sean-mcmanus:
Looks like this breaks the Azure Pipeline build...even though building works locally. Error: Cannot find module './out/tools/GenerateOptionsSchema'

WardenGnaw (commented):
I wouldn't webpack src/tools, since it is not used for the released VsCode extension.

I would add another agent build for testing webpacked builds and keep our normal just tsc builds since we do not want webpacking for regular dev workflow. 

Add Extension/dist/* to .gitignore

bobbrow:
When the Config UI goes in, we'll need to update this to make sure all the assets are in the right place.

Extension/tools/prepublish.js
```diff
@@ -10,23 +10,36 @@
...
 }
 
-// Compile the TypeScript code
-console.log(">> tsc -p ./");
-cp.execSync("tsc -p ./", {stdio:[0, 1, 2]});
+const tscCompileListStr = fs.readFileSync("./tscCompileList.txt").toString();
+
+tscCompileListStr.split(os.EOL).forEach(filePath => {
+    if (!filePath.startsWith("#") && /\S/.test(filePath))
+    {
```
bobbrow: move curly brace to previous line

Extension/tscCompileList.txt
```diff
@@ -0,0 +1,13 @@
+# This file is consumed in tools\prepublish.js
+# '#' at the start of the line is treated as a comment and is ignored.
+# whitespace empty lines are also ignores.
+#
+# This file is to be used if you do not wish to webpack a .ts file
+# If you do not want it to be webpacked:
+# 1. Please make sure it is not in the src directory
```
bobbrow, line 7: oh... I was thinking we would want to move the settings UI into the src folder at some point.  Not a big deal, I guess, but we can revisit this when we make that call.

Extension/package.json
```diff
@@ -1472,11 +1474,13 @@
...
     "mkdirp": "~0.5.1",
     "tmp": "~0.0.33",
-    "vscode-cpptools": "2.1.1",
+    "vscode-cpptools": "^2.1.2",
     "vscode-debugadapter": "~1.33.0",
     "vscode-debugprotocol": "~1.33.0",
     "vscode-extension-telemetry": "^0.1.1",
     "vscode-languageclient": "5.1.1",
+    "webpack": "^4.29.6",
+    "webpack-cli": "^3.2.3",
```
bobbrow, line 1483: Do these additions generate a new 3rd party licenses file that we need to check in?

bobbrow (approved):

Extension/package.json
WardenGnaw: These should actually be under devDependencies. I don't think we have to ship any webpack code.

Extension/tscCompileList.txt
WardenGnaw: IMO and with the addition of UI, we should move towards more of a [webpage](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/Dealing_with_files) styled folder structure. 

E.g.
```
dist/
   main.js
   *.html
   images/
   styles/
   scripts/
```

Extension/package.json
sean-mcmanus: Oops, I didn't see that this change wasn't made. I'll submit a fix.

philkuz:
Linking #3213 as it is fixed with this merge.

PR #3421, opened 2019-04-05
Default intelliSenseEngineFallback to Disabled.  Add additional settings value to errorSquiggles

Colengms:


Extension/package.json
```diff
@@ -135,7 +135,8 @@
           "type": "string",
           "enum": [
             "Enabled",
-            "Disabled"
+            "Disabled",
+            "MissingIncludesOnly"
```
bobbrow: Perhaps `EnabledWhenIncludesResolve`?  `MissingIncludesOnly` sounds like only `#include` squiggles will ever be shown.

Extension/package.json
```diff
@@ -135,7 +135,8 @@
...
           ],
           "default": "Enabled",
           "description": "Controls whether suspected compile errors detected by the IntelliSense engine will be reported back to the editor. Warnings about #includes that could not be located will always be reported to the editor. This setting is ignored by the Tag Parser engine.",
```
bobbrow: We'll need to update the description and change the default to the new enum value that we're adding.

Extension/src/LanguageServer/extension.ts
```diff
@@ -812,7 +812,7 @@ function onToggleSquiggles(): void {
     onActivationEvent();
     // This only applies to the active client.
     let settings: CppSettings = new CppSettings(clients.ActiveClient.RootUri);
-    settings.toggleSetting("errorSquiggles", "Enabled", "Disabled");
+    settings.toggleSetting("errorSquiggles", "Disabled", "Enabled");
```
bobbrow: Since we have 3 states, we need to decide which 2 we toggle, or if we need to rework this command.  (I don't think toggle will work if they have the other state set)

bobbrow (changes requested):

Extension/package.json
```diff
@@ -135,10 +135,11 @@
...
           "enum": [
             "Enabled",
-            "Disabled"
+            "Disabled",
+            "IfIncludesResolve"
           ],
-          "default": "Enabled",
-          "description": "Controls whether suspected compile errors detected by the IntelliSense engine will be reported back to the editor. Warnings about #includes that could not be located will always be reported to the editor. This setting is ignored by the Tag Parser engine.",
+          "default": "IfIncludesResolve",
+          "description": "Controls whether suspected compile errors detected by the IntelliSense engine will be reported back to the editor.",
```
bobbrow: can you keep the part about how the setting is ignored by the tag parser engine?

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
bobbrow: Maybe we just remove the toggle

Extension/package.nls.zh-cn.json
```diff
@@ -6,7 +6,8 @@
     "c_cpp.command.peekDeclaration.title": "查看声明",
     "c_cpp.command.switchHeaderSource.title": "切换头文件/源文件",
     "c_cpp.command.navigate.title": "导航...",
-    "c_cpp.command.toggleErrorSquiggles.title": "切换错误下划线",
+    "c_cpp.command.enableErrorSquiggles.title": "切换错误下划线",
```
bobbrow: We need an updated translation for these

bobbrow (changes requested):

bobbrow (approved):

PR #3424, opened 2019-04-06
Update FAQs.md

bobbrow:


michelleangela (approved):

michelleangela (dismissed):
It might be good to also add "How to disable IntelliSense Cache (ipch cached files)"

Documentation/FAQs.md
```diff
@@ -15,6 +16,20 @@ The file **launch.json** will now be open for editing with a new configuration.
 
 See the [**Documentation/Debugger**](https://github.com/Microsoft/vscode-cpptools/tree/master/Documentation/Debugger) folder in this repository for more in-depth documentation on how to configure the debugger.
 
+## What is the .vscode/ipch folder?
```
sean-mcmanus, line 20: I think this should be split into 2 additional questions that answer: 
1. How do I disable the ipch folder?
2. How do I move the ipch folder?
Which is similar to the sections describing the settings.

Oh, I didn't see Michelle's comment -- that was the main issue we were talking about.

Issue #3419, opened 2019-04-05
Clang_format_style doesn't work if trying to make custom style in settings box


a5ehren:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Oracle Linux 7.6
- VS Code Version: 1.33.0
- C/C++ Extension Version: 0.22.1
- Other extensions you installed (and if the issue persists after disabling them):
- Description: Settings dialog is changing everything to all-lower case but clang-format is case-sensitive

**To Reproduce**
1) Settings
2) C_Cpp: Clang_format_style
3) Enter "{BasedOnStyle: Google, ColumnLimit: 0}" and save
4) Try to format any C/C++ file with the new command

**Expected behavior**
The explanation text (taken straight from clang-format, I notice) implies that entering a custom style in the settings box should work, but it does not. I'd rather this work at the editor level than having to create a .clang-format file for this simple case.

**Additional context**
Error message from the console:
Formatting failed:
/home/ebendler/.vscode/extensions/ms-vscode.cpptools-0.22.1/bin/../LLVM/bin/clang-format -style="{basedonstyle: google, columnlimit: 0}" -fallback-style=google -assume-filename=<my path>
  Invalid value for -style

When I run clang-format from that location with the correct capitalization, it does not error out.

sean-mcmanus:
Remove the quotes around the `{}`. We've had other users hit the same problem. We should remove the `"` in the description to avoid confusing users. The problems got worse when VS Code starting switching to the UI instead of json for the setting, i.e. the quotes is part of the json and not used in the UI.

a5ehren:
Ah, yeah the UI text is what got me here. 

Doubly confusing because the clang-format help text has the quotes as well for shell-escaping reasons so it makes the UI text appear correct as well.

sean-mcmanus:
UPDATE: Our latest Insiders has the quotes removed and a better example: https://github.com/Microsoft/vscode-cpptools/releases/tag/0.22.2-insiders -- the example may not 100% match the Visual Studio style because some of the values are inherited dynamically based on other Visual Studio Code settings.

PR #3435, opened 2019-04-09
Seanmcm/small changes for insiders

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3419 -- previous users hit the same confusing issue with quotes. Also, changed the example to match the "Visual Studio" style, so it's easier for users to copy/paste that and add their own tweaks.


sean-mcmanus:
I'm checking this in now so I can build a vsix. If someone wants changes we can do that later.

Issue #3449, opened 2019-04-11
problem with c_cpp_properties.json


mamo5298:
Type: General

**Describe the bug**
- OS and Version: windows 10 1803
- VS Code Version: 133.0
- C/C++ Extension Version: from 0.22.0
- just as i open vs code a message appears says that there is error when parsing c_cpp_properties.json because of double backslash ended path string, and then intellisense will not work any more, when i edit the json file to avoid the error, the extension will automatically update the file after compiling the project and the problem comes back.
i am currently working on extension version 0.21.0 where everything is fine but i have to reinstall this version everytime i start vs code because the extension will update automatically to the latest version.

**To Reproduce**
1. install c/c++ v0.22.0 or later
2. use c_cpp_properties.json file with \\\\ ended include or browse path entry, example:
"C:\\\\Users\\\\mm\\\\Desktop\\\\Sampler32\\\\trunk\\\\common\\\\",
3. compile or simply open the project which has this file

**Expected behavior**
with the version v0.21.0 the json file works as it is, and everything is fine

**Screenshots**
![Unbenannt](https://user-images.githubusercontent.com/49508578/55934773-faa85700-5c31-11e9-8043-e08fbef9a87d.PNG)

**Additional context**
i am developing an embedded systems project, therefore i am using IAR Embedded extension for compiling, and cortex-debug extension for debugging as well as SVN extension for code syncing.

sean-mcmanus:
Sorry about that. We have a fix for our next release: https://github.com/Microsoft/vscode-cpptools/pull/3453 .

sean-mcmanus:
The parse error should be fixed with 0.23.0-insiders.

PR #3453, opened 2019-04-11
Seanmcm/fix json parse

sean-mcmanus:
Fix for 0.22.0 regression caused by squiggle work: https://github.com/Microsoft/vscode-cpptools/issues/3449 .


Extension/src/LanguageServer/configurations.ts
```diff
@@ -632,6 +632,7 @@ export class CppProperties {
             // Otherwise, the JSON.parse result will have the \<escape character> missing.
             readResults = readResults.replace(/\\/g, '\\\\');
             readResults = readResults.replace(/\\\\"/g, '\\"'); // Need to revert the change to \".
```
WardenGnaw: Is it possible to combine with the first replacement of `\"` with the correct number of `\` instead of having it done on a separate call?

sean-mcmanus: Yeah -- I thought of that afterwards.

sean-mcmanus: Let me know if you see any issue with the new regex.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -630,8 +630,7 @@ export class CppProperties {
 
             // Replace all \<escape character> with \\<character>.
             // Otherwise, the JSON.parse result will have the \<escape character> missing.
-            readResults = readResults.replace(/\\/g, '\\\\');
-            readResults = readResults.replace(/\\\\"/g, '\\"'); // Need to revert the change to \".
+            readResults = readResults.replace(/([^\\])\\([^\"\\])/g, '$1\\\\$2'); // Exclude \" and \\.
```
WardenGnaw: nit: An explanation of the complicated regex.

sean-mcmanus: There's an explanation at the top and the side. Do you want me to re-organize the comment or add more?

WardenGnaw: I guess its fine.

WardenGnaw (approved):

Extension/src/LanguageServer/configurations.ts
bobbrow: This doesn't fix any >2 slashes `\\\` or incorrect end of string slashes `\"`

sean-mcmanus: The cases you mentioned were not incorrect, but the `\\` case was broken, leading to no squiggles. I reverted it back to my original fix.

michelleangela (approved):

Issue #3452, opened 2019-04-11
Intellisense on split panes returns information about the currently active project, not the currently moused-over pane.


JarrettR:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10
- VS Code Version: 1.33.0
- C/C++ Extension Version: 0.22.1
- A clear and concise description of what the bug is:

Intellisense on split panes returns information about the currently active project, not the currently moused-over pane.

**To Reproduce**

1.  Set up a c++ project, copy the entire folder and name the two `project_left` and `project_right`
2. Open a new, empty workspace in VSCode
3. Add both folders to the workspace
4. Open up `main.cpp` for both projects in the left and right panes respectively, using the `Open to the side` or `split right` or whatever your preferred method is.
5. In another file (eg. main.h), `#define VAR 1` in the left project, and `#define VAR 2` on the right.
6. Use it in main() / compile, etc. Anything to get Intellisense to start reading the define. Might need to close and open VSCode.
7.  While the left pane is active, mousing over the `VAR` define will display "1" on both sides.
8. When the right pane is active (ie. click somewhere in the right pane's code window), `VAR` will display "2" on both sides

**Expected behavior**

The mouseover info for `VAR` should always display "1" on the left, and "2" on the right

**Screenshots**

#defines in the .h file of left and right projects:

![image](https://user-images.githubusercontent.com/6027644/55979310-b6738180-5c46-11e9-916e-dcdc06966ddc.png)

Test function of left and right projects:
![image](https://user-images.githubusercontent.com/6027644/55979370-dacf5e00-5c46-11e9-9ba2-b4a0fa6c9778.png)

Left pane is active, mouseover of left #define (This is proper and expected:
![image](https://user-images.githubusercontent.com/6027644/55979437-018d9480-5c47-11e9-88bd-ca4fcee1b0d4.png)

Left pane active, mouseover on right (This is wrong):

![image](https://user-images.githubusercontent.com/6027644/55979465-136f3780-5c47-11e9-8c63-a02d9017f7f7.png)

For completion, right pane is active, and mouseover on right is correct:

![image](https://user-images.githubusercontent.com/6027644/55979689-a314e600-5c47-11e9-8cea-225909d6ca8b.png)

And left is incorrect:

![image](https://user-images.githubusercontent.com/6027644/55979626-7e207300-5c47-11e9-8be3-5dad687f1b9c.png)

**Additional context**

[...]

PR #3455, opened 2019-04-12
Fix hover using the active client instead of the owning client.

sean-mcmanus:
Fix for https://github.com/Microsoft/vscode-cpptools/issues/3452 .


Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -49,7 +49,12 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
 
         provideCompletionItem: invoke4,
         resolveCompletionItem: invoke2,
-        provideHover: invoke3,
+        provideHover: (document, position, token, next: (document, position, token) => any) => {
+            if (clients.checkOwnership(me, document)) {
```
bobbrow, line 53: Are there any other lsp messages that need to be filtered in this way?

sean-mcmanus: Maybe later -- the formatting scenario is pretty busted with multi-root, but there are a bunch of issues on that for later (not worth dealing with now).

bobbrow (approved):

PR #3478, opened 2019-04-16
Seanmcm/fix reg exp in config name

sean-mcmanus:
Config squiggles would fail if the config name had special characters (I hit this when using "g++" in the name).


Extension/package.json
```diff
@@ -1455,7 +1455,8 @@
     "vscode-debugprotocol": "~1.33.0",
     "vscode-extension-telemetry": "^0.1.1",
     "vscode-languageclient": "5.1.1",
-    "yauzl": "~2.10.0"
+    "yauzl": "~2.10.0",
+    "escape-string-regexp": "1.0.5"
```
bobbrow: nit: I think we sort these

bobbrow (approved):

Extension/package.json
sean-mcmanus: I reordered it.

PR #3487, opened 2019-04-18
Add styling to configuration UI

michelleangela:


michelleangela:
dark
light
high-contrast

michelleangela:
light
high-contrast
dark

michelleangela:
Note: UI is currently only available in English

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+                <div style="font-size: 20px; font-weight: 400; line-height: 24px">Microsoft<br>C/C++ Extension</div>
+            </td>
+          </table>
+    </div>
+
+    <hr>
+    <div style="height: 30px; display: block"></div>
+
+    <div style="padding-right: 10px;">
+        Command palette to open this editor<br>
```
bobbrow: I wouldn't use the word "palette" in these descriptions.  I would say something like: "Open this editor by running the following command:"

(same goes for the (JSON) one)

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+        <div class="codeblock">C/C++: Edit configurations (UI)</div>
```
bobbrow: Should it just be `<code>` instead?  I was thinking you were going to use an outline or something for High Contrast mode, but this doesn't seem to do it based on the screenshots you shared.

bobbrow: Should they also be href'ed?

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+
+    <hr>
+    <div style="height: 30px; display: block"></div>
+
+    <!-- sections -->
+    <div class="section">
+      <div class="section-title">Configuration identifier</div>
+      <div class="section-text">
+        The identifier of the active configuration. <code>Mac</code>, <code>Linux</code>, and <code>Win32</code> are special identifiers for configurations that will be auto-selected on those platforms, but the identifier can be anything.
+      </div>
```
bobbrow, line 374: I'm wondering if we can href the command to change the active configuration here so that users can easily switch the config.  Would that break the UI?

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+        </div>
+        <div class="section-input">
+            <input id="compilerPath" style="width: 666px"></textarea> 
+        </div>
+    </div>
+
+    <div class="section">
+        <div class="section-title">IntelliSense mode</div>
+        <div class="section-text">
+            If set, it overrides the default mode used by the IntelliSense engine. Windows defaults to msvc-x64, Linux defaults to gcc-x64, and Mac default to clang-x64.
```
bobbrow: "Mac defaults"  (missing the 's')

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
-    </td>
-  </tr>
-  </table>
+        </div>
+    </div>
+
+    <div class="section">
+        <div class="section-title">Include path</div>
+        <div class="section-text">
+            A list of paths for the IntelliSense engine to use while searching for included headers. If a path ends with <code>/**</code> the IntelliSense engine will do a recursive search for includes starting from that directory.
```
bobbrow: In the second sentence: "includes" -> "headers"

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+        </div>
+        <div class="section-input">
+            <div class="section-note">One definition per line.</div>
+            <textarea id="defines" rows="4" cols="39"></textarea>
+        </div>
+    </div>
+
+    <div class="section">
+        <div class="section-title">C Standard</div>
+        <div class="section-text">Version of the C language standard to use for IntelliSense.</div>
```
bobbrow: "The version ..."

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+                <option value="c11">c11</option>
+                <option value="c99">c99</option>
+                <option value="c89">c89</option>
+              </select>
+        </div>
+    </div>
+
+    <div class="section">
+        <div class="section-title">C++ Standard</div>
+        <div class="section-text">Version of the C++ language standard to use for IntelliSense.</div>
```
bobbrow: "The version ..."

Extension/ui/settings.html
```diff
@@ -3,129 +3,411 @@
...
+    <hr>
+    <div style="height: 30px; display: block"></div>
+
+    <!-- sections -->
+    <div class="section">
+      <div class="section-title">Configuration identifier</div>
+      <div class="section-text">
+        The identifier of the active configuration. <code>Mac</code>, <code>Linux</code>, and <code>Win32</code> are special identifiers for configurations that will be auto-selected on those platforms, but the identifier can be anything.
+      </div>
+      <div class="section-input"><input id=activeConfig style="width: 290px"></div>
```
bobbrow: Are users allowed to rename the config here?

bobbrow (dismissed):

Extension/ui/settings.html
michelleangela: Yes, they can rename the config here the same as they can rename it in .json

Extension/ui/settings.html
michelleangela: `<code>` would be best for high contrast. Changing to `<code>`. I like the idea of href that opens the command palette for them. Similar to how handleConfigurationSelectCommand()

Extension/ui/settings.html
michelleangela: I added a simple link to the command to select a different configuration.

Extension/ui/settings.html
michelleangela: I check adding hrefs for release.

Extension/ui/settings.html
michelleangela: Changed this field to ready-only and added a link to select a different configuration. If user changes the name, the new name does not show up in the configuration list. Will need to add logic to update the list if there is a rename.

sean-mcmanus (approved):

Issue #3490, opened 2019-04-18
no environment file support for MSVC debug configuration


kejxu:
Type: Debugger

for a common launch configuration:
```json
{
    "name": "(msvc) Launch",
    "type": "cppvsdbg",
    "request": "launch",
    "program": "${workspaceFolder}/out.exe",
    "args": [],
    "stopAtEntry": true,
    "cwd": "${workspaceFolder}",
    "environment": [],
    "externalConsole": false
}
```

is there any plan to add support for environment variable definitions file (similar concept in [python debugger](https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file)) aside from the current `environment` key-value list?

sean-mcmanus:
Should be fixed with 0.23.0-insiders2: https://github.com/Microsoft/vscode-cpptools/releases/tag/0.23.0-insiders2

kejxu:
Thanks for updating on this @pieandcakes! We are currently staffing at //build, and will be testing this fix out afterwards. Would you want us to close this issue once we verify the fix?

pieandcakes:
@kejxu We will close once we ship the official release. I closed it prematurely.

sean-mcmanus:
@kejxu 0.23.0 shipped yesterday with the fix. If there's an issue still we could re-open this.

PR #3500, opened 2019-04-20
Adding support for .env files for cppvsdbg

WardenGnaw:
This PR allows users to have a .env file that listed as
KEY=VALUE or KEY="VALUE WITH SPACE" to be used as environment variables
when launching a process. Anything that begins with '#' is treated as a
comment. Blank lines are ignored.

This .env file will be used if envFile passed in with the launch.json.
It will also work with the existing environment variable, however
anything listed in the .env file will overwrite anything from the
environment field in the launch.json.

@gregg-miskelly

### DIFF
--- a/Extension/package-lock.json
+++ b/Extension/package-lock.json
@@ -2109,10 +2109,6 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-<<<<<<< HEAD
-=======
-          "dev": true,
->>>>>>> 98912ec... Adding support for .env files for cppvsdbg
           "optional": true
         },
         "aproba": {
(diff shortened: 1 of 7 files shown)


WardenGnaw:
Addresses https://github.com/Microsoft/vscode-cpptools/issues/3490

michelleangela:
Is this targeting 0.23.0-insiders?

Extension/package-lock.json
```diff
@@ -1569,7 +1569,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
sean-mcmanus: Why did this change?

WardenGnaw: No idea.

WardenGnaw:
@michelleangela there’s no big rush for this PR

chuckries (approved):

Extension/package.json
```diff
@@ -1201,6 +1201,11 @@
                 },
                 "default": []
               },
+              "envFile": {
+                "type": "string",
+                "description": "Environment variables passed to the program by a file.",
```
bobbrow: Should there be more description about the format of the file here, or are users expected to know what this file is?

WardenGnaw: We will need to update the vscode docs or add to launch.md with a similar description in https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file.

Any preference?

bobbrow: No preference. If you move it to the vscode docs, you may want to mention that in the description here.

WardenGnaw: Added to launch.md

sean-mcmanus:
@WardenGnaw Can you check this in by our 0.23.0-insiders2 on Friday? We want the next Insiders to be as close to the final release as possible.

PR #3519, opened 2019-04-24
Update Select Configurations command to offer Edit Configs JSON option

Colengms:
The "Select Configuration..." command will display a list of configurations to choose between, or an option to "Edit Configurations...".

Settings UI uses this Select Configurations command to allow the user to change which configuration is currently being edited.

Because the user is likely to select the option to edit configurations when the configuration they want to select is not on the list, and because settings UI does not yet allow new configurations to be added, we should offer "Edit Configurations (JSON)" in the selection list instead.

### DIFF
--- a/Extension/src/LanguageServer/ui.ts
+++ b/Extension/src/LanguageServer/ui.ts
@@ -171,7 +171,7 @@ export class UI {
         for (let i: number = 0; i < configurationNames.length; i++) {
             items.push({ label: configurationNames[i], description: "", index: i });
         }
-        items.push({ label: "Edit Configurations...", description: "", index: configurationNames.length });
+        items.push({ label: "Edit Configurations (JSON)", description: "", index: configurationNames.length });
 
         return vscode.window.showQuickPick(items, options)
             .then(selection => (selection) ? selection.index : -1);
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/ui.ts
```diff
@@ -171,7 +171,7 @@ export class UI {
         for (let i: number = 0; i < configurationNames.length; i++) {
             items.push({ label: configurationNames[i], description: "", index: i });
         }
-        items.push({ label: "Edit Configurations...", description: "", index: configurationNames.length });
+        items.push({ label: "Edit Configurations (JSON)", description: "", index: configurationNames.length });
```
bobbrow: This changes it for every invocation, though, right?  We don't want to force users to go to JSON when they click on the label in the status bar, or run the command to select a config manually.

bobbrow: I think we want to register a new private command (not in package.json).  An example:
https://github.com/Microsoft/vscode-cpptools/blob/f3a02fde4e466991230a6a55cc92456cb07bc5dd/Extension/src/LanguageServer/extension.ts#L680

The command will behave almost identically to the `select` function, but will open the JSON file instead of the UI regardless of the user's UI vs JSON setting.

bobbrow (changes requested):

bobbrow (approved):

Issue #3408, opened 2019-04-04
Switch to using the installExtension command for offline/insider vsix installing


sean-mcmanus:
They added it to 1.33.0.

vscode.commands.executeCommand('workbench.extensions.installExtension', vscode.Uri.file('path-to-vsix'));

The problem with the current method is that it has to locate the .exe, which may fail for various reasons (such as the .exe name not matching what we expect).

tara-raj:
Fixed in Insiders 0.23.0

PR #3521, opened 2019-04-24
Switch to using the installExtension command for offline/insider vsix installing

Colengms:
https://github.com/Microsoft/vscode-cpptools/issues/3408

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -17,11 +17,13 @@ import { CppSettings } from './settings';
 import { PersistentWorkspaceState, PersistentState } from './persistentState';
 import { getLanguageConfig } from './languageConfig';
 import { getCustomConfigProviders } from './customProviders';
+import { PlatformInformation } from '../platform';
 import { Range } from 'vscode-languageclient';
 import { ChildProcess, spawn, execSync } from 'child_process';
 import * as tmp from 'tmp';
 import { getTargetBuildInfo, BuildInfo } from '../githubAPI';
 import * as configs from './configurations';
+import { PackageVersion } from '../packageVersion';
 import { getTemporaryCommandRegistrarInstance } from '../commands';
 
 let prevCrashFile: string;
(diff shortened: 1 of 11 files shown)


Extension/src/githubAPI.ts
```diff
@@ -124,6 +124,11 @@ export interface BuildInfo {
     name: string;
 }
 
+/**
+ * Set this hook to true to force an update, to test the update flow.
+ */
+let testUpdateFlow: boolean = false;
```
bobbrow: how is this different from `testingInsidersVsixInstall`?  Do we need both variables?

Extension/src/LanguageServer/extension.ts
```diff
@@ -601,7 +496,7 @@ function applyUpdate(buildInfo: BuildInfo, updateChannel: string): Promise<void>
                 break;
             }
             try {
-                await installVsix(vsixPath, updateChannel);
+                await vscode.commands.executeCommand('workbench.extensions.installExtension', vscode.Uri.file(vsixPath));
```
bobbrow: Do we need to update package.json to use this, or is it already available in the version of VS Code we're targeting?

bobbrow (dismissed):

Extension/src/githubAPI.ts
```diff
@@ -141,8 +146,11 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
             // Allows testing pre-releases without accidentally downgrading to the latest version
             const userVersion: PackageVersion = new PackageVersion(util.packageJson.version);
             const latestVersion: PackageVersion = new PackageVersion(builds[0].name);
-            if (!testingInsidersVsixInstall && (userVersion.isGreaterThan(latestVersion) || (userVersion.suffix && userVersion.suffix !== 'insiders'))) {
-                return undefined;
+            if (!testUpdateFlow)
+            {
```
bobbrow: curly brace should be on previous line.

Extension/src/LanguageServer/extension.ts
Colengms: According to Sean (in the issue); They added it to 1.33.0.

Extension/src/githubAPI.ts
Colengms: Removed

sean-mcmanus:
I don't think we should check this in until we ship 0.23.0 due to the 1.33.0 requirement.

sean-mcmanus (approved):

PR #3527, opened 2019-04-24
Unescape backslashes in settings for UI

Colengms:
Add un-escaping of backslashes for display in settings UI.

Backslashes were already being added when saving strings from the UI.

### DIFF
--- a/Extension/ui/settings.ts
+++ b/Extension/ui/settings.ts
@@ -79,19 +79,23 @@ class SettingsApp {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Extension/ui/settings.ts
```diff
@@ -79,19 +79,23 @@ class SettingsApp {
         }
     }
 
+    private unescapteConfigString(s: string): string {
```
bobbrow: [sp] `unescape`?

bobbrow (approved):

PR #3529, opened 2019-04-24
Various settings UI load/save related changes

Colengms:
Integrated parsePropertiesFileAndHandleSquiggles into parsePropertiesFile(true).

Switched from delayed saving to saving immediately as setting change, using a new event.

Removed some escaping logic from parsePropertiesFile(false) not needed for settings UI.

Changed SettingsPanelViewStateChanged to SettingsPanelActivated, only triggered on activation.

Removed previous fix slash escaping fix.

### DIFF
--- a/Extension/src/LanguageServer/settingsPanel.ts
+++ b/Extension/src/LanguageServer/settingsPanel.ts
@@ -25,7 +25,7 @@ export class SettingsPanel {
     private configValues: config.Configuration;
     private isIntelliSenseModeDefined: boolean = false;
     private settingsPanelActivated = new vscode.EventEmitter<void>();
-    private settingsPanelStateChanged = new vscode.EventEmitter<void>();
+    private configValuesChanged = new vscode.EventEmitter<void>();
     private panel: vscode.WebviewPanel;
     private disposable: vscode.Disposable = undefined;
     private disposablesPanel: vscode.Disposable = undefined;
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/settingsPanel.ts
```diff
@@ -21,16 +21,11 @@ const elementId: { [key: string]: string } = {
...
-    isDirty: boolean;
-}
-
 export class SettingsPanel {
     private configValues: config.Configuration;
     private isIntelliSenseModeDefined: boolean = false;
-    private configDirty: boolean = false;
-    private settingsPanelViewStateChanged = new vscode.EventEmitter<ViewStateEvent>(); 
+    private settingsPanelActivated = new vscode.EventEmitter<void>();
+    private settingsPanelStateChanged = new vscode.EventEmitter<void>();
```
michelleangela: Rename to onConfigValuesChanged?

Extension/src/LanguageServer/settingsPanel.ts
```diff
@@ -40,7 +35,7 @@ export class SettingsPanel {
     constructor() {
         this.configValues = { name: undefined };
         this.disposable = vscode.Disposable.from(
-            this.settingsPanelViewStateChanged,
+            this.settingsPanelStateChanged,
```
michelleangela: Add this.settingsPanelActivated to this.disposable list

michelleangela (changes requested):

michelleangela:
Also, add this pull request to our internal "pending change log" list under bug fixes.

Extension/src/LanguageServer/settingsPanel.ts
Colengms: configValuesChanged?  (the other event is not using 'on')  Or, use 'on' for both?

sean-mcmanus (approved):

bobbrow:
Since this feature is not yet released in an official channel, any changelog updates would only apply to a future insiders release.  It wouldn't need to go in the changelog for 0.23.0.

michelleangela (approved):

Issue #2558, opened 2018-09-25
Prevent "*.C": "cpp" from getting added to files.associations and/or force "*.c" to use C IntelliSense


sean-mcmanus:
Due to bug https://github.com/Microsoft/vscode/issues/59369 .

...and force "*.C" files to be C++.

Issue #1989, opened 2018-05-13
Debug/Spell Check Error?


LinshuH:
I am running VS Code 1.23.1 Version and have extension C/C++ IntelliSense, debugging, and code browsing for C++.
My code can successfully compiled by g++, but VS Code keep using Red underline to remind me the error in the code.
For example, like the one below
![vs code2](https://user-images.githubusercontent.com/28817501/39964504-a1f8d02e-563a-11e8-9bd0-5bf390772d3c.png)
However, I do have "Image Struct" defined in my .h file.
Also in my .h file, while I define the class Image, it shows "function type is not allowed".
My .h file is shown below.
![vs code](https://user-images.githubusercontent.com/28817501/39964517-26c4c1e6-563b-11e8-9575-b0547b309b7b.png)

I am really confused, if the code is able to been compiled, why there is error like this exist? What it really means?

Thanks

sean-mcmanus:
It looks like you're using .C files but your struct has a constructor. Our IntelliSense parser is compiling the file as a .C file instead of C++. You can fix this via changing the language for the file to C++ in the bottom right status bar (which only lasts while the file is open) or via the files.associations setting:

```
    "files.associations": {
        "*.c": "cpp"
    }
```

If you still get problems you can't resolve, you can disable the errors via the C_Cpp.errorSquiggles setting.

bobbrow:
I believe GCC and friends treat capitalized `*.C` as c++ by default.  VS Code probably doesn't take this into account.

sean-mcmanus:
Unfortunately, it looks like VS Code's files.associations is case-insensitive, even on Linux. It's possible we could bypass VS Code's file associations and add our case-sensitive ".C" case.

LinshuH:
Hi all,
Thanks for all the answers. After change .C to .cpp, the issue looks like been solved. Not really sure why, but after opening it for a while, the underline does not show up.

bobbrow:
Fixed as part of #2558

PR #3536, opened 2019-04-25
Work around vscode not treating .C files as C++

Colengms:
".C" is associated with C++ by GNU, and Intel.

https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html
http://www.physics.udel.edu/~bnikolic/QTTG/shared/docs/intel_c_user_and_reference_guide.pdf

This works around the issue by adding file associations for any encountered .C file, by name.

This addresses part of: https://github.com/Microsoft/vscode-cpptools/issues/2558

### DIFF
--- a/Extension/src/LanguageServer/protocolFilter.ts
+++ b/Extension/src/LanguageServer/protocolFilter.ts
@@ -4,6 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 'use strict';
 
+import * as path from 'path';
 import { Middleware } from 'vscode-languageclient';
 import { ClientCollection } from './clientCollection';
 import { Client } from './client';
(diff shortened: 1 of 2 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,6 +23,14 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
         didOpen: (document, sendMessage) => {
             if (clients.checkOwnership(me, document)) {
                 me.TrackedDocuments.add(document);
+
+                // Work around vscode treating ".C" as c, by adding this file name to file associations as cpp
+                if (document.uri.path.endsWith(".C")) {
+                    let fileName: string = document.uri.fsPath.split('\\').pop().split('/').pop();
```
bobbrow: I was thinking there was a `path` npm module API that can get you the filename out of a path string.  I would rather we not reimplement this.

bobbrow (approved):

bobbrow:
Also addresses #1989

PR #3545, opened 2019-04-26
Work around issue with getting stale contents from openTextDocument

Colengms:
This addresses one of the scenarios discussed in this vscode issue: https://github.com/Microsoft/vscode/issues/72831

Using openTextDocument in response to a FileSystemWatcher.onDidChange may result in reading stale (cached) contents, if the file is already opened by vscode.  This can be worked around by instead waiting for onDidChangeTextDocument to fire for the file.


sean-mcmanus (approved):

michelleangela (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -681,11 +699,6 @@ export class CppProperties {
                 }
 
                 let fullPathToFile: string = path.join(this.configFolder, "c_cpp_properties.json");
-                let filePath: vscode.Uri = vscode.Uri.file(fullPathToFile).with({ scheme: "untitled" });
-
-                let document: vscode.TextDocument = await vscode.workspace.openTextDocument(filePath);
```
bobbrow, line 686: I was thinking that this would open and (potentially) focus the file in the editor.  Are we doing anything to maintain that behavior with this new approach?

bobbrow (approved):

PR #3573, opened 2019-05-01
Address issue with trailing backslashes in config path not squiggled

Colengms:
The prior approach using a RegEx was identifying a backslash followed by a quote as an escaped quote, even if that backslash was itself escaped.  Unclear how to solve that with RegEx.  Replaced with manual parsing.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -764,30 +764,7 @@ export class CppProperties {
             if (handleSquiggles) {
                 // Replace all \<escape character> with \\<character>, except for \"
                 // Otherwise, the JSON.parse result will have the \<escape character> missing.
-                let newResults: string = "";
-                let lastWasBackslash: Boolean = false;
-                let lastBackslashWasEscaped: Boolean = false;
-                for (let i: number = 0; i < readResults.length; i++) {
-                    if (readResults[i] === '\\') {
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -752,13 +752,34 @@ export class CppProperties {
...
-                // Replace all \<escape character> with \\<character>.
+                // Replace all \<escape character> with \\<character>, except for \"
                 // Otherwise, the JSON.parse result will have the \<escape character> missing.
-                readResults = readResults.replace(/\\/g, '\\\\'); 
-                readResults = readResults.replace(/\\\\"/g, '\\"'); // Need to revert the change to \". 
-                readResults = readResults.replace(/\\\\"/g, '\\"'); // Need to do it again for \\". 
+                let newResults: string = "";
+                let lastWasBackslash: Boolean = false;
+                let lastBackslashWasEscaped: Boolean = false;
+                for (let i: number = 0; i < readResults.length; i++) {
```
bobbrow: Can you factor this out into a separate function, and then add some unit tests for it?

sean-mcmanus: Can that go in a future pull request? We want to ship this fix today.

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #3581, opened 2019-05-04
Dev/waan/fix test run

sean-mcmanus:


sean-mcmanus:
The languageServer tests are passing on my (3) local machines...not sure why it's timing out in the Azure Pipeline.

WardenGnaw (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -413,7 +413,8 @@ class DefaultClient implements Client {
         let settings: CppSettings = new CppSettings(workspaceFolder ? workspaceFolder.uri : null);
         let other: OtherSettings = new OtherSettings(workspaceFolder ? workspaceFolder.uri : null);
 
-        let storagePath: string = util.extensionContext.storagePath;
+        let storagePath: string = util.extensionContext ? util.extensionContext.storagePath :
+            path.join((workspaceFolder ? workspaceFolder.uri.fsPath : ""), "/.vscode");
```
bobbrow, line 417: I'm not sure we want to change this.  The extension binary does something different when there is no storage path, right?

Extension/test/integrationTests/languageServer/languageServer.integration.test.ts
```diff
@@ -10,12 +10,16 @@ import * as util from '../../../src/common';
...
+import { initializeTemporaryCommandRegistrar } from '../../../src/commands';
+
+const defaultTimeout: number = 10000;
 
 suite("multiline comment setting tests", function(): void {
     suiteSetup(async function(): Promise<void> { 
-        let extension: vscode.Extension<any> = vscode.extensions.getExtension("ms-vscode.cpptools"); 
+        let extension: vscode.Extension<any> = vscode.extensions.getExtension("ms-vscode.cpptools");
+        util.setExtensionPath(extension.extensionPath);
+        initializeTemporaryCommandRegistrar();
```
bobbrow, line 22: Why do we need this now?

bobbrow (changes requested):

PR #3599, opened 2019-05-07
add getClients API

### LINKED ISSUE #3598: Feature request: custom language server middleware
Since version 0.23.0, vscode-cpptools is offered as a bundled file (dist\main.js) instead of separate files for each module. This makes wiring internal modules on-the-fly impossible. I am working on a plugin for vscode-cpptools to filter and sort the completion items in the completion list based on an AI-powered web service, which is practically what [Microsoft IntelliCode](https://github.com/microsoftDocs/intellicode) does.

Before version 0.23.0, I could do something like this:

```ts
const mscpp = vscode.extensions.getExtension("ms-vscode.cpptools");
assert(mscpp.isActive);
const _m = require("module");
const lsext = _m._cache[path.join(mscpp.extensionPath, "out", "src", "LanguageServer", "extension.js")];
const client = lsext.exports.getClients().ActiveClient.languageClient;
const oldProvideCompletionItems = client.clientOptions.middleware.provideCompletionItem;
client.clientOptions.middleware.provideCompletionItem = function(...) {
  const list = await oldProvideCompletionItems(...);
  // sort and filter the list
  return list;
}
```

Now all I got is a CppTools API object, which does not cover the functionality I need. I suggest we export a `getClientCollection` function in CppTools API, which calls `getClients` function in `LanguageServer\extension.ts`, so extensions can have a better control.

felixhao28:
Close #3598

### DIFF
--- a/Extension/src/cppTools.ts
+++ b/Extension/src/cppTools.ts
@@ -121,4 +121,8 @@ export class CppTools implements CppToolsTestApi {
     public getTestHook(): CppToolsTestHook {
         return test.getTestHook();
     }
+
+    public getClients() {
+        return LanguageServer.getClients();
+    }
 }


Extension/src/cppTools.ts
```diff
@@ -121,4 +121,8 @@ export class CppTools implements CppToolsTestApi {
     public getTestHook(): CppToolsTestHook {
         return test.getTestHook();
     }
+
+    public getClients() {
```
bobbrow, line 125: This is not functionality that we intend to expose to extension authors at this time and we would provide no guarantees that we won't break this behavior again in the future.  We do not want extensions to tamper with our completion results because we do intend to add support for IntelliCode in a future release.

PR #3601, opened 2019-05-07
Update TPN.

sean-mcmanus:


Extension/ThirdPartyNotices.txt
```diff
@@ -375,7 +425,7 @@ SOFTWARE.
 
 -------------------------------------------------------------------
 
-vscode-cpptools 2.1.1 - MIT
+vscode-cpptools 2.1.2 - MIT
```
pieandcakes, line 428: should this be vscode-cpptools-api?

sean-mcmanus: It's called "vscode-cpptools" in the package.json.

pieandcakes (approved):

PR #3602, opened 2019-05-07
Add build dependencies.

sean-mcmanus:
These were "hidden" dev dependencies that were installed during our vsix builds. Adding these to the devDependencies should allow us to get these removed from the ThirdPartyNotices.txt.


sean-mcmanus:
There might be a better way to deal with this via running an npm uninstall in our builds to prevent Component Governance from detecting the vsce/rimraf node modules. Investigating now...

sean-mcmanus:
Yeah, looks like doing npm uninstall in our builds fixes this. I'll be sending an updated TPN soon.

Issue #3610, opened 2019-05-08
Error: Missing binary Microsoft.VSCode.CPP.Extension.exe


si-jon:
Today, I suddenly couldn't use the usual C/C++ IntelliSense features in vscode, and I notice I got this error when starting:

`Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: Missing binary at c:\Users\MyName\.vscode\extensions\ms-vscode.cpptools-0.23.0\bin\Microsoft.VSCode.CPP.Extension.exe`

When I go to that folder, as it says the binary isn't there. I've tried to uninstall and reinstall the vscode-cpptools, but the error persist and the file is still missing.

I cannot give any more info on how to reproduce this: IntelliSense worked yesterday, but today, after a computer restart, it just doesn't any more.

si-jon:
Was able to fix this by installing the previous version, 0.22.1, and reload visual studio code. Now the binary has been added to \ms-vscode.cpptools-0.23.0\bin\ and IntelliSense is working again.

sean-mcmanus:
I don't know why the binary would fail to be there. Uninstall/reinstall of extensions with VS Code requires restarting VS Code after the uninstall or running the "Developer: Reinstall extension" command. I've filed a feature request on VS Code to make that command more discoverable: https://github.com/microsoft/vscode/issues/73485 .

Dellyjoe:
Hello everyone, 

I have a similar issue, I have a tinkerbaord S which I have installed VSC. from Headmelt. I have extensions C/C++ and platformIO installed. I use platformIO to do my coding in, which when I build it uses c/c++ to do some of the work.  

I'm not sure how to fix this, it never worked for me. Is this extension not able to work do to me being on an arm processor. ?

sean-mcmanus:
@Dellyjoe Yes, our binary isn't built for arm yet (we have another issue tracking that...it's around our 9th most requested feature).

crashdown0815:
Same issue here on windows. Since 0.23.0. Reinstall does also not work.
Is there another way to fetch the executables as work around?
E.g. from where is the extension fetching them?

bobbrow:
@crashdown0815 check your `%userprofile%/.vscode` folder and delete the `ms-vscode.cpptools-<ver>` folder.  Sometimes there is a `.obsolete` file there listing our extension that prevents reinstall.  Removing the entry that points to our folder can also fix it.

[...]

PR #3611, opened 2019-05-08
Add telemetry for missing language service extension binary

sean-mcmanus:
So we have a better understanding of how severe the missing binaries issues are, such as https://github.com/microsoft/vscode-cpptools/issues/3610 . We've had other missing binary reports in the past (such as from virus scan deletions).

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -402,6 +402,7 @@ class DefaultClient implements Client {
         let serverModule: string = getLanguageServerFileName();
         let exeExists: boolean = fs.existsSync(serverModule);
         if (!exeExists) {
+            telemetry.logLanguageServerEvent("missingBinary");
             throw String('Missing binary at ' + serverModule);
         }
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -402,6 +402,7 @@ class DefaultClient implements Client {
         let serverModule: string = getLanguageServerFileName();
         let exeExists: boolean = fs.existsSync(serverModule);
         if (!exeExists) {
+            telemetry.logLanguageServerEvent("missingBinary");
```
pieandcakes: Do you want to log which binary name specifically is "missing"? In this case since it is files that we need that we are creating, it might be useful to log the filename only (no path) so you can see how often it is missing.

sean-mcmanus: This telemetry is only shown for the Microsoft.VSCode.CPP.Extension.exe -- that's the main scenario I'm interested in.

michelleangela (approved):

Issue #3583, opened 2019-05-05
Unable to create 'launch.json' file inside the '.vscode' folder (Do not activate the extension when IntelliSense is disabled.).


daandemeyer:
Type: Debugger

- OS and Version: Windows 10 / Arch Linux
- VS Code Version: 1.33.1
- C/C++ Extension Version: 0.22.1
- Other extensions you installed (and if the issue persists after disabling them): 
  - vscode-clangd (persists after disable)
  - Others but I don't think they matter
- A clear and concise description of what the bug is:

When opening launch.json for the first time and selecting a debugger configuration, an error notification occurs containing the text in the issue title. This happens on both Windows with the VS debugger and Arch Linux with the GDB debugger.

I have intellisense disabled in the vscode-cpptools extension because I'm using vscode-clangd for intellisense instead. From the error message, I'm assuming VSCode tries to activate the extension when opening launch.json for the first time and generating a debugger configuration, which the extension doesn't allow because intellisense is disabled.

**To Reproduce**
1. Open VSCode with vscode-clangd and vscode-cpptools installed
2. Make sure vscode-cpptools intellisense is disabled (`"C_Cpp.intelliSenseEngine": "Disabled"`)
3. Execute the `Debug: Open launch.json` command and select any debugger.

pieandcakes:
@bobbrow @sean-mcmanus Looks like a bug introduced when intellisense is disabled:

```
Unable to create 'launch.json' file inside the '.vscode' folder (Do not activate the extension when IntelliSense is disabled.).
```

It is from [this](https://github.com/Microsoft/vscode-cpptools/blob/77f37476300171b7f8696c313e477f08901fe764/Extension/src/LanguageServer/extension.ts#L283) line of code. I think its blocking activation because it is throwing.

sean-mcmanus:
Should be fixed with 0.23.1.

PR #3612, opened 2019-05-08
Fix launch.json creation with intelliSenseEngine Disabled.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/3583

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -150,7 +150,12 @@ export async function getBuildTasks(returnComplerPath: boolean): Promise<vscode.
     // for the active file, remove duplicate compiler names, then finally add the user's compilerPath setting.
     let compilerPaths: string[];
     const isWindows: boolean = os.platform() === 'win32';
-    const activeClient: Client = getActiveClient();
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -150,7 +150,12 @@ export async function getBuildTasks(returnComplerPath: boolean): Promise<vscode.
     // for the active file, remove duplicate compiler names, then finally add the user's compilerPath setting.
     let compilerPaths: string[];
     const isWindows: boolean = os.platform() === 'win32';
-    const activeClient: Client = getActiveClient();
+    let activeClient: Client;
+    try {
+        activeClient = getActiveClient();
+    } catch (e) {
+        return []; // Language service features may be disabled.
```
pieandcakes, line 161: Do you want to log it to the output when you catch it as a warning in case it isn't the exception you are expecting?

sean-mcmanus: It's not something I'm currently concerned about, but it was easy to add a warning.

pieandcakes (approved):

Issue #3615, opened 2019-05-09
command C_Cpp.ConfigurationEditJSON not found


akshsin:
Issue Type: <b>Bug</b>

I am trying to use VS code on my windows machine for c++ development (with WSL).

Following the steps outlined in - https://code.visualstudio.com/docs/cpp/config-wsl

While trying to create the configuration json, using Ctrl+Shift+P, "Edit Configurations":

1. I don't see "Edit Configurations"

2. Do see "Edit Configurations (JSON)" though which I believe should work similarly, but getting this error "command 'C_Cpp.ConfigurationEditJSON' not found".

Extension version: 0.23.0
VS Code version: Code 1.33.1 (51b0b28134d51361cf996d2f0a1c698247aeabd8, 2019-04-11T08:27:14.102Z)
OS version: Windows_NT x64 10.0.17134

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz (8 x 1992)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|7.89GB (2.49GB free)|
|Process Argv|.|
|Screen Reader|no|
|VM|0%|

</details>

ReonLion:
Hello, i have the same problem, did you figure it out?

erezak:
The same happens with the UI configuration command.

ReonLion:
You should have a .cpp file first.

bobbrow:
Thank you for reporting this issue.  We are investigating the root cause.

ahmedkrmn:
A similar error message is also shown for all other extension commands!
I'm running the same VS Code and extension versions as OP.

bobbrow:
Yes, they are all affected.  You can work around this issue by creating and opening a C/C++ file in your new folder before running the command.

sean-mcmanus:
Should be fixed with 0.23.1.

abdoulayegk:
I have this problem on visual studio can anyone help me?
"{
[...]

PR #3622, opened 2019-05-09
Fix issue with extension context not being set before initializeTemporaryCommandRegistrar

Colengms:
Fix for https://github.com/microsoft/vscode-cpptools/issues/3615


michelleangela:
This is to fix issue https://github.com/microsoft/vscode-cpptools/issues/3615 so that commands will execute when extension has not been activated yet.

bobbrow (approved):

sean-mcmanus (approved):

michelleangela (approved):

Issue #3642, opened 2019-05-14
 Intelli Sense Cache Path Bug


Lennon925:
- C/C++ Extension Version: 0.23.1

I set  "Intelli Sense Cache Path" to "D:/storageLocation/${workspaceFolderBasename}/".
If I have three folders which have the same name 'cpptools-demo‘ and have different storage path，it goes wrong.

Open first folder, it created cpptools-demo folder in D:/storageLocation
Open second folder, it create untitled folder in D:/storageLocation
Open third folder, it didn't create folder in D:/storageLocation any more.

In addition, is there a way to set the cache path to where cpptools browse.db is storaged, ,"AppData\Roaming\Code\User\workspaceStorage"?

sean-mcmanus:
Yeah, I'm seeing the ${workspaceFolderBasename} incorrectly being resolved to "untitled".

Also, I'm seeing bogus temporary paths getting created while the setting is changed in the UI editor. I filed a bug at https://github.com/microsoft/vscode-cpptools/issues/3644 .

There isn't a way currently to set the location to the "workspaceStorage" location. Is there an issue on this already?

sean-mcmanus:
I have a fix at https://github.com/microsoft/vscode-cpptools/pull/3643 . I'm not able to repro the bugs after that.

Lennon925:
> Yeah, I'm seeing the ${workspaceFolderBasename} incorrectly being resolved to "untitled".
> 
> Also, I'm seeing bogus temporary paths getting created while the setting is changed in the UI editor. I filed a bug at #3644 .
> 
> There isn't a way currently to set the location to the "workspaceStorage" location. Is there an issue on this already?

I tried to set ${workspaceFolderBasename} to cpptools database storage path, there is only one cpptool-demo folder when I opened many folders which have the same name but have different storage path.

If I use ${workspaceFolderBasename}, folders which have the same name but have different storage path will be regarded as a same folder, right?

sean-mcmanus:
Yeah, using ${workspaceFolderBasename} just uses the name and not the storage path -- the storage path creates a hash for the workspace. However, I think we could create our own ${workspaceStorage} variable that resolves to the storage path and add that to the AdditionalEnvironment in the exact same we do do for the ${workspaceFolderBasename}, which would allow users to use that location for the intelliSenseCachePath.

sean-mcmanus:
[...]

PR #3643, opened 2019-05-14
Fix ${workspaceFolderBasename} resolving to "untitled", and add ${workspaceStorage}

sean-mcmanus:
Fix for an issue reported at https://github.com/microsoft/vscode-cpptools/issues/3642 .

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -340,11 +340,11 @@ class DefaultClient implements Client {
 
     constructor(allClients: ClientCollection, workspaceFolder?: vscode.WorkspaceFolder) {
         try {
+            this.rootFolder = workspaceFolder;
             let languageClient: LanguageClient = this.createLanguageClient(allClients, workspaceFolder);
             languageClient.registerProposedFeatures();
             languageClient.start();  // This returns Disposable, but doesn't need to be tracked because we call .stop() explicitly in our dispose()
             util.setProgress(util.getProgressExecutableStarted());
-            this.rootFolder = workspaceFolder;
             ui = getUI();
             ui.bind(this);
 
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -340,7 +340,6 @@ class DefaultClient implements Client {
 
     constructor(allClients: ClientCollection, workspaceFolder?: vscode.WorkspaceFolder) {
         try {
-            this.rootFolder = workspaceFolder;
             let languageClient: LanguageClient = this.createLanguageClient(allClients, workspaceFolder);
             languageClient.registerProposedFeatures();
             languageClient.start();  // This returns Disposable, but doesn't need to be tracked because we call .stop() explicitly in our dispose()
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -340,11 +340,11 @@ class DefaultClient implements Client {
 
     constructor(allClients: ClientCollection, workspaceFolder?: vscode.WorkspaceFolder) {
         try {
+            this.rootFolder = workspaceFolder;
```
pieandcakes: Is this to resolve "workspaceFolder" first or because setting "this.rootFolder" first resolves the value?

I see you moved it up but I don't understand how this fixes the issue. If its because of ordering, please add a comment inline after it that it needs to go first to prevent it from getting reordered accidently.

pieandcakes (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Code in in the constructor is using this.rootFolder, but it's undefined, so the ${workspaceFolderBasename} gets set to "untitled" instead. This used to work, but it must have regressed due to some other code moving around. I'm not sure what you mean by "reordered".

sean-mcmanus: Specifically, line 318 in client.ts is calling this.getName(this.rootFolder) in the Name() method, called by AdditionalEnvironment().

sean-mcmanus: I moved the setting into the creation method -- is that okay?

Extension/src/LanguageServer/client.ts
```diff
@@ -405,6 +404,7 @@ class DefaultClient implements Client {
             telemetry.logLanguageServerEvent("missingLanguageServerBinary");
             throw String('Missing binary at ' + serverModule);
         }
+        this.rootFolder = workspaceFolder;
```
pieandcakes: wait.. If you are already passing it in, why does where you set it matter? Why is it failing in the first place? I assume that since you pass it in, instead of using this.rootFolder, that you should be operating on workspaceFolder instead.

I feel like someone from your team needs to look at this.

sean-mcmanus: Code needs to references the this.rootFolder specifically, i.e. the lifetime of the workspaceFolder variable is only that specific function, so assigning to this.rootFolder extends the life of the variable when the method returns.

sean-mcmanus: I added the ${workspaceStorage} case too.

Extension/src/LanguageServer/client.ts
```diff
@@ -344,7 +345,6 @@ class DefaultClient implements Client {
             languageClient.registerProposedFeatures();
             languageClient.start();  // This returns Disposable, but doesn't need to be tracked because we call .stop() explicitly in our dispose()
             util.setProgress(util.getProgressExecutableStarted());
-            this.rootFolder = workspaceFolder;
```
bobbrow, line 347: remove this order of initialization dependency by initializing this first, and then don't pass the local variable in to `createLanguageClient`.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Uh...why does that matter?

bobbrow: because the intuitive place to initialize member variables is in the constructor.

sean-mcmanus: Ok

bobbrow (approved):

PR #3647, opened 2019-05-14
Add unlimited string size for gdb

### LINKED ISSUE #3646: Enable unlimited string size for gdb
Type: Debugger
**Input information below**
Perhaps it is a good idea to change the default configuration for Windows and Linux to enable unlimited string size for gdb.

rynz:
Closes #3646


WardenGnaw:
There’s a discussion about this: https://github.com/microsoft/vscode-cpptools/issues/3646

We don’t want to have the default size be unlimited, but rather 200. The question is would we want to have this in the default launch.json or have users add it themselves. VS Code prefers a minimal launch.json for users.

rynz:
No worries, love this extension! :)

PR #3651, opened 2019-05-14
Add colorization support to cpptools

Colengms:
Depends on changes on the native side.


Extension/src/LanguageServer/colorizationSettings.ts
```diff
@@ -0,0 +1,275 @@
...
+'use strict';
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { OtherSettings, TextMateRule, TextMateRuleSettings } from './settings';
+import * as fs from 'fs';
+import { loadavg } from 'os';
+
+export enum TokenKind {
+    // These need to match sematic_token_kind in edge_hlapi.h
```
bobbrow: I would just say these need to match the token_kind in the server.  Don't mention the internal header here.

Extension/src/LanguageServer/colorizationSettings.ts
```diff
@@ -0,0 +1,275 @@
...
+export class ThemeStyle {
+    foreground: string;
+    background: string;
+    fontStyle: string;
+}
+
+export class ColorizationSettings {
+
+    constructor(uri: vscode.Uri)
+    {
```
bobbrow: The TS linter will be strict about curly braces.  This needs to go on the previous line.

pieandcakes:
@Colengms Please fix the typescript linting issues. you can run `npm tslint` from your command line to verify.

Extension/src/LanguageServer/colorizationSettings.ts
```diff
@@ -0,0 +1,275 @@
...
+        this.findThemeStyleForScope(baseCStyle, baseCppStyle, scope, themeTextMateRules);
+        let otherSettings: OtherSettings = new OtherSettings(this.uri);
+
+        // Next in priority would be a global user override of token color of the equivilent scope
+        let colorTokenName: string | undefined = ColorizationSettings.scopeToTokenColorNameMap.get(scope);
+        if (colorTokenName) {
+            let settingValue: string = otherSettings.getCustomColorToken(colorTokenName);
+            if (settingValue) {
+                if (baseCStyle) {
+                    baseCStyle.foreground = settingValue;
```
bobbrow: Do we need to grab the other style properties (e.g. background/fontStyle)?

Extension/src/LanguageServer/colorizationSettings.ts
```diff
@@ -0,0 +1,275 @@
...
+    public reload(): void {
+        let otherSettings: OtherSettings = new OtherSettings(this.uri);
+        let themeName: string = otherSettings.colorTheme;
+        
+        // Enumerate through all extensions, looking for this theme.  (Themes are implemented as extensions - even the default ones)
+        // Crack open each package.json to check for a theme name patch
+        for (let i: number = 0; i < vscode.extensions.all.length; i++) {
+            let extensionPath: string = vscode.extensions.all[i].extensionPath;
+            let extensionPackageJsonPath: string = path.join(extensionPath, "package.json");
+            let packageJson: any = JSON.parse(fs.readFileSync(extensionPackageJsonPath).toString());
```
bobbrow: We should check the existence of the file first.  If running in the headless scenario, the file may not be there (we're not sure at this point).  In any case, we should have a graceful failure path in case any of our assumptions are invalid.

Extension/src/LanguageServer/colorizationSettings.ts
```diff
@@ -0,0 +1,275 @@
...
+                for (let j: number = 0; j < packageJson.contributes.themes.length; j++) {
+                    let themeId: string = packageJson.contributes.themes[j].id;
+                    let themeLabel: string = packageJson.contributes.themes[j].label;
+                    if (themeId === themeName || themeLabel === themeName) {
+                        // found currently selected theme
+
+                        let defaultStyle: ThemeStyle = new ThemeStyle();                        
+                        let themeRelativePath: string = packageJson.contributes.themes[j].path;
+                        let themeFullPath: string = path.join(extensionPath, themeRelativePath);
+                        let themeContent: any = JSON.parse(fs.readFileSync(themeFullPath).toString());
```
bobbrow: Check the existence of the file first.

Extension/package.json
```diff
@@ -495,6 +495,18 @@
           "default": true,
           "description": "If true, snippets are provided by the language server.",
           "scope": "resource"
+        },
+        "C_Cpp.textMateColorization": {
```
bobbrow: I don't think we should officially ship this setting.  You may still be able to use it in the extension for testing without declaring it here though...  There is code that iterates through the settings to see what's changed, and in that case we actually see everything in the JSON object representing the C_Cpp section whether it's declared in package.json or not.  You could potentially do something similar for this one.

bobbrow (dismissed):

Extension/src/LanguageServer/colorizationSettings.ts
Colengms: getCustomColorToken gets the color token settings ("types", "variables", etc.).  The values of these settings are foreground colors only.

Extension/package.json
Colengms: Are you sure we shouldn't ship with this setting?  The two systems are significantly different.  That's not apparent with the few color themes we've looked at, as very little that differs is colored.  However, if someone takes advantage of the ability to color the various items we allow to be colored, the output can be significantly different, resulting in scenarios where the color swapping is apparent and distracting.  (i.e. on making a file active - first textMate colors are applied, then ours).

Colengms: As discussed, kept the setting but not publishing it in the schema.

Extension/package.json
```diff
@@ -495,6 +495,12 @@
           "default": true,
           "description": "If true, snippets are provided by the language server.",
           "scope": "resource"
+        },
+        "C_Cpp.enhancedColorization": {
+          "type": "boolean",
+          "default": "true",
+          "description": "If true, code is colorized based on syntactic and semantic analysis.  This is disabled if IntelliSense is disabled.",
```
bobbrow: "This setting has no effect if IntelliSense is disabled"

Extension/src/LanguageServer/client.ts
```diff
@@ -521,22 +577,54 @@ class DefaultClient implements Client {
...
+            if (changedSettings["textMateColorization"]) {
+                this.updateGrammarsInPackageJson();
+            }
             this.configuration.onDidChangeSettings();
             telemetry.logLanguageServerEvent("CppSettingsChange", changedSettings, null);
         }
 
+        let settings: CppSettings = new CppSettings(this.RootUri);
+        if (settings.enhancedColorization) {
+            this.colorizationSettings.reload();
```
bobbrow: should we only do this if the setting changes?

Extension/src/LanguageServer/client.ts
```diff
@@ -521,22 +577,54 @@ class DefaultClient implements Client {
...
+            }
+        }
+
         return changedSettings;
     }
 
+    private editVersion: number = 0;
+
+    public onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): void {
+
```
bobbrow: style: we don't normally add an empty line after a `{`

Extension/package.json
```diff
@@ -495,6 +495,12 @@
           "default": true,
           "description": "If true, snippets are provided by the language server.",
           "scope": "resource"
+        },
+        "C_Cpp.enhancedColorization": {
+          "type": "boolean",
```
bobbrow: Sorry I didn't catch this earlier...  The name of the setting doesn't match having a boolean type since it isn't named like an action.  This name style would result in an enum string type with "Enabled"/"Disabled" values.

Extension/src/common.ts
```diff
@@ -103,6 +103,11 @@ export function fileIsCOrCppSource(file: string): boolean {
     return [".C", ".c", ".cpp", ".cc", ".cxx", ".mm", ".ino", ".inl"].some(ext => fileExtLower === ext);
 }
 
+export function isEditorFileCpp(file: string): boolean {
+    let editor: vscode.TextEditor = vscode.window.visibleTextEditors.find(e => e.document.uri.toString() === file);
+    return editor && editor.document.languageId === "cpp";
```
bobbrow, line 108: We should treat `.C` as `"cpp"` too, right?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+        if (!insertRange.isEmpty) {
+            for (let i: number = 0; i < ranges.length; i++) {
+                ranges[i] = this.shiftRangeAdded(ranges[i], insertRange);
+            }
+        }
+
+        return ranges;
+    }
+
+    private fixRanges1(originalRanges: vscode.Range[], changes: vscode.TextDocumentContentChangeEvent[]): vscode.Range[] {
```
bobbrow: style: I never saw a `fixRanges2`.  Should we rename this?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+                    newEndCharacter += insertRange.end.character;
+                }
+            }
+            return new vscode.Range(newStartLine, newStartCharacter, newEndLine, newEndCharacter);
+        }
+        return originalRange;
+    }
+
+    private textToRange(text: string, startPosition: vscode.Position): vscode.Range {
+        // Need to parse text into a range.  (TODO: Test if tab is a single character, etc.)
```
bobbrow: tab should be a single character, but may have a different column width.  Do the colorization ranges that come back from the language server use columns or characters as offsets?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+
+            let newStart: vscode.Position = originalRange.start.translate(lineDelta, startCharacterDelta);
+            let newEnd: vscode.Position = originalRange.end.translate(lineDelta, endCharacterDelta);
+            return new vscode.Range(newStart, newEnd);
+        }
+        return originalRange;
+    }
+
+    private shiftRangeAdded(originalRange: vscode.Range, insertRange: vscode.Range): vscode.Range {
+        if (insertRange.start.isBeforeOrEqual(originalRange.start)) {
```
bobbrow: what if it's inside the `originalRange`?  should this be checking against `originalRange.end`?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+        let isCpp: boolean = util.isEditorFileCpp(uri.toString());
+        this.createColorizationDecorations(isCpp);
+        let editors: vscode.TextEditor[] = vscode.window.visibleTextEditors.filter(e => e.document.uri === uri);
+        for (let e of editors) {
+            this.refreshColorizationRanges(e);
+        }
+    }
+
+    private shiftRangeRemoved(originalRange: vscode.Range, removedRange: vscode.Range): vscode.Range {
+        if (removedRange.start.isBeforeOrEqual(originalRange.start)) {
```
bobbrow: what if it's inside the `originalRange`? should this be checking against `originalRange.end`?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+
+    private inactiveDecoration: vscode.TextEditorDecorationType = null;
+    private inactiveRanges: vscode.Range[] = [];
+
+    private versionedEdits: VersionedEdits[] = [];
+    private lastSyntacticVersion: number = 0;
+    private lastSemanticVersion: number = 0;
+
+    public constructor(client: Client, colorizationSettings: ColorizationSettings) {
+        this.client = client;
```
bobbrow: Since you only need the client for the RootUri, consider storing that instead.  This will shut off access to other client methods that likely shouldn't be accessible from this class.  It appears that ColorizationSettings only stores the uri.

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+    public reload(): void {
+        let otherSettings: OtherSettings = new OtherSettings(this.uri);
+        let themeName: string = otherSettings.colorTheme;
+        
+        // Enumerate through all extensions, looking for this theme.  (Themes are implemented as extensions - even the default ones)
+        // Crack open each package.json to check for a theme name patch
+        for (let i: number = 0; i < vscode.extensions.all.length; i++) {
+            let extensionPath: string = vscode.extensions.all[i].extensionPath;
+            let extensionPackageJsonPath: string = path.join(extensionPath, "package.json");
+            if (!fs.existsSync(extensionPackageJsonPath)) {
```
bobbrow: In the future, we should consider if we can make this file access async so we can ensure that we're not blocking other operations such as getting the language server started.  What is the impact of doing this synchronously right now?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+                }
+                inactiveRanges = this.fixRanges1(inactiveRanges, edit.changes);
+            }
+        });
+
+        this.updateColorizationRanges(uri, null, semanticRanges, inactiveRanges);
+        this.lastSemanticVersion = editVersion;
+        this.purgeOldVersionedEdits();
+    }
+}
```
bobbrow, line 678: I didn't have time to look at all of this file.  I'll need someone else from the team to sign off here.

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+            accumulatedScope += ".";
+        }
+    }
+
+    public reload(): void {
+        let otherSettings: OtherSettings = new OtherSettings(this.uri);
+        let themeName: string = otherSettings.colorTheme;
+        
+        // Enumerate through all extensions, looking for this theme.  (Themes are implemented as extensions - even the default ones)
+        // Crack open each package.json to check for a theme name patch
```
michelleangela: This is open source code/comment. I would make comments as simple and clear as possible. I would remove "crack" and just say "open".

Extension/src/LanguageServer/client.ts
```diff
@@ -521,22 +577,54 @@ class DefaultClient implements Client {
...
+        if (settings.enhancedColorization) {
+            this.colorizationSettings.reload();
+        }
+
+        // Update colorization
+        let processedUris: vscode.Uri[] = [];
+        for (let e of vscode.window.visibleTextEditors) {
+            let uri: vscode.Uri = e.document.uri;
+            // Make sure we don't process the same file multiple times
+            if (!processedUris.find(e => e === uri)) {
```
michelleangela: Does this work when a file is split in two windows?

Extension/src/LanguageServer/colorization.ts
michelleangela: Is the RootUri only used for CppSettings? Do you need to create a new CppSettings everytime or can the one settings value that's queried be stored?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+            let extensionPackageJsonPath: string = path.join(extensionPath, "package.json");
+            if (!fs.existsSync(extensionPackageJsonPath)) {
+                continue;
+            }
+            let packageJson: any = JSON.parse(fs.readFileSync(extensionPackageJsonPath).toString());
+            if (packageJson.contributes && packageJson.contributes.themes) {
+                for (let j: number = 0; j < packageJson.contributes.themes.length; j++) {
+                    let themeId: string = packageJson.contributes.themes[j].id;
+                    let themeLabel: string = packageJson.contributes.themes[j].label;
+                    if (themeId === themeName || themeLabel === themeName) {
```
michelleangela: Is this for loop only looking for selected theme? Can you do a "find" or a "filter" on the array? https://stackoverflow.com/questions/42580100/typescript-take-object-out-of-array-based-on-attribute-value

michelleangela (changes requested):

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+                            return;
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    public static createDecorationFromThemeStyle(themeStyle: ThemeStyle): vscode.TextEditorDecorationType {
+        if (themeStyle) {
```
michelleangela: condense into one if statement?

Extension/src/LanguageServer/client.ts
```diff
@@ -521,22 +577,54 @@ class DefaultClient implements Client {
...
     }
 
+    private editVersion: number = 0;
+
+    public onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): void {
+
+        this.editVersion++;
+        try {
+            let colorizationState: ColorizationState = this.getColorizationState(textDocumentChangeEvent.document.uri.toString());
+            colorizationState.fixColorizationState(textDocumentChangeEvent.contentChanges, this.editVersion);
```
michelleangela: What does it mean to fix a colorization state?

Extension/src/LanguageServer/client.ts
```diff
@@ -506,6 +526,42 @@ class DefaultClient implements Client {
         return new LanguageClient(`cpptools: ${serverName}`, serverOptions, clientOptions);
     }
 
+    private updateGrammarsInPackageJson(): void {
```
michelleangela: Is this logic something that the client should be responsible for?

Extension/src/LanguageServer/client.ts
```diff
@@ -80,19 +81,30 @@ interface OutputNotificationBody {
...
 
 interface InputRegion {
     startLine: number;
     endLine: number;
 }
 
-interface DecorationRangesPair {
-    decoration: vscode.TextEditorDecorationType;
-    ranges: vscode.Range[];
+interface InputColorizationRegion {
```
michelleangela, line 103: Does InputColorizationRegion  belong in Client or Colorization class? What about SemanticColorizationRegionsParams  and SyntacticColorizationRegionsParams ?

Extension/src/LanguageServer/colorization.ts
```diff
@@ -0,0 +1,601 @@
...
+    public refreshColorizationRanges(e: vscode.TextEditor): void {
+        // Clear inactive regions
+        if (this.inactiveDecoration) {
+            e.setDecorations(this.inactiveDecoration, []);
+        }
+        let settings: CppSettings = new CppSettings(this.client.RootUri);
+        if (settings.enhancedColorization) {
+            for (let i: number = 0; i < TokenKind.Count; i++) {
+                if (this.decorations[i]) {
+
```
michelleangela: nit: remove space?

michelleangela (changes requested):

Extension/src/common.ts
Colengms: languageId should be getting set to "cpp" for .C files.

bobbrow: Do you change this somewhere?  VSCode doesn't look at casing and would normally assign "c" as the language ID to *.C" files.

Extension/src/LanguageServer/client.ts
Colengms: Yes.  (?).  This is calling onSettingsChanged for each file visible in an editor, but it should not do so multiple times for the same file.  onSettingsChanged ensures all visible text editors for that file get refreshed, after it's created the (same) set of decorators for that file.  (The set of decorators are shared by all visible instances of the file).

Extension/src/LanguageServer/client.ts
Colengms: These are for the messages between TS and native sides.  Those interfaces, and the requests, responses, notifications, and their handler functions, are all in client.ts.

Extension/src/LanguageServer/colorization.ts
Colengms: I changed this code pretty significantly.  Let me know if you have similar questions about the new iteration.

Extension/src/LanguageServer/colorization.ts
Colengms: I changed this code pretty significantly.  Let me know if you have similar questions about the new iteration.

Extension/src/LanguageServer/colorization.ts
Colengms: Made this async.  Also fixed(?) BlockingTask and moved it to utils.

Extension/src/LanguageServer/client.ts
Colengms: When the user has made edits, I need to adjust the stored ranges, to continue to colorize correctly until an updated set of ranges is available.  Changed name to updateAfterEdits, which is hopefully more clear.

Extension/src/LanguageServer/colorization.ts
Colengms: As discussed, when creating CppSettings it calls vscode.workspace.getConfiguration(), which essentially takes a snapshot which does not change as settings change.  While OtherSettings calls vscode.workspace.getConfiguration each time a setting is accessed, it's inexpensive as it's just storing the Uri.  As a rule, I think we should create these settings objects as they are needed rather than keep them around.

Extension/src/LanguageServer/settings.ts
```diff
@@ -99,7 +118,15 @@ export class OtherSettings {
     public get searchExclude(): vscode.WorkspaceConfiguration { return vscode.workspace.getConfiguration("search", this.resource).get("exclude"); }
     public get settingsEditor(): string { return vscode.workspace.getConfiguration("workbench.settings").get<string>("editor"); }
 
+    public get colorTheme(): string { return vscode.workspace.getConfiguration("workbench", this.resource).get<string>("colorTheme"); }
+
+    public getCustomColorToken(colorTokenName: string): string { return vscode.workspace.getConfiguration("editor.tokenColorCustomizations", this.resource).get<string>(colorTokenName); }
```
bobbrow: There are a bunch of warnings in the console regarding accessing the tokenColorCustomizations:

`[ms-vscode.cpptools] Accessing a window scoped configuration for a resource is not expected. To associate 'editor.tokenColorCustomizations' to a resource, define its scope to 'resource' in configuration contributions in 'package.json'.`

I think you need to pass `null` for the resource in these cases.

bobbrow: (or just don't pass a second parameter)

Extension/package.json
```diff
@@ -495,6 +495,16 @@
...
           "scope": "resource"
+        },
+        "C_Cpp.enhancedColorization": {
+          "type": "string",
+          "enum": [
+            "Enabled",
+            "Disabled"
+          ],
+          "default": "Enabled",
+          "description": "If enabled, code is colorized based on syntactic and semantic analysis.  This setting has no effect if IntelliSense is disabled.",
```
sean-mcmanus: I think "syntactic" should be "lexical" or "syntactic and semantic analysis" should be something like "IntelliSense results".

sean-mcmanus: I think we do only 1 space after sentences.

Extension/src/LanguageServer/client.ts
```diff
@@ -81,19 +82,30 @@ interface OutputNotificationBody {
     output: string;
 }
 
-interface InactiveRegionParams {
+interface SyntacticColorizationRegionsParams {
```
sean-mcmanus, line 85: I'm fine with "Syntactic" being in the code.

michelleangela (approved):

sean-mcmanus (approved):

Extension/package.json
```diff
@@ -1481,7 +1491,8 @@
     "vscode-debugprotocol": "^1.34.0",
     "vscode-extension-telemetry": "^0.1.1",
     "vscode-languageclient": "^5.2.1",
-    "yauzl": "^2.10.0"
+    "yauzl": "^2.10.0",
+    "plist": "^2.0.1"
```
sean-mcmanus: Oh -- can you alphabetize this?

sean-mcmanus (changes requested):

sean-mcmanus:
Should we add this feature's default setting to the ABTesting so we can dynamically change the default based on feedback? Seems like we should.

sean-mcmanus (approved):

pparuzel:
After latest update I noticed that I can't overwrite the highlight background color and for some themes it seems to be fixed and merges with the editor background making it as if the word wasn't selected.
image
image


Might this have been caused by the addition of colorization?

bobbrow:
Yes.  We'll be investigating this on Monday.  I'm going to turn off the new colorization by default in the meantime.

PR #3658, opened 2019-05-16
Add a command to rescan the workspace

bobbrow:
Addresses https://github.com/microsoft/vscode-cpptools-api/issues/11.
Should we call it "rescan folder" or "rescan workspace"?


sean-mcmanus:
rescan folder seems fine, since the command is just for a single workspace folder and not the entire workspace.

Extension/src/LanguageServer/extension.ts
```diff
@@ -912,6 +913,11 @@ function onLogDiagnostics(): void {
     clients.ActiveClient.logDiagnostics();
 }
 
+function onRescanFolder(): void {
+    onActivationEvent();
+    clients.ActiveClient.rescanFolder();
```
bobbrow: need to think about multi-root

Colengms (approved):

Extension/src/LanguageServer/extension.ts
bobbrow: I talked with @sherryyshi about this and we decided to go with "rescan workspace"

Extension/package.nls.zh-cn.json
```diff
@@ -18,5 +18,6 @@
     "c_cpp.command.showParsingCommands.title": "查看解析命令",
     "c_cpp.command.takeSurvey.title": "调查问卷",
     "c_cpp.command.buildAndDebugActiveFile.title": "生成和调试当前文件",
-    "c_cpp.command.logDiagnostics.title": "记录诊断"
+    "c_cpp.command.logDiagnostics.title": "记录诊断",
+    "c_cpp.command.rescanWorkspace.title": "重新分析工作区"
```
xisui-MSFT: I think "重新扫描工作区" would be a better translation.

bobbrow: thanks! :)

sean-mcmanus (changes requested):
The commands should all "title cased" to match the other VS Code command styles.

Some of our existing commands had the incorrect "sentence" casing and no one noticed.

sean-mcmanus (approved):

PR #3659, opened 2019-05-16
Fix unit tests

bobbrow:
Unit tests were running before the compile test step.  This should fix the Azure pipeline.


sean-mcmanus (approved):
Cool, good catch.

PR #3661, opened 2019-05-16
Add telemetry to configuration UI

michelleangela:
Add telemetry to get usage data on which IntelliSense settings are mostly modified on the UI.

### DIFF
--- a/Extension/src/LanguageServer/settingsPanel.ts
+++ b/Extension/src/LanguageServer/settingsPanel.ts
@@ -210,8 +210,9 @@ export class SettingsPanel {
     }
 
     private logTelementryForElement(elementId: string): void {
-        if(this.telemetry[elementId] === undefined)
+        if (this.telemetry[elementId] === undefined) {
             this.telemetry[elementId] = 0;
+        }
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/settingsPanel.ts
```diff
@@ -165,30 +166,37 @@ export class SettingsPanel {
         switch (message.key) {
             case elementId.activeConfig:
                 this.configValues.name = message.value;
+                telemetry.logLanguageServerEvent("ConfigUI", { "ValueChanged": "configName" }, null);
```
bobbrow: How much telemetry is this going to send?  Is there a way to tell when the user has stopped interacting with the UI (e.g. focus a different tab) and send some cached data instead?  Kind of like the way we batch up LSP messages when logging telemetry.

michelleangela: The data can be cached and sent whenever the webview panel is disposed (either when the config UI window was closed or VS code was closed).

bobbrow (approved):

PR #3680, opened 2019-05-22
Refresh IntelliSense time

andreeis:
Small change to add timing information about how long it takes for IntelliSense to refresh.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -33,6 +33,7 @@ import * as fs from 'fs';
 import * as os from 'os';
 
 let ui: UI;
+let timeStamp: number = 0;
 const configProviderTimeout: number = 2000;
 
 interface NavigationPayload {
@@ -1017,9 +1018,12 @@ class DefaultClient implements Client {
             this.model.isTagParsing.Value = true;
             testHook.updateStatus(Status.TagParsingBegun);
         } else if (message.endsWith("Updating IntelliSense...")) {
+            timeStamp = Date.now();
             this.model.isUpdatingIntelliSense.Value = true;
             testHook.updateStatus(Status.IntelliSenseCompiling);
         } else if (message.endsWith("IntelliSense Ready")) {
+            let duration = Date.now() - timeStamp;
+            console.log("Flame time (sec):", duration/1000);
             this.model.isUpdatingIntelliSense.Value = false;
             testHook.updateStatus(Status.IntelliSenseReady);
         } else if (message.endsWith("Ready")) { // Tag Parser Ready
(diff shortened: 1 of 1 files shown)


pieandcakes:
@andreeis Please run the `tslinter` to fix the failed builds.

pieandcakes:
Is this feature to to show users the time the flame is on? Should we be logging this to telemetry too(if we aren't doing it) so we can see what the trend is?

I don't know if `Flame on` should be the term for this as it has other connotations. @tara-raj Do you have a better term for this?

andreeis:
Small change to add timing information about how long it takes for IntelliSense to refresh.

Extension/src/LanguageServer/client.ts
```diff
@@ -1017,9 +1018,12 @@ class DefaultClient implements Client {
             this.model.isTagParsing.Value = true;
             testHook.updateStatus(Status.TagParsingBegun);
         } else if (message.endsWith("Updating IntelliSense...")) {
+            timeStamp = Date.now();
             this.model.isUpdatingIntelliSense.Value = true;
             testHook.updateStatus(Status.IntelliSenseCompiling);
         } else if (message.endsWith("IntelliSense Ready")) {
+            let duration: number = Date.now() - timeStamp;
+            console.log("Flame time (sec):", duration / 1000);
```
bobbrow: This should only be logged when the loggingLevel is "Debug".
Consider also rewording this to "Refresh IntelliSense time".

sean-mcmanus: I think it should be "Update IntelliSense time" to match the hover message, right?

andreeis: Done

andreeis:
> Is this feature to show users the time the flame is on? Should we be logging this to telemetry too(if we aren't doing it) so we can see what the trend is?

Currently we don't collect this information as telemetry. So far, this change is useful only for visual mapping of a particular timing value with the coding operation done in VSCode. We are thinking whether is worth to collect the same info via telemetry and what is the best way to do that (aggregation criterias, more info about the operation associated with a particular timing value).

Extension/src/LanguageServer/client.ts
```diff
@@ -1015,9 +1016,16 @@ class DefaultClient implements Client {
             this.model.isTagParsing.Value = true;
             testHook.updateStatus(Status.TagParsingBegun);
         } else if (message.endsWith("Updating IntelliSense...")) {
+            timeStamp = Date.now();
             this.model.isUpdatingIntelliSense.Value = true;
             testHook.updateStatus(Status.IntelliSenseCompiling);
         } else if (message.endsWith("IntelliSense Ready")) {
+            let settings: CppSettings = new CppSettings(this.RootUri);
```
bobbrow, line 1023: Did this sneak in?  I don't think you need `settings`, right?

bobbrow: my bad, sorry it's used on the next line.

bobbrow (approved):

michelleangela (approved):

PR #3708, opened 2019-05-28
Add list of detected compiler paths to IntelliSense config UI

michelleangela:
Change input field of compiler path to editable drop-down list. This list will show detected compiler paths.

**Screenshots**
compilerpathlist

compilerpathlis-editt

compilerpathlis-errort

### DIFF
--- a/Extension/ui/settings.ts
+++ b/Extension/ui/settings.ts
@@ -50,12 +50,9 @@ class SettingsApp {
     }
 
     private onKnownCompilerSelect(): void {
-        // todo: add telemetry
         const x: HTMLInputElement = <HTMLInputElement>document.getElementById(elementId.knownCompilers);
-        if (x.value !== "(no compilers detected)") {
-            (<HTMLInputElement>document.getElementById(elementId.compilerPath)).value = x.value;
-            this.onChanged(elementId.compilerPath);
-        }
+        (<HTMLInputElement>document.getElementById(elementId.compilerPath)).value = x.value;
+        this.onChanged(elementId.compilerPath);
     }
 
     private onChanged(id: string): void {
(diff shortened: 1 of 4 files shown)


michelleangela:
Verified UI on Linux and Mac

sean-mcmanus:
Can you normalize the `/` and `\\` in the dropdown list of compilers? Maybe using the "C_Cpp.preferredPathSeparator"?

sean-mcmanus:
Can you increase the top margin in the dropdown? It seems too little top margin in the screenshot.

Extension/ui/settings.ts
```diff
@@ -95,26 +107,50 @@ class SettingsApp {
...
+        let list: HTMLElement = document.getElementById(elementId.knownCompilers);
+
+        // No need to add items unless webview is reloaded, in which case it will not have any elements.
+        // Otherwise, add items again.
+        if (list.firstChild) {
+           return;
+        }
+
+        if (compilers.length === 0) {
+            const noCompilers: string = "(No other compiler paths detected)";
```
sean-mcmanus: Why is the word "other" used? Can that be removed?

sean-mcmanus (changes requested):
I left some comments.

Extension/ui/settings.ts
michelleangela: "other" was used in comparison to a specified path. This doesn't make sense though to use just for the list of detected paths. 

Changing to "(No compiler paths detected)"

michelleangela:
> Can you increase the top margin in the dropdown? It seems too little top margin in the screenshot.

Styling the dropdown is not straightforward. Haven't been able to figure out it. Will need to research and look into VS Code repo. The dropdown style is the same on the existing dropdowns.

sean-mcmanus:
> Styling the dropdown is not straightforward. Haven't been able to figure out it. Will need to research and look into VS Code repo. The dropdown style is the same on the existing dropdowns.

Isn't it just the style margins or something? I remember changing some of the margins with other UI a few weeks ago. Oh, looks like the existing dropdowns have the same issue, i.e. we've shipped like this already...I just never noticed.

michelleangela:
> > Styling the dropdown is not straightforward. Haven't been able to figure out it. Will need to research and look into VS Code repo. The dropdown style is the same on the existing dropdowns.
> 
> Isn't it just the style margins or something? I remember changing some of the margins with other UI a few weeks ago. Oh, looks like the existing dropdowns have the same issue, i.e. we've shipped like this already...I just never noticed.

The HTML `select` and `option` elements cannot be styled the same way as the other elements. The style depends on the OS, and requires more than CSS to modify the native style. As the current native style works for now, we can change the style later to look like VS Code after we've put in more features/functionalities on the config UI.

sean-mcmanus (approved):

Issue #3709, opened 2019-05-29
High CPU due to parse_file


sean-mcmanus:
Hiya,

<img width="882" alt="Screenshot 2019-05-24 at 16 08 29" src="https://user-images.githubusercontent.com/4742697/58360601-8cfd6680-7e3e-11e9-9a57-4238c181b12f.png">

Once again seeing high cpu usage recently, must have been the May update. 70% -> 30% battery in about 20 minutes!

Sample of process attached.
[sample-process.txt](https://github.com/microsoft/vscode-cpptools/files/3218833/sample-process.txt)

_Originally posted by @jamesadevine in https://github.com/microsoft/vscode-cpptools/issues/1816#issuecomment-495811437_

sean-mcmanus:
@jamesadevine Hmm...that is the 1st report we've seen of any sort of performance regression like this with our latest update, which doesn't really change any of that code in the sample-process.txt. It looks as if your database might have gotten invalided for some reason and it's re-parsing files. One thing to double check is that your database isn't getting clobbered via having multiple different projects pointing to the same file or opening the same folder in different VS Code instances -- in your screenshot I see 3 extension processes -- are you using 3 workspace folders?

If you wait for the data base parsing to finish and then re-open the workspace is everything back to normal? If you set your C_Cpp.loggingLevel to Debug you can check if we output a database invalidation message for some reason.

I think we need more repro info.

jamesadevine:
Hi @sean-mcmanus,

Yes I almost definitely had more than one VS code instance pointing to the same folder tree. As a user (with no understanding of the implementation) I wouldn't expect multiple processes to invalidate the database. I work on https://github.com/Microsoft/pxt/ and my workflow is usually:

1 x workspace containing all the files with folders containing CPP files.
1 x workspace containing project files with folders containing CPP files.
1 x workspace containing just the CPP files.

It looks a little like this:

```
Root (workspace)
    typscript files
    projects (workspace)
        more typescript
        CPP source files (workspace)
            libraries
                CPP...
            project files
```

sean-mcmanus:
What is your databaseFilename setting? The default/unchanged? If you're using 3 different workspace folders in the graph shown it should be creating 3 different browse.vc.db files for each workspace in your workspaceStorage location. If you don't need the C/C++ symbols in some workspaces you can add the C/C++ folders to your files.exclude setting for that particular workspace to avoid parsing the C/C++ files.

jamesadevine:
[...]

PR #3715, opened 2019-05-30
Change multi-root storagePath to use the index instead of name.

sean-mcmanus:
Fix 2 workspace folders with the same name using the same storage path: discovered in https://github.com/microsoft/vscode-cpptools/issues/3709


michelleangela (approved):

sean-mcmanus:
Bob says I should try to delete the old location...maybe I can just keep the previous naming and just add a new folder for the duplicates...

michelleangela (approved):

PR #3738, opened 2019-06-06
Update colorization support to only return tokens for currently visible ranges

Colengms:
Prior to this change, colorization would have issues with very large files consuming extensive resources on the TS side.  This change maintains the tokens on the native side, and returns tokens only for currently visible regions.  A return receipt is returned when handled, to allow throttling of colorization notifications.


Documentation/LanguageServer/colorization.md
```diff
@@ -0,0 +1,87 @@
+
+# VSCode C/C++ Extension - Enhanced Colorization
+
+The VSCode C/C++ extension now supports lexical/syntactic and semantic colorization, using IntelliSense.
+
+# Theming
+
+Colors can be associated using the existing support for theming and color customization in VSCode.  Documentation on Theming in VS Code can be found <a HREF="https://code.visualstudio.com/docs/getstarted/themes">here</a>.
```
michelleangela: Change "VSCode" to "VS Code"? See other "VSCode" in text.

Documentation/LanguageServer/colorization.md
```diff
@@ -0,0 +1,87 @@
...
+
+Colors are associated with <a HREF="https://macromates.com/manual/en/language_grammars#naming_conventions">TextMate scopes</a>.
+
+
+# All IntelliSense Tokens and Scopes
```
michelleangela: Maybe remove the word "All" unless there is a token that will never be listed.

Documentation/LanguageServer/colorization.md
```diff
@@ -0,0 +1,87 @@
+
+# VS Code C/C++ Extension - Enhanced Colorization
+
+The VS Code C/C++ extension now supports lexical/syntactic and semantic colorization, using IntelliSense.
+
```
sean-mcmanus, line 9: Is this comma supposed to be here? Should you explicitly say whether the lexical colorization requires IntelliSense or not?

Documentation/LanguageServer/colorization.md
```diff
@@ -0,0 +1,87 @@
...
+# Theming
+
+Colors can be associated using the existing support for theming and color customization in VS Code.  Documentation on Theming in VS Code can be found <a HREF="https://code.visualstudio.com/docs/getstarted/themes">here</a>.
```
sean-mcmanus: Is there a reason you are using an HTML anchor instead of a markdown link?

sean-mcmanus (approved):

Documentation/LanguageServer/colorization.md
Colengms: I think the comma is as intended, as the 'using' applies to more than just 'semantic colorization'.  I was being intentionally vague for simplicity, rather than explain the difference between lexical/syntactic and semantic approaches, as that distinction may not be meaningful to most people.  If IntelliSense is disabled, the intention was that enhancedColorization would be fully disabled, rather than provide only lexical/syntactic.  The setting documentation includes: "This setting has no effect if IntelliSense is disabled."  … That said, it looks like I'll need to address some issues with enabling/disabling IntelliSense, to ensure that enables/disables colorization properly.

Documentation/LanguageServer/colorization.md
Colengms: Nope, just inexperience with markdown.  :)

sean-mcmanus (approved):

Documentation/LanguageServer/colorization.md
michelleangela: Maybe say something like "The extension now supports feature B when feature A is enabled." or with the use of IntelliSense.

Extension/src/LanguageServer/client.ts
```diff
@@ -616,12 +622,31 @@ class DefaultClient implements Client {
...
+        } else {
+            isSame = ranges.length === 0;
+        }
+        if (!isSame) {
+            this.visibleRanges.set(uri.toString(), ranges);
+            let params: DidChangeVisibleRangesParams = {
+                uri: uri.toString(),
+                ranges: ranges
+            };
+            this.notifyWhenReady(() => this.languageClient.sendNotification(DidChangeVisibleRangesNotification, params));
```
sean-mcmanus, line 648: This doesn't work still. When I put a breakpoint here it starts working though, so there's some sort of race condition.

sean-mcmanus (changes requested):
Existing open files doesn't work still. When I put a breakpoint here it starts working though, so there's some sort of race condition.

Extension/src/LanguageServer/client.ts
Colengms: I'm not able to repro an issue here, and don't see where there could be a race.  When you get a chance, I'd like to see your repro.

Documentation/LanguageServer/colorization.md
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    "scope": "entity.name.type",
+                    "settings": {
+                        "foreground": "#FF0000",
+                        "fontStyle": "italic bold underline"
+                    }
+                }
+            ]    
+        }
+```
```
bobbrow, line 91: Let's add example overrides at the theme level  (e.g. dark theme overrides and light theme overrides)

Documentation/LanguageServer/colorization.md
```diff
@@ -89,3 +89,445 @@ Or, overridden on a per-theme basis:
             ]    
         }
 ```
+
+Use the following to augment the Visual Studio Dark theme to match what Visual Studio would display for C++ files.
```
bobbrow, line 93: 👍 👍 👍 👍

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -464,7 +480,7 @@ class DefaultClient implements Client {
                 errorSquiggles: settings.errorSquiggles,
                 dimInactiveRegions: settings.dimInactiveRegions,
                 textMateColorization: settings.textMateColorization,
-                enhancedColorization: settings.enhancedColorization,
+                enhancedColorization: abTestSettings.UseEnhancedColorization,
```
bobbrow: use `settings.enhancedColorization` here. We change the default, but should still use whatever VS Code reports to us here.

Issue #3664, opened 2019-05-18
Change the default intelliSenseCachePath/databaseFilename to "~/.cpptools" instead of ${workspaceFolder}/.vscode and ${workspaceStorage}


sean-mcmanus:
This would prevent the space used by all ipch across all workspaces to be bounded by the intelliSenseCacheSize instead of being unbounded (i.e. the size limit wouldn't be per workspace).

Lennon925:
I hope you can combine intelliSenseCachePath and cpptools database storage path to be one path which name may be “cpptoolsStoragePath”. And then you can add subfolders for intelliSenseCache and db storage path. If so, we just need to set this path 'cpptoolsStoragePath'.
It's strange to set database filename instead of database storage path.

Such as:
If I don't set cpptoolsStoragePath,it's like this:
![1](https://user-images.githubusercontent.com/39362073/58242036-b35ccc80-7d80-11e9-9ae8-49cb9f8814ac.PNG)
If I set cpptoolsStoragePath to "D:\VsCodeCpptoolsStoragePath"
![2](https://user-images.githubusercontent.com/39362073/58242063-be176180-7d80-11e9-9a34-bfd4097b2bb9.PNG)

sean-mcmanus:
Yeah, I think we should be able to use the same workspace storage hash for the per-workspace browse database files as well.

michelleangela:
<h3>Proposed changes to default path of IntelliSense cache path</h3>

We propose using the user directory as the new default IntelliSense cache path (see proposed new default path below). The proposed solution will, by default, reduce the disk space usage and alleviate implications on source control for users who have not configured the cache. 

As part of this proposed change, the extension will also automatically remove any caches previously added to the `${workspaceFolder}/.vscode` folder when the old default behavior was applied.


**Disk Space**
By using the user directory as the default path, it will create one cache location per user for the extension instead of creating a cache per workspace. As the cache size limit is applied to a cache location, having one cache location will also reduce the disk space usage of the cache for everyone using the default setting value. 


**Source Control**
Moving away from the workspace folder to the user directory as the default path will also address the concerns of the cache implications on source control for the workspace folder for everyone using the default path.

**New default path**
The new default path for `C_Cpp.intelliSenseCachePath` setting will be the following:

On Windows
`%USERPROFILE%/.cpptools`

On Linux and Mac
`~/.cpptools`

<h3>Please provide us with your feedback on this proposal.</h3>

[...]

PR #3747, opened 2019-06-10
Change cache path default

michelleangela:
Change the default cache path from workspace to user folder.

See details in #3664

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -294,7 +294,7 @@
         "C_Cpp.intelliSenseCachePath": {
           "type": "string",
           "default": null,
-          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%USERPROFILE%\/.cpptools\" on Windows and \"~/.cpptools\" on Linux and Mac. The default path will be used if no path is specified or if a specified path is invalid.",
+          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%LocalAppData%\/Microsoft/vscode-cpptools\" on Windows and \"~/.vscode-cpptools\" on Linux and Mac. The default path will be used if no path is specified or if a specified path is invalid.",
           "scope": "resource"
         },
         "C_Cpp.intelliSenseCacheSize": {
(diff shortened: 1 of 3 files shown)


Extension/src/common.ts
```diff
@@ -292,6 +292,17 @@ export function isOptionalArrayOfString(input: any): input is string[]|undefined
     return input === undefined || isArrayOfString(input);
 }
 
+export function resolveCachePath(input: string, additionalEnvironment: {[key: string]: string | string[]}): string {
+    let resolvedPath: string = (process.platform === 'win32') ? process.env.USERPROFILE : process.env.HOME;
+    if (!input || input === "") {
```
WardenGnaw: "" (Empty string) should already be a [falsy](https://basarat.gitbooks.io/typescript/docs/javascript/truthy.html). Where are you doing the check to see if specified path is invalid?

E.g. just
`if (!input)`

Extension/package.json
```diff
@@ -293,8 +293,8 @@
         },
         "C_Cpp.intelliSenseCachePath": {
           "type": "string",
-          "default": "${workspaceFolder}/.vscode",
-          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default path \"${workspaceFolder}/.vscode\" will be used if a specified path is invalid.",
+          "default": null,
+          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%USERPROFILE%\/.cpptools\" on Windows and \"~/.cpptools\" on Linux and Mac. The default path will be used if no path is specified or if a specified path is invalid.",
```
sean-mcmanus: I think we wanted to change this to something more specific like ms-vscode.cpptools.

Extension/src/common.ts
michelleangela: The actual verification if any path exists will continue to remain in the C++ code.

Extension/package.json
michelleangela: I agree. I'll wait for more feedback from the users and then we can decided on the name and final location.

bobbrow: I think we're going to go with `%LocalAppData%\Microsoft\vscode-cpptools`

Issue #3637, opened 2019-05-13
intelliSenseMode clang-x64 in incompatible with compilerPath


Kri-Ol:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10 1809 x64
- VS Code Version: 1.33.1
- C/C++ Extension Version: 0.23 insides
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is.
Have Intellisense mode set to "clang-x64", and "compilerPath": "C:/Program Files/LLVM/bin/clang-cl.exe", which is LLVM 8.0. Compiler works fine, but there is warning message which is not right, I believe
**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Kri-Ol:
"windowsSdkVersion": "10.0.17763.0",
            "compilerPath": "C:/Program Files/LLVM/bin/clang-cl.exe",
            "cStandard": "c11",
            "cppStandard": "c++17",
            "intelliSenseMode": "clang-x64",

Clang-cl from LLVM v8.0 is main compiler used alongside with Visual C++ 2019 Pro

minkir014:
Remove `compiler path` setting intellisense works great without it.

michelleangela:
Hi @Kri-Ol, the usage of compiler `clang-cl.exe` with IntelliSense mode `clang-x64` should not show an incompatibility error message. You can use `clang-x64` with `clang-cl.exe` and just ignore the error. I'll change the logic that validates the compatibility; the fix will be available in a future release.

minkir014:
I say if he removed that setting that will work. It works for me greatly.

sean-mcmanus:
Fixed with 0.24.0-insiders (available by setting "C_Cpp.updateChannel" to "Insiders").

PR #3752, opened 2019-06-11
Update IntelliSense mode and compiler compatibility validation

michelleangela:
Get the full compiler name, if successful in getting compiler path and extracting any arguments. Then use full compiler name to checked against IntelliSense mode compatibility.

Fix for #3637 (intelliSenseMode clang-x64 in incompatible with compilerPath).

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -808,8 +808,11 @@ export function extractCompilerPathAndArgs(inputCompilerPath: string): CompilerP
                     additionalArgs = compilerPath.substr(spaceStart + 1).split(" ");
                     additionalArgs = additionalArgs.filter((arg: string) => { return arg.trim().length !== 0; }); // Remove empty args.
                     compilerPath = potentialCompilerPath;
-                    compilerName = compilerPath.replace(/^.*(\\|\/|\:)/, '');
                 }
+            }
+            // Get compiler name if there are no args but path is valid or a valid path was found with args.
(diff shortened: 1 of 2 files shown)


Extension/src/common.ts
```diff
@@ -785,6 +787,7 @@ export function extractCompilerPathAndArgs(inputCompilerPath: string): CompilerP
                 additionalArgs = compilerPath.substr(endQuote + 1).split(" ");
                 additionalArgs = additionalArgs.filter((arg: string) => { return arg.trim().length !== 0; }); // Remove empty args.
                 compilerPath = compilerPath.substr(1, endQuote - 1);
+                compilerName = compilerPath.replace(/^.*(\\|\/|\:)/, '');
```
sean-mcmanus, line 790: Does this work when a path is specified as an arg to the compiler path?

michelleangela: The  `compilerPath.replace(/^.*(\\|\/|\:)/, '');`  will only be applied if compilerPath was extracted with quotes around it, so the regex will not be applied to args.

I also moved the other `compilerPath.replace(/^.*(\\|\/|\:)/, '');`  line after trying to get a potentialCompilerPath. It now only gets the compiler if a valid compilerPath was found, meaning that is when compilerPath only contains the compiler path and it doesn't have any args. So `compilerPath.replace(/^.*(\\|\/|\:)/, '');` will only be applied to a compiler path without args.

sean-mcmanus (approved):

PR #3770, opened 2019-06-13
Enable enhancedColorization by default.  Fix refresh race in colorization.ts

Colengms:


Extension/src/LanguageServer/colorization.ts
```diff
@@ -442,7 +442,7 @@ export class ColorizationState {
         this.disposeColorizationDecorations();
     }
 
-    public refresh(e: vscode.TextEditor): void {
+    private refresh_inner(e: vscode.TextEditor): void {
```
bobbrow: nit: this is a non-standard naming convention for TS.

sean-mcmanus (approved):

PR #3778, opened 2019-06-15
Make handleSquiggles no longer dependent on special escaping in parsePropertiesFile

Colengms:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -865,7 +860,7 @@ export class CppProperties {
             success = false;
         }
 
-        if (handleSquiggles && success) {
+        if (success) {
```
michelleangela, line 863: so squiggles are always handled now? Does this work for when edits are done to the UI and then user goes to json file?

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
michelleangela: Oh I see, it only applies escapeForSquiggles on a copy of config and not the main copy.

michelleangela (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1044,11 +1039,18 @@ export class CppProperties {
...
+
+            // Replace all \<escape character> with \\<character>, except for \"
+            // Otherwise, the JSON.parse result will have the \<escape character> missing.
+            let configurationsText: string = util.escapeForSquiggles(curText);
+            let configurations: ConfigurationJson = JSON.parse(configurationsText);
+            let currentConfiguration: Configuration = configurations.configurations[this.CurrentConfigurationIndex];
+
             let curTextStartOffset: number = 0;
-            if (!this.CurrentConfiguration.name) {
+            if (!currentConfiguration.name) {
```
bobbrow: I'm pretty sure `this.CurrentConfiguration` is used as a safety net that is guaranteed not to be undefined.  Please check before changing this.

PR #3791, opened 2019-06-17
Vcpkg integration

grdowns:
- Adds `VcpkgClipboardInstallSuggested` and `VcpkgOnlineHelpSuggested` commands/code actions
    - `VcpkgClipboardInstallSuggested`
        - From command palette: globs up all missing #include errors and matches each to an include in vcpkg and copies a `vcpkg install` line to the user's clipboard. E.g. `vcpkg install boost-hana:x64-linux`
        - From code action context menu: Matches the current line to an include in vcpkg and copies a `vcpkg install` line to the user's clipboard
    - `VcpkgOnlineHelpSuggested`
        - Sends the user to [aka.ms/vcpkg](https://aka.ms/vcpkg)
        - Function does not differ between command palette and code action context menu

The C++ side of the PR will be coming soon. This PR requires these changes in the C++ side:
- Add a `source` and `code` to missing #include diagnostics
- Copy over `VCPkgHeaderDatabase.zip` using CMake's `install` function


Extension/src/LanguageServer/extension.ts
```diff
@@ -933,6 +1036,79 @@ function onTakeSurvey(): void {
...
+            break;
+        case 'darwin':
+            triplets = ['x64-osx'];
+            break;
+        default:
+            triplets = ['x64-linux'];
+    }
+
+    let installCommand: string = 'vcpkg install';
+    ports.forEach(port => {
```
michelleangela: Will there always be at least one install to return? Checking if there's a case when it will only return "vcpkg install"

grdowns: `ports` is checked prior, but it is possible that it will be defined but the length is 0. Thanks!

Extension/package.json
```diff
@@ -294,7 +294,7 @@
         "C_Cpp.intelliSenseCachePath": {
           "type": "string",
           "default": null,
-          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%LocalAppData%\/Microsoft/vscode-cpptools\" on Windows and \"~/.vscode-cpptools\" on Linux and Mac. The default path will be used if no path is specified or if a specified path is invalid.",
+          "description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%LocalAppData%/Microsoft/vscode-cpptools\" on Windows and \"~/.vscode-cpptools\" on Linux and Mac. The default path will be used if no path is specified or if a specified path is invalid.",
```
grdowns: This change likes to sneak in. Does this get rewritten when the extension activates? It appears to be a typo. @michelleangela

Extension/src/LanguageServer/extension.ts
```diff
@@ -933,6 +1039,80 @@ function onTakeSurvey(): void {
...
+        });
+        ports = [].concat(...(await Promise.all(portsPromises)));
+        if (!ports.length) {
+            return Promise.resolve();
+        }
+        ports = ports.filter((port: string, index: number) => { return ports.indexOf(port) === index; });
+    }
+
+    let triplets: string[];
+    const platformInfo: PlatformInformation = await PlatformInformation.GetPlatformInformation();
```
WardenGnaw: Unless you need arch and distro, use `os.platform()` to get the platform instead.

Extension/package.nls.json
```diff
@@ -19,5 +19,7 @@
     "c_cpp.command.takeSurvey.title": "Take Survey",
     "c_cpp.command.buildAndDebugActiveFile.title": "Build and Debug Active File",
     "c_cpp.command.logDiagnostics.title": "Log Diagnostics",
-    "c_cpp.command.rescanWorkspace.title": "Rescan Workspace"
+    "c_cpp.command.rescanWorkspace.title": "Rescan Workspace",
+    "c_cpp.command.VcpkgClipboardInstallSuggested.title": "Copy vcpkg install command to clipboard",
```
sean-mcmanus: This should be "title" case. And translated for zh-ch and it languages. I think Bob might know how to translate.

Extension/package.json
grdowns: Resolved after a merge w/ master

Extension/package.json
```diff
@@ -506,6 +506,15 @@
           "default": "Disabled",
           "description": "If enabled, code is colorized based on IntelliSense. This setting has no effect if IntelliSense is disabled.",
           "scope": "resource"
+        },
+        "C_Cpp.dependencyManager": {
```
sean-mcmanus: Is "dependencyManager" the correct terminology? I thought it was packageManager.

Extension/package.json
```diff
@@ -506,6 +506,15 @@
...
+          "type": "string",
+          "enum": [
+            "vcpkg",
+            "none"
+          ],
+          "default": "vcpkg",
+          "markdownDescription": "Enable integration services for the [vcpkg dependency manager](https://aka.ms/vcpkg/)."
```
sean-mcmanus: @michelleangela @bobbrow @tara-raj It would be nice to add more markup to our other settings descriptions to make it easier to read and with more links to docs (not in this pull request).

grdowns: Yes! I found it yesterday after a bit of experimenting. It's undocumented AFAIK

Extension/package.json
grdowns: The preferred terminology has changed. Vcpkg is still recognized as a package manager, but **officially** it is a "library manager" or a bit more generally than that a "dependency manager" :man_shrugging:

grdowns:
All feedback has been responded to, translations should be all that remains. @bobbrow this PR is ready for review.

Extension/src/commands.ts
```diff
@@ -36,6 +36,7 @@ class TemporaryCommandRegistrar {
         "C_Cpp.TakeSurvey",
         "C_Cpp.LogDiagnostics",
         "C_Cpp.RescanWorkspace",
+        "C_Cpp.VcpkgClipboardInstallSuggested"
     ];
```
bobbrow, line 41: Missing `C_Cpp.VcpkgOnlineHelpSuggested` or this that command internal?

Extension/src/main.ts
```diff
@@ -369,6 +369,7 @@ function rewriteManifest(): Promise<void> {
         "onCommand:C_Cpp.TakeSurvey",
         "onCommand:C_Cpp.LogDiagnostics",
         "onCommand:C_Cpp.RescanWorkspace",
+        "onCommand:C_Cpp.VcpkgClipboardInstallSuggested",
         "onDebug",
```
bobbrow, line 374: Missing `C_Cpp.VcpkgOnlineHelpSuggested` or is that command internal?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -351,6 +351,9 @@ export class CppProperties {
             }
         } catch (error) {} finally {
             this.vcpkgPathReady = true;
+            if (this.vcpkgIncludes.length) {
+                telemetry.logLanguageServerEvent('vcpkgUserPlatform', { 'platform': await os.platform() });
```
bobbrow: I don't believe this was on our telemetry list.  In any case the OS info is automatically appended to all events and it is unnecessary to add.

Extension/src/LanguageServer/extension.ts
```diff
@@ -85,6 +165,32 @@ export function activate(activationEventOccurred: boolean): void {
         }
     });
 
+    vcpkgDbPromise = initVcpkgDatabase();
```
bobbrow: I don't believe this belongs here.  You should only read the vcpkg database in the `realActivation` function.

Extension/src/LanguageServer/client.ts
```diff
@@ -1597,6 +1608,8 @@ class NullClient implements Client {
     rescanFolder(): Promise<void> { return Promise.resolve(); }
     getCurrentConfigName(): Thenable<string> { return Promise.resolve(""); }
     getCompilerPath(): Thenable<string> { return Promise.resolve(""); }
+    getVcpkgInstalled(): Thenable<boolean> { return Promise.resolve(false); }
+    getDependencyManager(): Thenable<string> { return Promise.resolve('vcpkg'); }
```
bobbrow: I would put 'none' here.

Extension/src/LanguageServer/client.ts
```diff
@@ -877,6 +879,15 @@ class DefaultClient implements Client {
         return this.queueTask(() => Promise.resolve(this.configuration.CompilerPath));
     }
 
+    public getVcpkgInstalled(): Thenable<boolean> {
+        return this.queueTask(() => Promise.resolve(this.configuration.VcpkgInstalled));
+    }
+
+    public getDependencyManager(): Thenable<string> {
+        const cppSettings: CppSettings = new CppSettings(this.RootUri);
+        return this.queueTask(() => Promise.resolve(cppSettings.dependencyManager));
```
bobbrow: You don't need to queue this one because it doesn't depend on the configuration.

Extension/package.json
```diff
@@ -614,6 +623,16 @@
...
         "title": "%c_cpp.command.rescanWorkspace.title%",
         "category": "C/C++"
+      },
+      {
+        "command": "C_Cpp.VcpkgClipboardInstallSuggested",
+        "title": "%c_cpp.command.vcpkgClipboardInstallSuggested.title%",
+        "category": "C/C++"
+      },
+      {
+        "command": "C_Cpp.VcpkgOnlineHelpSuggested",
```
bobbrow, line 643: I'm unclear as to whether this is meant to be an internal command or not.

bobbrow (changes requested):
I'm not done yet, but here is some feedback in the interim.

Extension/package.json
grdowns: I've added it to the other necessary places

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
 const taskSourceStr: string = "C/C++";
 const cppInstallVsixStr: string = 'C/C++: Install vsix -- ';
 let taskProvider: vscode.Disposable;
+let codeActionProvider: vscode.Disposable;
 const intelliSenseDisabledError: string = "Do not activate the extension when IntelliSense is disabled.";
 
+type vcpkgDatabase = { [key: string]: string[] }; // Stored as <header file entry> -> [<port name>]
+let vcpkgDbPromise: Promise<vcpkgDatabase>;
+async function initVcpkgDatabase(): Promise<vcpkgDatabase> {
```
bobbrow: `async` features are not used in this function. You should remove `async`

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+    return new Promise((resolve, reject) => {
+        yauzl.open(util.getExtensionFilePath('VCPkgHeadersDatabase.zip'), { lazyEntries: true }, async (err? : Error, zipfile?: yauzl.ZipFile) => {
```
bobbrow: `async` features are not used in this function. You should remove `async`.

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+            if (err) {
+                return resolve({});
```
bobbrow: This function has no return type: `void`.
```
resolve({});
return;
```

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+            }
+            zipfile.readEntry();
+            zipfile.on('entry', entry => {
+                if (entry.fileName !== 'VCPkgHeadersDatabase.txt') {
+                    return resolve({});
```
bobbrow: same as above. (no return type)

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+                        const relativeHeader: string = portFilePair[1];
+
+                        if (!database[relativeHeader]) {
+                            database[relativeHeader] = [];
+                        }
+
+                        database[relativeHeader].push(portName);
+                    });
+                    reader.on('close', () => {
+                        return resolve(database);
```
bobbrow: same as above. (no return type)

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+                });
+            });
+        });
+    });
+}
+
+function getVcpkgHelpAction(): vscode.CodeAction {
+    const dummy: any[] = [{}]; // To distinguish between entry from CodeActions and the command palette
+    return {
+        command: { title: 'vcpkgOnlineHelpSuggested', command: 'C_Cpp.VcpkgOnlineHelpSuggested', arguments: dummy },
```
bobbrow, line 106: Is `title` correct?  I thought this was a localized string.

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+        title: "Learn how to install a library for this header with vcpkg",
+        kind: vscode.CodeActionKind.QuickFix
+    };
+}
+
+function getVcpkgClipboardInstallAction(port: string): vscode.CodeAction {
+    return {
+        command: { title: 'vcpkgClipboardInstallSuggested', command: 'C_Cpp.VcpkgClipboardInstallSuggested', arguments: [[port]] },
```
bobbrow, line 114: Is `title` correct?

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,86 @@ let buildInfoCache: BuildInfo | null = null;
...
+        title: `Copy vcpkg command to install '${port}' to the clipboard`,
+        kind: vscode.CodeActionKind.QuickFix
+    };
+}
+
+async function lookupIncludeInVcpkg(document: vscode.TextDocument, line: number): Promise<string[]> {
```
bobbrow, line 120: This function mixes `async` and `Promise` semantics.  Don't return `Promise<string[]>`, just `string[]`.  `async` generates the promises for you.

Extension/src/LanguageServer/extension.ts
```diff
@@ -85,6 +165,31 @@ export function activate(activationEventOccurred: boolean): void {
         }
     });
 
+    const selector: vscode.DocumentSelector = [
+        { scheme: 'file', language: 'cpp' },
+        { scheme: 'file', language: 'c' }
+    ];
+    codeActionProvider = vscode.languages.registerCodeActionsProvider(selector, {
+        provideCodeActions: async (document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext, token: vscode.CancellationToken): Promise<vscode.CodeAction[]> => {
```
bobbrow, line 186: This function mixes `async` and `Promise` semantics. Don't return `Promise<T[]>`, just `T[]`. async generates the promises for you.

Extension/src/LanguageServer/extension.ts
```diff
@@ -85,6 +165,31 @@ export function activate(activationEventOccurred: boolean): void {
...
+                return Promise.resolve([]);
+            }
+
+            telemetry.logLanguageServerEvent('codeActionsProvided', { "source": "vcpkg" });
+
+            if (!await clients.ActiveClient.getVcpkgInstalled()) {
+                return Promise.resolve([getVcpkgHelpAction()]);
+            }
+
+            return lookupIncludeInVcpkg(document, range.start.line).then(ports => ports.map<vscode.CodeAction>(getVcpkgClipboardInstallAction));
```
bobbrow: use `await` here instead of `then`

Extension/src/LanguageServer/extension.ts
```diff
@@ -934,6 +1043,67 @@ function onTakeSurvey(): void {
     vscode.commands.executeCommand('vscode.open', uri);
 }
 
+async function onVcpkgOnlineHelpSuggested(dummy?: any): Promise<void> {
```
bobbrow: This function doesn't need to be async nor does it need to return a promise.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
grdowns: This function asynchronously reads a text file. Are you sure this is safe to remove?

Extension/src/LanguageServer/extension.ts
grdowns: Thanks! TIL

Extension/src/LanguageServer/extension.ts
bobbrow: `async` is a keyword that tells the compiler to generate Promises from return values. In this lambda you are explicitly resolving a previously created Promise, so you don't need this.

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,89 @@ let buildInfoCache: BuildInfo | null = null;
...
+    return new Promise((resolve, reject) => {
+        yauzl.open(util.getExtensionFilePath('VCPkgHeadersDatabase.zip'), { lazyEntries: true }, async (err? : Error, zipfile?: yauzl.ZipFile) => {
+            if (err) {
+                resolve({});
+                return;
+            }
+            zipfile.readEntry();
+            zipfile.on('entry', entry => {
+                if (entry.fileName !== 'VCPkgHeadersDatabase.txt') {
+                    resolve({});
```
bobbrow: Actually, you shouldn't `resolve` the promise here. You should just ignore the entry.  But you probably want a handler for 'end' in case the entry was never encountered so you can resolve the promise as empty.

Extension/src/LanguageServer/extension.ts
```diff
@@ -42,8 +44,89 @@ let buildInfoCache: BuildInfo | null = null;
...
+
+                        if (!database[relativeHeader]) {
+                            database[relativeHeader] = [];
+                        }
+
+                        database[relativeHeader].push(portName);
+                    });
+                    reader.on('close', () => {
+                        resolve(database);
+                        return;
```
bobbrow: Explicit `return` is not necessary here.

grdowns:
As a heads up, I'm dealing with the command title localization right now. VS Code has a `vscode-nls` localization module

Extension/src/LanguageServer/extension.ts
```diff
@@ -934,6 +1048,67 @@ function onTakeSurvey(): void {
...
+            return;
+        }
+        ports = ports.filter((port: string, index: number) => { return ports.indexOf(port) === index; });
+    }
+
+    let installCommand: string = 'vcpkg install';
+    ports.forEach(port => installCommand += ` ${port}`);
+    telemetry.logLanguageServerEvent('vcpkgAction', { 'source': source, 'action': 'vcpkgClipboardInstallSuggested', 'ports': ports.toString() });
+
+    return vscode.env.clipboard.writeText(installCommand);
```
bobbrow: `onVcpkgClipboardInstallSuggested` returns Thenable<void>.  The right thing to do is `await` the `writeText` function call if you want this function to return a promise that fires after `writeText` completes.

Extension/src/LanguageServer/extension.ts
grdowns: Oh! I see what you mean now. Thanks.

Extension/src/LanguageServer/extension.ts
grdowns: Good catch!

Extension/src/LanguageServer/extension.ts
```diff
@@ -82,10 +81,13 @@ async function initVcpkgDatabase(): Promise<vcpkgDatabase> {
                     });
                     reader.on('close', () => {
                         resolve(database);
-                        return;
                     });
                 });
             });
+            zipfile.on('end', () => { // In case the header db was not found
+                resolve({});
```
bobbrow: You only want to resolve (or reject) if the promise has not been previously resolved.

Extension/src/LanguageServer/extension.ts
grdowns: Discussed offline. Localization on these strings will be done as a part #3963

bobbrow (approved):

grdowns:
`TS2741: Property 'getCompilerPath' is missing in type 'DefaultClient' but required in type 'Client'`

grdowns:
I'm handling it -- one moment

Issue #3781, opened 2019-06-15
0.24.0-insiders - Noticeable latency with enhanced colorization


notAlaanor:
**Type: LanguageService**

With Enhanced Colorization enabled there is a noticeable amount of latency when typing and simply moving around the document.

Disabling Enhanced Colorization immediately fixes this issue.

Note: Is this a flaw of using the Decorations API?

sean-mcmanus:
Fixed with 0.24.0-insiders2. Set the "C_Cpp.updateChannel" setting to "Insiders" to auto-install it.

Let us know if you're still seeing lag issues. After scrolling > 150 lines, it's expected there might be a slight lag in the semantic colors being updated.

notAlaanor:
The performance has improved tremendously (scrolling through a 600+ line file).

Thanks for the work on this extension.

Issue #3780, opened 2019-06-15
0.24.0-insiders - Syntax colorization showing incorrect background color for highlighted lines


bobbrow:
**Type: LanguageService**

Some themes highlight the selected line with a color that doesn't match the unselected lines' background color.  In these cases the new syntax highlighting doesn't honor the theme.  This affects multiple themes.

It gets more visible when you select things:
![image](https://user-images.githubusercontent.com/12818240/59547339-40172800-8ef2-11e9-9a3a-81e9de085389.png)

To disable the new syntax highlighting, set `"C_Cpp.enhancedColorization": "Disabled"` and you will get the old behavior again.

Themes with known issues:
* Abyss
* Monokai
* Monokai Dimmed
* One Monokai
* Red
* Quiet Light
* Solarized Dark
* Solarized Light
* Tomorrow Night Blue

heimskr:
I'm having this issue too for the same set of themes. It also affects my usual theme, unfortunately. I have a lot of other themes installed too (too many, really—still trying to find the perfect theme) and I went through them to check which ones are affected. I first started experiencing the issue within the past few hours or so. Before then, everything worked fine.

Affected:
- Abyss
- Clymate Pop Neon
- Flatland Monokai
- Kimbie Dark
- Material Theme (default, Darker, Darker High Contrast, High Contrast, Ocean, Ocean High Contrast, Palenight, Palenight High Contrast)
- Monokai
- Monokai Dark Soda
- Monokai Dimmed
- Monokai++
- One Monokai
- Palenight (Italic, Operator, Theme)
- Quango Classic
- Red
- Reed Alert
- sandstorm - classic
- Solarized (Dark, Light)
- Spirited Away (my usual theme :-/)
- Tomorrow Night Blue


Not affected:
- Atom One Dark
- Clymate, Clymate Monochrome, Clymate Vintage
- Cyberpunk+
- Dark (Visual Studio)
- Dark Wiwa
- Dark+ (default dark)
- Electron
- Hackz, Hackz Darker
- Henna
- Radical
- sandstorm - darker
- Sidious
- skGRTT
- Spotify
- High Contrast

Thanks for the workaround!

Colengms:
Please up-vote the following issue: https://github.com/microsoft/vscode/issues/75642

sean-mcmanus:
Fixed with 0.24.0-insiders2. Set the "C_Cpp.updateChannel" setting to "Insiders" to auto-install it.

PR #3793, opened 2019-06-17
Avoid applying edits to cached colorization tokens until needed

Colengms:
Avoid applying edits to cached tokens unless/until they need to be reapplied.  (Similar fix applied to native side)

Avoid setting background color of decoration if same as editor background, to avoid selection highlighting issue.

Ensure editVersion is always increment per onDidChangeTextDocument, to ensure in sync with native side.  (Has corresponding change on native side)

### DIFF
--- a/Extension/src/LanguageServer/colorization.ts
+++ b/Extension/src/LanguageServer/colorization.ts
@@ -71,6 +71,7 @@ class ThemeStyle {
 export class ColorizationSettings {
     private uri: vscode.Uri;
     private pendingTask: util.BlockingTask<any>;
+    private editorBackground: string;
 
     public themeStyleCMap: ThemeStyle[] = [];
     public themeStyleCppMap: ThemeStyle[] = [];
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/colorization.ts
```diff
@@ -251,7 +252,7 @@ export class ColorizationSettings {
                             e.scope = e.scope.split(',').map((s: string) => s.trim());
                         }
                     });
-                }
+                }
```
sean-mcmanus: Space added after } ?

sean-mcmanus (approved):

Colengms:
This will address https://github.com/microsoft/vscode-cpptools/issues/3780 for themes with background colors equal to the editor background color.  Selection highlighting will still fail for background colors that differs from the editor background color.

Colengms:
This may improve https://github.com/microsoft/vscode-cpptools/issues/3781 .  There is definitely overhead involved for various reasons.  But, this change may improve performance when editing as it avoids adjusted cached token ranges while editing, deferring that work until/unless cache tokens need to be reapplied (i.e. change editor visibility).

PR #3812, opened 2019-06-21
Warn against multi-root workspaces

TheGreatRambler:
This ruins many parts of the extension.


sean-mcmanus (approved):

sean-mcmanus:
Maybe we could suggest setting C_Cpp.formatting to "Disabled" and using the "clang-format" extension for formatting, but I haven't verified yet if that actually works around the issue, and in scenarios were formatting is broken, it means our internal document has incorrect data, so other language service features like hover and go to def are likely to give incorrect results as well, even if formatting via another extension works.

PR #3824, opened 2019-06-25
Enable enhanced colorization by default

Colengms:


Extension/package.json
```diff
@@ -503,7 +503,7 @@
             "Enabled",
             "Disabled"
           ],
-          "default": "Disabled",
+          "default": "Enabled",
```
bobbrow, line 506: You should also update `cpptools.json` or else this will be immediately overridden by the A/B testing logic.

bobbrow (approved):

PR #3843, opened 2019-06-27
Decrease F5 workflow time w/ VSCode

WardenGnaw:
webpack is being called twice. Once with --mode production and then
rewritten with --mode development. This causes long delays. 

Adjusting tasks and launch so it is only called once.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1459,14 +1459,15 @@
   },
   "scripts": {
     "compile": "npm run vscode:prepublish",
-    "compileDev": "node ./tools/prepublish.js && webpack --mode development",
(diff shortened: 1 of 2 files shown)


Extension/package.json
```diff
@@ -1459,7 +1459,7 @@
   },
   "scripts": {
     "compile": "npm run vscode:prepublish",
-    "compileDev": "webpack --mode development",
+    "compileDev": "node ./tools/prepublish.js && webpack --mode development",
```
pieandcakes: Can we refactor out the `node ./tools/prepublish.js` so if that commandchanges we don't have to remember to change it in two places?

pieandcakes (approved):

PR #3847, opened 2019-06-27
Add more linting.

sean-mcmanus:
I hit a bug in my FindAllReferences branch that would have been caught by more linting. Most of the changes are whitespace, or other trivial/quick changes.


Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "ban-comma-operator": true,
+        "binary-expression-operand-order": true,
+        "callable-types": true,
+        "class-name": true,
+        "comment-format": true,
+        "completed-docs": false,
+        "no-console": false,
         "curly": true,
+        "cyclomatic-complexity": false,
+        "deprecation": false,
```
sean-mcmanus: Some of these "false" ones we should probably fix later on.

Colengms (approved):

pieandcakes:
@sean-mcmanus Why are you disabling a lot of items in tslint when you say you are "Adding more linting"? I think you need to explain more about why certain things are being turned off. Like `variable-name` and `array-type` are not trivial and should be enforced.

WardenGnaw (changes requested):
There is a lot of lint changes that are arguably should be enabled.

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
 {
     "rules": {
+        "adjacent-overload-signatures": true,
+        "align": true,
+        "array-type": false,
```
WardenGnaw: Why are we not checking types for arrays?

https://palantir.github.io/tslint/rules/array-type/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "arrow-parens": false,
+        "arrow-return-shorthand": true,
+        "await-promise": false,
```
WardenGnaw: > Warns for an awaited value that is not a Promise.

Why disable this?
https://palantir.github.io/tslint/rules/await-promise/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+            true,
+            "variable-declaration",
+            "call-signature"
+        ],
+        "typedef-whitespace": true,
+        "typeof-compare": true,
+        "unified-signatures": true,
+        "use-default-type-parameter": true,
+        "use-isnan": true,
+        "variable-name": false,
```
WardenGnaw: Delete this line. This should be handled by `ext-variable-name`

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "array-type": false,
+        "arrow-parens": false,
+        "arrow-return-shorthand": true,
+        "await-promise": false,
+        "ban-comma-operator": true,
+        "binary-expression-operand-order": true,
+        "callable-types": true,
+        "class-name": true,
+        "comment-format": true,
+        "completed-docs": false,
```
WardenGnaw: Does this need to be explicitly marked as false?

https://palantir.github.io/tslint/rules/completed-docs/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "no-console": false,
```
WardenGnaw: I'm pro `no-console` since the extension should be writing to its own output window and not the developer console for electron. 
https://palantir.github.io/tslint/rules/no-console/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
         "curly": true,
+        "cyclomatic-complexity": false,
```
WardenGnaw: Why include if false?

https://palantir.github.io/tslint/rules/cyclomatic-complexity/

Extension/tslint.json
WardenGnaw: This is scary. We should know when things are deprecated. I would not mark this as false to ignore.

sean-mcmanus:
> There is a lot of lint changes that are arguably should be enabled.

Yeah, but we were failing so I disabled them for now, but they're listed as "false" so they're easier to discover. If there are any in particular, you think we should try to fix now, I could investigate more.

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "indent": [
+            true,
+            "spaces",
+            4
+        ],
+        "interface-name": false,
+        "interface-over-type-literal": false,
+        "jsdoc-format": false,
+        "label-position": true,
+        "linebreak-style": false,
```
WardenGnaw: This should be true, and git should also make sure our *.ts files are consistent.

sean-mcmanus:
> @sean-mcmanus Why are you disabling a lot of items when you say you are "Adding more linting"

Did I disable existing linting errors? I don't see any. The intention was to explicitly add the things that are disabled so we could revisit them later.

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "match-default-export-name": true,
+        "member-access": false,
```
WardenGnaw: Not a fan of this being `false`. Makes it harder to read. 

https://palantir.github.io/tslint/rules/member-access/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "member-ordering": true,
+        "newline-before-return": false,
```
WardenGnaw: nit: I like my spaces before returns 

https://palantir.github.io/tslint/rules/newline-before-return/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "quotemark": false,
+        "radix": false,
+        "restrict-plus-operands": false,
+        "return-undefined": false,
         "semicolon": true,
+        "space-before-function-paren": false,
+        "space-within-parens": true,
+        "strict-boolean-expressions": false,
+        "strict-type-predicates": true,
+        "switch-default": false,
```
WardenGnaw: Having a default case is good to catch errors. 

https://palantir.github.io/tslint/rules/switch-default/

Extension/tslint.json
```diff
@@ -1,31 +1,184 @@
...
+        "prefer-readonly": false,
+        "prefer-switch": false,
+        "prefer-template": false,
+        "prefer-while": true,
+        "promise-function-async": false,
         "promise-must-complete": true,
+        "quotemark": false,
+        "radix": false,
+        "restrict-plus-operands": false,
+        "return-undefined": false,
```
WardenGnaw: Would we want `return undefined`?

Extension/tslint.json
sean-mcmanus: What is ext-variable-name? It doesn't appear to be supported by my TsLint extension.

WardenGnaw: Part of [vrsource-tslint-rules](https://github.com/vrsource/vrsource-tslint-rules#ext-variable-name).

Pulled in as a package and added as part of the rules directory.

Extension/tslint.json
sean-mcmanus: Do you know how to get "true" to work on Windows? It gives me "Expected linebreak to be 'LF'" on every line. Are you saying my VS Code git settings aren't configured correctly?

Extension/tslint.json
sean-mcmanus: I could use "array-type": [true, "array"] (not 100% yet).

Extension/tslint.json
WardenGnaw: Sorry, what I mean is that we should be checking out as with the OS default EOL type and checking is as just LF. E.g. `Checkout Windows (or Unix)-style, commit Unix-style`

https://help.github.com/en/articles/dealing-with-line-endings

sean-mcmanus:
> @sean-mcmanus Why are you disabling a lot of items in tslint when you say you are "Adding more linting"? I think you need to explain more about why certain things are being turned off. Like variable-name and array-type are not trivial and should be enforced.

I think I fixed those now.

sean-mcmanus:
> There is a lot of lint changes that are arguably should be enabled.

I added the TODO's to our internal docs and removed the "false" rules.

Extension/tslint.json
```diff
@@ -1,31 +1,122 @@
...
+            "check-open-brace",
+            "check-whitespace"
+        ],
+        "one-variable-per-declaration": true,
+        "prefer-method-signature": true,
+        "prefer-object-spread": true,
+        "prefer-while": true,
         "promise-must-complete": true,
         "semicolon": true,
+        "space-within-parens": true,
```
WardenGnaw, line 93: Not sure if this is doing anything:
https://palantir.github.io/tslint/rules/space-within-parens/

Extension/tslint.json
```diff
@@ -1,31 +1,122 @@
...
+            true,
+            "spaces",
+            4
+        ],
+        "label-position": true,
+        "match-default-export-name": true,
+        "member-ordering": true,
         "new-parens": true,
+        "no-arg": true,
+        "no-bitwise": true,
```
WardenGnaw, line 41: Why no bitwise?

https://palantir.github.io/tslint/rules/no-bitwise/

sean-mcmanus: Because we don't use them and probably won't need to. More chance for errors.

WardenGnaw (approved):

Extension/tslint.json
sean-mcmanus: No, this is correct. The docs are wrong. Or...it's correct with my version of TS Lint, which enforces the 0 space rule when "true" is used and doesn't accept a number like the docs say.

PR #3863, opened 2019-07-01
Update dev dependencies and ThirdPartyNotices.txt

sean-mcmanus:


Extension/package.json
```diff
@@ -1471,26 +1471,26 @@
...
     "vrsource-tslint-rules": "^6.0.0",
-    "vscode": "^1.1.34",
-    "webpack": "^4.30.0",
-    "webpack-cli": "^3.3.2"
+    "vscode": "^1.1.35",
+    "webpack": "^4.35.2",
+    "webpack-cli": "^3.3.5"
   },
   "dependencies": {
-    "@types/minimatch": "^3.0.3",
```
sean-mcmanus, line 1493: This was incorrectly being marked as a dependency, potentially causing our ThirdPartyNotices.txt to be incorrect (I'll update that later).

sean-mcmanus: I updated the ThirdPartyNotices.txt too.

pieandcakes (approved):

PR #3865, opened 2019-07-01
Add the binaryVersion to Mac crash telemetry.

sean-mcmanus:


Extension/src/LanguageServer/extension.ts
```diff
@@ -1003,6 +1003,16 @@ function handleCrashFileRead(err: NodeJS.ErrnoException, data: string): void {
...
     }
 
+    // Extract the crashing process version, because the version might not match
+    // if multiple VS Codes are running with different extension versions.
+    let binaryVersion: string = "";
+    let startVersion: number = data.indexOf("Version:");
+    if (startVersion >= 0) {
+        data = data.substr(startVersion);
+        const binaryVersionMatches: string[] = data.match(/^Version:\s*(\d|\d*\.\d*\.\d*\.\d*)/);
+        binaryVersion = binaryVersionMatches && binaryVersionMatches.length > 1 ? binaryVersionMatches[1] : "";
```
pieandcakes, line 1013: instead of "" do you want some diagnostic logging or "unknown"  ?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1028,10 +1038,10 @@ function handleCrashFileRead(err: NodeJS.ErrnoException, data: string): void {
     const process2: string = "Microsoft.VSCode.CPP.Extension.darwin\t";
     if (data.includes(process1)) {
         data = data.replace(new RegExp(process1, "g"), "");
-        data = process1 + "\n" + data;
+        data = `${process1}${binaryVersion}\n${data}`;
```
pieandcakes, line 1041: do you want to separate your binary name with your version string with a `.` so its easier to parse?

pieandcakes (approved):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Sure.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: It's already separated by '\t'.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Umm...actually, if the version is missing for some unexpected reason we'll see it in the crash logs itself, so having a separate telemetry message seems unnecessary, e.g. I would just query for crashes that didn't have the expected version like `<process>\t0.` and if we get results back that will let us know we have a problem.

sean-mcmanus:
I'm not checking this in until we've added the Mac version to our binary -- currently it's 0.

PR #3896, opened 2019-07-08
Improve scope names used by Enhanced Colorization

Colengms:
Updating scope names based on @jeff-hykin 's feedback, in  https://github.com/microsoft/vscode-cpptools/pull/3832 .


sean-mcmanus (approved):

Colengms:
A potential issue with the Type token:  VS IntelliSense uses this token for all types (except primitives).  Both a class and a typedef'ed int will get this token.

Perhaps we should use entity.name.type.class for the Type token?  That would (incorrectly?) color a typedef'ed int as a class, but wouldn't risk replacing the color of class types with entity.name.type that would have been colored with entity.name.type.class .

sean-mcmanus (approved):

jeff-hykin:
It might be more accurate if typedef was given `entity.name.type.typedef` so that user's could decide whether they want to assume it's a class or a primitive.

Currently when a type is being inherited from, its given the scope `entity.name.type.inherited` since its unknown whether it is a class or a struct.

The C++ syntax uses `entity.name.type.class` in contrast with
- `entity.name.type.enum`
- `entity.name.type.struct`
- `entity.name.type.union`

In general I believe themes would have
- `entity` or `entity.name` set for general/unknown entities (like namespaces or operators)
- and then `entity.name.type` for everything that is a typedef/class/struct/enum/union
- then if a theme wanted to color classes different they would set entity.name.type.class

Colengms:
@jeff-hykin , It looks like we should be able to break up the "Type" token into separate tokens for class, enum, struct, and union.

Would it be sufficient to refer to templates of unions or structs as "Class Template", or do you think those require distinct token types?  (I think a single token type for template types should be ok).

Are there any other semantic tokens we may need to split into multiple?

jeff-hykin:
Breaking it up would be great if thats possible. 

Adding 
`entity.name.type.class.templated`
`entity.name.type.struct.templated`
`entity.name.type.union.templated`
`entity.name.type.enum.templated`
would be the most accurate way, but it would probably also probably make it more annoying on theme makers. I think defaulting to `entity.name.type.class.templated` isn't a great solution and sadly this is part of the challenge of the TextMate scopes :/

One technique that can be used is to turn `entity.name.type.class.templated`
into `variation.templated entity.name.type.class`

Inside a theme this would allow 
`variation.templated entity.name.type` to color all of the templated types (class/struct/enum/union)
And `variation.templated entity.name.type.class` to only color the templated classes

This might work well for functions too since they could be tagged as
`variation.templated entity.name.function`

This could potentially work well for reference types and generic types as well.
`variation.reference entity.name.type`
`variation.generic entity.name.type`

#### The Catch
The only catch is if multiple "variations" are used simultaneously they have to be in the original order. I've talked with the other C++ contributor about potential ways of sorting them so that they're always in a particular order, but we haven't come up with anything conclusively.

For example, lets say there is class that is tagged as
`variation.abstract variation.templated entity.name.type.class`
✅`variation.abstract entity.name.type`
✅`variation.templated entity.name.type`
✅`variation.abstract variation.templated entity.name.type`
❌ `variation.templated variation.abstract entity.name.type`

Colengms:
It looks like breaking up the Type semantic token into multiple tokens (and separate templated tokens) is not going to happen quickly.  I'd like to go ahead with the current changes to use entity.name.type.class to color all types (so classes get colored correctly) and follow up when/if we can get them separated.

jeff-hykin:
Okay, sounds good to me. Hopefully in the future there will be an alternative to textmate scopes that don't require these kinds of tradeoffs.

jeff-hykin:
Also, I want to say thank you for even asking for input at all. I really appreciate you guys working with the rest of the community. It's part of what keeps me pushing late night commits for open source projects.

Colengms:
Likewise, thank you for your help.  :)  We really appreciate it.

Colengms:
Hi @jeff-hykin .  One question.  Currently I have entity.name.namespace for namespace tokens.  Should this instead be entity.name.type.namespace ?  Using the Inspect TM Scopes command, I see that entity.name.type.namespace is used when a namespace is declared.

jeff-hykin:
The old atom syntax used `entity.name.type.namespace` but the scope was updated to `entity.name.namespace` back in May. There was an issue about it [here](https://github.com/jeff-hykin/cpp-textmate-grammar/issues/101#issuecomment-483971100) that covers the details

Colengms:
I'm seeing this behavior in VS Code 1.36.1.  Though, with Better C++ Syntax enabled, it is entity.name.namespace.  Perhaps I misunderstood, but I thought your grammar was being incorporated directly into VS Code.  Is that not the case?

jeff-hykin:
Actually that was my understanding as well. We'll have to ask @alexr00

alexr00:
in 1.36.1 we have an older version of the grammar, which does still have `entity.name.type.namespace`. In insiders we are much more up-to-date and have a version that has `entity.name.namespace`. If all goes well, the 1.37.0 release will have this update.

Issue #3448, opened 2019-04-11
Add C++20 to the supported versions


sean-mcmanus:
Our compiler has some usable C++20 features implemented we should enable (i.e. to match what VS 2019 supports).

PR #3905, opened 2019-07-09
Add c++20 to enums

Jasdriel:
This are the files in the Typescript part that I found to add the enums, is ok with this changes or there are other parts to update?

For issue #3448.


Extension/c_cpp_properties.schema.json
```diff
@@ -37,6 +37,8 @@
                             "c++11",
                             "c++14",
                             "c++17",
+                            "c++20",
+                            "c++Latest",
```
sean-mcmanus: We shouldn't use c++20 latest -- it should just be c++20.

sean-mcmanus (changes requested):
see my other comment

michelleangela:
Also add "c++20" as another `<option>` to Config UI in [settings.html](https://github.com/microsoft/vscode-cpptools/blob/53107bc57ca9c0e772a8d73746e1caf4612568db/Extension/ui/settings.html#L562)

And to [package.json](https://github.com/microsoft/vscode-cpptools/blob/master/Extension/package.json#L421)

sean-mcmanus:
The msvc version needs to be changed in the .json files: https://github.com/microsoft/vscode-cpptools/blob/master/Extension/bin/msvc.32.intel.json and the other 2 .msvc files.

Extension/c_cpp_properties.schema.json
Jasdriel: ok, fixed

sean-mcmanus:
Okay, this seems good, but don't check it in until the C++ side is checked in.

sean-mcmanus (approved):

michelleangela:
Also need to add c++20 to the API https://github.com/microsoft/vscode-cpptools-api. You'll have to create separate pull request on that repo.

Issue #3878, opened 2019-07-04
Intellisense stopped working after update


krzikalla:
**Type: LanguageService**

**Describe the bug**
- OS and Version: SUSE Linux Enterprise Desktop 12 SP2
- VS Code Version: 1.36
- C/C++ Extension Version: 0.24.0
- A clear and concise description of what the bug is.
After updating VS Code from 1.26 and the C/C++ Extension some symbols can't be found anymore.
I am unable to pinpoint the nature of symbols. Apparently it only affects symbols related to templates. Some templates are found though.

bobbrow:
Can you provide a little more information about the problem and maybe a small sample where it doesn't work?  We did integrate the latest set of updates for IntelliSense from Visual Studio, so it's possible that something regressed, but this isn't enough information for us to work with.

krzikalla:
I couldn't create a small sample. The affected project is large and contains a lot of templates.  
It appears to happen always inside class definitions and often also in namespace scope. Toggling Intellisense options seems of no help. Neither switching to Tag Parser nor taking clang-x64 instead of msvc-x64 makes a difference. No problems are shown in the "problems" tab.
(there is an error on startup: "/lib64/libdbus-1.so.3: no version information available" but I guess, it is unrelated).

sean-mcmanus:
Can you give more info about what "symbols can't be found" means? Does the symbol not appear in the Outline view after making an edit to the file? (we have a known issue where symbols won't appear in the Outline view before an edit is made). Is hover or go to definition failing to find the symbol? Is the symbol not found even in the file that defines it? Does it give an error squiggle about something being undefined?

krzikalla:
In the outline view appears nothing - even after editing. Hover and "go to definition" finds very few symbols - and apparently only after editing. The kind of found symbols does not depend on the file location. There are no error squiggles anywhere - not even at the includes (which can't be opened too).

If I have compiler errors in the problem tab, then I can't open them and get the pop-up (Error: File not found (PATH)). PATH is correct though. I am not sure if this is related, since the "cannot-navigate-to-compiler-errors-bug" is actually an old one which I actually hoped to get fixed meanwhile (that was actually the sole reason for me updating to 1.36).

sean-mcmanus:
Do you know what version the C/C++ extension you were using previously? Are you able to use Install Another Version... in the Extension view to use 0.23.1 or an earlier version that works?

[...]

PR #3916, opened 2019-07-11
Use autoDetect problemMatcher.

sean-mcmanus:


michelleangela:
Is there a work item for this or reason for using autoDetect now?

michelleangela (approved):

sean-mcmanus:
> Is there a work item for this or reason for using autoDetect now?

They just added it in the most recent release of VS Code -- and the default of "relative" is bad because it fails in the default case where absolute paths are used so users have to modify the problem matcher (a user recently mentioned hitting this problem in another bug a few days ago).

bobbrow:
mentioned in this issue: #3878

HackerBaloo:
I still have this problem in 1.39.2 with ms-vscode.cpptools 0.26.1
I experience this problem when I compile with clang on Linux.
And I have tried to create a custom problem matcher like this:
"problemMatcher": {
        "owner": "cpp",
        "fileLocation": ["autoDetect", "${workspaceFolder}"],
        "pattern": {
          "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
          "file": 1,
          "line": 2,
          "column": 3,
          "severity": 4,
          "message": 5
        }
      }

Issue #3864, opened 2019-07-01
doesn't pick up changes in compile_commands.json at runtime.


mdzahidh:
Extension Version: 0.24.0-insiders3
Vscode version: 1.36.0-insider
OS: Linux, Fedora

We have a script that changes the compile_commands.json while vscode/cpptool is running, but unfortunately the cpptools doesn't seem to pick up the changes until we restart vscode (rescan workspace doesn't seem to handle this case). Are there ways to trigger this ? It would be really nice if cpptools picked up the changes automatically, especially when you are on a remote session (restarting vscode means reconnecting to the server and etc).

sean-mcmanus:
Yeah, we have code that handles the change to the compile commands and updates the IntelliSense args, but somehow it's not working so we need to debug it further.

Lennon925:
> Extension Version: 0.24.0-insiders3
> Vscode version: 1.36.0-insider
> OS: Linux, Fedora
> 
> We have a script that changes the compile_commands.json while vscode/cpptool is running, but unfortunately the cpptools doesn't seem to pick up the changes until we restart vscode (rescan workspace doesn't seem to handle this case). Are there ways to trigger this ? It would be really nice if cpptools picked up the changes automatically, especially when you are on a remote session (restarting vscode means reconnecting to the server and etc).

Hi, I am intersted in your mechanism how to change to compile_command.json when vscode/cpptools is running. Could you share it with me? Thanks a lot.

sean-mcmanus:
@Lennon925 Who are you asking? Usually, your build system will regenerate the compile_commands.json (e.g. CMake). This issue was fixed already via https://github.com/microsoft/vscode-cpptools/pull/3938 and a language server change.

Lennon925:
> @Lennon925 Who are you asking? Usually, your build system will regenerate the compile_commands.json (e.g. CMake). This issue was fixed already via #3938 and a language server change.

OK，Thanks for your reply!

sean-mcmanus:
Should be fixed with 0.25.0-insiders: https://github.com/microsoft/vscode-cpptools/releases/tag/0.25.0-insiders .

PR #3938, opened 2019-07-17
file watcher compiler commands

Jasdriel:
Changes to resolve issue https://github.com/microsoft/vscode-cpptools/issues/3864

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -586,8 +586,8 @@ export class CppProperties {
         this.compileCommandFileWatchers = []; //reset it
         let filePaths: Set<string> = new Set<string>();
         this.configurationJson.configurations.forEach(c => {
-            if (c.compileCommands !== undefined && fs.existsSync(this.resolvePath(c.compileCommands, os.platform()==="win32"))) {
-                filePaths.add(this.resolvePath(c.compileCommands, os.platform()==="win32"));
+            if (c.compileCommands !== undefined && fs.existsSync(this.resolvePath(c.compileCommands, os.platform() === "win32"))) {
+                filePaths.add(this.resolvePath(c.compileCommands, os.platform() === "win32"));
             }
         });
         try {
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -586,8 +586,8 @@ export class CppProperties {
         this.compileCommandFileWatchers = []; //reset it
         let filePaths: Set<string> = new Set<string>();
         this.configurationJson.configurations.forEach(c => {
-            if (c.compileCommands !== undefined && fs.existsSync(c.compileCommands)) {
-                filePaths.add(c.compileCommands);
+            if (c.compileCommands !== undefined && fs.existsSync(this.resolvePath(c.compileCommands, os.platform() === "win32"))) {
+                filePaths.add(this.resolvePath(c.compileCommands, os.platform() === "win32"));
```
michelleangela: Save the result of `this.resolvePath(c.compileCommands, os.platform() === "win32")` and then use in fs.existsSync(...) and filePaths.add(...)? This way, it does not have to go resolvePath twice.

michelleangela (changes requested):

Extension/src/LanguageServer/configurations.ts
Jasdriel: you're right, changed thanks

Extension/src/LanguageServer/configurations.ts
```diff
@@ -586,8 +586,9 @@ export class CppProperties {
         this.compileCommandFileWatchers = []; //reset it
         let filePaths: Set<string> = new Set<string>();
         this.configurationJson.configurations.forEach(c => {
-            if (c.compileCommands !== undefined && fs.existsSync(c.compileCommands)) {
-                filePaths.add(c.compileCommands);
+            let fileSystemCompileCommandsPath = this.resolvePath(c.compileCommands, os.platform() === "win32");
+            if (c.compileCommands !== undefined && fs.existsSync(fileSystemCompileCommandsPath)) {
```
michelleangela: Check `c.compileCommands !== undefined` first before resolving the path? Or you can check the return value of resolvePath depending on the desired output value that's needed.

michelleangela:
The build is failing with lint errors. You may want to install TSLint extension.

Extension/src/LanguageServer/configurations.ts
Jasdriel: I may have to do the undefined validation first because, resolving the path returns an empty string when path is undefined or ${default}

michelleangela (approved):

PR #3963, opened 2019-07-23
Enable string localization

Colengms:
Added use of vscode-nls and vscode-nls-dev. vscode-nls-dev involves a source rewrite step to convert key lookups into index lookup (presumably for efficiency).  Since we use webpack, this step is done by a webpack loader.

This includes scripts to generate a file to hand off to the loc team, and to import what we get back from them.  Loc hand-off will be a somewhat manual process, in which we export a file, send it to the loc team, and wait for a file back.  (The SLA is weekly. Monday out, Friday in.)  Import will result in an i18n directory structure.  My preference is to check the generated i18n directory structure into the repo.  Alternatively, we could check in the xlf file we get from the loc team and generate the i18n directory structure when generating an official build (my second choice).  Or, we could check in the nls files generated from the i18n directory.  Since some of those files need to be built to the dist directory, I'd prefer we avoid the later, as we currently don't check in anything in dist, and it's useful to delete it to build cleanly.  Also, the i18n files are potentially serviceable if necessary.

I also went ahead and integrated all obviously user-facing strings to use this approach.  Those instances look something like:

const askLater: string = localize("ask.me.later.button", "Ask Me Later");

The first string is a unique key used to look up the string in a string table.  The second is the English string.  The loc export script will scan the code for all calls to localize, to include in the xlf file to hand off to loc.

We will need to use a different approach to handle localization on the native side, of course.


Extension/gulpfile.js
```diff
@@ -11,6 +11,33 @@ const tslint = require('gulp-tslint');
 const mocha = require('gulp-mocha');
 const fs = require('fs');
 const optionsSchemaGenerator = require('./out/tools/GenerateOptionsSchema');
+const nls = require('vscode-nls-dev');
+const path = require("path");
```
sean-mcmanus: inconsistent '' and "".

Extension/src/githubAPI.ts
```diff
@@ -274,13 +276,13 @@ async function getReleaseJson(): Promise<Build[]> {
...
-        throw new Error('Failed to parse release JSON');
+        throw new Error(localize("failed.parse.release", 'Failed to parse release JSON'));
     }
 
     // Type check
     if (isArrayOfBuilds(releaseJson)) {
         return releaseJson;
     } else {
-        throw new Error('Release JSON is not of type Build[]');
+        throw new Error(localize('failed.release.type', 'Release JSON is not of type {0}', "Build[]"));
```
sean-mcmanus: These errors get logged in telemetry -- I think we need to translate them later on after they haven't been logged to telemetry.

Extension/src/common.ts
```diff
@@ -218,11 +221,11 @@ export function setProgress(progress: number): void {
         let telemetryProperties: { [key: string]: string } = {};
         let progressName: string;
         switch (progress) {
-            case 0: progressName = "install started"; break;
-            case progressInstallSuccess: progressName = "install succeeded"; break;
-            case progressExecutableStarted: progressName = "executable started"; break;
-            case progressExecutableSuccess: progressName = "executable succeeded"; break;
-            case progressParseRootSuccess: progressName = "parse root succeeded"; break;
+            case 0: progressName = localize("install.started", "install started"); break;
```
bobbrow: These strings are for telemetry only. They should not be localized.  Perhaps we should set up a convention for marking strings that should not be localized.

Extension/src/common.ts
```diff
@@ -254,7 +257,7 @@ export async function showReleaseNotes(): Promise<void> {
     if (releaseNotesPanel) {
         releaseNotesPanel.reveal();
     } else {
-        releaseNotesPanel = vscode.window.createWebviewPanel('releaseNotes', "C/C++ Extension Release Notes", vscode.ViewColumn.One);
+        releaseNotesPanel = vscode.window.createWebviewPanel('releaseNotes', localize("extension.release.notes", "C/C++ Extension Release Notes"), vscode.ViewColumn.One);
```
bobbrow: This is interesting...  ReleaseNotes.html would probably need to be localized (if we decide to keep it).

Extension/src/githubAPI.ts
bobbrow: I don't believe any of these errors in this file are user-facing.  They're just for telemetry.

Extension/src/packageVersion.ts
```diff
@@ -14,7 +18,7 @@ export class PackageVersion {
     constructor(version: string) {
         const tokens: string[] = version.split(new RegExp('[-\\.]', 'g')); // Match against dots and dashes
         if (tokens.length < 3) {
-            throw new Error('Failed to parse version string: ' + version);
+            throw new Error(localize("version.string.parse.failed", "Failed to parse version string: {0}", version));
```
bobbrow: I believe that these errors are only for telemetry (in connection with `githubapi.ts` for the Insiders updateChannel)

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,7 +501,7 @@ function installVsix(vsixLocation: string): Thenable<void> {
             }
         }(platformInfo);
         if (!vsCodeScriptPath) {
-            return Promise.reject(new Error('Failed to find VS Code script'));
+            return Promise.reject(new Error(localize('failed.to.find.scipt', 'Failed to find VS Code script')));
```
bobbrow: Most of the strings related to processing the Insiders update are not user-facing.  They are for telemetry and should not be localized.

bobbrow (changes requested):
Some general questions:
* Are there any strings in the Debugger code that we need to add to the loc files?
* Should we pre-loc all strings in each file into `const` variables so that we don't have to repeatedly look them up? (maybe only for code paths that are exercised multiple times)
* Were you going to include the xlf or i18n files as part of this commit?  I think it would be good for community members to be able to submit pull requests against translations if they think they could be improved.

Extension/src/common.ts
Colengms: > Are there any strings in the Debugger code that we need to add to the loc files?

(Adding these in next push)

> Should we pre-loc all strings in each file into const variables so that we don't have to repeatedly look them up? (maybe only for code paths that are exercised multiple times)

If the concern is about efficiency, I don't believe that would be an issue.  vscode-nls-dev does a source rewrite step which converts localize calls to indexed table lookups, so they should be fast.

> Were you going to include the xlf or i18n files as part of this commit?  I think it would be good for community members to be able to submit pull requests against translations if they think they could be improved.

I plan to include the i18n files.  Once these changes are close to done, I will export a file for loc, and include what I get back from them in this PR.  I may wait a while to submit this PR until we've dealt with the native side, as we've discuss potentially routing them via the TS side.

Extension/bin/edge_strings.cs.json
```diff
@@ -0,0 +1 @@
...ou desetinnou čárku je změna znaménka.","Celočíselná hodnota se nevejde do požadovaného typu s pevnou desetinnou čárkou.","Výsledek operace s pevnou desetinnou čárkou je mimo rozsah.","několik pojmenovaných adresních prostorů","Proměnná s automatickým trváním úložiště nemůže být uložená v pojmenovaném adresním prostoru. ","Typ se nemůže kvalifikovat pomocí pojmenovaného adresního prostoru.","Typ funkce se nemůže kvalifikovat pomocí pojmenovaného adresního prostoru.","Typ pole se nemůže kvalifikovat pomocí pojmenovaného adresního prostoru.","Hodnota s pevnou desetinnou čárkou se nevejde do požadovaného typu s plovoucí desetinnou čárkou.","Hodnota s pevnou desetinnou čárkou se nevejde do požadovaného celočíselného typu.","Hodnota se nevejde do požadovaného typu s pevnou desetinnou čárkou.","Možnost named_registers se dá použít jenom při kompilaci C.","Třída úložiště pojmenovaného registru tady není povolená.","Došlo ke změně deklarace %nd s nekompatibilní třídou úložiště pojmenovaného registru.","Třída úložiště pojmenovaného registru se nedá specifikovat pro proměnnou alias.","Specifikátor úložiště pojmenovaného registru se už používá.","Možnost embedded_c se nedá kombinovat s možnos
```
sean-mcmanus: Can these JSON files be formatted with spaces/newlines?

Colengms: I'd rather not.  These are generated, and not intended to be edited by hand.  Also, they are rather large, and formatting would increase their size.

Extension/ui/settings.html
```diff
@@ -648,37 +627,37 @@
         </div>
 
         <div class="section">
-            <div class="section-title">Browse: path</div>
+            <div class="section-title" data-loc-id="browse.path">Browse: path</div>
             <div class="section-text">
-                A list of paths for the Tag Parser to search for headers included by your source files. If omitted, <code>includePath</code> will be used as the <code>path</code>. 
-                Searching on these paths is recursive by default. Specify <code>*</code> to indicate non-recursive search. 
-                For example: <code>/usr/include</code> will search through all subdirectories while <code>/usr/include/*</code> will not.
+                <span data-loc-id="browse.path.description1">A list of paths for the Tag Parser to search for headers included by your source files.</span>
```
michelleangela: >browse.path.description1 [](start = 35, length = 24)

What is the process when the descriptions need to be modified? Is it a new description# per sentence?

And will the descriptions get translated per sentence or as a paragraph and still retain the context as a whole? #Resolved

Colengms: I broke these into separate sentences as a contention that loc should be per sentence.  But, you're right, these should probably be locallized as an entire paragraph so the whole context gets translated, if necessary.  (Though, the contents of the code blocks will not be localized, and will be {0} and {1}, etc., in the string.)   #Resolved

Colengms: As far as process; Any node just needs a "data-loc-id", or any attribute needs a "data-loc-id-<attribute-name>".  If an existing string is modified (with the same id) it may get the original localized strings until the next loc pass, whereas using a new ID will result in the English string until the next loc pass. #Resolved

michelleangela:
Instructions to add or modify loc strings can be added to https://github.com/microsoft/vscode-cpptools/blob/master/CONTRIBUTING.md

Extension/.gitignore
```diff
@@ -19,3 +19,17 @@ install.lock
...
 # ignore vscode test
 .vscode-test/
+
+# ignore generated localization file
+**/nls.*.json
+**/*.nls.json
+**/*.nls.*.json
+
+# ignore exported localization xlf directory
+localization-export
```
bobbrow: Some of these entries will also need to be added to .vscodeignore so that they don't end up in the final VSIX.

Extension/package.json
```diff
@@ -1433,16 +1433,21 @@
     }
   },
   "scripts": {
-    "compile": "npm run vscode:prepublish",
-    "compileDev": "npm run prepublishjs && webpack --mode development",
+    "compile": "npm run prepublishjs && npm run generateOptionsSchema && npm run generate-native-strings && npm run localization-generate && webpack --mode production --vscode-nls",
+    "compileDev": "npm run prepublishjs && npm run generateOptionsSchema && npm run generate-native-strings && webpack --mode development",
     "generateOptionsSchema": "gulp generateOptionsSchema",
+    "generate-native-strings": "gulp generate-native-strings",
+    "localization-export.temp": "npm run generate-native-strings && gulp localization-export",
```
bobbrow: Is this script still needed?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+        return path;
+    }
+    if (!path.startsWith(prefix)) {
+        return path;
+    }
+    var ch = prefix.charAt(prefix.length - 1);
+    if (ch === '/' || ch === '\\') {
+        return path.substr(prefix.length);
+    }
+    else {
```
bobbrow: nit: `else` on line above.

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+        return path.substr(prefix.length + 1);
```
bobbrow, line 143: Should you guard against mid-path matches, in the event that path = `/some/path2/src` and prefix = `/some/path` so you don't end up with `2/src` as a result (which is invalid)?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+    // Consoldate them into nls.metadata.json, which the xlf is built from.
+    .pipe(nls.bundleMetaDataFiles('ms-vscode.cpptools', 'dist'))
+
+    // filter down to just the resulting metadata files
+    .pipe(filter(['**/nls.metadata.header.json', '**/nls.metadata.json']))
+
+    // Add package.nls.json, used to localized package.json
+    .pipe(gulp.src(["package.nls.json"]))
+
+    // package.nsl.json and nls.metadata.json are used to generate the xlf file
```
bobbrow: typo: `nsl` -> `nls`

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+gulp.task("localization-import", (done) => {
+    var options = minimist(process.argv.slice(2), {
+        string: "location",
+        default: {
+            location: "../vscode-cpptools-localization-import"
+        }
+    });
+    es.merge(languages.map((language) => {
+        let id = language.transifexId || language.id;
+        // This path needs to be revisited once we iron out the process for receiving this xlf and running this scripts.
```
bobbrow: grammar?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+    }));
+});
+
+
+// ****************************
+// Command: localization-generate
+// The following is used to import an i18n directory structure and generate files used at runtime.
+// ****************************
+
+// TODO: Make sure json files for HTML files are filtered out, as we've already create fully localized HTML files for those.
```
bobbrow: Is this addressed?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+// Outputs to root path, as these nls files need to be along side package.json
+const generatedAdditionalLocFiles = () => {
+    return gulp.src(['package.nls.json'])
+        .pipe(nls.createAdditionalLanguageFiles(languages, 'i18n'))
+        .pipe(gulp.dest('.'));
+};
+
+// // Generates ./dist/<src_path>/<filename>.nls.<language_id>.json, from files in ./i18n/**/<src_path>/<filename>.i18n.json
+// // Localized strings are read from these files at runtime.
+// const generatedSrcLocFiles = () => {
```
bobbrow: Is this function still needed?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+        languages.map((language) => {
+            let stringTable = {};
+            // Try to open i18n file for this HTML file
+            let relativePath = removePathPrefix(file.path, file.cwd);
+            let locFile = path.join("./i18n", language.folderName, relativePath + ".i18n.json");
+            if (fs.existsSync(locFile)) {
+                stringTable = JSON.parse(fs.readFileSync(locFile).toString());
+            }
+            // Entire file is scanned and modified, then serialized for that language.
+            // Even if no localization is available, we still write new files to dist/html/...
```
bobbrow: What happens in this case? Does it write out English strings?

Extension/gulpfile.js
```diff
@@ -70,6 +105,388 @@ gulp.task('pr-check', (done) => {
...
+// Generate localized versions of HTML files
+// Check for cooresponding localized json file in i18n
+// Generate new version of the HTML file in dist/<language_id>/<path>
+const generateHtmlLoc = () => {
+    return gulp.src(htmlFilesPatterns)
+        .pipe(generateLocalizedHtmlFile())
+        .pipe(gulp.dest('dist'));
+};
+
+//gulp.task('localization-generate', gulp.series(generatedAdditionalLocFiles, generatedSrcLocFiles, generateHtmlLoc));
```
bobbrow: Is this still needed?

Extension/import_edge_strings.js
```diff
@@ -0,0 +1,84 @@
...
+let localizeRepoPath = process.argv[2];
+let cpptoolsRepoPath = process.argv[3];
+
+if (!localizeRepoPath || !cpptoolsRepoPath) {
+    console.error(`ERROR: Usage: ${path.parse(process.argv[0]).base} ${path.parse(process.argv[1]).base} <Localize repo path> <vscode-cpptools repo path>`);
+    return;
+}
+
+console.log("Importing EDGE strings from Localize repo: " + localizeRepoPath);
+console.log("Writing to cpptols repo: " + cpptoolsRepoPath);
```
bobbrow: typo: `cpptols` -> `cpptools`

Extension/import_edge_strings.js
```diff
@@ -0,0 +1,84 @@
...
+    { id: "hu", folderName: "hun" }, // VS Code supports Hungarian, but VS is not currently localized for those languages.
+    { id: "pt-br", folderName: "ptb", transifexId: "pt-BR" },
+    { id: "tr", folderName: "trk" },
+    { id: "cs", folderName: "csy" },
+    { id: "pl", folderName: "plk" }
+];
+
+var locFolderNames = fs.readdirSync(localizeRepoPath).filter(f => fs.lstatSync(path.join(localizeRepoPath, f)).isDirectory());
+locFolderNames.forEach((locFolderName) => {
+    let lclPath = path.join(localizeRepoPath, locFolderName, "vc/vc/cpfeui.dll.lcl");
```
bobbrow, line 43: Not sure how I feel about exposing internal path details in public.  This could also change whenever.  Perhaps keep the file name but require specifying the full path to the .lcl file as argv[2].

Extension/import_edge_strings.js
```diff
@@ -0,0 +1,84 @@
...
+                        if (highestValue < itemId) {
+                            highestValue = itemId;
+                        }
+                    }
+                });
+            }
+        });
+    });
+
+    fs.writeFileSync(outputPath, JSON.stringify(resultArray), 'utf8');
```
bobbrow: Sometimes it's nice to have human-readable formatting for these.  You could try `JSON.stringify(resultArray, null, 2)` to see if it will put one entry on each line.

Extension/src/packageManager.ts
```diff
@@ -303,18 +315,18 @@ export class PackageManager {
     }
 
     private InstallPackage(pkg: IPackage, progressCount: string, progress: vscode.Progress<{message?: string; increment?: number}>): Promise<void> {
-        this.AppendLineChannel(`Installing package '${pkg.description}'`);
+        this.AppendLineChannel(localize("installing.package", "Installing package '{0}'", pkg.description));
 
         progress.report({message: `Installing ${progressCount}: ${pkg.description}`});
```
bobbrow, line 320: Does this one need localization?

Extension/src/packageManager.ts
```diff
@@ -303,18 +315,18 @@ export class PackageManager {
...
 
         return new Promise<void>((resolve, reject) => {
             if (!pkg.tmpFile || pkg.tmpFile.fd === 0) {
-                return reject(new PackageManagerError('Downloaded file unavailable', 'InstallPackage', pkg));
+                return reject(new PackageManagerError('Downloaded file unavailable', localize("downloaded.unavailable", 'Downloaded file unavailable'), 'InstallPackage', pkg));
```
bobbrow, line 324: I don't think any PackageManagerErrors need localization.  I only see this object used for telemetry.

Extension/src/LanguageServer/client.ts
```diff
@@ -456,6 +500,57 @@ class DefaultClient implements Client {
...
+                                    };
+                                    resultCommands.push(vscodeCommand);
+                                });
+
+                                return resultCommands;
+                            });
+                        }
+                    }
+
+                    this.disposables.push(vscode.languages.registerCodeActionsProvider(documentSelector, new CodeActionProvider(this), null));
```
bobbrow, line 539: Does this cause any interference with the vcpkg CodeActionProvider?

Extension/src/LanguageServer/client.ts
```diff
@@ -456,6 +500,57 @@ class DefaultClient implements Client {
...
+                                r = Range.create(Position.create(range.start.line, range.start.character), Position.create(range.end.line, range.end.character));
+                            }
+
+                            let params: GetCodeActionsRequestParams = {
+                                range: r,
+                                uri: document.uri.toString()
+                            };
+
+                            return this.client.languageClient.sendRequest(GetCodeActionsRequest, params)
+                            .then((commands) => {
```
bobbrow: nit: if `.then` is on a new line, it should be indented.

bobbrow (approved):
I didn't do a full review of string replacements but I approve of the general flow.

Extension/.gitignore
Colengms: The nls files need to be in the VSIX, but are generated so should not be checked in.  I've since modes the localization-export directory to outside of the Extension directory, so moving this to parent .gitignore.

Extension/import_edge_strings.js
Colengms: Sean suggested making the generated files more easily readable as well.  My only concern is that the edge string tables are not intended to be user serviceable and are relatively large (for text files).

Extension/src/packageManager.ts
Colengms: Doesn't it?  It's being output to the main output channel.

Extension/src/packageManager.ts
Colengms: These appear to be logged to the output channel.  i.e. handleError() in main.ts, where errorMessage is passed to outputChannelLogger.appendLine();

Extension/bin/edge_strings.cs.json
Colengms: Bob suggested the same.  I made them more readable.

Extension/package.json
Colengms: Removed

Extension/src/LanguageServer/client.ts
Colengms: I've confirmed that having 2 CodeActionProviders works properly.

Extension/tools/OptionsSchema.json
```diff
@@ -7,80 +7,80 @@
   "definitions": {
     "PipeTransport": {
       "type": "object",
-      "description": "When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the MI-enabled debugger backend executable (such as gdb).",
+      "description": "%c_cpp.debuggers.pipeTransport.description%",
```
WardenGnaw, line 10: Can we not have the ```%c_cpp.debuggers.*``` in this file? I would prefer that we have the original English text in the file. Have them write to the nls.json and then replace with the ```%..%``` when written to the package.json.

This schema file makes it easier for us to read all the flags we have for the debugger without the duplicates.

Colengms: To do what you're suggesting would require some mechanism to map English strings (or nodes) in OptionsSchema.json to the appropriate %% variables.  That seems much more complicated (and error prone) than using the %% vars and maintaining these strings in package.nls.json instead of OptionsSchema.json.

pieandcakes: Since this file isn't shipped but is used to generate the package.json anyways, could we have it generate a .nls.json file and then after the loc has happened, build the package.json?

Colengms: I'm open to a reasonable justification for doing this differently, but so far the complexity of what you suggesting does not seem a fair trade off for the sake of being able to keep English strings in this file.

WardenGnaw: I'll make a separate PR after this is checked in to handle how we generate OptionsSchema.json and replace it with the resource strings.

Extension/bin/messages/cs/messages.json
```diff
@@ -0,0 +1,3194 @@
+[
+  "Neznámá chyba",
+  "poslední řádek souboru končí bez nového řádku",
+  "poslední řádek souboru končí zpětným lomítkem",
+  "Soubor #include %sq obsahuje sám sebe.",
+  "nedostatek paměti",
+  null,
```
bobbrow, line 7: Why is it `null` here?  What happens in this case?

Extension/bin/messages/pt-br/messages.json
```diff
@@ -0,0 +1,3194 @@
...
+  "faltando aspas de fechamento",
+  "comentários aninhados não são permitidos",
+  "'#' não esperado aqui",
+  "diretivas de pré-processamento não reconhecidas",
+  "análise reinicia aqui, depois do erro de sintaxe anterior",
+  "esperado nome de arquivo",
+  "texto extra depois do esperado fim da diretiva de pré-processamento",
+  null,
+  null,
+  "esperado um ']5D;'",
```
bobbrow: This looks incorrect.

CONTRIBUTING.md
```diff
@@ -18,3 +18,18 @@
...
+import * as nls from 'vscode-nls';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+```
+* For each user-facing string, wrap the string in a call to localize:
+```typescript
+const readmeMessage: string = localize("refer.read.me", "Please refer to {0} for troubleshooting information. Issues can be created at {1}", readmePath, "https://github.com/Microsoft/vscode-cpptools/issues");
+```
+* The first parameter to localize should be a unique key for that string.  The second parameter is the string to localize.  Both of these parameters must be string literals.  Tokens such as {0} and {1} are supported in the localizable string, with replacement values passed as additional parameters to localize().
```
bobbrow: Please add a note about where the 'refer.read.me' key should be defined.

bobbrow (commented):
Why are the strings for the TypeScript files split across multiple `json` files?  I thought they were going to go into one uber `json` file.

Extension/bin/messages/cs/messages.json
bobbrow: nevermind, I think those messages have been removed but the ids stayed the same.

CONTRIBUTING.md
Colengms: I'm not sure I follow..  It does not need to be defined anywhere else.  As it says, "The first parameter to localize should be a unique key for that string.".  I will try to make that more clear.

Extension/bin/messages/pt-br/messages.json
Colengms: Weird.  If this is standard notation (for escaping an end bracket instead of interpreting it as ending a CDATA section), then the XML parser should have converted it.  I can convert it myself.  I don't see any other cases of escaping that aren't handled...

sean-mcmanus (approved):

Issue #3888, opened 2019-07-07
Error after installing Microsoft C/C++ Extension


ArtsJrn:
As title states, the extension from **Microsoft** *C/C++* in the visual code's recommended extensions list keeps returning an error every time I try to install it.

The error goes like this: `Failed to parse "\.vscode\c_cpp_properties.json": ENOENT: no such file or directory, open '\.vscode\c_cpp_properties.json'`

Now I also keep getting these error messages from the Problem section of the lower panel (outputs, terminal...etc): 

- `#include errors detected. Please update your includePath. Squiggles are disabled for this translation unit (C:\Users\Jordan\Documents\sample.cpp).`

- `cannot open source file "iostream"`

I've tried reinstalling the extension & visual code multiple times...restarting my computer multiple times, but nothing seems to work. It used to work before I did a clean reset for my windows 10 computer, but now it's just acting up. Please fix this error so I can continue to create applications programmed in C++... :) 

*P.S, these are additional images in case anyone wants a look at them themselves:*

![image](https://user-images.githubusercontent.com/31326199/60763743-e1ece900-a048-11e9-8ea3-d07b67f1b1db.png)

![image](https://user-images.githubusercontent.com/31326199/60763746-f92bd680-a048-11e9-9eb1-fd42ef326145.png)
![image](https://user-images.githubusercontent.com/31326199/60763751-1b255900-a049-11e9-935c-26650342b1de.png)
![image](https://user-images.githubusercontent.com/31326199/60763756-31331980-a049-11e9-893e-dca85f283738.png)

sean-mcmanus:
We need more repro info. Are you opening a folder? Does the folder contain a ".vscode" folder and a "c_cpp_properties.json" file? Can you do an Edit Configuration command to create that file?

TheThinker01:
After installing Microsoft's C/C++ extension whenever I try opening a C file (only a single file, from a folder containing many other non related files) the error 
Failed to parse "/.vscode/c_cpp_properties.json": ENOENT: no such file or directory, open '/.vscode/c_cpp_properties.json'
keeps on popping up as a notification toast. 
Barring this irritating notification, intelliSense seems to be working fine.   
Screenshots:
![Screenshot from 2019-07-26 18-40-49](https://user-images.githubusercontent.com/42699688/61953859-15ca8700-afd5-11e9-8051-49a450c60d7d.png)

sean-mcmanus:
[...]

PR #3986, opened 2019-07-26
Fix error in open file scenario.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/3888 .

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -178,7 +178,7 @@ export class CppProperties {
         // defaultPaths is only used when there isn't a c_cpp_properties.json, but we don't send the configuration changed event
         // to the language server until the default include paths and frameworks have been sent.
         let configFilePath: string = path.join(this.configFolder, "c_cpp_properties.json");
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -178,7 +178,7 @@ export class CppProperties {
         // defaultPaths is only used when there isn't a c_cpp_properties.json, but we don't send the configuration changed event
         // to the language server until the default include paths and frameworks have been sent.
         let configFilePath: string = path.join(this.configFolder, "c_cpp_properties.json");
-        if (fs.existsSync(configFilePath)) {
+        if (fs.existsSync(configFilePath) && this.rootUri !== null) {
```
michelleangela: Would it be better to check for `this.rootUri !== null` first so it doesn't have to do existsSync if it's not true?

michelleangela (approved):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Sure.

PR #3987, opened 2019-07-26
Support for the V3 API

bobbrow:
I need to add tests too...


michelleangela (approved):

Extension/package-lock.json
```diff
@@ -2511,8 +2499,7 @@
         "number-is-nan": {
           "version": "1.0.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
```
michelleangela, line 2515: Is `"optional" : true` supposed to be removed?

PR #3988, opened 2019-07-26
Unsupported error message

andreeis:
Implement additional error message for unsupported scenarios, like any architecture other than x86 and amd64 or alpine containers.
Tested on raspberry pie and on a hello world alpine container.


bobbrow (dismissed):

pieandcakes:
@andreeis Can you move this to overall extension activation so it blocks entirely? I don't know if it makes sense for there to be debugging before language services.

andreeis:
I submitted another pull request with the same message implemented higher up, following the recommendation. I had some issues with git and this PR and just to skip fixing those issues I am closing this PR and sending a new one.

PR #4014, opened 2019-08-01
Typo and grammar fixes in documentation

ghost:
This is a PR that originate in VSCode. However, because some affected files also reside on this repository, I created this PR.

This breaks no code, and the only behavior changed is the user-facing code, which was changed to be correct in grammar and spelling. There are also typos in the comments I corrected as well.

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -68,11 +68,11 @@ async function processRuntimeDependencies(): Promise<void> {
                 // Send the failure telemetry since postInstall will not be called.
                 sendTelemetry(await PlatformInformation.GetPlatformInformation());
             }
-        // The extension have been installed and activated before.
+            // The extension have been installed and activated before.
         } else {
             await finalizeExtensionActivation();
         }
-    // No lock file, need to download and install dependencies.
+        // No lock file, need to download and install dependencies.
     } else {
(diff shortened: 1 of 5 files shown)


Extension/src/main.ts
```diff
@@ -68,11 +68,11 @@ async function processRuntimeDependencies(): Promise<void> {
                 // Send the failure telemetry since postInstall will not be called.
                 sendTelemetry(await PlatformInformation.GetPlatformInformation());
             }
-        // The extension have been installed and activated before.
+            // The extension have been installed and activated before.
```
bobbrow: Can you move this one inside the `else` block?  And fix `have` --> `has`?

Extension/src/main.ts
```diff
@@ -68,11 +68,11 @@ async function processRuntimeDependencies(): Promise<void> {
...
         } else {
             await finalizeExtensionActivation();
         }
-    // No lock file, need to download and install dependencies.
+        // No lock file, need to download and install dependencies.
```
bobbrow: Can you move this one inside the `else` block?

bobbrow (approved):
Thank you for the PR.

ghost:
Fixed!

PR #4019, opened 2019-08-03
VS Code themes that match VS Light and Dark, including semantic tokens

Colengms:
Attempts to more closely match VS (2017) themes.
Does not yet include new VS 2019 changes to Light and Dark.


Themes/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "cpptools-themes",
+  "displayName": "C/C++ Themes",
+  "description": "UI Themes for C/C++ extension.",
+  "version": "0.24.1",
```
bobbrow: Update the version.

Themes/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "publisher": "ms-vscode",
+  "preview": true,
```
bobbrow: Does this token tell VS Code to preview the theme when the extension is installed?  If you set this to false, can we bundle it with our extension after all?

bobbrow (changes requested):

Themes/package.json
Colengms: This does not appear to be related to themes.

https://code.visualstudio.com/api/references/extension-manifest


preview |   | boolean | Sets the extension to be flagged as a Preview in the Marketplace.
-- | -- | -- | --

In emails with @aeschli , he seemed to imply this behavior is not currently optional.

Themes/package.json
Colengms: Should this have its own version number?  These would be in their own vsix, and we probably don't want to have to update its version every time we update cpptools.

bobbrow: Sorry, that's what I meant.  This would be version 1.0.0 probably.

Themes/package.json
bobbrow: Ok.  Then we should set this to false since the theme extension will not be marked as "Preview" in the Marketplace.

bobbrow (approved):

PR #4027, opened 2019-08-06
Implement not supported message for arm platform and alpine containers

andreeis:
Implement additional error message for unsupported scenarios, like any architecture other than x86 and amd64 or alpine containers.
Tested on raspberry pie and on a hello world alpine container.


michelleangela (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -70,6 +70,17 @@ export function activate(activationEventOccurred: boolean): void {
         return;
     }
 
+    let errMsg: string;
```
bobbrow: I thought we were going to put these changes in main.ts so that debugging will be disabled too.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
andreeis: I didn't realize there was an activation in main.ts. Originally the message was in client.ts and then I moved it higher up in extension.ts so I thought that was it. True, activate in main.ts is the correct place. I made the correction, retested and pushed.

bobbrow (approved):

PR #4068, opened 2019-08-12
Minor references fixes

sean-mcmanus:


Colengms (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1953,12 +1952,7 @@ class DefaultClient implements Client {
 
     private processReferencesResult(referencesResult: ReferencesResult): void {
         this.referencesViewFindPending = false;
-        if (!this.referencesChannel) {
-            this.referencesChannel = vscode.window.createOutputChannel("C/C++ References");
-            this.disposables.push(this.referencesChannel);
-        } else {
-            this.referencesChannel.clear();
-        }
+        this.referencesChannel.clear();
```
bobbrow, line 1955: Is there a guarantee that this channel is available?  Is `handleReferencesProgress` always called first?

sean-mcmanus: Yes, it's guaranteed. That's why I removed it.

Issue #4058, opened 2019-08-10
updating compile_commands.json cause error message and indexing freezes.


mdzahidh:
**Describe the bug**
- OS and Version: Linux and Windows
- VS Code Version: 1.37.0-insider
- C/C++ Extension Version: 0.25.0-insiders

**To Reproduce**
When you update compile_commands.json while vscode is running, cpptools gives out error saying ""${workspaceFolder}/compile_commands.json" could not be parsed. 'includePath' from c_cpp_properties.json will be used instead.", sometimes multiple times, and the indexing either takes very long or freezes (the red flame turns on indefinitely), which also freezes all code intellisense.

**Expected behavior**
cpptools should pick up changes in compile_commands.json after the file has been fully updated and re-indexing shouldn't freeze.

sean-mcmanus:
This sounds related to the changes https://github.com/microsoft/vscode-cpptools/issues/2306 and https://github.com/microsoft/vscode-cpptools/issues/3864 .

mdzahidh:
While I really want to thank the team for addressing this bug in 0.25.0-insiders, but I think there is still something that's causing issues. Perhaps, cpptools is picking up the changes while the compile_command.json is in the middle of being updated ? For example, our compile_commands.json is fairly large and it takes a little long to write it out.

mdzahidh:
Could you repro this on your end ? fixing this would be such a tremendous productivity boost for us

sean-mcmanus:
I'm not sure if we've reproed it yet -- but the fix might be to just set a timer (like we do with our mac crash file watcher) and/or later switch to using a different file watcher API -- VS Code added a vscode.workspace.fs API with 1.37, but that would require us taking a dependency on that newer version (but that API might have the same issue?).

sean-mcmanus:
Are you able to check how many seconds it takes for your compile_commands.json to be written? Do you think waiting 5 seconds before processing the change would work? Would possibly delaying the update a few seconds longer than necessary be an issue for you?

sean-mcmanus:
We have a potential fix at https://github.com/microsoft/vscode-cpptools/pull/4081 .

mdzahidh:
The fix looks great given there is no other event such as "closed" being triggered by the fileWatcher. I checked your branch, and it does seem to resolve the problem. Thanks a lot! can't wait for this to see in a released version.

sean-mcmanus:
[...]

PR #4081, opened 2019-08-13
Fix bug with compile_commands.json updating after a change.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/4058

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -581,6 +581,8 @@ export class CppProperties {
         }
     }
 
+    private compileCommandsFileWatcherChangedPending: boolean = false;
+
     // Dispose existing and loop through cpp and populate with each file (exists or not) as you go.
     // paths are expected to have variables resolved already
     public updateCompileCommandsFileWatchers(): void {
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -598,9 +601,22 @@ export class CppProperties {
...
-                    if (event !== "rename") {
-                        this.onCompileCommandsChanged(path);
+                    if (event === "rename") {
+                        return;
+                    }
+                    // Wait 1 second after a change to allow time for the write to finish.
+                    if (this.compileCommandsFileWatcherTimer) {
+                        clearInterval(this.compileCommandsFileWatcherTimer);
                     }
+                    this.compileCommandsFileWatcherFiles.add(path);
```
mdzahidh: looks like compileCommandsFileWatcherFiles may get multiple copies of "path" if file change events are trigger within quick successions. Maybe we need to check if "path" already exists or simply clear the list in the same clock as clearInterval above ?

mdzahidh: o never mind, i didn't notice it was a Set<string> :)

bobbrow (approved):

Issue #4098, opened 2019-08-16
Click the search icon to preview results - which search icon?


Timmmm:
Minor issue using 0.25.0-insiders2: I tried the new "Find All References..." function, and it starts "confirming" files (weird verb by the way - surely "processing" or "parsing" makes more sense?). Anyway I get this message:

> Find All References: 171/176 files confirmed. Click the search icon to preview results.

I have no idea which "search icon" it is referring to. I tried a few and none of them showed me any preview results.

michelleangela:
The "search icon" to preview results is on the right side of the status bar. 

![image](https://user-images.githubusercontent.com/38734287/63185168-1f607100-c00e-11e9-869b-9bcaab9afcf2.png)

The preview results get displayed on the "REFERENCES:RESULTS" panel.

You can click on the "search icon" again to refresh the preview results. Once the search is complete, the results panel should auto update with the final results.

sean-mcmanus:
Where do you see the verb "confirming"?

The VS Code UI APIs don't allow us to add buttons for commands to the progress UI -- if they fix that we could add a "Preview Results" button. We also can't show an image of the icon in the progress UI unless we use the Unicode search icon, which looks odd.

Is there a reason you didn't notice the search icon in the task bar?

Timmmm:
Ah! I didn't look in the status bar.. I guess because I was expecting the button to preview the results to be near where the results are.

Anyway I guess you could just change it to "Click the search icon in the status bar to preview results." Dunno about "task bar". Maybe that's what it's called but I bet most users don't know that.

It isn't literalled "confirming" but "confirmed",  as in "171/176 files confirmed".

tokyophil:
I had the same problem. I tried the search icon on the left, which i usually use for searching. I did not expect another search icon to click and did not notice the one in the status bar.
Also, i clicked the settings icon, because i thought that maybe the settings icon in the message is representing the icon i should click... (see screenshot)
![Searchicon](https://user-images.githubusercontent.com/19667554/63261785-97d06780-c284-11e9-9ce3-947c937c67d7.JPG)

maybe a message like "Click the search icon in the status bar to preview results" would help others too.

michelleangela:
Would it be helpful if the search icon is a different color so that it is more noticeable? Like the "flame" icon for IntelliSense parsing is red?

tokyophil:
[...]

PR #4106, opened 2019-08-19
Improve the search icon message.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/4098

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1775,7 +1775,7 @@ class DefaultClient implements Client {
     private readonly referencesProgressDelayInterval: number = 2000;
 
     private reportReferencesProgress(progress: vscode.Progress<{message?: string; increment?: number }>, forceUpdate: boolean): void {
-        const helpMessage: string = this.model.referencesCommandMode.Value === ReferencesCommandMode.Peek ? "" : " Click the search icon to preview results.";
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -1775,7 +1775,7 @@ class DefaultClient implements Client {
     private readonly referencesProgressDelayInterval: number = 2000;
 
     private reportReferencesProgress(progress: vscode.Progress<{message?: string; increment?: number }>, forceUpdate: boolean): void {
-        const helpMessage: string = this.model.referencesCommandMode.Value === ReferencesCommandMode.Peek ? "" : " Click the search icon to preview results.";
+        const helpMessage: string = this.model.referencesCommandMode.Value === ReferencesCommandMode.Peek ? "" : " Click the search icon in the status bar to preview results.";
```
michelleangela, line 1778: is it possible to use the actual search icon in the text?

bobbrow: We can't use the same icon in the text as the one that shows up in the status bar.  The unicode character renders differently too.

bobbrow (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: We can if we use the search Unicode character, but it looks "odd" with colors. The status bar has a special image. We could probably do something better if they supported markup or buttons in the UI.

Issue #4141, opened 2019-08-26
new input verification of configurations.intelliSenseMode fails to verify msvc-x{86|64} if configurations.compilerPath contains a variable


mistersandman:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows (any version)
- VS Code Version: any
- C/C++ Extension Version: 0.25.0
- Other extensions you installed (and if the issue persists after disabling them): none
- A clear and concise description of what the bug is.

**To Reproduce**
Example configuration in `c_cpp_properties.json` triggering the issue:
```
{
    "env": {
        "vc.path": "<path_to_your_msvc_installation>",
        ...
    },
    "configurations": [
        {
            ...,
            "compilerPath": "${vc.path}\\bin\\HostX64\\x64\\cl.exe",
            "intelliSenseMode": "msvc-x64",
            ...
        }
    ]
}
```
Problems reported for the above configuration: `intelliSenseMode msvc-x64 is incompatible with compilerPath.`

**Expected behavior**
No problems reported, as the expanded `compilerPath` is a valid compiler for `msvc-x64` mode.

**Additional context**
In function `isCompilerIntelliSenseModeCompatible`, `util.extractCompilerPathAndArgs` is called with the unexpanded compiler path, which only expects expanded paths.
https://github.com/microsoft/vscode-cpptools/blob/3251ee599ba677d95e00df0dc62019a2337d4713/Extension/src/LanguageServer/configurations.ts#L406

sean-mcmanus:
The fix is available in 0.25.1 now.

PR #4142, opened 2019-08-26
Resolve compiler path before verification (fixes issue #4141)

mistersandman:
Fixes issue #4141


sean-mcmanus (approved):

michelleangela (approved):
Thanks for catching and fixing this.

Issue #3632, opened 2019-05-11
Cannot revert to older version if newest one crashes


Cannedfood:
Type: General

**Describe the bug**
- OS and Version: N/A - Platform independent
- VS Code Version: 1.33.1
- C/C++ Extension Version: <= 0.23.0
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is:

The extension updates itself to the newest release, ignoring vscode's settings regarding updates.
That way, if the newest version crashes for you, all you can do is wait for it to get fixed.

**To Reproduce**
1. Go to settings
2. Disable automatic extension updates (Features > Extension Viewlet > Auto Update)
3. Downgrade the extension ([The cog next to the extension] > Install another version)
4. Restart vscode
4. See the extension upgrade to the newest version anyways

**Expected behavior**
The extension should rely on vscode's update mechanisms.
OR
Allow you to lag at least one version behind somehow, so you're not blocked until an issue is fixed.

pieandcakes:
@bobbrow @sean-mcmanus this should be working for @Cannedfood right?

sean-mcmanus:
@Cannedfood You also need to change the C_Cpp.updateChannel to "Default" instead of "Insiders".

Also, we're releasing a fix for a high hitting crash today with 0.23.1. If you're still getting a crash with 0.23.1, we could fix the crash if a call stack can be obtained.

bobbrow:
We should change the behavior and let the vscode setting override ours.  We were doing the opposite previously.

sean-mcmanus:
This should be fixed starting with 0.25.1 (as in, we check the extensions.autoUpdate before updating) -- however, if you revert to a version older than 0.25.0, then it won't have that fix so, so the benefit would only start when we release 0.26.0-insiders and you change to 0.25.1.

PR #4151, opened 2019-08-27
Check extensions.autoUpdate.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/3632


sean-mcmanus:
I tested it via creating a vsix with an older 0.24.1-insiders name and it correctly doesn't auto-update to the 0.25.0 after this fix (with autoUpdate set to false and Insiders channel set), and updates when autoUpdate and Insiders updateChannel is set.

Colengms (approved):

PR #4167, opened 2019-08-29
Move references code to separate class/file

michelleangela:
This is initial change set for refactoring code for "Find all references".


sean-mcmanus (approved):

sean-mcmanus (changes requested):
This needs to updated/merged with the localization changes.

sean-mcmanus (approved):

PR #4187, opened 2019-09-03
Remove navigation

Jasdriel:
Remove code related to navigation bar, but if I missing something please let me know


sean-mcmanus (changes requested):
I think there are other references to C_Cpp.Navigate that should be removed.

sean-mcmanus:
Also, I still see references to navigationStatusBarItem.

Extension/package.json
```diff
@@ -185,12 +185,6 @@
           "description": "%c_cpp.configuration.formatting.description%",
           "scope": "resource"
         },
-        "C_Cpp.navigation.length": {
-          "type": "number",
-          "default": 60,
-          "description": "%c_cpp.configuration.navigation.length.description%",
```
Colengms, line 191: Could you also remove this string from package.nls.json?  Same with any other string (placeholders) in package.json removed and no longer referenced.  (c_cpp.command.navigate.title)

Colengms (dismissed):

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #4214, opened 2019-09-06
Joreyesp/remove navigation

Jasdriel:
change nativeString and modify glupfile to work with empty strings


Colengms:
The native and typescript are out of sync right now, so we need to get this in ASAP to get those string indexes to line up again.

Colengms (approved):

PR #4234, opened 2019-09-12
Localize strings for document symbols and workspace symbols

Jasdriel:
Changes to localize some strings that are shown in the outline and in the symbol search


Extension/src/LanguageServer/client.ts
```diff
@@ -435,44 +461,117 @@ export class DefaultClient implements Client {
...
+                                let r: Range;
+                                if (range instanceof vscode.Selection) {
+                                    if (range.active.isBefore(range.anchor)) {
+                                        r = Range.create(Position.create(range.active.line, range.active.character), Position.create(range.anchor.line, range.anchor.character));
+                                    } else {
+                                        r = Range.create(Position.create(range.anchor.line, range.anchor.character), Position.create(range.active.line, range.active.character));
+                                    }
                                 } else {
-                                    r = Range.create(Position.create(range.anchor.line, range.anchor.character), Position.create(range.active.line, range.active.character));
+                                    r = Range.create(Position.create(range.start.line, range.start.character), Position.create(range.end.line, range.end.character));
```
Colengms, line 475: I think this part needs to be reverted.  'range' is either a vscode.Range or a vscode.Selection (range: vscode.Range | vscode.Selection).  This else cause is for vscode.Selection, so needs to use the vscode.Selection members.

Extension/src/LanguageServer/client.ts
```diff
@@ -1869,6 +1968,7 @@ class NullClient implements Client {
     TrackedDocuments = new Set<vscode.TextDocument>();
     onDidChangeSettings(event: vscode.ConfigurationChangeEvent): { [key: string] : string } { return {}; }
     onDidOpenTextDocument(document: vscode.TextDocument): void {}
+    afterDidOpenTextDocument(): void {}
```
Colengms: This can be removed?

Colengms (approved):

Issue #4233, opened 2019-09-12
[cppvsdbg] requireExactSource option is not recognized by launch.json


guusw:
Type: Debugger

Sometimes when debugging my application (UE4 project) the symbols seem to not match the source files. Although I would still like to debug my application.

The debug  console suggests that I can disable this source file matching in my launch.json by adding `"requireExactSource":false`, however the settings' json linting tells me this option is not allowed.

Extension version: 0.25.1
VS Code version: Code 1.37.0 (036a6b1d3ac84e5ca96a17a44e63a87971f8fcc8, 2019-08-08T02:33:50.993Z)
OS version: Windows_NT x64 10.0.18362

pieandcakes:
@guusw We will take a look. thanks

PR #4235, opened 2019-09-12
Adding requireExactSource flag to vsdbg schema

WardenGnaw:
Addresses https://github.com/microsoft/vscode-cpptools/issues/4233#event-2630065026

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1301,6 +1301,11 @@
                     "default": false
                   }
                 }
+              },
+              "requireExactSource": {
+                "type": "boolean",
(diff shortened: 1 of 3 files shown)


pieandcakes (approved):

Extension/package.json
```diff
@@ -1333,6 +1338,11 @@
                     "default": false
                   }
                 }
+              },
```
pieandcakes, line 1381: i thought you said it was only used in launch ?

pieandcakes (approved):

Extension/package.json
WardenGnaw: Nope. Also used in attach.

PR #4237, opened 2019-09-13
Add custom view for other references results

michelleangela:
- contribute a new view to references-view container that displays other references results
- add icons for references types


Extension/src/LanguageServer/extension.ts
```diff
@@ -1131,6 +1133,13 @@ function onRescanWorkspace(): void {
     clients.forEach(client => client.rescanFolder());
 }
 
+function onShowRefCommand(arg?: ReferenceItem): void {
+    const { location } = arg;
+    vscode.window.showTextDocument(location.uri, {
+        selection: location.range.with({ end: location.range.start })
```
bobbrow: arg is an optional param.  Should it be validated before it is dereferenced?

Extension/src/LanguageServer/references.ts
```diff
@@ -129,11 +131,26 @@ export class ProgressHandler {
         this.client = client;
     }
 
+    initilizeViews(): void {
```
bobbrow: [sp] `initialize`

Extension/src/LanguageServer/references.ts
```diff
@@ -93,12 +94,14 @@ function convertReferenceTypeToString(referenceType: ReferenceType, isReferences
...
 }
-
 export class ProgressHandler {
     private client: DefaultClient;
     private disposables: vscode.Disposable[] = [];
 
+    // TODO: move views to class that manages view
     private referencesChannel: vscode.OutputChannel;
+    private findAllRefsView: FindAllRefsView;
+    private viewsInitilized: boolean = false;
```
bobbrow: [sp] `viewsInitialized`

Extension/src/LanguageServer/referencesModel.ts
```diff
@@ -0,0 +1,94 @@
...
+                fileItem.addReference(reference);
+            }
+        }
+    }
+}
+
+export class ReferenceTypeItem {
+    private files: FileItem[] = [];
+
+    constructor(
```
bobbrow: formatting?

Extension/src/LanguageServer/referencesModel.ts
```diff
@@ -0,0 +1,94 @@
...
+            file = new FileItem(uri, fileName);
+            this.files.push(file);
+        }
+        return file;
+    }
+
+    private indexOfFile(fileName: string): number {
+        return this.files.findIndex(function(item): boolean {
+            return item.name === fileName;
+         });
```
bobbrow: nit: indent looks off by one

Extension/src/LanguageServer/referencesProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+import * as vscode from 'vscode';
+import * as util from '../common';
+import { Model, FileItem, ReferenceItem, ReferenceTypeItem } from './referencesModel';
+import { ReferenceInfo, ReferenceType, convertReferenceTypeToString } from './references';
+
+function getReferneceTypeIconPath(referenceType: ReferenceType): vscode.ThemeIcon {
```
bobbrow: [sp] `getReferenceTypeIconPath`

Colengms (approved):

Extension/src/LanguageServer/referencesProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        if (element instanceof ReferenceTypeItem) {
+            return element.getFiles();
+        }
+
+        return this.references.items;
+    }
+}
+
+// TODO: add provider for rename
```
bobbrow, line 95: Would this be the right file to put 'rename' stuff in?  This is referencesProvider.ts.  Perhaps renaming the file or putting the provider in a different file would be better.

Extension/src/LanguageServer/referencesView.ts
```diff
@@ -0,0 +1,38 @@
...
+    setData(results: ReferenceInfo[]): void {
+        this.referenceViewProvider.setModel(results);
+    }
+
+    clearData(): void {
+        this.referenceViewProvider.clear();
+    }
+}
+
+// TODO: create class to manage displaying data to rename view
```
bobbrow, line 36: Similar comment as above.  'rename' stuff should probably get its own file.

bobbrow (commented):
General feedback.

Extension/src/LanguageServer/extension.ts
michelleangela: There should be check on "arg"

Extension/src/LanguageServer/referencesProvider.ts
michelleangela: Pending design of rename and separating of current code that's used for reference results and rename. Planning to check this in first for Colen and I to collaborate on.

Extension/src/LanguageServer/referencesView.ts
michelleangela: Pending design for rename and further separate of code for references results and rename.

Extension/package.json
```diff
@@ -43,6 +43,15 @@
   ],
   "main": "./dist/main",
   "contributes": {
+    "views": {
+      "references-view": [
+        {
+          "id": "CppReferencesView",
+          "name": "%c_cpp.contributes.views.cppReferencesView.title%",
+          "when": "cppReferenceTypes:hasResults && reference-list.hasResult"
```
sean-mcmanus, line 51: I think this should be ||, right?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1131,6 +1133,15 @@ function onRescanWorkspace(): void {
     clients.forEach(client => client.rescanFolder());
 }
 
+function onShowRefCommand(arg?: ReferenceItem): void {
+    if (arg) {
+        const { location } = arg;
+        vscode.window.showTextDocument(location.uri, {
+            selection: location.range.with({ end: location.range.start })
```
sean-mcmanus, line 1140: We could send back the end or the source string to get the full range later.

Extension/src/LanguageServer/references.ts
```diff
@@ -218,58 +235,65 @@ export class ProgressHandler {
...
+
         switch (notificationBody.referencesProgress) {
             case ReferencesProgress.StartedRename:
             case ReferencesProgress.Started:
-                this.referencesStartedWhileTagParsing = this.client.IsTagParsing;
-                let mode: ReferencesCommandMode = notificationBody.referencesProgress === ReferencesProgress.StartedRename ? ReferencesCommandMode.Rename :
-                    (this.visibleRangesDecreased && (Date.now() - this.visibleRangesDecreasedTicks < this.ticksForDetectingPeek) ?
-                    ReferencesCommandMode.Peek : ReferencesCommandMode.Find);
-                this.client.setReferencesCommandMode(mode);
                 if (this.client.ReferencesCommandMode === ReferencesCommandMode.Peek) {
```
sean-mcmanus, line 293: Why is this not in the handleProgressStarted method?

Extension/src/LanguageServer/referencesProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+import * as vscode from 'vscode';
+import * as util from '../common';
+import { Model, FileItem, ReferenceItem, ReferenceTypeItem } from './referencesModel';
+import { ReferenceInfo, ReferenceType, convertReferenceTypeToString } from './references';
+
+function getReferenceTypeIconPath(referenceType: ReferenceType): vscode.ThemeIcon {
+    // TODO: return icon path for light and dark themes based on reference type
+    switch (referenceType) {
```
sean-mcmanus, line 13: ConfirmationInProgress is missing. The icon should change depending on if the operation is canceled or not. VS doesn't appear to use any icon for this though, even though it's a type.

sean-mcmanus:
Do you know where we are going to expose the "Warning message" about the tag parsing not being finished? I see that it's still shown in the C/C++ references output window -- can the warning appear in the tree view if that is used? If not, it seems like it should be moved to the regular C/C++ output window if the references output isn't being used.

sean-mcmanus (approved):

Extension/package.json
sean-mcmanus: Or just cppReferenceTypes:hasResults ? If there isn't any reason to show the cppReferenceTypes if there are only confirmed references.

sean-mcmanus (changes requested):
"when": "cppReferenceTypes:hasResults ?

michelleangela:
The custom view is only shown when there are confirmed results and there are other reference types. Sometimes, there are only confirmed results, so there is nothing to output in the custom references view.

Extension/src/LanguageServer/referencesModel.ts
```diff
@@ -0,0 +1,93 @@
...
+            if (index < 0) {
+                const uri: vscode.Uri = vscode.Uri.file(r.file);
+                fileItem = new FileItem(uri, r.file);
+                this.items.push(fileItem);
+            } else {
+                fileItem = this.items[index];
+            }
+
+            // Add reference to file
+            let isFileReference: boolean = (r.position.line === 0 && r.position.character === 0);
```
sean-mcmanus, line 29: Can you handle "file references"? Those are cases where the line/column are not known yet, in which case, it seems like it should add a link to just the start of the file.

Oh, it looks like it adds the file, but just leaves out the reference...that seems fine.

michelleangela:
> 's still shown in the C/C++ references output window -- can the warning appear in the tree view if that is used? If not, it seems like it should be moved to the regular C/C

I don't know where to place this warning yet.

Extension/src/LanguageServer/referencesProvider.ts
michelleangela: Will add icon for ConfirmationInProgress later.

sean-mcmanus (approved):

Extension/package.json
michelleangela: Verified that reference-list.hasResult is also true if there are no confirmed references or if you cancel "find all references" while in progress.

Extension/src/LanguageServer/extension.ts
michelleangela: Having the full range of the symbol result would mean we will highlight in the editor. A selection of a confirmed reference already highlights the symbol. Should we highlight two different selections?

Extension/src/LanguageServer/references.ts
michelleangela: I wanted to separate telemetry stuff from the logic that processes the data.

Extension/src/LanguageServer/referencesModel.ts
michelleangela: We still need to handle file references.

Issue #4052, opened 2019-08-09
"Cancel" for Find All References can take "too long" when it's in the "Processing source" state


sean-mcmanus:
The reason "Cancel" gets "stuck" is that it is doing the normal "IntelliSense operation" code without a separate "cancel" thread that can kill the IntelliSense process (which is what enables faster canceling when it's in the "confirming" state). It should be fixable...not sure if it's worth the effort though.

bobbrow:
Not seeing a lot of feedback on this.  Closing...

sean-mcmanus:
I'm seeing this happen when confirming references too (repros with call hierarchy too).

PR #4314, opened 2019-09-23
Work on FAR/Rename w/UI reliablity with cancellation

Colengms:


Extension/src/LanguageServer/client.ts
```diff
@@ -674,47 +689,63 @@ export class DefaultClient implements Client {
...
+                                        newName: newName,
+                                        position: Position.create(position.line, position.character),
+                                        textDocument: this.client.languageClient.code2ProtocolConverter.asTextDocumentIdentifier(document)
+                                    };
+                                    referencesParams = params;
+                                    this.client.notifyWhenReady(() => {
+                                        // The current request is represented by referencesParams.  If a request detects
+                                        // referencesParams does not match the object used when creating the request, abort it.
+                                        if (params !== referencesParams) {
+                                            reject();
```
bobbrow, line 708: `reject` does not return.  So unless you want to run the rest of the code in this function after rejecting, you should early return here.

Colengms: I do want it to run the rest of the code in that function.

michelleangela:
Does this address https://github.com/microsoft/vscode-cpptools/issues/4052 ?

Colengms:
> Does this address #4052 ?

I don't believe so.

michelleangela (approved):

Extension/src/LanguageServer/client.ts
Colengms: Ack, you were correct, this one should have returned.

bobbrow (approved):

PR #4324, opened 2019-09-25
Fix loc strings.

sean-mcmanus:


Extension/package.json
```diff
@@ -1630,9 +1630,9 @@
   },
   "scripts": {
     "vscode:prepublish": "yarn run compile",
-    "compile": "yarn run precompile && webpack --mode production --vscode-nls",
+    "compile": "yarn run precompile && yarn run translations-generate && webpack --mode production --vscode-nls",
     "compile-dev": "yarn run precompile && webpack --mode development",
-    "compile-watch": "yarn run precompile && webpack --mode production --vscode-nls --watch --info-verbosity verbose",
+    "compile-watch": "yarn run precompile && yarn run translations-generate && webpack --mode production --vscode-nls --watch --info-verbosity verbose",
     "compile-dev-watch": "yarn run precompile && webpack --mode development --watch --info-verbosity verbose",
```
sean-mcmanus, line 1630: This could also go in precompile (not sure what is preferred).

Colengms: I don't think this is a good idea.  compile-watch is not for a production build, so shouldn't incur the overhead of generating translations.  If you need translations, use the production build.

Colengms: Er, it looks like it's needed on compile, but I'd prefer to don't add it to compile-watch.

sean-mcmanus: Ok

Extension/package.json
```diff
@@ -1624,7 +1624,7 @@
   },
   "scripts": {
     "vscode:prepublish": "yarn run compile",
-    "compile": "yarn run precompile && webpack --mode production --vscode-nls",
+    "compile": "yarn run precompile && yarn run translations-generate && webpack --mode production --vscode-nls",
     "compile-dev": "yarn run precompile && webpack --mode development",
     "compile-watch": "yarn run precompile && webpack --mode production --vscode-nls --watch --info-verbosity verbose",
```
Colengms: Sorry, I gave you bad feedback on this.  Obviously, 'compile-watch' is for production and 'compile-dev-watch' is not.  Could you put the 'yarn run translations-generate' back on 'compile-watch'.   :]

sean-mcmanus: Sure

Colengms (approved):

PR #4337, opened 2019-09-25
Fix cancel not working after a preview.

sean-mcmanus:


Colengms (approved):

Colengms:
Disregard my last email about this change.  I think this should work fine.

PR #4341, opened 2019-09-26
Fix cancel references.

sean-mcmanus:


Extension/src/LanguageServer/client.ts
```diff
@@ -631,30 +631,25 @@ export class DefaultClient implements Client {
...
-                                                        locations.push(new vscode.Location(uri, range));
-                                                    }
-                                                });
-                                                resolve(locations);
+                                                let pendingCancel: ReferencesCancellationState = referencesPendingCancellations[0];
+                                                referencesPendingCancellations.pop();
+                                                pendingCancel.callback();
                                             }
+                                            let locations: vscode.Location[] = [];
+                                            result.referenceInfos.forEach(referenceInfo => {
```
sean-mcmanus, line 645: Why is this being called 2 times when a cancel occurs? I'm still investigating the difference between the "final" and non-final cases.

sean-mcmanus: I added a check to prevent it from resolving 2 times.

Colengms: Please don't check in until I can review.  Final is present to ensure a race condition does not occur due to the final result being received, but the Finished message not yet being received.  It needs to delay completion until the Finished message is received to avoid a subsequent operation thinking that Finish is its own. #Closed

Colengms: (VPN is horrible right now, sorry for the slow responses).  false is passed for final in only 1 place.  The problem is that (right there) we can't trigger the next request waiting in the cancellation queue until the Finished message is also received, or the next request WILL see it and complete too soon.  But, we don't know if the Finished message is going to be received (after the results).  We need a way to deliver the results to the callback but defer actually completing the promise until the Finished message is received IF it is not due to a preview.  If the results are due to a preview, we DO need to complete that request without waiting for Finished.

---
In reply to: [328821621](https://github.com/Microsoft/vscode-cpptools/pull/4341#discussion_r328821621) [](ancestors = 328821621)

sean-mcmanus: Are you saying this isn't the correct fix? I haven't noticed a bug or race condition yet. Is there a scenario I should test better?

Colengms: To test: Issue a FAR while another FAR is already in progress. (The scenario I've been trying to fix)  "Final" was originally intended to postpone the completion of the first operation until Finish is received.  Otherwise, the second operation will get quickly queued up and see the Finish message as its own, and fail to do anything. #Closed

Colengms: I tried to be detailed in my prior comment about what needs to be fixed.  We need to postpone the completion of the promise until Finished is received, if not due to a preview. #Closed

Colengms: Perhaps a clean fix might be to get rid of the 'finished' message, now that we are always sending a set of final results, and replace it with a 'finished' bool in the result message. #Closed

sean-mcmanus: I did that...still testing though.

Colengms (approved):
:shipit:

PR #4342, opened 2019-09-26
Fix rename cancel and subsequent renames

michelleangela:
**Issue 1:**
When doing subsequent renames, as a set of two, the second rename does nothing.

Repro steps:
1. do rename
2. commit rename --> operation completes
3. do another rename
4. commit rename --> nothing happens
5. do another rename 
6. commit rename --> operation completes
7. do another rename
8. commit rename --> nothing happens

Solution:
When the first rename completes, the client gets onDidChangeTextDocument(...) callback, which will cause a cancel on a rename operation if there is one in pending. However, the flag `renamePending` does not get reset when the first rename completes, therefore any second rename will get canceled. Solution is to set `renamePending` to false whenever the current rename session is completed.

**Issue 2:**
Canceling rename while searching for references is in progress does not cancel the blue progress bar on editor to stop.

Solution:
In `processResults(...)`, handle state of when `this.referencesCanceled` is true for `this.client.ReferencesCommandMode === ReferencesCommandMode.Rename`. That is ensure `resultsCallback` is called to complete the rename operation.

**Misc:**
Update "remove" icon from https://github.com/microsoft/vscode-icons/blob/master/icons/light/dash.svg to https://github.com/microsoft/vscode-icons/blob/master/icons/light/remove.svg

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2155,13 +2155,11 @@ export class DefaultClient implements Client {
     public cancelReferences(): void {
         referencesParams = null;
         renamePending = false;
-        if (referencesRequestPending) {
-            let cancelling: boolean = referencesPendingCancellations.length > 0;
-            if (!cancelling) {
-                referencesPendingCancellations.push({ reject: () => {}, callback: () => {} });
-                this.languageClient.sendNotification(CancelReferencesNotification);
-                this.references.closeRenameUI();
-            }
+        let cancelling: boolean = referencesPendingCancellations.length > 0;
+        if (!cancelling) {
+            referencesPendingCancellations.push({ reject: () => {}, callback: () => {} });
+            this.languageClient.sendNotification(CancelReferencesNotification);
+            this.references.closeRenameUI();
         }
     }
 
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -712,7 +712,6 @@ export class DefaultClient implements Client {
                                         this.client.languageClient.sendNotification(RenameNotification, params);
                                         this.client.references.setResultsCallback((final, referencesResult) => {
                                             referencesRequestPending = false;
-                                            renamePending = false;
                                             let workspaceEdit: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
                                             let cancelling: boolean = referencesPendingCancellations.length > 0;
                                             if (cancelling) {
@@ -2155,11 +2154,13 @@ export class DefaultClient implements Client {
     public cancelReferences(): void {
         referencesParams = null;
         renamePending = false;
-        let cancelling: boolean = referencesPendingCancellations.length > 0;
-        if (!cancelling) {
-            referencesPendingCancellations.push({ reject: () => {}, callback: () => {} });
-            this.languageClient.sendNotification(CancelReferencesNotification);
-            this.references.closeRenameUI();
+        if (referencesRequestPending) {
+            let cancelling: boolean = referencesPendingCancellations.length > 0;
+            if (!cancelling) {
+                referencesPendingCancellations.push({ reject: () => {}, callback: () => {} });
+                this.languageClient.sendNotification(CancelReferencesNotification);
+                this.references.closeRenameUI();
+            }
         }
     }
 
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -712,6 +712,7 @@ export class DefaultClient implements Client {
                                         this.client.languageClient.sendNotification(RenameNotification, params);
                                         this.client.references.setResultsCallback((final, referencesResult) => {
                                             referencesRequestPending = false;
+                                            renamePending = false;
                                             let workspaceEdit: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
                                             let cancelling: boolean = referencesPendingCancellations.length > 0;
                                             if (cancelling) {
@@ -744,6 +745,7 @@ export class DefaultClient implements Client {
                                     let cancelling: boolean = referencesPendingCancellations.length > 0;
                                     referencesPendingCancellations.push({ reject, callback });
                                     if (!cancelling) {
+                                        renamePending = false;
                                         this.client.languageClient.sendNotification(CancelReferencesNotification);
                                         this.client.references.closeRenameUI();
                                     }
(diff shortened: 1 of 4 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -2153,11 +2155,13 @@ export class DefaultClient implements Client {
     public cancelReferences(): void {
         referencesParams = null;
         renamePending = false;
-        let cancelling: boolean = referencesPendingCancellations.length > 0;
-        if (!cancelling) {
-            referencesPendingCancellations.push({ reject: () => {}, callback: () => {} });
-            this.languageClient.sendNotification(CancelReferencesNotification);
-            this.references.closeRenameUI();
+        if (referencesRequestPending) {
```
Colengc: By adding this check back in, you are actually undoing a previous fix made by Sean.  Cancel needs to go through even if there is no pending request, for FAR after a preview.  (In this state, the request is no longer pending, rather when final results arrive or another preview is request, it reissues the request). #Closed

Colengc (changes requested):

Extension/src/LanguageServer/client.ts
michelleangela: I'll revert this change. #Resolved

Colengms:
>         renamePending = false;

On thing I'm unclear on.  Why isn't it sufficient to clear rename to false here (and elsewhere) where cancellation occurs?  It does need to be cleared early, not when the cancelation completes.  Was there another place cancellation was occuring that I missed?

---
Refers to: Extension/src/LanguageServer/client.ts:2157 in 17f1727. [](commit_id = 17f1727942d202d6a569228da457320bb9808e59, deletion_comment = False)

Extension/src/LanguageServer/client.ts
```diff
@@ -712,6 +712,7 @@ export class DefaultClient implements Client {
                                         this.client.languageClient.sendNotification(RenameNotification, params);
                                         this.client.references.setResultsCallback((final, referencesResult) => {
                                             referencesRequestPending = false;
+                                            renamePending = false;
```
Colengms: >renamePending = false; [](start = 44, length = 22)

Could you remove this one?  It should be cleared as soon as cancellation is requested.  If this one is actually needed, let me know.  That might point to another problem. #Resolved

michelleangela: This one is needed, otherwise it will issue a cancel after a "commit rename" is completed. The cancel will get issued from `onDidChangeTextDocument(...)` because `renamePending` is true.

Colengms: Actually, this one is a problem.  It will clear renamePending, but it's possible for another rename to be queued after this that has already set renamePending to true, and this will stomp on it.

michelleangela:
> > ```
> >     renamePending = false;
> > ```
> 
> On thing I'm unclear on. Why isn't it sufficient to clear renamePending to false here (and elsewhere) where cancellation occurs? It does need to be cleared early, not when the cancelation completes. Was there another place cancellation was occuring that I missed?
> 
> Refers to: Extension/src/LanguageServer/client.ts:2157 in [17f1727](https://github.com/microsoft/vscode-cpptools/commit/17f1727942d202d6a569228da457320bb9808e59). [](commit_id = [17f1727](https://github.com/microsoft/vscode-cpptools/commit/17f1727942d202d6a569228da457320bb9808e59), deletion_comment = False)

The line Extension/src/LanguageServer/client.ts:2157 is for canceling which has no issue. The issue addressed in this PR is related to sending a cancel after a rename is completed when it's not supposed to send a cancel.

Extension/src/LanguageServer/client.ts
Colengms: Prior to removing the check for whether an request is really in progress in cancelReferences(), this was not an issue.  I think we need that check restored, but also check for the state in which a preview occurred instead, rather than always issue that cancallation.

Colengms: (In response to my deleted comment you replied to).  I think my above 2 responses cover the issue.

michelleangela: > but it's possible for another rename to be queued after this that has already set renamePending to true, and this will stomp on it.

this `renamePending` does get stomped on when an edit happens during rename pending state.

> I think we need that check restored, but also check for the state in which a preview occurred instead, rather than always issue that cancallation.

Will try checking for a "preview occurred"

Colengms: I was referring to it being set to true on line 695.  By setting it to false again on line 715, you are stomping on a subsequent Rename operation (not the one that was cancelled).  renamePending gets set outside of the 'queue' of pending operations.  As long as there is any rename pending, it should be true, and should only be cleared when a cancel occurs (as any pending rename should be cancelled by that).

michelleangela: Yep, that's what I ran into when I debugged the scenario:

1. do a rename
2. while a rename is pending, send another rename request

Extension/src/LanguageServer/client.ts
michelleangela: Concluded that this flag is needed to address subsequent renames to work. Need to create a different solution for FAR preview.

Extension/src/LanguageServer/client.ts
```diff
@@ -744,6 +745,7 @@ export class DefaultClient implements Client {
                                     let cancelling: boolean = referencesPendingCancellations.length > 0;
                                     referencesPendingCancellations.push({ reject, callback });
                                     if (!cancelling) {
+                                        renamePending = false;
```
Colengms: >renamePending = false; [](start = 40, length = 22)

Could you remove this?  Although it's cancelling, it's also queueing up a new rename, so should not clear the renamePending flag. #Closed

Extension/src/LanguageServer/client.ts
Colengms: I think there is still a problem here, even if we restore the check that was removed from cancelReferences.  We need to ensure that pendingRename gets cleared after the operation is completed successfully (only if there are no other renames pending), or onDidChangeTextDocument might wind up cancelling the next FAR operation thinking it's a rename.  ...  

---
In reply to: [328849839](https://github.com/Microsoft/vscode-cpptools/pull/4342#discussion_r328849839) [](ancestors = 328849839)

Colengms:
>             this.languageClient.sendNotification(CancelReferencesNotification);

This actually needs to happen before the IF.  To ensure we prevent the last thing on the referencesPendingCanallations queue from being started when the cancellation completes. 

---
Refers to: Extension/src/LanguageServer/client.ts:2161 in 17f1727. [](commit_id = 17f1727942d202d6a569228da457320bb9808e59, deletion_comment = False)

Extension/src/LanguageServer/client.ts
Colengms: I think we could address this with a counter that is incremented at the very start of each rename, and decremented on each path out of rename (whenever rejected or completed).  If decremented to 0, clear renamePending.   I could create a separate PR to do these things I'm suggesting, if you would prefer.

---
In reply to: [328870043](https://github.com/Microsoft/vscode-cpptools/pull/4342#discussion_r328870043) [](ancestors = 328870043,328849839)

Extension/src/LanguageServer/client.ts
michelleangela: - Counter from [#4344](https://github.com/microsoft/vscode-cpptools/pull/4344) was added to this changeset.
- Added flag to keep track of symbolSearchInProgress 

Some issues still rename though:

- Rename does not cancel if a file was edited when rename is pending (that is rename UI is shown waiting for user to take action).

Overall, canceling rename correctly still needs work.

PR #4344, opened 2019-09-27
Ensure renamePending gets reset when operation is complete

Colengms:
These are the changes I suggested on https://github.com/microsoft/vscode-cpptools/pull/4342

This does not address the condition we've discussed needing to be restored in cancelReferences.  I think that may not actually be needed.  As long as pendingRename is getting properly reset as the (last) rename operation is completed successfully, we shouldn't see unwanted calls to cancelReferences() after a completed rename.

### DIFF
--- a/Extension/src/LanguageServer/references.ts
+++ b/Extension/src/LanguageServer/references.ts
@@ -344,8 +344,20 @@ export class ReferencesManager {
             this.referencesChannel.show(true);
         }
 
-        if (this.client.ReferencesCommandMode === ReferencesCommandMode.Rename) {
+        let currentReferenceCommandMode: ReferencesCommandMode = this.client.ReferencesCommandMode;
+        if (referencesResult.isFinished) {
             this.symbolSearchInProgress = false;
+            clearInterval(this.referencesDelayProgress);
+            if (this.currentUpdateProgressTimer) {
+                clearInterval(this.currentUpdateProgressTimer);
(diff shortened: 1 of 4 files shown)


michelleangela:
Note: added this changeset to #4342 and it doesn't handle the case:

1. Do a rename with unconfirmed references
2. edit file to cause rename to cancel
Result: does not cancel rename

Investigating still...

sean-mcmanus (approved):

Extension/src/LanguageServer/references.ts
```diff
@@ -343,7 +344,20 @@ export class ReferencesManager {
...
+            if (this.currentUpdateProgressTimer) {
+                clearInterval(this.currentUpdateProgressTimer);
+                this.currentUpdateProgressResolve();
+                this.currentUpdateProgressResolve = null;
+                this.currentUpdateProgressTimer = null;
+            }
+            this.client.setReferencesCommandMode(ReferencesCommandMode.None);
+        }
+
+        if (currentReferenceCommandMode === ReferencesCommandMode.Rename) {
```
sean-mcmanus, line 360: Can this be an "else if"?

sean-mcmanus: Looks like it can't...

Colengms: Are you suggesting if isFinished, don't do the rest?  I think the rest is still needed even if isFinished.

Issue #4332, opened 2019-09-25
Change "C/C++ References" Output panel to "C++ Peek References" and only create it when Peek occurs


sean-mcmanus:
...this is with 0.26.0-insiders, due to the UI tree view being used.

The "Warning" messages can be sent to the C/C++ window instead (but with Peek it should still be shown in the C/C++ Peek References window, or it won't be visible).

PR #4363, opened 2019-10-01
Change "C/C++ References" output panel to "C/C++ Peek References" and create it when Peek occurs

Jasdriel:
The "Warning" messages can be sent to the C/C++ window instead (but with Peek it should still be shown in the C/C++ Peek References window, or it won't be visible). (https://github.com/microsoft/vscode-cpptools/issues/4332)

### DIFF
--- a/Extension/src/LanguageServer/references.ts
+++ b/Extension/src/LanguageServer/references.ts
@@ -9,6 +9,7 @@ import { FindAllRefsView } from './referencesView';
 import * as telemetry from '../telemetry';
 import * as nls from 'vscode-nls';
 import { RenameView } from './renameView';
+import * as logger from '../logger';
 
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/references.ts
```diff
@@ -159,7 +160,7 @@ export class ReferencesManager {
 
     initializeViews(): void {
         if (!this.viewsInitialized) {
-            this.referencesChannel = vscode.window.createOutputChannel(localize("c.cpp.references", "C/C++ References"));
+            this.referencesChannel = vscode.window.createOutputChannel(localize("c.cpp.peek.references", "C/C++ Peek References"));
```
bobbrow: This should only create the output channel when a "peek references" command happens.  For the normal "find all references" command, this should not be created.

bobbrow (dismissed):

sean-mcmanus (approved):

PR #4385, opened 2019-10-04
Use previously received browse config if invalid/empty browse config is received from provider

Colengms:


Extension/src/LanguageServer/client.ts
```diff
@@ -2001,40 +2001,59 @@ export class DefaultClient implements Client {
...
     private sendCustomBrowseConfiguration(config: any): Thenable<void> {
-        // config is marked as 'any' because it is untrusted data coming from a 3rd-party. We need to sanitize it before sending it to the language server.
-        if (!config || config instanceof Array) {
-            console.warn("discarding invalid WorkspaceBrowseConfiguration: " + config);
-            return Promise.resolve();
-        }
+        let lastCustomBrowseConfiguration: PersistentFolderState<WorkspaceBrowseConfiguration> = new PersistentFolderState<WorkspaceBrowseConfiguration>("CPP.lastCustomBrowseConfiguration", null, this.RootPath);
+
+        let sanitized: util.Mutable<WorkspaceBrowseConfiguration>;
+        while (true) {
```
bobbrow, line 2031: Just checking... This `while (true)` exists just so you can break out early when the config is set?  Can you add a brief comment to that effect?

Extension/src/LanguageServer/client.ts
```diff
@@ -2001,40 +2001,59 @@ export class DefaultClient implements Client {
...
-            return Promise.resolve();
-        }
+            sanitized = {...<WorkspaceBrowseConfiguration>config};
+            if (!util.isArrayOfString(sanitized.browsePath) ||
+                !util.isOptionalString(sanitized.compilerPath) ||
+                !util.isOptionalArrayOfString(sanitized.compilerArgs) ||
+                !util.isOptionalString(sanitized.standard) ||
+                !util.isOptionalString(sanitized.windowsSdkVersion)) {
+                console.warn("discarding invalid WorkspaceBrowseConfiguration: " + config);
+                if (lastCustomBrowseConfiguration.Value !== null) {
```
bobbrow, line 2050: I wasn't thinking we'd need to do this, but ok.

bobbrow (approved):

PR #4423, opened 2019-10-13
improve i18n chs clang-format translations

zhangyafeikimi:


michelleangela:
Referencing issue #4422

michelleangela:
The process to update translation strings is initiated internally, where our internal localization team manages and creates the GitHub pull requests to update the strings when needed.

We filed an internal issue for the localization team to update the string if approved. The string will then get updated by their pull requests.

Issue #4464, opened 2019-10-19
With non-EN display languages: Unable to create 'launch.json' file inside the '.vscode' folder (Cannot read property 'name' of undefined).


koutoftimer:
Type: Debugger

**Describe the bug**
- OS and Version: Win 10
- VS Code Version: 1.39.2
- C/C++ Extension Version: 0.26.0
- Other extensions you installed (and if the issue persists after disabling them): no other extensions installed

Following [Get Started with C++ and Mingw-w64][1] and stopped at "Configure debug settings" because of an error message "Unable to create 'launch.json' file inside the '.vscode' folder (Cannot read property 'name' of undefined).".

**To Reproduce**
1. From the Command Palette, type "launch" and then choose Debug: Open launch.json. Next, choose the GDB/LLDB environment.
2. Get an error mesage "Unable to create 'launch.json' file inside the '.vscode' folder (Cannot read property 'name' of undefined)."

**Additional context**
I have installed VS Code few minutes ago from "system setup". It is still "fresh" without any extensions, except "C/C++ for Visual Studio Code"

[1]: https://code.visualstudio.com/docs/cpp/config-mingw#_configure-debug-settings

sky3141:
Same problem, and v0.25.1 works well.

AsceticBoy:
Same Problem

md-software:
Same problem for me!

ebubekirtrkr:
same problem for me

sean-mcmanus:
We're having problems reproing this bug. Can someone use Help->Toggle Developer Tools and provide the call stack of the undefined reference? You have a mingw compiler installed, right?

koutoftimer:
@sean-mcmanus 
[console.ts_137.log](https://github.com/microsoft/vscode-cpptools/files/3753889/console.ts_137.log) and [notificationsAlerts.ts_38.log](https://github.com/microsoft/vscode-cpptools/files/3753890/notificationsAlerts.ts_38.log) appears in Developer Tools console. I think you are looking for this

```
notificationsAlerts.ts:38 Error: Unable to create 'launch.json' file inside the '.vscode' folder (Cannot read property 'name' of undefined).
    at fileService.readFile.then.then.e (debugConfigurationManager.ts:583)
    at process._tickCallback (internal/process/next_tick.js:68)
onDidNotificationChange @ notificationsAlerts.ts:38
_register.model.onDidNotificationChange.e @ notificationsAlerts.ts:26
```

[...]

PR #4480, opened 2019-10-22
Fix `launch.json` creation when using non-English display languages.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/4464 .

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -41,7 +41,7 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
 
     async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
         const configs: vscode.DebugConfiguration[] = await this.underlyingProvider.provideDebugConfigurations(folder, token);
-        const defaultConfig: vscode.DebugConfiguration = configs.find(config => isDebugLaunchStr(config.name));
+        const defaultConfig: vscode.DebugConfiguration = configs.find(config => isDebugLaunchStr(config.name) && config.request === "launch");
         console.assert(defaultConfig);
         const editor: vscode.TextEditor = vscode.window.activeTextEditor;
         if (!editor || !util.fileIsCOrCppSource(editor.document.fileName) || configs.length <= 1) {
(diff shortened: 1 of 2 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -24,7 +24,7 @@ nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFo
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 function isDebugLaunchStr(str: string): boolean {
-    return str === "(gdb) Launch" || str === "(lldb) Launch" || str === "(Windows) Launch";
+    return str.startsWith("(gdb) ") || str.startsWith("(lldb) ") || str.startsWith("(Windows) ");
```
pieandcakes, line 27: I don't know if this is the right fix. You could end up with the Attach script. you need to also check that the `request` field is `launch`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -114,7 +114,7 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
 
         // Filter out build tasks that don't match the currently selectede debug configuration type.
```
pieandcakes: `selected` has an extra letter `e`.

Same here, also should check defaultConfig.request is `launch`

pieandcakes (changes requested):

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: I couldn't get a repro where "Attach" was hit by this code. Do you know a repro or is it in case the isDebugLaunchStr is used incorrectly later on in a different case?

sean-mcmanus: I added code to check the request field...although I still don't know how to hit a case where the request is not "launch".

pieandcakes: I thought about this some more. Is your goal to get the first `launch` configuration? I think you can do this without doing string comparisons with the title.

You should be able to use `type` as `cppdbg` (or `cppvsdbg` for VC++) and the `request` field and make sure that `pipeTransport` object is null. 

I don't know why Griffin originally chose the name, probably because it was easier for his case.

sean-mcmanus: That doesn't work. Our build and debug tasks have the same type, etc.

pieandcakes (approved):

sean-mcmanus:
...waiting for debugger changelog info...

PR #4511, opened 2019-10-30
Seanmcm/new multi root

sean-mcmanus:
This PR is a work in progress and not ready for review.

I can't get the non-resource scoped settings to work -- it keeps generating the error `Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '${key}', provide the URI of a resource or 'null' for any resource.` Does anyone see what I'm doing wrong?

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -160,7 +160,7 @@
           ],
           "default": "Default",
           "description": "%c_cpp.configuration.autocomplete.description%",
-          "scope": "resource"
+          "scope": "window"
         },
         "C_Cpp.errorSquiggles": {
           "type": "string",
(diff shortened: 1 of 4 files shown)


sean-mcmanus:
> This PR is a work in progress and not ready for review.
> I can't get the non-resource scoped settings to work -- it keeps generating the error Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '${key}', provide the URI of a resource or 'null' for any resource. Does anyone see what I'm doing wrong?

Ah, I think I figured it out -- looks like it's coming from the collectSettings() -- I got confused because my breakpoint was incorrectly being hit at another spot where there resource was correctly null.

Issue #4510, opened 2019-10-29
Rename (F2) function not working in certain scenarios


janwilmans:
Rename works for a local variable, dubble click on the name, edit the value and press enter.
![image](https://user-images.githubusercontent.com/5933444/67785672-e3b24080-fa6d-11e9-9e5b-407c6700d7d7.png)

Becomes :

![image](https://user-images.githubusercontent.com/5933444/67785696-ef056c00-fa6d-11e9-90a2-6cf6ca7878f4.png)

So far so good.

However, if the variable can't be looked up, like here:

![image](https://user-images.githubusercontent.com/5933444/67785906-473c6e00-fa6e-11e9-89aa-a7df13413843.png)

rename doesn't work, no message, just nothing.
so I'm guessing this a the scenario many people run into, if you have not configured the build system right (cmake?), the symbol lookup fails and just nothing happens, which is very confusing to people.

looks like https://github.com/microsoft/vscode-go/issues/1954
or https://github.com/microsoft/vscode-python/issues/1992

Colengms:
Hi @janwilmans .  I believe what you are seeing is that the selected symbol has not been detected by IntelliSense.  Since there are no 'unconfirmed' matches, we do not display the Rename UI.  Instead, we immediately apply the rename.  In this case, we are applying the rename to all 0 of the found references.  We should probably display a message in this case that IntelliSense was unable to detect a symbol to rename.

Note that it shouldn't be necessary to configure a build system.  It should be sufficient that IntelliSense is configured in the C/C++ Extension.

sean-mcmanus:
Yeah, VS shows "A definition for the symbol 'duck' could not be located."

janwilmans:
Yes, vs2019 shows: 
![image](https://user-images.githubusercontent.com/5933444/67803890-33a0ff80-fa8e-11e9-8c23-b621c7087648.png)

when you try to 

![image](https://user-images.githubusercontent.com/5933444/67803954-54695500-fa8e-11e9-8580-d29916642f59.png)

doesn't even popup the rename edit box, which makes it at least very clear that nothing is going to be renamed :)

michelleangela:
Rename experience improved in version 0.26.2-insiders, available at https://github.com/microsoft/vscode-cpptools/releases/tag/0.26.2-insiders.

platocrat:
For anyone wondering about how to pass the _Rename Refactoring_ tutorial on Mac, first highlight the function or variable name, then press CMD + F2.

Issue #4504, opened 2019-10-28
Rename UI fails to rename symbols after the document cursor moves in VS Code 1.40.0-insiders


Colengms:
This is related to:
https://github.com/Microsoft/vscode/issues/82146
https://github.com/microsoft/vscode/issues/83459

In VS Code 1.40.0-insiders, Rename operations that require disambiguation (display rename UI due to unconfirmed references found) will likely fail to apply. We are working with the VS Code to identify the proper solution.

sean-mcmanus:
It only repros after the document cursor moves (although clicking the leaf reference in the tree view will trigger the cursor change as well).

michelleangela:
Fixed in version 0.26.2-insiders, available at https://github.com/microsoft/vscode-cpptools/releases/tag/0.26.2-insiders.

PR #4517, opened 2019-10-30
Dismiss progress indicator when in Rename UI.  Add error dialog when no symbol is found.

Colengms:
This addresses: 

https://github.com/microsoft/vscode-cpptools/issues/4510
https://github.com/microsoft/vscode-cpptools/issues/4504

### DIFF
--- a/Extension/src/LanguageServer/references.ts
+++ b/Extension/src/LanguageServer/references.ts
@@ -420,6 +420,8 @@ export class ReferencesManager {
                 if (!foundUnconfirmed) {
                     this.resultsCallback(referencesResult, true);
                 } else {
+                    // Passing a null result and doResult of true to resultsCallback will cause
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/references.ts
```diff
@@ -420,9 +420,10 @@ export class ReferencesManager {
                 if (!foundUnconfirmed) {
                     this.resultsCallback(referencesResult, true);
                 } else {
+                    this.resultsCallback(null, true);
```
michelleangela, line 425: >this.resultsCallback(null, true); [](start = 20, length = 33)

Add comment for calling this.resultsCallback(null, true) ? Does this cancel the progress bar?

sean-mcmanus (approved):

PR #4519, opened 2019-10-31
Improve TU selection

Colengms:
This corresponds with filename cache / TU selection improvements in the native extension process.

On the TypeScript, the change is basically to use a list of candidate TU files, instead of just one, when loading custom configuration.

Other changes:
 - Explicitly clear custom configurations, rather than use update of multiple configurations to imply the need to clear them.
 - Does not clone the source file's custom configuration to the header file.  This behavior resulted in the header never again opening within the TU of the source, as the existence of the custom configuration for the header circumvented lookup of TU source file.


Colengms:
Tests are failing due to the change to QueryTranslationUnitSourceResult, which is received when opening a file.  Tests pass locally.

sean-mcmanus (approved):

PR #4525, opened 2019-11-01
Add test project folder for automated and manual tests

michelleangela:
This test project folder will be used to store files for automated and manual testing.

Initial files will be used for "find all references" feature.

### DIFF
--- a/Extension/test/integrationTests/testAssets/SingleRootProject/references.cpp
+++ b/Extension/test/integrationTests/testAssets/SingleRootProject/references.cpp
@@ -1,0 +1,54 @@
+namespace foo {
+	namespace bar {
+		namespace baz {
+			int qux = 42;
+		}
+	}
+}
+
+namespace foo::bar::baz
+{
+}
+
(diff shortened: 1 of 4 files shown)


Extension/test/integrationTests/testAssets/SingleRootProject/references.cpp
```diff
@@ -0,0 +1,54 @@
+namespace foo {
+	namespace bar {
+		namespace baz {
+			int qux = 42;
+		}
+	}
+}
```
sean-mcmanus, line 7: I got this test case code from https://en.cppreference.com/w/cpp/language/namespace_alias -- looks like there's some copyright restrictions -- maybe we should change the test case? https://en.cppreference.com/w/Cppreference:Copyright/CC-BY-SA

michelleangela: Moving these test assets into an internal repro so we don't to publish it.

PR #4538, opened 2019-11-05
Fix squiggle config settings

michelleangela:
Issue:

Could not enable error squiggles in c_cpp_properties.json unless setting was set in both c_cpp_properties.json and settings.json.

Fix:
Fix logic that checks settings in c_cpp_properties.json and settings.json. Setting in c_cpp_properties.json takes precedence if it is not undefined.


Colengms (approved):

michelleangela:
Fixed in version 0.26.2-insiders, available at https://github.com/microsoft/vscode-cpptools/releases/tag/0.26.2-insiders.

PR #4546, opened 2019-11-06
Add optional identifier validation to RenameProvider

Colengms:
Added isValidIdentifier which follows the rules defined [here](https://en.cppreference.com/w/cpp/language/identifiers) for identifier character validation.  (But does not verify things such as use of reserved keywords, etc..  Do we need that?)  We might actually want to adopt this on the native side, as it would seem more rigorous than the is_upper/is_lower check currently being used there.

Adds a setting to enable/disable identifier validation.


Extension/package.json
```diff
@@ -562,6 +562,12 @@
           "default": true,
           "markdownDescription": "%c_cpp.configuration.vcpkg.enabled.markdownDescription%",
           "scope": "resource"
+        },
+        "C_Cpp.renameRequiresIdentifier": {
+          "type": "boolean",
+          "default": true,
+          "description": "%c_cpp.configuration.renameRequiresIdentifier.description%",
+          "scope": "resource"
```
sean-mcmanus: I think this should this be "application" scope (code that passes in a scope would need to change too).

Extension/src/LanguageServer/client.ts
```diff
@@ -697,6 +697,12 @@ export class DefaultClient implements Client {
                             this.client = client;
                         }
                         public async provideRenameEdits(document: vscode.TextDocument, position: vscode.Position, newName: string, token: vscode.CancellationToken): Promise<vscode.WorkspaceEdit> {
+                            let settings: CppSettings = new CppSettings(this.client.RootUri);
+                            if (settings.renameRequiresIdentifier && !util.isValidIdentifier(newName)) {
+                                vscode.window.showErrorMessage(localize("invalid.identifier.for.rename", "Invalid identifier provided for Rename Symbol operation"));
+                                let workspaceEdit: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
```
sean-mcmanus, line 703: Should this be "for the Rename Symbol operation."?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #4570, opened 2019-11-07
Fix issue with English schema files not being found

Colengms:


Extension/src/LanguageServer/extension.ts
```diff
@@ -474,12 +474,11 @@ function realActivation(): void {
             let fileName: string = uri.fsPath;
             let locale: string = util.getLocaleId();
             let localizedFilePath: string = util.getExtensionFilePath(path.join("dist/schema/", locale, fileName));
```
bobbrow, line 477: To be consistent we should probably do the substr here too.  Let's also protect ourselves and only do it if the string starts with a slash.

bobbrow (approved):

PR #4585, opened 2019-11-11
enable breakpoints set in cu and cuh files

trxcllnt:
This PR adds a `"language": "cuda"` entry to package.json breakpoints list to allow setting breakpoints for the `vscode-cpptools` plugin in `.cu` and `.cuh` files.

This new entry shouldn't have any effect on regular users, but should enable breakpoints in `.cu` and `.cuh` files for users with the [`vscode-cudacpp`](https://marketplace.visualstudio.com/items?itemName=kriegalex.vscode-cudacpp) plugin installed. That plugin contributes a CUDA language definition (with `"languageId": "cuda"`), and handles some of the cuda-specific syntax like the `<<< >>>` kernel launch parameters.

This PR opts for the minimum changes necessary to get these two plugins working together. An alternate approach is to include the `vscode-cudacpp` CUDA language definitions in `vscode-cpptools` itself, which would enable `vscode-cpptools` to support CUDA out of the box.

In my testing, debugging CUDA is largely already supported by adding the following to any `"type": "cppdbg"` launch.json debugging configuration:
```
"MIMode": "gdb",
"miDebuggerPath": "/usr/local/cuda/bin/cuda-gdb",
```

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1516,6 +1516,9 @@
       },
       {
         "language": "cpp"
+      },
+      {
+        "language": "cuda"
       }
     ],
     "jsonValidation": [


sean-mcmanus (approved):

WardenGnaw (approved):

pieandcakes:
@trxcllnt Thanks for your PR. We have had some users asking questions about how to configure CUDA debugging. is the `vscode-cudacpp` extension a good place for them to go and ask questions? I don't have experience with CUDA debugging at all.

trxcllnt:
@pieandcakes I can't say one way or the other. `vscode-cudacpp` doesn't seem actively maintained, but it does still work for CUDA 10.1 syntax highlighting.

I use `vscode-cudacpp` for syntax highlighting, `vscode-clangd` (and `clangd-10`) for C/C++/CUDA intellisense, and `vscode-cpptools` for debugging. `vscode-clangd` and `vscode-cpptools` intellisense tend to clash, so I disable `vscode-cpptools` intellisense with these options:
```
// Place this in either `settings.json` or `some_workspace.vscode-workspace`
        "C_Cpp.formatting": "Disabled",
        "C_Cpp.autocomplete": "Disabled",
        "C_Cpp.errorSquiggles": "Disabled",
        "C_Cpp.intelliSenseEngine": "Disabled",
        "C_Cpp.configurationWarnings": "Disabled",
        "C_Cpp.autoAddFileAssociations": false,
        "C_Cpp.vcpkg.enabled": false,
```

After that, I have a custom build step to generate a `.vscode/launch.json` file that looks like this:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug C++ and CUDA tests",
            "type": "cppdbg",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "envFile": "${workspaceFolder}/.env",
            "program": "${workspaceFolder}/build/gtests/${input:TEST_NAME}",
            "MIMode": "gdb", "miDebuggerPath": "/usr/local/cuda/bin/cuda-gdb",
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ]
        },
    ],
    "inputs": [
        {
            "id": "TEST_NAME",
            "type": "pickString",
            "description": "Please select a test to run",
            "options": ["TEST_1","TEST_2","TEST_3","TEST_4"]
        }
    ]
}
```

Having this as part of the build means the test names are automatically inserted into the `"inputs"` list, so clicking the debugger button opens up the dropdown to select a test, rather than having to manually type each test name by hand.

Hope this helps,
Paul

vakokako:
Hey @trxcllnt, idk if it's a right place for it, but could you share how you use clangd with cuda files, since cmake doesn't provide compile_commands.json for .cu files?

trxcllnt:
@friendnick CMake should include .cu files in the compilation database if you do `enable_language(CUDA)` (or add `LANGUAGES CUDA` to your `project()` declaration).

vakokako:
@trxcllnt do you tweak cuda compiler setting in cmake to use clang then? 

by default cmake uses nvcc and then clangd fails to translate that to clang, giving `"Language no recognized: 'cu'" error`

trxcllnt:
@friendnick ah right. No, we use gcc and nvcc in CMake (with the `-GNinja` cmake generator), but we have to massage the `compile_commands.json` output after the configure step to translate a few of the nvcc-specific flags to ones that clang understands.

Here's a quick run-down of the setup:

* First, I recommend using the clangd dev releases (currently `clangd-11`) from [apt.llvm.org](http://apt.llvm.org/).
  Here's the TLRD [install command](https://github.com/trxcllnt/ubuntu-setup/blob/f195d33ac73e1040cafbc2688fb342dad2e3767d/scripts/06-vscode.sh#L8-L15) for Debian/Ubuntu.
* See the `cpp/CMakeLists.txt` in any of the [RAPIDS](http://rapids.ai/) [repositories](https://github.com/rapidsai) for general guidance on our CMake and CUDA configurations ([cuDF's](https://github.com/rapidsai/cudf/blob/branch-0.14/cpp/CMakeLists.txt) is the most complex).
* [Here's the bash function](https://github.com/trxcllnt/rapids-compose/blob/f2e3451928f9b856c741136f26035ecc59d39290/etc/bash-utils.sh#L852-L931) of `sed` commands we run post-configure to fix `compile_commands.json`. CMake prior to 3.17 has a bug with the Ninja generator where it put the entire `$PRE_LINK && /usr/bin/g++ ...` command into `compile_commands.json`, but this script corrects that too.

That bash script also has some other things in it you may find useful like [CCaching for nvcc](https://github.com/trxcllnt/rapids-compose/blob/f2e3451928f9b856c741136f26035ecc59d39290/etc/bash-utils.sh#L747-L753). You have to use CMake>=3.17.0 and ccache 4.0 (currently only available if built from master), then set the compiler symlinks manually (unfortunately using `update-alternatives` to switch to ccache doesn't work for nvcc like it does for gcc/g++).

PR #4621, opened 2019-11-16
Address issue with TU not being created for file already open when activated

Colengms:
Basically, this restores the original didOpen handling if the file is already visible, in which case we won't necessarily receive a call to onDidChangeVisibleTextEditors().

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -572,7 +572,6 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
                 if (clients.checkOwnership(client, editor.document)) {
                     if (!client.TrackedDocuments.has(editor.document)) {
                         // If not yet tracked, process as a newly opened file.  (didOpen is sent to server in client.takeOwnership()).
-                        client.TrackedDocuments.add(editor.document);
                         // Work around vscode treating ".C" as c, by adding this file name to file associations as cpp
                         if (editor.document.uri.path.endsWith(".C") && editor.document.languageId === "c") {
                             let cppSettings: CppSettings = new CppSettings(client.RootUri);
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -20,11 +23,37 @@ export function createProtocolFilter(me: Client, clients: ClientCollection): Mid
...
     return {
-        didOpen: (_document, _sendMessage) => {
-            // NO-OP
-            // didOpen handling has been moved to onDidChangeVisibleTextEditors in extension.ts.
-            // A file is only loaded when it is actually opened in the editor (not in response to
-            // control-hover, which sends a didOpen), and first becomes visible.
+        didOpen: (document, sendMessage) => {
+            let editor: vscode.TextEditor = vscode.window.visibleTextEditors.find(e => e.document === document);
+            if (editor) {
+                // If the file was is visible editor when we were activated, we will not get a call to
```
michelleangela: Fix wording? "If the file was is visible editor when we were activated"

Extension/src/LanguageServer/extension.ts
```diff
@@ -572,7 +572,6 @@ function onDidChangeVisibleTextEditors(editors: vscode.TextEditor[]): void {
                 if (clients.checkOwnership(client, editor.document)) {
                     if (!client.TrackedDocuments.has(editor.document)) {
                         // If not yet tracked, process as a newly opened file.  (didOpen is sent to server in client.takeOwnership()).
```
michelleangela: Remove comment for deleted statement?

Colengms: The comment is still correct, and is in the context of the IF before it.  The line that was removed was redundant, as the same thing happens in takeOwnership().  However, because that happens asynchronously (after notifyWhenReady), I should put this back to ensure there isn't a race condition here that could cause us to try to open the file twice.

sean-mcmanus (approved):

michelleangela (approved):

PR #4658, opened 2019-11-25
Update common.ts

### LINKED ISSUE #4643: Extension doesn't run if not owned by user
Type: General

**Describe the bug**
- OS and Version: Redhat Linux 7.5
- VS Code Version: 1.40.0
- C/C++ Extension Version: 0.26.1
- Other extensions you installed: 
  - python 2019.10.44104
  - xml 0.9.1
  - yaml 0.5.3
The issue persists after disabling them.
- Description:
  - I'm attempting to install code and the extensions listed above in an automated manner on new VMs. The idea is that a developer can spin up a new VM and have code and the extensions already installed on the VM (thus they never install an extension). I went by the docs for [Portable Mode](https://code.visualstudio.com/docs/editor/portable) to create the `/usr/share/code/data/` area. Code and the extensions get installed via ansible on boot and the group and file permissions are updated after installing (so the files are `root:swdev`).  Code launches fine and all of the extensions work except this one, `cpptools`.  There is no completion, syntax highlighting, and etc... for c++ and c files.

**To Reproduce**
Steps to reproduce the behavior (assuming cpptools is not yet installed):
  ```bash
  sudo /usr/bin/code --extensions-dir /usr/share/code/data/extensions --install-extension cpptools-0.26.1-linux.vsix
  sudo chgrp swdev -R /usr/share/code/data/extensions
  sudo chmod -R u+rwx,g+rwx,o+rX /usr/share/code/data/extensions
  /usr/bin/code 
  ```

**Expected behavior**
Code launches and the cpptools extension loads; providing completion, syntax highlighting, and etc... for c++ and c files.

**Additional context**
- I've created a launch.json with the following:
  ```json
  {
     "logging": {
        "trace": true,
        "traceResponse": true,
        "engineLogging": true
     }
  }
  ```
  In order to get the following log messages from the output window:
  > The installation of the C/C++ extension failed. Please see the output window for more information

- Reinstalling the the extension as the user via `/usr/bin/code --extensions-dir /usr/share/code/data/extensions --install-extension cpptools-0.26.1-linux.vsix` resolves the issue of the extension not working as desired.  So this works for us as a workaround for the time being.
- The user can also run `sudo chown -R $USER /usr/share/code/data/extensions/ms-vscode.cpptools-0.26.1` to resolve the issue.
- Running a diff between the `root:swdev` and the `$USER:swdev` folder `ms-vscode.cpptools-0.26.1` shows the only diff to be:
  - `ms-vscode.cpptools-0.26.1/package.json` 
    - The `activationEvents` in the `root` owned only lists `"*"`
    - The `activationEvents` in the `$USER` owned lists `"onLanguage:cpp",` ....and a few more.

dx-coding:
This change resolves #4643 and only attempts to update the file permissions if they are not already executable.


Extension/src/common.ts
```diff
@@ -632,11 +632,23 @@ export function spawnChildProcess(process: string, args: string[], workingDirect
...
+        });
+    });
+}
+
 export function allowExecution(file: string): Promise<void> {
     return new Promise<void>((resolve, reject) => {
         if (process.platform !== 'win32') {
             checkFileExists(file).then((exists: boolean) => {
-                if (exists) {
+                if (exists && !isExecutable(file)) {
```
Colengms: Since isExecutable() returns a Promise, isn't this "!" checking whether that Promise is null?  I think you'll want something like this:

```
                if (exists) {
                    isExecutable(file).then((isExec: boolean) => {
                        if (isExec) {
                            resolve();
                        } else {
                            fs.chmod(file, '755', (err: NodeJS.ErrnoException) => {
                                if (err) {
                                    reject(err);
                                    return;
                                }
                                resolve();
                            });
                        }
                    });
                } else {
```

It looks like our native binaries are not executable by default in our offline vsix.  (I'll look at fixing that).  Are you setting them as executable manually?

Colengms (changes requested):

Extension/src/common.ts
dx-coding: Good point on the Promise, first time looking at JavaScript/Typescript.  Updating, testing and will commit if all appears well.

And yes, in my case I am setting the binaries to executable.

dx-coding: Tested and updated

sean-mcmanus (approved):

Extension/src/common.ts
sean-mcmanus: Thanks

Colengms (approved):

Issue #4649, opened 2019-11-22
Configuring config provider does not update TU's of open header files


Colengms:
To repro this with cmake-tools,
1. Configure cmake-tools (...)
2. Open a header file. (Use `C/C++: Log Diagnostics` to see how it's configured)
3. Change CMake kit.
4. Run `CMake: Configure` or `CMake: Delete Cache and Reconfigure`.
5. Wait for cpptools to catch up
6. Use `C/C++: Log Diagnostics` to inspect the TU again.

Seeing: Although it was refreshed, it is not using the correct configuration

The cause: In response to didChangeCustomConfiguration(), updateCustomConfigurations() uses trackedDocuments to build of list of files to request new custom configurations for.  However, trackedDocuments is a list of open files, not the source files used to generate TU for headers.  The configuration provider may not provide configurations for headers.

It looks like we need to either keep track of the source file previously selected to configure the TU for the header and re-request a config for it.  Or, use a QueryTranslationUnitSourceRequest again to get the ranked list of sources to try to get custom configurations for (or call provideCustomConfiguration() from updateCustomConfigurations() to do the same).  The later seems best, as we might get different results from the config provider for the ranked sources.

Issue #3985, opened 2019-07-26
cpptools does not ask the configuration provider for the configuration of headers files


Mossop:
**Type: LanguageService**

**Describe the bug**
- OS and Version: OSX 10.14.5
- VS Code Version: 1.37.0-insider Commit: 6d82826e990407364d5c829b73602a19ecdf7d00

- C/C++ Extension Version: 0.24.1
- Other extensions you installed (and if the issue persists after disabling them): https://github.com/FractalBrew/vscode-mozillacpp
- A clear and concise description of what the bug is.

I have registered a custom configuration and it is correctly servicing requests for the configuration for .cpp files. However when I have a header file open I see no requests for the configuration for the header file and then the notification appears:

```
'Mozilla' is unable to provide IntelliSense configuration information for '/Users/dave/mozilla/source/trunk/dom/events/KeyEventHandler.h'. Settings from the 'Mac' configuration will be used instead.
```

**To Reproduce**

1. Have a source tree for Firefox built (sorry)
2. Install the referenced extension and select it as the configuration provider.
3. Set the extension's log level to "always" in settings.
4. Open a header file.
5. See error notification

**Expected behavior**
Expected to see the configuration provider called to ask for configuration for the header file.

**Screenshots**

**Additional context**

bobbrow:
Since many providers were unable to provide configs for headers, we decided to try to find the matching .cpp file and query your provider for that instead.  If your extension is capable of providing configs for headers, we could consider adding an API to disable this logic in cpptools.

I'm not exactly sure whether this is just a bug and a bad error message though. My suspicion is that cpptools **thinks** that some foo.cpp includes KeyEventHandler.h and when it asks you for the config for foo.cpp, your extension says it doesn't have a config for that.

Mossop:
Ah ok. I think what is actually going on is that cpptools is failing to find a cpp file for the header.

If I open vscode, clear the intellisense database and then open KeyEventHandler.h then the error notification is immediately shown, my configuration provider receives no request for configuration for any files.

If I open vscode, clear the intellisense database, open KeyEventHandler.cpp and let it finish parsing that then when I open KeyEventHandler.h then no error is shown and intellisense seems to work. Perhaps worth noting that my configuration provider doesn't receive any additional request for configuration for any files after the initial one for KeyEventHandler.cpp, presumably because it is cached at this point.

bobbrow:
[...]

PR #4671, opened 2019-11-27
Fix issue in which source file selection for header TU's does not occur when reconfiguring custom config provider

Colengms:
Also fixes an issue in which settings are not refreshed after a configuration provider has changed.

Also updated provideCustomConfiguration to use this.configurationProvider instead of looking it up again based on the setting..

queueTaskWithTimeout was removed to avoid an error, as it's no longer called.

This (and a corresponding PR against the native processes) addresses https://github.com/microsoft/vscode-cpptools/issues/4649 .  They should also address https://github.com/microsoft/vscode-cpptools/issues/3985


sean-mcmanus (approved):

sean-mcmanus:
This is for 0.26.3-insiders, right?

PR #4690, opened 2019-12-03
Update package.json for 0.26.3

michelleangela:
- updated version
- updated links for binaries


Colengms (commented):
Why are the linkid's changed in master but not release?  Did wrong (not updated) links go out in release?  Until master diverges significantly enough to require new offline binaries for tests to pass, shouldn't master be using the same linkid's as release?

michelleangela:
> Why are the linkid's changed in master but not release? Did wrong (not updated) links go out in release? Until master diverges significantly enough to require new offline binaries for tests to pass, shouldn't master be using the same linkid's as release?

These new linkids are for the next release but point to the same binaries are the current release. The current release has the correct linkids.

Colengms (approved):

PR #4820, opened 2020-01-07
Fix integration tests timeouts and events

michelleangela:
**Issues:**

1. Updating IntelliSense may take more than the test timeout value on macOS. 
2. The returned IntelliSense status for tests may not be synchronized with the document that is being tested. For example, a document `ABC.cpp` was opened, and then a second document `123.cpp` is opened and the test is waiting to verify IntelliSense status of `123.cpp`, but receives the status for `ABC.cpp` and uses this status to verify `123.cpp` which results in an incorrect verification.
3. Extra instance of extension gets created when checking for cppProperty changes in tests, that is when `getActiveClient()` is called in test code.

**Fix:**
1. Increase the timeout value of test to account for the slow IntelliSense update time on macOS.
2. Update the testHook API to include the name of the active document to the IntelliSense status (see pull request (https://github.com/microsoft/vscode-cpptools-api/pull/23). This will allow correct verification of the IntelliSense status for an active document that is in test.
3. Remove code creates extra instance of extension. This code is not needed to check for cppProperty changes, but sleep timer is needed for file watcher.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1762,7 +1762,7 @@
     "mkdirp": "^0.5.1",
     "plist": "^3.0.1",
     "tmp": "^0.1.0",
-    "vscode-cpptools": "^3.0.1",
+    "vscode-cpptools": "^3.1.0",
     "vscode-debugadapter": "^1.35.0",
     "vscode-debugprotocol": "^1.35.0",
     "vscode-extension-telemetry": "^0.1.2",
--- a/Extension/src/testHook.ts
+++ b/Extension/src/testHook.ts
@@ -4,25 +4,33 @@
  * ------------------------------------------------------------------------------------------ */
 'use strict';
 
-import { CppToolsTestHook, IntelliSenseStatus } from 'vscode-cpptools/out/testApi';
+import { CppToolsTestHook, Status, IntelliSenseStatus } from 'vscode-cpptools/out/testApi';
 import * as vscode from 'vscode';
 
 export class TestHook implements CppToolsTestHook {
-    private statusChangedEvent: vscode.EventEmitter<IntelliSenseStatus> = new vscode.EventEmitter<IntelliSenseStatus>();
+    private intelliSenseStatusChangedEvent: vscode.EventEmitter<IntelliSenseStatus> = new vscode.EventEmitter<IntelliSenseStatus>();
+    private statusChangedEvent: vscode.EventEmitter<Status> = new vscode.EventEmitter<Status>();
 
-    public get StatusChanged(): vscode.Event<IntelliSenseStatus> {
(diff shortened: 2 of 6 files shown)


sean-mcmanus (approved):

Extension/package.json
```diff
@@ -1762,7 +1762,7 @@
     "mkdirp": "^0.5.1",
     "plist": "^3.0.1",
     "tmp": "^0.1.0",
-    "vscode-cpptools": "^3.0.1",
+    "vscode-cpptools": "^3.1.0",
```
Colengms, line 1765: Shouldn't this cause an update to yarn.lock also?

FYI, the API change was a breaking change.  The "^" implies "use at least this version, up until the next major version number".  So, anyone already using "^3.x.x" (which is probably the common approach) will receive the new version.  Can we withdraw 3.1.0 and republish as 4.0?

Colengms (changes requested):

Extension/package.json
michelleangela: Since there is a new event in [CppToolsTestHook](https://github.com/microsoft/vscode-cpptools-api/blob/ca992e4637cf5f0e576d9bb721fea792e97935d4/testApi.ts#L37), any class that implements it and depends on "^3.x.x" will need to use the new event as well. 

If the c/c++ extension is the only extension that's using `CppToolsTestHook`, then I think it's okay to keep the API as 3.1.0 and update the c/c++ extension's dependency to "^3.1.0". However, to not break those using the CppToolsTestHook API, then the new event should be published under 4.0. 

@bobbrow, thoughts?

michelleangela: I'll update yarn.lock when we decide on what version of API to publish and use.

michelleangela: Discussed offline. Only c/c++ extension is using `CppToolsTestHook`, so only need to update dependency to "^3.1.0".

Colengms (approved):

Extension/src/testHook.ts
```diff
@@ -4,25 +4,33 @@
...
+
     public get valid(): boolean {
-        return !!this.statusChangedEvent;
+        return !!this.intelliSenseStatusChangedEvent;
     }
 
-    public updateStatus(status: Status): void {
-        this.statusChangedEvent.fire(status);
+    public updateStatus(status: IntelliSenseStatus): void {
+        this.intelliSenseStatusChangedEvent.fire(status);
```
bobbrow, line 28: @michelleangela 
Note, that you need to fire both events here or else the old one will be totally broken.  I don't know if anyone took a dependency on it or not, but it's best not to break it until at least the next major version.

Issue #4802, opened 2019-12-31
Some files in the database get unnecessarily removed on startup with a configuration provider


sean-mcmanus:
Using CMake Tools, I'm seeing around 2k files being removed (unnecessarily) from the database on startup and then reparsed. Reproed with 0.26.2 and 0.26.3-insiders2. Might be a regression.

Colengms:
This is because we haven't heard from the custom configuration provider yet.  Other issues related to this scenario that resulted in incorrectly configured TU's have been addressed be ensuring the TU is corrected when the new configuration arrives.  The next step is to give the provider some time to start up before falling back to the cpptools configuration.

sean-mcmanus:
@Colengms I still see this reproing with 0.26.3-insiders3 and our latest code -- from the logging it looks like the correct browse.path is getting used, but the parse_root is operating on a copy of old data.

PR #4846, opened 2020-01-14
Fix issue with didOpen processed before initial custom browse config

Colengms:
Should address incomplete fix to #4802 

Updates sendCustomBrowseConfiguration to avoid using notifyWhenReady(), in order to ensure the 'last seen' custom browse configuration is delivered on start up before any pending didOpen is processed.  The other places this is called are already within a notifyWhenReady().

Fixes an extension ID check to consider  `vector-of-bool.cmake-tools` and `ms-vscode.cmake-tools` to be equivalent.

Removed a no longer used arg and related code from notifyWhenReady.


sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1941,30 +1937,27 @@ export class DefaultClient implements Client {
...
-            let newProvider: string = this.configuration.CurrentConfigurationProvider;
-            if (!isSameProviderExtensionId(newProvider, this.configurationProvider)) {
-                this.configurationProvider = newProvider;
-                this.updateCustomBrowseConfiguration();
-                this.updateCustomConfigurations();
-            }
-        });
+            this.doneInitialCustomBrowseConfigurationCheck = true;
+        }
+        this.languageClient.sendNotification(ChangeFolderSettingsNotification, params);
```
bobbrow, line 1962: notifyWhenReady ensures that the languageClient is in a state that it can be used.  What mechanism are you putting in place to ensure that you don't crash here?

bobbrow (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: It wasn't crashing when I tested it out...was my testing insufficient?

FYI, this fixes the 100% repro of the database clearing with a file open, but I was getting a 2nd harder to repro race condition case still.

Colengms: The code in question (sending the initial folder settings) needs to happen before the language client can be used (particularly didOpen needs to be delayed), so this needs to precede anything else that was queued using notifyWhenReady().  When this is called for the first time, it's triggered by the setting of CompilerDefaults, which occurs in a continuation after a prior call to languageClient.sendRequest() (roughly line 817).  That would seem to imply it's OK to use languageClient to send messages at that point.   All other calls that could wind up here appear to be protected by notifyWhenReady(), or are otherwise only possible if the client is ready.  Please let me know if you can identify any cases in which we could wind up here when the client is not ready.  Is the client ready at the point in realActivation() where we install callbacks?

bobbrow: The code has changed a lot since I wrote it, so if you're confident there is no race here, then you can move forward with it.  I'm just voicing concerns based on past problems we experienced with crashing while trying to send messages before the language client was ready.  As a result, I tried to wrap everything with the *whenReady functions.  If this is safe, please add a comment here stating why a *whenReady isn't needed.  I used to do periodic checks for `sendNotification` calls outside of the *whenReady context.

Colengms: I have traced all existing calls back to either use of notifyWhenReady, or other evidence that it should be safe (such occurring near the end of the client constructor, where we have an existing call to `languageClient.sendRequest`.  I added use of notifyWhenReady to `onDidChangeSettings`, as it might be needed there.

There is one remaining code path that I'm having trouble with.  In configurations.ts, buildVcpkgIncludePath() is invoking this.handleConfigurationChange() in a finally block.  buildVcpkgIncludePath() is called from the constructor of CppProperties.  It's unclear to me if it's possible it might (as it's async) trigger a onConfigurationsChanged somewhere in the middle of the onReady() block in the client constructor.  But, since all of that is non-async and makes use of languageClient.sendRequest(), it might actually be safe.

Issue #3569, opened 2019-04-30
Prefer the clang-format in PATH over the one bundled with the extension


sean-mcmanus:
The right thing to do is to pick the clang-format in PATH over the VSCode one.

_Originally posted by @ncihnegn in https://github.com/Microsoft/vscode-cpptools/issues/2887#issuecomment-487843841_

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/0.26.3 .

PR #4855, opened 2020-01-16
Look for clang-format installed on system before defaulting to bundled version

Colengms:
Addresses https://github.com/microsoft/vscode-cpptools/issues/3569


Extension/src/LanguageServer/settings.ts
```diff
@@ -30,7 +31,14 @@ export class CppSettings extends Settings {
         super("C_Cpp", resource);
     }
 
-    public get clangFormatPath(): string { return super.Section.get<string>("clang_format_path"); }
+    public get clangFormatPath(): string {
+        let path: string = super.Section.get<string>("clang_format_path");
+        if (!path) {
+            path = which.sync('clang-format', {nothrow: true});
```
michelleangela, line 37: Is it possible to get the bundled version without taking a dependency on `which`?

michelleangela (changes requested):

Extension/src/LanguageServer/settings.ts
Colengms: I'm not sure I follow.  Why not use `which`?

Colengms: Bob suggested `which` specifically.  :)  https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/223355

michelleangela: I was wondering if there was a way to reduce dependencies if possible. But if it's easier to use module `which`, then I agree.

michelleangela (approved):

sean-mcmanus (changes requested):
Does it fallback to the shipped clang-format if it's not found on the path? Also, I think we should have a way to specify to "use the clang-format shipped with the C/C++ extension", because a hardcoded path won't work because the version will change every new release. Maybe some `{default}` or `{bundled}` variable or the empty string `""`? Also, maybe the description of the clang_format_path should be updated?

Colengms:
> Does it fallback to the shipped clang-format if it's not found on the path? Also, I think we should have a way to specify to "use the clang-format shipped with the C/C++ extension", because a hardcoded path won't work because the version will change every new release. Maybe some `{default}` or `{bundled}` variable? Also, maybe the description of the clang_format_path should be updated?

I'm not convinced that falling back to the bundled version if someone has incorrectly configured their path, would be the right thing to do, as it would silently fail to use the setting they provided.  IMHO, it would be better to fail.  Also, I'm not convinced a new setting to override the behavior makes sense.  I think using the version installed on the system should be preferred, and our version only used as a fallback.  If the clang-format-path setting is not set, the path to the executable will be queried with `which` on launch.  The only hard-coded path involved would be if the user specified one, which I think should always take precedence.

Colengms:
> Does it fallback to the shipped clang-format if it's not found on the path?

Just to make sure I understand your comment correctly; I had assumed you meant if not found on a user-specified path.  My understanding is that `which` would not return a path unless the executable was 'found' on that path.

sean-mcmanus:
> I'm not convinced that falling back to the bundled version if someone has incorrectly configured their path, would be the right thing to do, as it would silently fail to use the setting they provided.  IMHO, it would be better to fail.  Also, I'm not convinced a new setting to override the behavior makes sense.  I think using the version installed on the system should be preferred, and our version only used as a fallback.  If the clang-format-path setting is not set, the path to the executable will be queried with which on launch.  The only hard-coded path involved would be if the user specified one, which I think should always take precedence.

If they don't have any clang_format_path set, shouldn't it fall back to the bundled one? I think that's what your code does currently. How does the user overwrite the defaults (PATH version) to get the bundled one if they want it?

sean-mcmanus:
> Does it fallback to the shipped clang-format if it's not found on the path?
> 
> Just to make sure I understand your comment correctly; I had assumed you meant if not found on a user-specified path.  My understanding is that which would not return a path unless the executable was found on that path.

Yeah, I wasn't talking about an explicitly set clang_format_path, I was talking about the PATH found from "which".

sean-mcmanus (approved):

sean-mcmanus (changes requested):
Can you update the description?

sean-mcmanus:
If this is just TypeScript changes can it go into 0.26.3-insiders4?

Colengms:
> If this is just TypeScript changes can it go into 0.26.3-insiders4?

Yes.  I'll update the changelog.

sean-mcmanus (approved):

Extension/src/LanguageServer/settings.ts
```diff
@@ -30,7 +31,14 @@ export class CppSettings extends Settings {
         super("C_Cpp", resource);
     }
 
-    public get clangFormatPath(): string { return super.Section.get<string>("clang_format_path"); }
+    public get clangFormatPath(): string {
```
bobbrow, line 34: This will fix the initial configuration sent to the extension, but will it work if someone changes the setting and then clears it again? I don't believe the settings notifications that come from VS Code and go to the language server call into this function.

bobbrow: In other words, you might need to communicate the clangFormatPath that comes from $PATH down to the language server.

sean-mcmanus: Code just needs to be added near where it sends the UpdateClangFormatPathNotification in the Typescript.

sean-mcmanus: Oh, no, it looks fine -- the update code calls into settings.clangFormatPath.

bobbrow: I forgot we already had a notification for this.  Thanks for checking!

PR #4874, opened 2020-01-21
Merge changelog for 0.26.3.

sean-mcmanus:


sean-mcmanus:
I'm waiting for a new ThirdPartyNotices.txt for "which" and its "isext" dependency, which I'll probably add to this PR.

Colengms (approved):

sean-mcmanus:
> I'm waiting for a new ThirdPartyNotices.txt for "which" and its "isext" dependency, which I'll probably add to this PR.

Looks like the generator is broken/buggy...maybe we can ship anyway and then update it later?

Issue #4870, opened 2020-01-20
possibility to set cpptools.activeConfigName from a different extension


aliaksei-siarhei:
Type: General

I'm trying to make an extension for VS Code and possibility to control activeConfigName would be highly useful.

So far I only found how to bring "Select configuration" dropdown on the screen:
`vscode.commands.executeCommand('C_Cpp.ConfigurationSelect')`

Is there any hack/possibility to select configuration without user interaction?

sean-mcmanus:
The ConfigurationSelect command doesn't take an argument, so we'd need to add a command that takes an argument for the configuration to select (cpptools.activeConfigName is read only). I don't know of a workaround.

aliaksei-siarhei:
@sean-mcmanus , thanks for the info!
I've submitted a PR to add `cpptools.setActiveConfigName` command. Please take a look at #4893

sean-mcmanus:
cpptools.setActiveConfigName is available in our Insiders release: https://github.com/microsoft/vscode-cpptools/releases/tag/0.27.0-insiders

aliaksei-siarhei:
awesome! Works great)
Thanks for the help, will close the issue now.

PR #4893, opened 2020-01-24
add support for a new command: cpptools.setActiveConfigName

aliaksei-siarhei:
relates to #4870 

New command allows other extensions to control **activeConfigName**.
Can be used like in a code snippet below:
```typescript
async function func() {
  await vscode.commands.executeCommand("cpptools.setActiveConfigName", "config_A");
}
```
---
I can also update the documentation but need help to figure out an appropriate place for this.


Extension/src/LanguageServer/client.ts
```diff
@@ -1395,6 +1396,20 @@ export class DefaultClient implements Client {
...
+    public setCurrentConfigName(configurationName: string): Thenable<void> {
+        return this.queueTask(() => new Promise((resolve, reject) => {
+            let configurations : configs.Configuration[] = this.configuration.Configurations || [];
+            let configurationIndex : number = configurations.findIndex((config) => config.name === configurationName);
+
+            if (configurationIndex !== -1) {
+                this.configuration.select(configurationIndex);
+                resolve();
+            } else {
+                reject(new Error(`requested configuration name is not found: ${configurationName}`));
```
Colengms: To get this string localized, could you wrap this string in a call to localize() ?  The first string passed to localize() just needs to be a unique identifier (maybe "config.not.found"?).  With localize(), you'll need to use "{0}" and pass in configurationName as an additional argument, instead of using string interpolation.

Colengms (changes requested):

Colengms:
> I can also update the documentation but need help to figure out an appropriate place for this.

Could you add the command to `package.json`, in the `commands` section?  There is a variable replacement pattern you'll see used there (i.e. `%c_cpp.command.configurationSelect.title%`).  Those refer to entries in `package.nls.json`.  Could you add a new entry there?  Only an English entry is needed.  Other language files will be automatically generated by our Localization team.

That would be for the title.  Our docs are [here](https://github.com/microsoft/vscode-docs/tree/master/docs/cpp), but I'm not sure if we have verbose doc for commands.  (@sean-mcmanus Do we have doc for commands?)

sean-mcmanus:
> I can also update the documentation but need help to figure out an appropriate place for this.
> 
> Could you add the command to package.json, in the commands section?  There is a variable replacement pattern you'll see used there (i.e. %c_cpp.command.configurationSelect.title%).  Those refer to entries in package.nls.json.  Could you add a new entry there?  Only an English entry is needed.  Other language files will be automatically generated by our Localization team.
> That would be for the title.  Our docs are here, but I'm not sure if we have verbose doc for commands.  (@sean-mcmanus Do we have doc for commands?)

We never added any docs for the "cpptools.activeConfigName" command either. Maybe we could just add an issue to track that.

Extension/src/LanguageServer/client.ts
aliaksei-siarhei: ack. will do via `localize` wrapper.

aliaksei-siarhei:
> > I can also update the documentation but need help to figure out an appropriate place for this.
> 
> Could you add the command to `package.json`, in the `commands` section? There is a variable replacement pattern you'll see used there (i.e. `%c_cpp.command.configurationSelect.title%`). Those refer to entries in `package.nls.json`. Could you add a new entry there? Only an English entry is needed. Other language files will be automatically generated by our Localization team.
> 
> That would be for the title. Our docs are [here](https://github.com/microsoft/vscode-docs/tree/master/docs/cpp), but I'm not sure if we have verbose doc for commands. (@sean-mcmanus Do we have doc for commands?)

@Colengms , I encounted these problems while trying to implement your suggestion:
- `c_cpp.command.configurationSelect.title` already exists and is used to choose a configuration **via UI dropdown**. Adding another 'configuration set/select' command **might confuse users of cpptools** extension.
- I tried to add c_cpp.command.configurationSelect**ByName**.title... Unluckily didn't find a way to inform vscode that command requires user to provide an argument. As result, attempt to execute 
new command fails with "The requested configuration not found: undefined" message.

---
After some thinking can suggest these alternatives:
- Accept current implementation. `cpptools.activeConfigName` and `cpptools.setActiveConfigName` stay outside of **commands** in package.json. Documentation will be created via a separate issue as suggested by @sean-mcmanus 

- instead of introducing new command, we might try to modify behavior of already existing `onSelectConfiguration(...)`.  Function might accept an optional parameter (e.g. configurationName) and show the UI dropdown only if parameter is empty.
  **Cons:** existing `onSelectConfiguration(...)` doesn't return a Promise; refactoring is necessary to make things working properly. Implementation would differ much from other UI-related onSomething handlers.

- make `cpptools.activeConfigName` and `cpptools.setActiveConfigName` part of [vscode-cpptools-api](https://github.com/microsoft/vscode-cpptools-api) . Since the goal of PR is to allow other extensions to control currentConfigName, public API might be a good candidate.

sean-mcmanus:
> I can also update the documentation but need help to figure out an appropriate place for this.
> 
> Could you add the command to package.json, in the commands section?  There is a variable replacement pattern you'll see used there (i.e. %c_cpp.command.configurationSelect.title%).  Those refer to entries in package.nls.json.  Could you add a new entry there?  Only an English entry is needed.  Other language files will be automatically generated by our Localization team.
> That would be for the title.  Our docs are here, but I'm not sure if we have verbose doc for commands.  (@sean-mcmanus Do we have doc for commands?)

I don't think package.json needs to be changed to add the command there -- those are just for commands that appear in the command prompt or UI.

sean-mcmanus (approved):

Colengms (approved):

michelleangela (approved):

PR #4911, opened 2020-01-30
Validate dependencies for lldb-mi

WardenGnaw:
If a user does not provide a custom lldb-mi on macOS running in LLDB
mode. We check to see if the XCode/XCode CLI LLDB.framework dependencies
exist.

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -230,9 +230,67 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
                     config.pipeTransport.pipeProgram = replacedPipeProgram;
                 }
             }
+
+            const macOSMIMode: string = config.osx?.MIMode ?? config.MIMode;
+            const macOSMIDebuggerPath: string = config.osx?.miDebuggerPath ?? config.miDebuggerPath;
+
+            const lldb_mi_10_x_path: string = path.join(util.extensionPath, "debugAdapters", "lldb-mi", "bin", "lldb-mi");
+
+            // Validate LLDB-MI
+            if (os.platform() === 'darwin' && // Check for macOS
+                fs.existsSync(lldb_mi_10_x_path) && // lldb-mi 10.x exists
+                (!macOSMIMode || macOSMIMode === 'lldb') &&
+                !macOSMIDebuggerPath // User did not provide custom lldb-mi
+            ) {
+                const frameworkPath: string = this.getLLDBFrameworkPath();
+
+                if (!frameworkPath) {
+                    const moreInfoButton: string = localize("lldb.framework.install.xcode", "More Info");
+                    const LLDBFrameworkMissingMessage: string = localize("lldb.framework.not.found", "Unable to locate 'LLDB.framework' for lldb-mi. Please install XCode or XCode Command Line Tools.");
+
+                    vscode.window.showErrorMessage(LLDBFrameworkMissingMessage, moreInfoButton)
+                        .then(value => {
+                            if (value === moreInfoButton) {
+                                let helpURL: string = "https://aka.ms/vscode-cpptools/LLDBFrameworkNotFound";
+                                vscode.env.openExternal(vscode.Uri.parse(helpURL));
+                            }
+                        });
+
+                    return undefined;
+                }
+            }
         }
         // if config or type is not specified, return null to trigger VS Code to open a configuration file https://github.com/Microsoft/vscode/issues/54213
         return config && config.type ? config : null;
+    }
+
+    private getLLDBFrameworkPath(): string {
+        const LLDBFramework: string = "LLDB.framework";
+        // Note: When adding more search paths, make sure the shipped lldb-mi also has it. See Build/lldb-mi.yml and 'install_name_tool' commands.
+        let searchPaths: string[] = [
+            "/Library/Developer/CommandLineTools/Library/PrivateFrameworks", // XCode CLI
+            "/Applications/Xcode.app/Contents/SharedFrameworks" // App Store XCode
+        ];
+
+        for (const searchPath of searchPaths) {
+            if (fs.existsSync(path.join(searchPath, LLDBFramework))) {
+                // Found a framework that 'lldb-mi' can use.
+                return searchPath;
+            }
+        }
+
+        const outputChannel: logger.Logger = logger.getOutputChannelLogger();
+
+        outputChannel.appendLine(localize("lldb.find.failed", "Missing dependency '{0}' for lldb-mi executable.", LLDBFramework));
+        outputChannel.appendLine(localize("lldb.search.paths", "Searched in:"));
+        searchPaths.forEach(searchPath => {
+            outputChannel.appendLine(`\t${searchPath}`);
+        });
+        const xcodeCLIInstallCmd: string = "xcode-select --install";
+        outputChannel.appendLine(localize("lldb.install.help", "To resolve this issue, either install XCode through the Apple App Store or install the XCode Command Line Tools by running '{0}' in a Terminal window.", xcodeCLIInstallCmd));
+        logger.showOutputChannel();
+
+        return null;
     }
 
     private resolveEnvFile(config: vscode.DebugConfiguration, folder: vscode.WorkspaceFolder): void {
--- a/Extension/src/packageManager.ts
+++ b/Extension/src/packageManager.ts
@@ -36,6 +36,13 @@ export interface IPackage {
     // Architectures for which the package is applicable
     architectures: string[];
 
+    // OS Version regex to check if package is applicable
+    versionRegex: string;
+
+    // A flag to indicate if 'versionRegex' should match or not match.
+    // Required if versionRegex is used. Default is false.
+    matchVersion: boolean;
+
     // Binaries in the package that should be executable when deployed
     binaries: string[];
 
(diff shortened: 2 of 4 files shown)


WardenGnaw:
Draft mode to prevent merging until we have a new LLDB-MI package.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+                !this.hasLLDBFramework()
+                ) {
+
+                const moreInfoMessage: string = localize("lldb.framework.more.info", "More Info");
+                const LLDBFrameworkMissingMessage: string = localize("ldlb.framework.not.found", "Missing 'LLDB.framework' for lldb-mi.");
+
+                vscode.window.showErrorMessage(LLDBFrameworkMissingMessage, moreInfoMessage)
+                 .then(value => {
+                    if (value === moreInfoMessage) {
+                        let helpURL: string = "https://aka.ms/VSCode-CPP-LLDBFrameworkNotFoundHelp";
```
pieandcakes: can we go with https://aka.ms/vscode-cpptools/LLDBFrameworkNotFound

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+
+            // Validate LLDB-MI
+            if (os.platform() === 'darwin' && // Check for macOS
+                (!macOSMIMode || macOSMIMode === 'lldb') &&
+                !macOSMIDebuggerPath && // User did not provide custom lldb-mi
+                !this.hasLLDBFramework()
+                ) {
+
+                const moreInfoMessage: string = localize("lldb.framework.more.info", "More Info");
+                const LLDBFrameworkMissingMessage: string = localize("ldlb.framework.not.found", "Missing 'LLDB.framework' for lldb-mi.");
```
pieandcakes: `Unable to locate...`

pieandcakes: also `ldlb`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
         // if config or type is not specified, return null to trigger VS Code to open a configuration file https://github.com/Microsoft/vscode/issues/54213
         return config && config.type ? config : null;
     }
 
+    private hasLLDBFramework(): boolean {
+        const LLDBFramework: string = "LLDB.framework";
+        let searchPaths: string[] = [
+            "/Library/Developer/CommandLineTools/Library/PrivateFrameworks", // XCode CLI
+            "/Applications/Xcode.app/Contents/SharedFrameworks" // App Store XCode
+        ];
```
pieandcakes, line 273: What if the user has it next to lldb-mi in the extensions folder or maps it? should we also check there just in case?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+        }
+
+        const outputChannel: logger.Logger = logger.getOutputChannelLogger();
+
+        outputChannel.appendLine(localize("lldb.find.failed", "Missing dependency '{0}' for lldb-mi executable.", LLDBFramework));
+        outputChannel.appendLine(localize("lldb.search.paths", "Searched in:"));
+        searchPaths.forEach(searchPath => {
+            outputChannel.appendLine(`\t${searchPath}`);
+        });
+        outputChannel.appendLine(localize("lldb.install.help", "To resolve this issue, please either install XCode through the App Store or install XCode CLI with 'xcode-select --install'."));
```
pieandcakes: `To resolve this issue, either install XCode through the Apple App Store or install the XCode Command Line Tools by running 'xcode-select --install' in a Terminal window.`

pieandcakes (approved):

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: Then they would point to their own custom `lldb-mi` and does not go through this function.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+
+            // Validate LLDB-MI
+            if (os.platform() === 'darwin' && // Check for macOS
+                (!macOSMIMode || macOSMIMode === 'lldb') &&
+                !macOSMIDebuggerPath && // User did not provide custom lldb-mi
+                !this.hasLLDBFramework()
+                ) {
+
+                const moreInfoMessage: string = localize("lldb.framework.more.info", "More Info");
+                const LLDBFrameworkMissingMessage: string = localize("lldb.framework.not.found", "Unable to locate 'LLDB.framework' for lldb-mi.");
```
bobbrow: I feel like there should be a bit more information in this message.  I think users are trained (at least I am) to be disappointed in "More Info" promises and I'm worried there's not enough incentive in this message to convince them to click the button before giving up or opening an issue.

Maybe adding something like "Please install XCode or the XCode command line tools." would be enough bait.

bobbrow: Oh, it looks like you are also logging to the output window.  Perhaps redirecting them there would be good here instead.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,61 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+        }
+
+        const outputChannel: logger.Logger = logger.getOutputChannelLogger();
+
+        outputChannel.appendLine(localize("lldb.find.failed", "Missing dependency '{0}' for lldb-mi executable.", LLDBFramework));
+        outputChannel.appendLine(localize("lldb.search.paths", "Searched in:"));
+        searchPaths.forEach(searchPath => {
+            outputChannel.appendLine(`\t${searchPath}`);
+        });
+        outputChannel.appendLine(localize("lldb.install.help", "To resolve this issue, either install XCode through the Apple App Store or install the XCode Command Line Tools by running 'xcode-select --install' in a Terminal window."));
```
bobbrow: The localizers might be able to determine that 'xcode-select --install' is a command and should not be translated.  To be sure, you could put {0} here and pass in the string for substitution.

Extension/src/packageManager.ts
```diff
@@ -36,6 +36,13 @@ export interface IPackage {
     // Architectures for which the package is applicable
     architectures: string[];
 
+    // OS Version regex to check if package is applicable
+    versionRegex: string;
+
+    // a flag to indicate if 'versionRegex' should match or not match.
+    // Required if versionRegex is used. Default is false.
```
pieandcakes, line 43: nit: `A`

Extension/src/packageManager.ts
```diff
@@ -150,7 +157,12 @@ export class PackageManager {
             .then((list) => {
                 return list.filter((value, index, array) => {
                     return (!value.architectures || value.architectures.indexOf(this.platformInfo.architecture) !== -1) &&
-                        (!value.platforms || value.platforms.indexOf(this.platformInfo.platform) !== -1);
+                        (!value.platforms || value.platforms.indexOf(this.platformInfo.platform) !== -1) &&
+                        (!this.platformInfo.version || !value.versionRegex || (
```
pieandcakes: I think this statement isn't right. Right now it is treating a package with a versionRegex and a match that is required but no platform version to allow that download so for some reasonif you weren't able to get a platform.version you end up with both packages? we may have to rethink what we want to happen as i don't believe that we'd want both. An assert or an output message + choosing one of the paths to go down might be better.

Extension/src/platform.ts
```diff
@@ -69,4 +73,11 @@ export class PlatformInformation {
                 return null;
             });
     }
+
+    private static GetDarwinVersion(): Promise<string> {
+        const DARWIN_SYSTEM_VERSION_PLIST: string = "/System/Library/CoreServices/SystemVersion.plist";
+        const systemVersionPListBuffer: Buffer = fs.readFileSync(DARWIN_SYSTEM_VERSION_PLIST);
```
pieandcakes: Should you do a file check to 100% make sure that file is there and if it isn't write an error out? at least then its more visible as a "problem"

pieandcakes (approved):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,69 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
                     config.pipeTransport.pipeProgram = replacedPipeProgram;
                 }
             }
+
+            const macOSMIMode: string = config.osx?.MIMode ?? config.MIMode;
+            const macOSMIDebuggerPath: string = config.osx?.miDebuggerPath ?? config.miDebuggerPath;
+
+            const lldb_10_x_path: string = path.join(util.extensionPath, "debugAdapters", "lldb-mi", "bin", "lldb-mi");
```
pieandcakes: nit: lldb-mi ? its not a path to lldb

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -231,11 +231,69 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+            // Validate LLDB-MI
+            if (os.platform() === 'darwin' && // Check for macOS
+                fs.existsSync(lldb_10_x_path) && // lldb-mi 10.x exists
+                (!macOSMIMode || macOSMIMode === 'lldb') &&
+                !macOSMIDebuggerPath // User did not provide custom lldb-mi
+            ) {
+                const frameworkPath: string = this.getLLDBFrameworkPath();
+
+                if (!frameworkPath) {
+                    const installButton: string = localize("lldb.framework.install.xcode", "Installing XCode");
```
pieandcakes: @bobbrow This button doesn't install XCode, it takes you to a doc that tells you what the problem is and how the user can solve it. `More Info` or `More Details` would make more sense or something saying `read more` or `How to Install LLDB.Framework`

bobbrow (approved):
Looks good to me!

Issue #4879, opened 2020-01-22
"Compiler arguments" is not localized


sean-mcmanus:
It's missing the data-loc-id like the other cases have.

Issue #4727, opened 2019-12-10
Configuration UI disables compiler path if no default compilers are found


michelleangela:
The input field on the configuration UI should be enabled even if no compiler paths are found on user's machine so that user can specify a custom path.

PR #4912, opened 2020-01-30
Fix issue with disabled compiler path in config UI if none detected

Colengms:
Addresses #4727, #4879

Keeps the hidden span in the HTML file for the purpose of localization, but builds a new option to display in the drop-down.  (For some reason, it does not work to hide/show an option in the HTML file).


michelleangela:
For future and to make validation easier, would be nice to have screenshots of fix if there are UI changes.

michelleangela (approved):

sean-mcmanus (approved):

Issue #4963, opened 2020-02-13
Add a minimum-required version check before using system installed clang-format


anjiawei1991:
my clang-format's version is 3.4.2 and has no option of -fallback-style.  but the latest cpptools extension use it.  and I can hardly upgrade clang-format based on my enviroment.

Colengms:
Hi @anjiawei1991 .  Let's use this issue to track adding a minimum version check before trying to use the installed version of clang-format.  In the meanwhile, try  setting `clang_format_path` to the clang-format binary bundled with the extension:

`"C_Cpp.clang_format_path": "<path_to_the_CppTools_extension>\\LLVM\\bin\\clang-format.exe"`

sean-mcmanus:
Setting the hardcoded "<path_to_the_cpptools>" will break after a new version of the extension gets installed. Should we add a setting or variable that enables using the extension's clang-format or do we think a "minimum version" setting is sufficient?

sean-mcmanus:
Fixed in our latest Insiders release: https://github.com/microsoft/vscode-cpptools/releases/tag/0.27.0-insiders

PR #4968, opened 2020-02-14
Use system installed clang-format only if newer, or bundled version is not executable

Colengms:
If a path to clang-format is configured explicitly, we'll use it.  Otherwise, we will check for one on the system.  If we find one on the system, we will:
    - Try to invoke our own, to see if we can successfully do so, and to query its version.
        - If we fail to invoke our own, use the one found on the system.
        - If we successfully execute our own, try to invoke the one found on the system as well, to compare versions.
            - If it fails to execute, use our own.
            - If both are executable, use the one with the greater version.

This should address: https://github.com/microsoft/vscode-cpptools/issues/4963

### DIFF
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -73,7 +73,7 @@ export class CppSettings extends Settings {
                 try {
                     let exePath: string = getExtensionFilePath("./LLVM/bin/" + this.clangFormatName);
                     let output: string[] = execSync(exePath + " --version").toString().split(" ");
-                    if (output.length < 3 || output[0] != "clang-format" || output[1] != "version" || !semver.valid(output[2])) {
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/settings.ts
```diff
@@ -47,10 +51,47 @@ export class CppSettings extends Settings {
...
+                    }
+                    clangFormatVersion = output[2];
+                } catch (e) {
+                    // Unable to invoke our own clang-format.  Use the system installed clang-format.
+                    return path;
+                }
+
+                // Invoke the version on the system to compare versions.  Use ours if it's more recent.
+                try {
+                    let output: string[] = execSync("\"" + path + "\" --version").toString().split(" ");
```
sean-mcmanus: I think the preferred style is 
```
`"${path}" --version`
```
unless it's a single string concatenation.

sean-mcmanus (changes requested):
Can you update the package.json version to 0.27.0-master?

sean-mcmanus (approved):

PR #4986, opened 2020-02-19
Get a stable yarn lock

Colengms:
Yarn.lock was unstable (was changing when running yarn install).


Colengms:
Commiting directly, as there are no code changes.

PR #5010, opened 2020-02-26
Fix config switching after adding folder.

sean-mcmanus:


Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -163,7 +163,9 @@ export class ClientCollection {
                 let path: string = util.asFolder(folder.uri);
                 let client: cpptools.Client = this.languageClients.get(path);
                 if (!client) {
-                    this.languageClients.set(path, cpptools.createClient(this, folder));
+                    let newClient: cpptools.Client = cpptools.createClient(this, folder);
```
michelleangela, line 166: Is there a way to create client without needing to deactivate it in the case it doesn't need to be activated at creation? Would it be better to:
1. create client
2. initialize it
3. active it only when needed?

Or will this not work for single folder scenario?

michelleangela (approved):

Extension/src/LanguageServer/clientCollection.ts
sean-mcmanus: Yeah, it could probably be changed to not be activated on creation, but I haven't found a good enough reason to change that yet. Deactivatinging clients that shouldn't be activated on creation seems okay so far...seemed less likely to regress the existing code.

PR #5030, opened 2020-02-29
Enable strictNullChecks

Colengms:
Also changes all string settings (except for compilerPath) such that edit fields will be used for them in settings UI, instead of just a link to "Edit settings.json".  This also means that setting to blank strings is the same as setting to null, or removing the setting.  `compilerPath` is an exception, as a blank string has a special meaning.  We will correct this by adding a separate bypass setting when other settings rename work is done.


Extension/src/LanguageServer/configurations.ts
```diff
@@ -53,7 +53,7 @@ export interface ConfigurationJson {
 }
 
 export interface Configuration {
-    name: string;
+    name?: string;
```
bobbrow: This field is required.  We should figure out where in the code is not expecting a name and fix it.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -318,7 +328,7 @@ export class CppProperties {
 
     private get ExtendedEnvironment(): Environment {
         let result: Environment = {};
-        if (this.configurationJson.env) {
+        if (this.configurationJson && this.configurationJson.env) {
```
sean-mcmanus: ?.env

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -14,18 +14,18 @@ import { onDidChangeActiveTextEditor, processDelayedDidOpen } from './extension'
 
 export function createProtocolFilter(clients: ClientCollection): Middleware {
     // Disabling lint for invoke handlers
-    let defaultHandler: (data: any, callback: (data: any) => void) => void = (data, callback: (data) => void) => { clients.ActiveClient.notifyWhenReady(() => callback(data)); };
+    let defaultHandler: (data: any, callback: (data: any) => void) => void = (data, callback: (data: any) => void) => { clients.ActiveClient.notifyWhenReady(() => callback(data)); };
     /* tslint:disable */
```
sean-mcmanus: Is this still needed?

Extension/src/LanguageServer/languageConfig.ts
```diff
@@ -302,23 +302,57 @@ export function getLanguageConfigFromPatterns(languageId: string, patterns: (str
 
 function constructCommentRules(comment: CommentPattern, languageId: string): Rules {
     if (comment && comment.begin && comment.begin.startsWith('/*') && (languageId === 'c' || languageId === 'cpp')) {
```
sean-mcmanus: comment?.begin?.startsWith

Extension/src/LanguageServer/configurations.ts
```diff
@@ -151,21 +150,25 @@ export class CppProperties {
...
     public get CurrentConfigurationIndex(): number { return this.currentConfigurationIndex.Value; }
-    public get CurrentConfiguration(): Configuration { return this.Configurations ? this.Configurations[this.CurrentConfigurationIndex] : null; }
-    public get KnownCompiler(): KnownCompiler[] { return this.knownCompilers; }
+    public get CurrentConfiguration(): Configuration | undefined { return this.Configurations ? this.Configurations[this.CurrentConfigurationIndex] : undefined; }
+    public get KnownCompiler(): KnownCompiler[] | undefined { return this.knownCompilers; }
 
-    public get CurrentConfigurationProvider(): string | null {
-        if (this.CurrentConfiguration.configurationProvider) {
+    public get CurrentConfigurationProvider(): string | undefined {
+        if (this.CurrentConfiguration && this.CurrentConfiguration.configurationProvider) {
```
sean-mcmanus: Does this.CurrentConfiguration?.configurationProvider work?

Extension/src/LanguageServer/customProviders.ts
```diff
@@ -247,7 +247,7 @@ export function getCustomConfigProviders(): CustomConfigurationProviderCollectio
     return providerCollection;
 }
 
-export function isSameProviderExtensionId(settingExtensionId: string, providerExtensionId: string): boolean {
+export function isSameProviderExtensionId(settingExtensionId?: string, providerExtensionId?: string): boolean {
```
sean-mcmanus: Shouldn't 2 args always be passed? Would string | undefined be better?

sean-mcmanus (changes requested):
If we're going to ship 0.27.0-insiders on Monday or Tuesday, then maybe this should go into 0.27.0-insiders2?

Extension/src/LanguageServer/protocolFilter.ts
Colengms: Is what still needed?  defaultHandler?  It is still referenced below.

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -76,12 +80,10 @@ export class ClientCollection {
     /**
      * get a handle to a language client. returns null if the client was not found.
```
michelleangela: >null [](start = 50, length = 4)

update comments too? #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -952,6 +1014,7 @@ export class DefaultClient implements Client {
                         vscode.window.showErrorMessage(localize("unable.to.start", "Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: {0}", String(err)));
                     }
                 }));
+            this.colorizationSettings.reload();
```
michelleangela, line 1013: >this.colorizationSettings.reload(); [](start = 12, length = 35)

is this a bug fix? #Resolved

Colengms: I move this out of the constructor, so that it could be constructed earlier and stored without using `| undefined`. #Resolved

Extension/src/common.ts
```diff
@@ -86,24 +87,33 @@ export async function ensureBuildTaskExists(taskName: string): Promise<void> {
...
+    let selectedTask: vscode.Task | undefined = rawTasksJson.tasks.find((task: any) => task.label && task.label === task);
     if (selectedTask) {
         return;
     }
 
     const buildTasks: vscode.Task[] = await getBuildTasks(false);
     selectedTask = buildTasks.find(task => task.name === taskName);
     console.assert(selectedTask);
+    if (!selectedTask) {
+        throw new Error("Failed to get selectedTask in ensureBuildTaskExists()");
```
bobbrow, line 99: do these Error strings need to be localized?

Extension/src/common.ts
```diff
@@ -301,12 +329,15 @@ export function resolveCachePath(input: string, additionalEnvironment: {[key: st
     return resolvedPath;
 }
 
-export function resolveVariables(input: string, additionalEnvironment: {[key: string]: string | string[]}): string {
+export function resolveVariables(input: string | undefined, additionalEnvironment: undefined | {[key: string]: string | string[]}): string {
     if (!input) {
         return "";
     }
+    let additionalEnvironment2: {[key: string]: string | string[]};
```
bobbrow: I thought this shouldn't be necessary.  TypeScript should be able to promote away the undefined typing after it is assigned.  Does that not work?

Extension/src/common.ts
```diff
@@ -318,19 +349,19 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
...
             switch (varType) {
                 case "env": {
-                    let v: string | string[] = additionalEnvironment[name];
+                    let v: string | string[] = additionalEnvironment2[name];
                     if (isString(v)) {
                         newValue = v;
                     } else if (input === match && isArrayOfString(v)) {
                         newValue = v.join(";");
                     }
-                    if (!isString(newValue)) {
```
bobbrow, line 333: why is this changing?  `isString` should be able to remove the undefined typing.

Extension/src/common.ts
```diff
@@ -356,15 +387,15 @@ export function resolveVariables(input: string, additionalEnvironment: {[key: st
                 }
                 default: { assert.fail("unknown varType matched"); }
             }
-            return (isString(newValue)) ? newValue : match;
```
bobbrow, line 359: again, not sure why this one needs to change

Extension/src/common.ts
```diff
@@ -613,13 +644,19 @@ export function spawnChildProcess(process: string, args: string[], workingDirect
     return new Promise<void>(function (resolve, reject): void {
         const child: child_process.ChildProcess = child_process.spawn(process, args, { cwd: workingDirectory });
 
-        child.stdout.on('data', (data) => {
-            dataCallback(`${data}`);
-        });
+        let stdout: Readable | null = child.stdout;
```
bobbrow, line 643: It would be simpler to use `child.stdout?.on(...)`

Extension/src/common.ts
```diff
@@ -613,13 +644,19 @@ export function spawnChildProcess(process: string, args: string[], workingDirect
...
+        if (stdout) {
+            stdout.on('data', (data) => {
+                dataCallback(`${data}`);
+            });
+        }
 
-        child.stderr.on('data', (data) => {
-            errorCallback(`${data}`);
-        });
+        let stderr: Readable | null = child.stderr;
```
bobbrow, line 650: It would be simpler to use `child.stderr?.on(...)`

Extension/src/common.ts
```diff
@@ -797,13 +840,13 @@ export function downloadFileToStr(urlStr: string, headers?: OutgoingHttpHeaders)
 }
 
 export interface CompilerPathAndArgs {
-    compilerPath: string;
+    compilerPath: string | undefined;
```
bobbrow: rather than `| undefined`, use `compilerPath?`

Extension/src/common.ts
```diff
@@ -797,13 +840,13 @@ export function downloadFileToStr(urlStr: string, headers?: OutgoingHttpHeaders)
...
     compilerName: string;
     additionalArgs: string[];
 }
 
-export function extractCompilerPathAndArgs(inputCompilerPath: string, inputCompilerArgs?: string[]): CompilerPathAndArgs {
-    let compilerPath: string = inputCompilerPath;
+export function extractCompilerPathAndArgs(inputCompilerPath: string | undefined, inputCompilerArgs?: string[]): CompilerPathAndArgs {
```
bobbrow: rather than `| undefined`, use `inputCompilerPath?`

Extension/src/cppTools.ts
```diff
@@ -58,22 +58,25 @@ export class CppTools implements CppToolsTestApi {
     public registerCustomConfigurationProvider(provider: CustomConfigurationProvider): void {
         let providers: CustomConfigurationProviderCollection = getCustomConfigProviders();
         if (providers.add(provider, this.version)) {
-            let added: CustomConfigurationProvider1 = providers.get(provider);
-            let settings: CppSettings = new CppSettings();
-            if (settings.loggingLevel === "Information" || settings.loggingLevel === "Debug") {
-                getOutputChannel().appendLine(localize("provider.registered", "Custom configuration provider '{0}' registered", added.name));
+            let addedMayBeUndefined: CustomConfigurationProvider1 | undefined = providers.get(provider);
```
bobbrow: if the check above succeeds, shouldn't we be able to assert that this is non-null/undefined? (note the `!` at the end below)

```
let added: CustomConfigurationProvider1 = providers.get(provider)!;
```

Extension/src/githubAPI.ts
```diff
@@ -248,15 +248,18 @@ async function getRateLimit(): Promise<RateLimit> {
 }
 
 async function rateLimitExceeded(): Promise<boolean> {
-    const rateLimit: RateLimit = await getRateLimit();
-    return rateLimit && rateLimit.rate.remaining <= 0;
```
bobbrow, line 252: I don't see why this line needs to change at all.  Is there a bug with the boolean expression?

Extension/src/main.ts
```diff
@@ -204,7 +204,7 @@ function touchInstallLockFile(): Promise<void> {
 function handleError(error: any): void {
     let installationInformation: InstallationInformation = getInstallationInformation();
     installationInformation.hasError = true;
-    installationInformation.telemetryProperties['stage'] = installationInformation.stage;
+    installationInformation.telemetryProperties['stage'] = installationInformation.stage ? installationInformation.stage : "";
```
bobbrow: `installationInformation.stage || ""` is a short form of `foo ? foo : bar`

Extension/src/packageManager.ts
```diff
@@ -160,25 +161,23 @@ export class PackageManager {
     }
 
     private ArchitecturesMatch(value: IPackage): boolean {
-        return !value.architectures || value.architectures.indexOf(this.platformInfo.architecture) !== -1;
+        return !value.architectures || (this.platformInfo.architecture !== undefined && value.architectures.indexOf(this.platformInfo.architecture) !== -1);
```
bobbrow, line 164: Is there a reason to test against undefined specifically, or can you remove the `!== undefined` part?

bobbrow:
>         let c: CommentPattern = (typeof pattern === "string") ? { begin: pattern, continue: pattern.startsWith('/*') ? " * " : pattern } : <CommentPattern>pattern;

This should probably use the `isString` function from common.ts and it should also do proper checking that "pattern" is a `CommentPattern` now that you removed the checking everywhere else.

---
Refers to: Extension/src/LanguageServer/languageConfig.ts:239 in 6c5bace. [](commit_id = 6c5bace716c7ba5a0f120be77af05dc46070d25d, deletion_comment = False)

Extension/src/platform.ts
```diff
@@ -15,13 +15,13 @@ nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFo
...
+    constructor(public platform: string, public architecture?: string, public distribution?: LinuxDistribution, public version?: string) { }
 
     public static GetPlatformInformation(): Promise<PlatformInformation> {
         let platform: string = os.platform();
-        let architecturePromise: Promise<string>;
-        let distributionPromise: Promise<LinuxDistribution> = Promise.resolve<LinuxDistribution>(null);
-        let versionPromise: Promise<string> = Promise.resolve<string>(null);
+        let architecturePromise: Promise<string | undefined>;
+        let distributionPromise: Promise<LinuxDistribution | undefined> = Promise.resolve<LinuxDistribution | undefined>(undefined);
+        let versionPromise: Promise<string | undefined> = Promise.resolve<string | undefined>(undefined);
```
bobbrow: It looks like this one doesn't need `| undefined`

Extension/src/main.ts
sean-mcmanus: We should switch to using C# style `foo ?? bar`.

Extension/src/common.ts
Colengms: Unfortunately, the promotion does not extend into a lambda.  :/  Will move the promotion into the lambda.

Extension/src/common.ts
Colengms: newValue will only be either undefined, or a string.  Use of isString() here was superfluous.  Removed it for simplicity.

Extension/src/common.ts
Colengms: Same answer as above.  newValue can only be a string, or undefined.  So removed isString() for simplicity.

Extension/src/common.ts
Colengms: That trips lint for [no-unused-expressions](https://eslint.org/docs/rules/no-unused-expressions), for some reason.

Extension/src/common.ts
Colengms: That trips lint for [no-unused-expressions](https://eslint.org/docs/rules/no-unused-expressions), for some reason.

Extension/src/cppTools.ts
Colengms: That would work, if we didn't have `no-non-null-assertion` enabled.  The check here is only necessary due to use of a lambda in a forEach(), which loses its promotion.  It looks like using 'const' instead of 'let' is a way around that issue.

Extension/src/githubAPI.ts
Colengms: Yes.  After adding `| undefined` to the type, it does not like performing a check for the null/undefined and a boolean expression in the same statement.  The following:

```
return rateLimit && rateLimit.rate.remaining <= 0;
```

Results in:

```
Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.
```

And the following:

```
return rateLimit?.rate.remaining <= 0;
```

Results in:

```
Object is possibly 'undefined'.
```

As the result of the left hand side may be undefined, which cannot be compared to 0.

Extension/src/packageManager.ts
Colengms: Like referred to above, in rateLimitExceeded(), testing the potentially null/undefined value is not compatible with boolean operators, as it does not result in a boolean value.  This is actually a cleaner approach (to use the !== undefined to convert it to a boolean), so I will use this in rateLimitExceeded().

Colengms:
> This should probably use the isString function from common.ts and it should also do proper checking that "pattern" is a CommentPattern now that you removed the checking everywhere else.

I'm not sure I follow.  The type is `string | CommentPattern`.  So, if it's not a string, then it's a CommentPattern.  Isn't it sufficient to check for only one of those 2 types?  It's not null or undefined - that's the whole point of this PR.  :)

Extension/src/common.ts
Colengms: The only 'throw new Error's that I've added are for scenarios that should not occur, so shouldn't generally be diagnosable to the user.  My opinion is no, we shouldn't localize them.  In fact, the more of these that can be removed by designing around them, the better.

Extension/src/main.ts
```diff
@@ -261,7 +261,7 @@ function sendTelemetry(info: PlatformInformation): boolean {
         util.setProgress(util.getProgressInstallSuccess());
     }
 
-    installBlob.telemetryProperties['osArchitecture'] = info.architecture;
+    installBlob.telemetryProperties['osArchitecture'] = info.architecture ? info.architecture : "";
```
sean-mcmanus: info.architecture ?? ""

Extension/src/main.ts
```diff
@@ -204,7 +204,7 @@ function touchInstallLockFile(): Promise<void> {
 function handleError(error: any): void {
     let installationInformation: InstallationInformation = getInstallationInformation();
     installationInformation.hasError = true;
-    installationInformation.telemetryProperties['stage'] = installationInformation.stage;
+    installationInformation.telemetryProperties['stage'] = installationInformation.stage || "";
```
sean-mcmanus: ?? ""

sean-mcmanus (approved):

sean-mcmanus (approved):

Veetaha:
Wow, nice to see such a big cleanup, big thanks to Colengms!

PR #5053, opened 2020-03-04
Optionally enable backtick command substitution for pickRemoteProcess

### LINKED ISSUE #4015: {command:pickRemoteProcess} doesn't work if remote user's default shell is CShell
Type: Debugger
**Input information below**

**Please review existing issues and our documentation at https://github.com/Microsoft/vscode-cpptools/tree/master/Documentation prior to filing an issue.**

**Describe the bug**
- OS and Version: Linux (RHEL 7.6)
- VS Code Version: 1.36.1
- C/C++ Extension Version: 0.24.1
- Other extensions you installed (and if the issue persists after disabling them): It does
- The Illegal Variable Name error is displayed when trying to pick a remote process on a host which has CShell as the default shell. This seems to stem from the fact that bash is the assumed shell and $(uname) isn't valid in cshell. (the equivalent would be \`uname\`)

Unfortunately it's not possible to change the default shell on the remote host to bash, since other processes rely on the default shell being CShell.

**To Reproduce**
*Please include a code sample and `launch.json` configuration.*
1. Configure launch.json to connect to a remote host with csh as default shell
2. Attempt to connect
3. Error appears:
Command failed: "/usr/bin/ssh" "-i" "~/.ssh/private_key" "user@remote-host" 'sh -c "uname && if [ $(uname) = \"Linux\" ] ; then ps axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= ; elif [ $(uname) = \"Darwin\" ] ; then ps axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= -c; fi"'
Illegal variable name.

My launch.json (with some stuff genericised):
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "debug process remote_server",
            "type": "cppdbg",
            "request": "attach",
            "program": "${workspaceRoot}/debug64/bin/process",
            "processId": "${command:pickRemoteProcess}",
            "MIMode": "gdb",
            "pipeTransport": {
                "pipeCwd": "/usr/bin",
                "pipeProgram": "/usr/bin/ssh",
                "pipeArgs": ["-i", "~/.ssh/private_key", "user@remote_server"],
                "debuggerPath": "/usr/bin/gdb"
            },
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ]
        }
    ]
}
```

**Additional context**
*If applicable, please include logging by adding "logging": { "engineLogging": true, "trace": true, "traceResponse": true } in your `launch.json`*
Add any other context about the problem here including log or error messages in your Debug Console or Output windows.

Helloimbob:
This adds an option to use the obsolete backtick (`) command substitution in the debugger's pickRemoteProcess shell command for use with remote shells (such as csh) which don't support $() substitution. 

This fixes #4015


Extension/src/Debugger/attachToProcess.ts
```diff
@@ -108,6 +109,16 @@ export class RemoteAttachPicker {
     private getRemoteProcessCommand(): string {
         let innerQuote: string = `'`;
         let outerQuote: string = `"`;
+        let substitutionBegin: string = `$(`;
+        let substitutionEnd: string = `)`;
+        let escapedQuote: string = `\\\"`;
+
+        let settings: CppSettings = new CppSettings();
```
sean-mcmanus, line 116: The new setting is "resource" scoped (i.e. per-workspace folder), but no folder uri is passed to CppSettings(). Should the scope of the setting be "window" (i.e. workspace) or is there a way to get the current workspace folder uri?

sean-mcmanus: Seems like just making the setting "window" scope would be good enough.

sean-mcmanus (approved):
Seems fine to me, but someone from the debugger team needs to sign off too.

Extension/src/Debugger/attachToProcess.ts
pieandcakes: This should be per workspace. Is that `window` scope?

Extension/package.json
```diff
@@ -568,6 +568,12 @@
           "default": true,
           "description": "%c_cpp.configuration.renameRequiresIdentifier.description%",
           "scope": "application"
+        },
+        "C_Cpp.useBacktickCommandSubstitution": {
+          "type": "boolean",
+          "default": false,
```
pieandcakes, line 545: @sean-mcmanus @bobbrow Are you ok with adding this to the c_cpp_properties settings? 

@WardenGnaw FYI

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -108,6 +109,16 @@ export class RemoteAttachPicker {
...
+        if (settings.useBacktickCommandSubstitution) {
+            substitutionBegin = `\``;
+            substitutionEnd = `\``;
+            escapedQuote = `\"`;
```
pieandcakes, line 120: Does backtick substitution mean that we don't need to escape the `\` symbol?

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -108,6 +109,16 @@ export class RemoteAttachPicker {
     private getRemoteProcessCommand(): string {
         let innerQuote: string = `'`;
         let outerQuote: string = `"`;
+        let substitutionBegin: string = `$(`;
```
pieandcakes: rename as `parameterBegin` and `parameterEnd`

pieandcakes (approved):

Extension/src/Debugger/attachToProcess.ts
Helloimbob: Originally with the escaped `\` the command failed with an unmatched `"` error. I've just looked into this again and it seems to actually be a quirk in csh where stuff in single quotes doesn't get expanded: [Stack Overflow answer](https://stackoverflow.com/questions/10761838/getting-unmatched-on-a-simple-csh-ssh-script)

Extension/src/Debugger/attachToProcess.ts
sean-mcmanus: window scope is workspace scope (strange naming).

Extension/package.json
sean-mcmanus: It's not a c_cpp_properties.json setting. It's a settings.json. Where else would the setting go?

sean-mcmanus:
We should probably delay checking this in until 0.27.0-insiders3 or 0.27.0.

Extension/package.json
pieandcakes: ok. thats fine. Just checking. We don't usually use settings, so i wanted to make sure we aren't stepping on anyone.

sean-mcmanus:
@bobbrow @WardenGnaw  Should we name the setting something like C_Cpp.debugging.useBacktickCommandSubstitution or C_Cpp.debugger.useBacktickCommandSubstitution ? It seems like that would help avoid confusing it with the other settings that are mostly language service related.

sean-mcmanus:
@Helloimbob We just checked in a big "enable strictNullChecks" change, which is probably causing the merge conflict.

sean-mcmanus:
@WardenGnaw Can this be checked in for 0.27.0-insiders3? Should we rename the setting?

WardenGnaw:
I agree that it should be renamed to `C_Cpp.debugger.useBacktickCommandSubstitution`.

WardenGnaw (approved):

PR #5064, opened 2020-03-05
Adding the arm/arm64 packages for mono runtime

pieandcakes:
Adding the arm/arm64 packages for the mono runtime. This is to enable
remote-ssh extension debugging.


Extension/package.json
```diff
@@ -1954,5 +1980,8 @@
       ],
       "binaries": []
     }
-  ]
+  ],
+  "dependencies": {
+    "vsce": "^1.74.0"
```
sean-mcmanus: Why was this added?

sean-mcmanus: It seems like it should be removed (along with the yarn.lock changes?).

sean-mcmanus (changes requested):
...

WardenGnaw (approved):

sean-mcmanus (approved):

Extension/package.json
```diff
@@ -1882,6 +1882,32 @@
         "./debugAdapters/mono.linux-x86_64"
       ]
     },
+    {
+      "description": "Mono Runtime (Linux / armhf)",
+      "url": "https://aka.ms/mono-runtime-linux-arm",
```
bobbrow, line 1698: Should these be fwlinks instead of aka.ms links?  I don't have a strong preference, but I think it's easier to manage links when they're all on the same dashboard.

bobbrow (approved):
I'm ok to submit this whenever you're ready.  I just put one comment in the PR about the link type, but that's non-blocking.

Extension/package.json
pieandcakes: I was thinking it would be easier to be more descriptive with aka.ms links since then you know what it points to but i also didn't want to update the rest of them today.

Issue #5062, opened 2020-03-05
Add logging for file watcher events


bobbrow:
You might see some error in the Toggle Developer Tools console if VS Code hits a file watcher error.

You could also set your C_Cpp.loggingLevel to "Debug" to and look for a "fileChanged" logging message (but unfortunately we don't log the file that changed yet).

_Originally posted by @sean-mcmanus in https://github.com/microsoft/vscode-cpptools/issues/5052#issuecomment-595406679_

We should log the relevant file path along with the file watcher events.

PR #5084, opened 2020-03-09
Log file watcher events

Colengms:
Add log messages for file watcher events that identify the file.

Addresses: https://github.com/microsoft/vscode-cpptools/issues/5062

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1880,14 +1880,18 @@ export class DefaultClient implements Client {
                 false /* ignoreChangeEvents */,
                 false /* ignoreDeleteEvents */);
 
+            let settings: CppSettings = new CppSettings(this.RootUri);
             this.rootPathFileWatcher.onDidCreate((uri) => {
+                if (settings.loggingLevel === "Debug") {
+                    log(localize("file.creation.detected", "File creation detected: {0}", uri.toString()));
+                }
                 this.languageClient.sendNotification(FileCreatedNotification, { uri: uri.toString() });
             });
 
             // TODO: Handle new associations without a reload.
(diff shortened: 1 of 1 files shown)


michelleangela (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1880,14 +1880,18 @@ export class DefaultClient implements Client {
                 false /* ignoreChangeEvents */,
                 false /* ignoreDeleteEvents */);
 
+            let settings: CppSettings = new CppSettings(this.RootUri);
             this.rootPathFileWatcher.onDidCreate((uri) => {
+                if (settings.loggingLevel === "Debug") {
+                    log(localize("file.creation.detected", "File creation detected: {0}", uri.toString()));
```
bobbrow, line 1886: This will make two messages for debug now, right?

I thought we would add the message in the native code and append it to the "cpptools/fileCreated" message.  Alternately, we can filter out the old messages

Colengms: The original message is due to logging of _all_ method names received by the native process.  Adding logging of the file name seems to make more sense on the TypeScript side, as localization is done on the TypesScript side.  (Otherwise, I'll need to plumb up some new 'native strings').  My thought was to leave the per-method logging in place, as it also indicates that the message was successfully received on the native side.  Also, it seemed cleaner to leave the generic logging for each method name in place, rather than add special behavior to it for just these 3 messages.

sean-mcmanus: It should be easy to filter out the cpptools/fileCreated message. That might make it less cluttery after switching branches.

Colengms: Based on a conversation with Bob, I'm going to go a different route, and print out the uri on the native side, for any (LSP) method that has a uri associated with it.

Issue #5090, opened 2020-03-11
0.27.0-insiders2 IntelliSense crashes on Mac 10.12 or earlier after a settings change


sean-mcmanus:
Is anyone using our latest Insiders release able to provide any info on the crash of cpptools-srv that contains the call stack frame

```
dyld                          	dyld::fastBindLazySymbol(ImageLoader**, unsigned long) + 139
```

The crash log might be under `~/Library/Logs/DiagnosticReports` on Mac -- is there a "Dyld Error Message" in the log?

It seems like the Mac crash logger is not giving us the full crash frame or it's some how an abnormal crash.

sean-mcmanus:
It looks like our AutoPch code is incorrectly getting run for users of Mac 10.12 or older.

PR #5091, opened 2020-03-11
Add "Dyld Error Message" crash telemetry, and bypass system call stack frame filtering in that case.

sean-mcmanus:
In response to the issue at https://github.com/microsoft/vscode-cpptools/issues/5090

An example of the "Dyld Error Message" telemetry might be something like:
```
    Dyld Error Message:
    Symbol not found: _IOBluetoothPreferenceGetRemoteWakeEnabled
    Referenced from: /System/Library/PreferencePanes/Bluetooth.prefPane/Contents/MacOS/Bluetooth
    Expected in: /System/Library/Frameworks/IOBluetooth.framework/Versions/A/IOBluetooth
    
    Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
```


sean-mcmanus:
This change seems unnecessary now that the root cause is figured out...I'll abandon this for now....

Issue #5078, opened 2020-03-07
C_Cpp.intelliSenseCachePath description has an incorrect strikeout in the description


sean-mcmanus:
Usage of `~` appears to cause VS Code to render text as a strikeout:

![image](https://user-images.githubusercontent.com/19859882/76131573-d4b34000-5fc3-11ea-9161-2c157cd21a09.png)

 Ignore the extra `\` in the screenshot (I added that).

PR #5108, opened 2020-03-13
Remove use of tilde from setting descriptions

Colengms:
This should address: https://github.com/microsoft/vscode-cpptools/issues/5078

Tilde is a markdown for strike-out.  Switching to using $HOME instead.


sean-mcmanus (approved):

michelleangela (approved):

bobbrow:
I think we can also just escape the \~.  But I thought we didn't use the markdown descriptions.  Is this a vscode bug?

Colengms:
Curiously, there is no way to escape the tilde in markdown.  (Also, escaping would itself be specific to markdown, and should probably be avoided, so translators don't trip over it).  I found reports of people with the same issue on the web, and using $HOME instead seemed to be the preferred solution.

PR #5114, opened 2020-03-13
Fix error when renaming

Colengms:
For enum values, the "!" operator will indicate true for enum values that are 0, in additional to null or undefined.

Switch checks to use `=== undefined` instead.

### DIFF
--- a/Extension/src/LanguageServer/referencesTreeDataProvider.ts
+++ b/Extension/src/LanguageServer/referencesTreeDataProvider.ts
@@ -118,7 +118,7 @@ export class ReferencesTreeDataProvider implements vscode.TreeDataProvider<TreeN
                 return resultFile;
 
             case NodeType.reference:
-                if (!element.referenceText) {
+                if (element.referenceText === undefined) {
                     throw new Error("Null or undefined referenceText in getTreeItem()");
                 }
                 if (element.referenceType === undefined) {
--- a/Extension/src/LanguageServer/referencesTreeDataProvider.ts
+++ b/Extension/src/LanguageServer/referencesTreeDataProvider.ts
@@ -76,13 +76,13 @@ export class ReferencesTreeDataProvider implements vscode.TreeDataProvider<TreeN
 
     getTreeItem(element: TreeNode): vscode.TreeItem {
         if (this.referencesModel === undefined) {
-            throw new Error("Null or undefined RefrencesModel in getTreeItem()");
+            throw new Error("Undefined RefrencesModel in getTreeItem()");
         }
 
         switch (element.node) {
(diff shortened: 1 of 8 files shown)


Extension/src/LanguageServer/referencesTreeDataProvider.ts
```diff
@@ -121,7 +121,7 @@ export class ReferencesTreeDataProvider implements vscode.TreeDataProvider<TreeN
                 if (!element.referenceText) {
```
bobbrow: empty string will also evaluate to false. Is that something else we should be checking?

Extension/src/LanguageServer/referencesTreeDataProvider.ts
```diff
@@ -121,7 +121,7 @@ export class ReferencesTreeDataProvider implements vscode.TreeDataProvider<TreeN
...
                     throw new Error("Null or undefined referenceText in getTreeItem()");
                 }
-                if (!element.referenceType) {
+                if (element.referenceType === undefined) {
                     throw new Error("Null or undefined referenceType in getTreeItem()");
```
bobbrow: Can you assign null to a `?` property?  I think it can only be `undefined`

sean-mcmanus: Maybe we could enable https://palantir.github.io/tslint/rules/no-null-keyword/ later on.

sean-mcmanus (approved):

sean-mcmanus (changes requested):
Could you update the changelog with the new date?

Extension/src/LanguageServer/referencesTreeDataProvider.ts
Colengms: I don't think we have a lot of cases in which a blank string would be valid.  I addressed some potential cases.

Extension/src/LanguageServer/referencesTreeDataProvider.ts
Colengms: I corrected those error messages.  Hypothetically, it's possible for external code (i.e. VS Code APIs) to return null or undefined from APIs that do not indicate they might.  (i.e. for templates  like get<type> in settings, if we support a null setting and don't use get<type | null> when we call it, we may still get null back).

Colengms:
> Could you update the changelog with the new date?

Sure.  I'll update it with the new year, also.  :)

sean-mcmanus (approved):

sean-mcmanus:
> Could you update the changelog with the new date?
> 
> Sure.  I'll update it with the new year, also.  :)

Good catch!

PR #5116, opened 2020-03-13
Undo bad fix for check in getVcpkgRoot

Colengms:


sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -172,9 +172,9 @@ export function getVcpkgPathDescriptorFile(): string {
     }
 }
 
-let vcpkgRoot: string;
+let vcpkgRoot: string | undefined;
 export function getVcpkgRoot(): string {
-    if (!vcpkgRoot) {
+    if (!vcpkgRoot && vcpkgRoot !== "") {
```
sean-mcmanus, line 177: Yeah, I noticed that when I code reviewed...I was thinking about asking what that was for :)

Colengms: It looks like it's using a value of "" to indicate that it has attempted and failed, so does not attempt again.

PR #5126, opened 2020-03-16
Fix some settings so they are editable in settings UI

Colengms:
Fix some settings so they are editable in settings UI

Fixes the definitions of the following settings, such that they can be edited using the settings UI in VS Code.
* C_Cpp.default.enableConfigurationSquiggles
* C_Cpp.inactiveRegionForegroundColor
* C_Cpp.inactiveRegionBackgroundColor

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -483,11 +483,8 @@
           "scope": "resource"
         },
         "C_Cpp.default.enableConfigurationSquiggles": {
-          "type": [
-            "boolean",
-            "null"
-          ],
-          "default": null,
+          "type": "boolean",
+          "default": false,
           "description": "%c_cpp.configuration.default.enableConfigurationSquiggles.description%",
           "scope": "resource"
         },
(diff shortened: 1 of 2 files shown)


michelleangela (approved):

Extension/package.json
```diff
@@ -483,11 +475,8 @@
           "scope": "resource"
         },
         "C_Cpp.default.enableConfigurationSquiggles": {
-          "type": [
-            "boolean",
-            "null"
-          ],
-          "default": null,
+          "type": "boolean",
+          "default": false,
```
sean-mcmanus: I thought "null" was equivalent to false, i.e. shouldn't this setting be "true" by default?

sean-mcmanus (changes requested):

Extension/package.json
bobbrow: This should be `true`

Colengms: FYI, the prior behavior was for this to be null by default, which was equivalent to false.  I'll change the default to true.

sean-mcmanus (approved):

sean-mcmanus:
I believe this change also fixes https://github.com/microsoft/vscode-cpptools/issues/5131

Issue #4990, opened 2020-02-20
Switch to using new VS Code WorkspaceEditMetadata API for Rename


Colengms:
VS Code has added a new API for previewing refactor operations.  https://github.com/microsoft/vscode/issues/77728

 Creating this issue to track switching to using it instead of our own Rename UI.

elahehrashedi:
This issue is fixed in 0.28.0.
0.28.0 release: https://github.com/microsoft/vscode-cpptools/releases/tag/0.28.0

Issue #4131, opened 2019-08-23
Switch to vscode-test instead of vscode/lib/testrunner?


sean-mcmanus:
VS Code has deprecated our current TypeScript testing infrastructure and has introduced vscode-test: https://code.visualstudio.com/api/working-with-extensions/testing-extension .

PR #5133, opened 2020-03-16
Remove Rename UI in favor of new WorkspaceEdit Preview API

Colengms:
Removes code for Rename UI. #4990
Switches to using WorkspaceEdit Preview API for rename confirmations.

image

Note that there does not appear to be a way for us to control the sort order of groups (when viewing by group).  VS Code appears to intentionally put our group containing confirmed references last.

Note that the icons for reference type are displayed only on groups, when viewing by group.

Also addresses https://github.com/microsoft/vscode-cpptools/issues/4131 , as updating to the new VS Code API's require switching to `vscode-test`.


Colengms:
Still reworking the tests...

Extension/test/integrationTests/IntelliSenseFeatures/index.ts
```diff
@@ -1,42 +1,38 @@
...
+  // Create the mocha test
+  const mocha = new Mocha({
+    ui: 'tdd'
+  });
+  mocha.useColors(true);
+
+  const testsRoot = path.resolve(__dirname, '.');
+
+  return new Promise((c, e) => {
+    glob('**/**.test.js', { cwd: testsRoot }, (err, files) => {
```
bobbrow: nit: There seems to be inconsistent spacing in the tests.  I'm seeing 2-space and tab indented files.  Can you normalize them all to use 4-space indents?

Colengms: FYI, the test runner (new index.ts) and runTest.ts were boiler-plate files provided by the vscode-test migration documentation.  I could change them, but there would also be value in keeping them as similar to the boiler-plate files as possible.

sean-mcmanus:
Can we get VS Code to add some sort order so we can force the Confirmed Reference type to come first? Otherwise, that seems like a major UI regression, since the confirmed references are the most important ones and could otherwise be scrolled off the screen.

Colengms:
> Can we get VS Code to add some sort order so we can force the Confirmed Reference type to come first? Otherwise, that seems like a major UI regression, since the confirmed references are the most important ones and could otherwise be scrolled off the screen.

They appear to be intentionally putting confirmed groups last, so it may be by-design.  Perhaps the most important ones are the ones that need attention (unconfirmed), so perhaps this is better?  Since this is a shipped feature, perhaps open an issue in their repo?

sean-mcmanus:
> Can we get VS Code to add some sort order so we can force the Confirmed Reference type to come first? Otherwise, that seems like a major UI regression, since the confirmed references are the most important ones and could otherwise be scrolled off the screen.
> 
> They appear to be intentionally putting confirmed groups last, so it may be by-design.  Perhaps the most important ones are the ones that need attention (unconfirmed), so perhaps this is better?  Since this is a shipped feature, perhaps open an issue in their repo?

I filed issue https://github.com/microsoft/vscode/issues/94210

Extension/package.json
```diff
@@ -1719,21 +1527,22 @@
     "minimatch": "^3.0.4",
     "minimist": "^1.2.5",
     "mkdirp": "^0.5.1",
+    "mocha": "4",
```
sean-mcmanus: Why isn't this something like "^4.0.0"?

Extension/src/LanguageServer/referencesModel.ts
```diff
@@ -6,35 +6,19 @@
...
-        } else {
-            results = resultsInput.referenceInfos.filter(r => r.type !== ReferenceType.Confirmed);
-        }
+        let results: ReferenceInfo[] = resultsInput.referenceInfos.filter(r => r.type !== ReferenceType.Confirmed);
 
         // Build a single flat list of all leaf nodes
-        // Currently, the hierachy is build each time referencesTreeDataProvider requests nodes.
-        // When moving between pending and candidate views in rename, the hierachy gets rebuilt.
-        // TODO: If that is a performance issue, we could change this to always build the tree up front.
+        // Currently, the hierachy is built each time referencesTreeDataProvider requests nodes.
```
sean-mcmanus: hierarchy

Extension/src/LanguageServer/referencesModel.ts
```diff
@@ -78,15 +61,13 @@ export class ReferencesModel {
         return result;
     }
 
-    getFileNodes(refType: ReferenceType | undefined, isRenameCandidate: boolean): TreeNode[] {
+    getFileNodes(refType: ReferenceType | undefined): TreeNode[] {
```
sean-mcmanus: Can this be refType? now?

sean-mcmanus (changes requested):
...

sean-mcmanus (approved):

PR #5137, opened 2020-03-17
Fix open file scenario bugs due to struct null changes

Colengms:
Should address: https://github.com/microsoft/vscode-cpptools/issues/5136

Strict null checking changes resulted in an assumption that an undefined/null root uri was an issue.  It's, in fact, expected for the 'open file' scenario.

Also fixed a bug in resolveVariables that caused blank string values to be confused with undefined values.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -105,11 +105,11 @@ export interface CompilerDefaults {
 }
 
 export class CppProperties {
-    private rootUri: vscode.Uri;
+    private rootUri: vscode.Uri | undefined;
     private propertiesFile: vscode.Uri | undefined | null = undefined; // undefined and null values are handled differently
     private readonly configFolder: string;
     private configurationJson?: ConfigurationJson;
-    private currentConfigurationIndex: PersistentFolderState<number>;
+    private currentConfigurationIndex: PersistentFolderState<number> | undefined;
     private configFileWatcher: vscode.FileSystemWatcher | null = null;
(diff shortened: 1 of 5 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -137,10 +137,12 @@ export class CppProperties {
     // we want to track when the default includes have been added to it.
     private configurationIncomplete: boolean = true;
 
-    constructor(rootUri: vscode.Uri, workspaceFolder: vscode.WorkspaceFolder) {
+    constructor(rootUri: vscode.Uri | undefined, workspaceFolder: vscode.WorkspaceFolder | undefined) {
```
bobbrow, line 140: Do we usually use `?` instead of `| undefined`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -975,11 +987,13 @@ export class CppProperties {
 
         // first resolve variables
         result = util.resolveVariables(path, this.ExtendedEnvironment);
-        if (result.includes("${workspaceFolder}")) {
-            result = result.replace("${workspaceFolder}", this.rootUri.fsPath);
-        }
-        if (result.includes("${workspaceRoot}")) {
-            result = result.replace("${workspaceRoot}", this.rootUri.fsPath);
+        if (this.rootUri) {
+            if (result.includes("${workspaceFolder}")) {
```
bobbrow, line 991: For the future (not 0.27.0): `workspaceFolder` and `workspaceRoot` could be added to the `ExtendedEnvironment` for this resolve. (not sure if they should be added for all resolves)

PR #5140, opened 2020-03-17
Use optional arg in client constructor instead of | undefined

Colengms:
A minor style fix, based on Bob's feedback on https://github.com/microsoft/vscode-cpptools/pull/5137 after it was committed.


bobbrow:
I think there were other places where `| undefined` was used in a method signature.  I didn't mark them all.

Colengms:
> I think there were other places where | undefined was used in a method signature.  I didn't mark them all.

I checked.  I did not see any that were not intentional.  Note that the ? only works if it's the last arg, or all trailing args are using it.

sean-mcmanus (approved):

PR #5153, opened 2020-03-19
Add cleanUnusedBinaries stage

WardenGnaw:
Our offline installer will include two versions of lldb.
./debugAdapters/lldb is lldb-mi 3.8 and is used for High-Sierra and older
./debugAdapters/lldb-mi is lldb-mi 10.x and is used for Mojave and
higher.

This clean up the folder that is unused. 

Refactored package platform checks to be unified in one area.

### DIFF
--- a/Extension/src/packageManager.ts
+++ b/Extension/src/packageManager.ts
@@ -153,41 +153,11 @@ export class PackageManager {
         return this.GetPackageList()
             .then((list) =>
                 list.filter((value, index, array) =>
-                    this.ArchitecturesMatch(value) &&
-                        this.PlatformsMatch(value) &&
-                        this.VersionsMatch(value)
+                    ArchitecturesMatch(value, this.platformInfo) &&
+                        PlatformsMatch(value, this.platformInfo) &&
+                        VersionsMatch(value, this.platformInfo)
                 )
             );
-    }
-
-    private ArchitecturesMatch(value: IPackage): boolean {
-        return !value.architectures || (this.platformInfo.architecture !== undefined && value.architectures.indexOf(this.platformInfo.architecture) !== -1);
-    }
-
-    private PlatformsMatch(value: IPackage): boolean {
-        return !value.platforms || value.platforms.indexOf(this.platformInfo.platform) !== -1;
-    }
-
-    private VersionsMatch(value: IPackage): boolean {
-        if (value.versionRegex) {
-            // If we have a versionRegex but did not get a platformVersion
-            if (!this.platformInfo.version) {
-                // If we are expecting to match the versionRegex, return false since there was no version found.
-                //
-                // If we are expecting to not match the versionRegex, return true since we are expecting to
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
This just removes the binaries listed in the package.json but keeps both lldb and lldb-mi folders -- is that sufficient for the correct version to get picked up?

WardenGnaw:
Yes. It is good enough to just remove the binaries. We check the existence via full path.

https://github.com/microsoft/MIEngine/blob/50a932ddbf2f5a10e64aaaf9747a72accca4741e/src/OpenDebugAD7/MILaunchOptions.cs#L466

sean-mcmanus (approved):

Colengms (changes requested):
If the user is running an older version of the macOS, then upgrades, will we upgrade to the correct binaries?

sean-mcmanus:
> If the user is running an older version of the macOS, then upgrades, will we upgrade to the correct binaries?

No, but I think that should be easy to fix -- just add code that uses PersistentState<string> to save and check the macOS version -- if it has changed from High Sierrer or older to older, then run the onlineInstallation(), which I believe should detect the missing binaries and install them (and ignore the existing binaries that already exist). Or it could be just be a bool for whether it's an lldb versus lldb-mi OS.

Extension/src/packageManager.ts
```diff
@@ -467,3 +437,33 @@ export class PackageManager {
         }
     }
 }
+
+export function VersionsMatch(pkg: IPackage, info: PlatformInformation): boolean {
```
bobbrow, line 441: Functions should start with a lower case letter.  I know these were previously wrong.  Can you fix them while you're in here?

sean-mcmanus:
We could add a fix for the "update OS" scenario later on if it can't be done by 0.27.0-insiders4 (on Monday probably).

Extension/src/packageManager.ts
```diff
@@ -467,3 +437,33 @@ export class PackageManager {
...
+            regex.test(info.version) :
+            !regex.test(info.version)
+        );
+    }
+
+    // No versionRegex provided.
+    return true;
+}
+
+export function ArchitecturesMatch(value: IPackage, info: PlatformInformation): boolean {
```
michelleangela, line 463: nit: rename "value" to something more distinguishable, like pkg.

michelleangela:
> We could add a fix for the "update OS" scenario later on if it can't be done by 0.27.0-insiders4 (on Monday probably).

Is the workaround to uninstall and reinstall? Or will something get persisted even if the extension is uninstalled?

sean-mcmanus:
> We could add a fix for the "update OS" scenario later on if it can't be done by 0.27.0-insiders4 (on Monday probably).
> 
> Is the workaround to uninstall and reinstall? Or will something get persisted even if the extension is uninstalled?

Yeah, uninstall/reinstall of our extension would be a workaround (or updating to a newer version).

PR #5186, opened 2020-03-25
Fix database when no folder is open.

### LINKED ISSUE #5176: 0.27.0-insiders4: "Open File..." scenarios (without a workspace) shows "Tag Parser initializing" (all symbol operations fail)
Go to Def, Find All Refs, Outline, etc. don't work.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/5176 . The language server side expects an empty storage path for the no workspace folder case.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -595,7 +595,7 @@ export class DefaultClient implements Client {
         }
 
         if (!storagePath) {
-            storagePath = path.join(this.RootPath, "/.vscode");
+            storagePath = this.RootPath ? path.join(this.RootPath, "/.vscode") : "";
         }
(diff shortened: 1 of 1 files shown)


michelleangela:
Does this work if you remove a folder or close a workspace? Trying to think of other cases that's similar to this.

Colengms (approved):

sean-mcmanus:
> Does this work if you remove a folder or close a workspace?

Oh, good catch -- there's some bug in that scenario. Investigating now...

sean-mcmanus:
> Does this work if you remove a folder or close a workspace?
> 
> Oh, good catch -- there's some bug in that scenario. Investigating now...

That appears to be a completely different issue I'm seeing -- looks like some bug with Find All References. I'm going to check this fix in, but I'll continue to investigate why Find All References is failing after a workspace folder is closed and a file is opened.

In other words, the database itself appears correct in that case.

PR #5195, opened 2020-03-25
Fix error string.

sean-mcmanus:


michelleangela (approved):

Colengms (approved):
You'll want to run `yarn run generate-native-strings` to regenerate the header for the native side.

sean-mcmanus:
> You'll want to run yarn run generate-native-strings to regenerate the header for the native side.

Okay, I'll wait for 1.0 to check this in.

Issue #5171, opened 2020-03-23
Add a check if macOS > High Sierra and the lldb-mi folder is missing


sean-mcmanus:
Follow up to PR https://github.com/microsoft/vscode-cpptools/pull/5153 .

PR #5197, opened 2020-03-26
Add darwin OS version change check for lldb-mi

WardenGnaw:
If the darwin version changes and is upgraded to Mojave or higher, users
will need to use ./debugAdapters/lldb-mi. These changes will cleanup the
old files and cause the online installation process to run again.


michelleangela (approved):

sean-mcmanus (approved):

bobbrow:
#5171

PR #5217, opened 2020-03-27
Fix incorrect compiler path with spaces and args squiggle.

### LINKED ISSUE #5215: Compiler path with spaces produces error
Related topic https://community.platformio.org/t/platform-io-compiler-error/12684

1) We provide a full path to the compiler using `compilerPath` option
2) This a path option, so we do not do any modifications because arguments are passed to `compilerArgs`

Yes, we can escape `compilerPath` by default but it looks like a bug.

Thanks!

/cc @valeros @sean-mcmanus

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/5215

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1036,7 +1036,7 @@ export class CppProperties {
             // Error when the compiler's path has spaces without quotes but args are used.
             // Except, exclude cl.exe paths because it could be for an older preview build.
             let compilerPathNeedsQuotes: boolean =
-                compilerPathAndArgs.additionalArgs &&
+                (compilerPathAndArgs.additionalArgs && compilerPathAndArgs.additionalArgs.length > 0) &&
                 !resolvedCompilerPath.startsWith('"') &&
                 compilerPathAndArgs.compilerPath !== undefined &&
                 compilerPathAndArgs.compilerPath.includes(" ");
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1315,7 +1315,7 @@ export class CppProperties {
                         continue;
                     }
                     // Squiggle when the compiler's path has spaces without quotes but args are used.
-                    compilerPathNeedsQuotes = compilerPathAndArgs.additionalArgs
+                    compilerPathNeedsQuotes = (compilerPathAndArgs.additionalArgs && compilerPathAndArgs.additionalArgs.length > 0)
```
sean-mcmanus, line 1318: This same type of error has occurred before. It would be nice if there were a way to track down any other cases.

michelleangela: Can you also add the "compilerPathAndArgs.additionalArgs.length >0" check after line 1039?

michelleangela (changes requested):

sean-mcmanus:
Yeah, good catch -- without the change the UI shows the error still.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: > Can you also add the "compilerPathAndArgs.additionalArgs.length >0" check after line 1039?

Fixed.

michelleangela (approved):

Issue #3940, opened 2019-07-18
Request to load $CPATH automatically


lisanhu:
**Type: LanguageService**

This is a feature request that is requesting to parse system environment variable automatically and put the paths in includePath, and hope it will also work for remote development if this feature is implemented.

**Describe the bug**
- OS and Version: MacOS
- VS Code Version: Version: 1.37.0-insider-commit-fd5985489982fe14381dd56e6f0b29c53d5f991b
- C/C++ Extension Version: 0.24.0
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is.
The plug-in reports an error about not able to find the include files. However, the files are just in my $CPATH system environment. The system is a remote cluster using environment modules and it has my libraries loaded as modules. Those modules are changing the $CPATH environment variable and the extension could not find the headers. So I'm requesting the extension to parse the environment variables, which would be really great.

bobbrow:
What is the format of $CPATH?  The extension already supports environment variables in the `includePath`.  If the list of paths is semicolon separated, you can just put
```
    "includePath": "${CPATH}"
```
in your `c_cpp_properties.json` file.

lisanhu:
The $CPATH is a list of folders separated by  colons ":", for example like this:
/dir1:/dir2:/dir3
Sorry, I can't post the exact folder here, I'm now after work and losing access to the clusters.

This environment variable is used by gcc and clang to search for header files, so you have installed a library in your home folder, and if you set up this environment variable, gcc would search the folders in $CPATH (colon-separated). That's also why I wish the includePath will parse this environment variable.

Thank you very much!

bobbrow:
The function that manages this is here:
https://github.com/microsoft/vscode-cpptools/blob/2cde4d6f1a2e567e77d9c64b1361a6ced1243b94/Extension/src/LanguageServer/configurations.ts#L491

We'd only want to add `:` processing on non-Windows OS'es though.

lisanhu:
Oh, great! Thank you!

elahehrashedi:
###Bug Fixes
* processing $CPATH on non-Windows OS [#3940](https://github.com/microsoft/vscode-cpptools/issues/3940)

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/0.28.0-insiders2

elahehrashedi:
This issue is fixed in 0.28.0.
[...]

PR #5229, opened 2020-03-30
define path delimiter based on windows/mac config

elahehrashedi:
parsing variables whose content is split by ":" instead of ";"
https://github.com/microsoft/vscode-cpptools/issues/3940


Extension/src/LanguageServer/configurations.ts
```diff
@@ -523,8 +523,9 @@ export class CppProperties {
         let result: string[] = [];
         if (paths) {
             paths = this.resolveDefaults(paths, defaultValue);
+            let delimiter: string = (this.ConfigurationNames && !(this.ConfigurationNames[0].trim().toLowerCase() === "win32")) ? ":" : ";";
```
bobbrow: You don't want to make this decision based on the configuration name.  You want to make it based on the `process.platform`

bobbrow (approved):

sean-mcmanus (approved):

PR #5234, opened 2020-03-31

csigs:


Extension/i18n/cht/package.i18n.json
```diff
@@ -23,7 +23,7 @@
...
 	"c_cpp.configuration.clang_format_path.description": "Clang 格式可執行檔的完整路徑。如果未指定，且在環境路徑中可用 Clang 格式，即會使用該格式。如果在環境路徑中找不到，則會使用延伸模組所配備的 Clang 格式複本。",
-	"c_cpp.configuration.clang_format_style.description": "編碼樣式，目前支援: Visual Studio、LLVM、Google、Chromium、Mozilla、WebKit。使用「檔案」從目前目錄或父目錄的 .clang-format 檔案載入樣式。使用 {key: value, ...} 來設定特定參數。例如，\"Visual Studio\" 樣式類似於: { BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Allman, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4 }",
+	"c_cpp.configuration.clang_format_style.description": "Coding style, currently supports: Visual Studio, LLVM, Google, Chromium, Mozilla, WebKit. Use \"file\" to load the style from a .clang-format file in the current or parent directory. Use {key: value, ...} to set specific parameters. For example, the \"Visual Studio\" style is similar to: { BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Allman, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All }",
```
sean-mcmanus: It appears the original string had \"file\" incorrectly translated, since it's a literal setting enum. Is there a way to prevent that?

sean-mcmanus: Some of the other language translations have the same issue (others do not).

Colengms: There is guidance in the OneNote for reporting bugs against the localized strings.  That is the best way to get the string fixed upstream.  But, as we change strings they get re-localized, so it may need to happen again whenever we make changes.  I have the following bug open against allowing us to add 'hints' to the exported loc xlf file: https://github.com/microsoft/vscode-nls-dev/issues/21

sean-mcmanus (approved):

sean-mcmanus:
I see .clang-format incorrectly translated to .clang 格式文件中加载样式

PR #5252, opened 2020-04-04
Fix test linter issues.

sean-mcmanus:


Colengms (approved):
index.js and runTest.js are not our sources, per se.  They are boiler-plate source files provided by vscode.  IMHO, there is no point is diverging from the boiler-plate, especially if they need to be updated in the future with new boilder-plater.  But, ok.

michelleangela (approved):

sean-mcmanus:
> index.js and runTest.js are not our sources, per se.  They are boiler-plate source files provided by vscode.  IMHO, there is no point is diverging from the boiler-plate, especially if they need to be updated in the future with new boilder-plater.  But, ok.

Sure, I added index.ts to the .eslintignore, but I kept the runTest.ts linting, because it appears like that is a file that could be edited per test.

sean-mcmanus:
> index.js and runTest.js are not our sources, per se.  They are boiler-plate source files provided by vscode.  IMHO, there is no point is diverging from the boiler-plate, especially if they need to be updated in the future with new boilder-plater.  But, ok.
> 
> Sure, I added index.ts to the .eslintignore, but I kept the runTest.ts linting, because it appears like that is a file that could be edited per test.

I added runTest.ts to the .eslintignore.

PR #5263, opened 2020-04-07
Remove workaround for v3 providers.

sean-mcmanus:


Extension/src/LanguageServer/client.ts
```diff
@@ -1508,15 +1508,19 @@ export class DefaultClient implements Client {
...
+                    for (let c of config.browsePath) {
+                        if (vscode.workspace.getWorkspaceFolder(vscode.Uri.file(c)) === this.RootFolder) {
+                            foundMatch = true;
+                            break;
+                        }
                     }
-                }
-                if (foundMatch) {
+                    if (foundMatch) {
+                        this.sendCustomBrowseConfiguration(config, currentProvider.extensionId);
```
michelleangela: replace line 1516 with 1521 and remove `foundMatch` flag?

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, good catch.

michelleangela (approved):

Issue #407, opened 2016-12-13
Code folding didn't work as expected in directives and block comment


LeThiHyVong:
In my code there are many places look like this

```
#if SOME_CONST_EXP
;//       some code
#else
      // some code
#endif
```
or 

```
if SOME_CONST_EXP
;//      some code
#endif


```
or

```
/*************************************
//
//
//
//
*************************************/
```

EDIT (@bobbrow): (inserting request for folding on multiple consecutive single-line comments from #1443)
```
// Comment
// Comment
// Comment
// Comment

```
The code folding only work correctly between #else and #endif.
From #if to next directive (#else in 1st case or #endif in 2nd case), if all line is started with ';//', the code folding didn't work. 
From #endif to next directive, it automatically group the code outside the directive for all time. 
The coding folder didn't work for 3rd, too. 

![codefolding](https://cloud.githubusercontent.com/assets/19317679/21126909/b68e2070-c121-11e6-89b8-80e807c949f4.png)

System spec:
- Windows 7 SP1 x86
- VSCode 1.7.2
- Plugin 0.9.3

Please correct this.

sean-mcmanus:
Yes, the code folding is buggy, but it's not a bug with the C/C++ extension. We don't implement any code folding. Can you move this to the vscode team?

sean-mcmanus:
Actually, it sounds like the folding behavior you are seeing is "by design". The fixing of this is being tracked at https://github.com/Microsoft/vscode/issues/3422 . VS Code implements the folding behavior and there isn't an API yet to allow language-aware folding.

LeThiHyVong:
According to https://github.com/Microsoft/vscode/issues/3422#event-1582353384, please re-open this issue. 
Thanks.

hoffmael:
This would be a great feature and bug fix to have and essential for the code base I work on which relies on many preprocessor switches to support multiple builds. Diming the inactive code with "C_Cpp.dimInactiveRegions": true is a good first step, but having the ability to collapse and hide the inactive code region altogether is even better.
Eclipse even has the option to collapse inactive regions by default when opening a file by default.

Here's a screenshot from 0.17.4 showing the various issues with folding currently.
![image](https://user-images.githubusercontent.com/24258055/40857672-123a5d86-6599-11e8-9e60-2e82de2fd12d.png)

sean-mcmanus:
+1 I hit this issue sometimes with our cross platform code.

intijk:
[...]

PR #5307, opened 2020-04-16
Code Folding

Colengms:
Addresses https://github.com/microsoft/vscode-cpptools/issues/407 

Most of the logic is in the native process.  The TypeScript side proxies between a FoldingRangeProvider and native requests.

Because it's not possible to get the ID of a request (to use to cancel it later using LSP's `$/cancelRequest`), I've added a new abort mechanism that accepts an arbitrary ID.  This makes it possible to cancel a request (which LSP maps a response to) instead of having to use independent notifications in either direction and map up requests with responses ourselves.


Extension/src/LanguageServer/client.ts
```diff
@@ -335,6 +335,20 @@ interface DidChangeConfigurationParams extends WorkspaceFolderParams {
     settings: any;
 }
 
+interface GetFoldingRangesParams {
```
sean-mcmanus, line 340: Should this inherit from AbortRequestsParams?

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
Colengms: It could, to consolidate the common field.  But, they aren't technically related.  The are params for different messages.

sean-mcmanus:
Can you increase the version to 0.28.0? Since this is the 1st major "feature".

PR #5309, opened 2020-04-16
insiders update/downgrade improvements

elahehrashedi:
Adding tests for upgrading or downgrading the extension's version for Insiders.
Checking the availability of insider's builds' assets on GitHub.


Extension/src/githubAPI.ts
```diff
@@ -145,7 +146,18 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
                 return undefined;
             }
 
-            return getTargetBuild(builds, userVersion, updateChannel);
+            const targetBuild: Build | undefined = getTargetBuild(builds, userVersion, updateChannel);
+            if (targetBuild === undefined) {
+                // no oction
```
bobbrow: [sp] "action"

Extension/src/githubAPI.ts
```diff
@@ -145,7 +146,18 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
...
+                telemetry.logLanguageServerEvent("NoUPgradeDowngrade");
```
bobbrow: I think we should use the same event name for all upgrade responses, but add different "properties" for the action that was taken. This also allows us to log additional information as necessary.

For example:
```
telemetry.logLanguageServerEvent("UpgradeCheck", { "action": "none" })
telemetry.logLanguageServerEvent("UpgradeCheck", { "action": "upgrade", "version": targetBuild.name })
telemetry.logLanguageServerEvent("UpgradeCheck", { "action": "downgrade", "version": targetBuild.name })
```

Extension/src/githubAPI.ts
```diff
@@ -172,19 +184,25 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
  * @param updateChannel The user's updateChannel setting.
  * @return The Build if the user should update to it, otherwise undefined.
  */
-function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string): Build | undefined {
+export function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string): Build | undefined {
     if (!vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
         return undefined;
     }
+    const latestVersion: PackageVersion = new PackageVersion(builds[0].name);
+    if ((updateChannel === 'Insiders') && (userVersion.suffix && userVersion.suffix === 'insiders') && (userVersion.isGreaterThan(latestVersion))) {
```
bobbrow: Do you need to check the suffix? I think just using `isGreaterThan` is better. You don't want to accidentally downgrade from 0.27.0 to 0.27.0-insiders (which would also be an internal testing case which I think was missing from the table)

Please also add a comment that explains why we do this here instead of as part of `useBuild` (so we don't have to check every build in the `builds` array).

Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -0,0 +1,141 @@
...
+                    name: insider2, assets: four_assets}, {
+                    name: insider1, assets: four_assets}, {
+                    name: release0, assets: four_assets}];
+
+                const userVersion: PackageVersion = new PackageVersion(insider3);
+                const targetBuild: Build | undefined = getTargetBuild(builds, userVersion, updateChannel);
+                assert.equal(targetBuild, undefined);
+            });
+        });
+        suite("Insider Users, Downgrade", () => {
```
bobbrow: Can you also add a test that downgrades past more than one recalled release? (e.g. 0.27.1-insiders2 -> 0.27.0 or 0.27.1-insiders3 -> 0.27.1-insiders)

Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+import * as assert from "assert";
+import { resolveVariables, escapeForSquiggles } from "../../src/common";
+import { Build, Asset, getTargetBuild } from "../../src/githubAPI";
+import { PackageVersion } from '../../src/packageVersion';
+
+
+suite("UpgradeDowngrade", () => {
+
+    const updateChannel: string = "Insiders";
```
bobbrow: Could you also add some tests for the "Default" channel?

Extension/src/githubAPI.ts
```diff
@@ -9,13 +9,14 @@ import * as util from './common';
 import { PlatformInformation } from './platform';
 import { OutgoingHttpHeaders } from 'http';
 import * as vscode from 'vscode';
+import * as telemetry from './telemetry';
 
 const testingInsidersVsixInstall: boolean = false; // Change this to true to enable testing of the Insiders vsix installation.
```
bobbrow, line 14: @sean-mcmanus do we need this variable anymore?

Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const release1: string = "0.27.1";
+    const insider3: string = "0.27.1-insiders3";
+    const insider2: string = "0.27.1-insiders2";
+    const insider1: string = "0.27.1-insiders";
+    const release0: string = "0.27.0";
+
+    suite("Downgrade", () => {
+        suite("Internal Testing, no Downgrade", () => {
+            test("Insider to Release", () => {
```
bobbrow: Can you add a test for "Release to Insider"?  This was not in the table.  e.g. internal testing of 0.27.1, but latest online build is 0.27.1-insiders

bobbrow (changes requested):
Looks pretty good.  Just a few suggestions.  Thanks!

Extension/src/githubAPI.ts
sean-mcmanus: Maybe. Might be useful for manual testing/debugging of the code to force an install.

Extension/src/githubAPI.ts
```diff
@@ -172,19 +183,26 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
...
+        return undefined;
+    }
 
     // Get predicates to determine the build to install, if any
     let needsUpdate: (installed: PackageVersion, target: PackageVersion) => boolean;
     let useBuild: (build: Build) => boolean;
     if (updateChannel === 'Insiders') {
         needsUpdate = (installed: PackageVersion, target: PackageVersion) => testingInsidersVsixInstall || (!target.isEqual(installed));
-        useBuild = (build: Build): boolean => true;
+        // check if the assetes are available (insider)
```
sean-mcmanus: assetes -> assets

sean-mcmanus (approved):

michelleangela (approved):

Extension/src/githubAPI.ts
```diff
@@ -201,7 +201,7 @@ export function getTargetBuild(builds: Build[], userVersion: PackageVersion, upd
         // check if the assetes are available (insider)
```
bobbrow: [sp] `assets`

Extension/src/githubAPI.ts
```diff
@@ -201,7 +201,7 @@ export function getTargetBuild(builds: Build[], userVersion: PackageVersion, upd
...
         useBuild = isBuild;
     } else if (updateChannel === 'Default') {
-        needsUpdate = function(installed: PackageVersion, target: PackageVersion): boolean { return installed.isGreaterThan(target); };
+        needsUpdate = function(installed: PackageVersion, target: PackageVersion): boolean { return target.isGreaterThan(installed); };
```
bobbrow: I think this needs to be reverted (and a comment should be added).  When this function is called with the "Default" channel it's because the channel used to be "Insiders".  That means the user is leaving the insiders channel and needs to downgrade back to the latest release build.  So a comment here stating that the user is exiting the insiders channel and needs to downgrade would help us avoid changing this in the future.

Extension/src/githubAPI.ts
```diff
@@ -218,10 +218,13 @@ export function getTargetBuild(builds: Build[], userVersion: PackageVersion, upd
     const targetVersion: PackageVersion = new PackageVersion(targetBuild.name);
     if (needsUpdate(userVersion, targetVersion)) {
         if (targetVersion.isEqual(userVersion)) {
```
bobbrow: I don't think a check for `isEqual` should be needed here.  I believe `getTargetBuildInfo` will exit early if the versions are the same.

Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -25,117 +22,156 @@ suite("UpgradeDowngrade", () => {
...
-        });
-    });
-
-    suite("Upgrade", () => {
+    suite("DefaultChannel", () => {
+        const updateChannel: string = "Default";
+        const extra_insider2: string = "0.27.2-insiders2";
+        const extra_insider1: string = "0.27.2-insiders";
         suite("Automatic Upgrade", () => {
             test("Release to Release", () => {
```
bobbrow: I may have unintentionally led you astray here. This function will only be called with "Default" as the release channel if you are exiting the "Insiders" program.  In this case the only action that should happen is "Automatic Downgrade" from "Insiders to Release".

bobbrow (changes requested):

Extension/src/githubAPI.ts
elahehrashedi: I checked it, that's true!

bobbrow (approved):

Issue #2782, opened 2018-11-07
Support gnu standard versions for querying compiler


invedj:
Hi, I am receiving the error `identifier "CLOCK_MONOTONIC" is undefined` from cpptools for the below code, which I don't receive when compiling with gcc. If I include `"defines": ["_POSIX_C_SOURCE 199309L"]` in `c_cpp_properties.json` then the issue is resolved and cpptools doesn't complain.

My question is, why is this behavior different when I compile with gcc? I believe the default C standard for gcc is gnu11, which doesn't require `_POSIX_C_SOURCE` to be defined. Is cpptools using a different standard, and if so is there a way to force using gnu11? I've tried adding `-std=gnu11` to the compiler path, but that doesn't seem to solve the issue.

- OS and Version: Ubuntu 16.04.3 LTS
- VS Code Version: 1.28.2
- C/C++ Extension Version: 0.20.1
- Other extensions: none

```c
#include <stdint.h>
#include <time.h>

uint64_t timer_diff(struct timespec *start)
{
	uint64_t diff;
	struct timespec now;

	clock_gettime(CLOCK_MONOTONIC, &now);

	diff = UINT64_C(1000000000) * (now.tv_sec - start->tv_sec) +
		   now.tv_nsec - start->tv_nsec;

	return diff;
}
```

```json
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "compilerPath": "/usr/lib/gcc",
            "intelliSenseMode": "gcc-x64",
            "browse": {
                "path": [
                    "${workspaceFolder}/**"
                ],
                "limitSymbolsToIncludedHeaders": true,
                "databaseFilename": ""
            },
            "cStandard": "c11",
            "cppStandard": "c++17"
        }
    ],
    "version": 4
}
```

```
initialized
workspace/didChangeConfiguration
IntelliSense Engine = Default.
The extension will use the Tag Parser for IntelliSense when #includes don't resolve.
Autocomplete is enabled.
Error squiggles are enabled.
File exclude: **/.git
File exclude: **/.svn
File exclude: **/.hg
File exclude: **/CVS
File exclude: **/.DS_Store
File exclude: **/.vscode
Search exclude: **/node_modules
Search exclude: **/bower_components
Search exclude: **/.vscode
textDocument/didOpen
cpptools/queryCompilerDefaults
[...]

PR #5321, opened 2020-04-17
Add GNU standards and C18 standard for querying compiler

michelleangela:
Adds the following GNU standards to query compiler to get #defines for GNU standards.
(issue https://github.com/microsoft/vscode-cpptools/issues/2782)

Note: if GNU standard is specified in IntelliSense configuration, IntelliSense will emulate the equivalent C or C++ standard version.

GNU C++ standard
- gnu++20
- gnu++17
- gnu++14
- gnu++11
- gnu++03
- gnu++98

GNU C standard
- gnu18
- gnu11
- gnu99
- gnu89

Also, add C18 standard.


sean-mcmanus:
It seems like we should use c18 instead of c17 https://en.wikipedia.org/wiki/C18_(C_standard_revision)

The backend should probably handle both on the command line (compilerArgs) though.

michelleangela:
I don't see `c18` for Clang at https://clang.llvm.org/docs/CommandGuide/clang.html

`c18` is supported by GNU https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html

sean-mcmanus:
> I don't see c18 for Clang at https://clang.llvm.org/docs/CommandGuide/clang.html
> c18 is supported by GNU https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html

c18 with clang works for me when I try it via https://godbolt.org/ . The docs might be out of date.

Extension/c_cpp_properties.schema.json
```diff
@@ -26,17 +26,22 @@
                         }
                     },
                     "cStandard": {
-                        "description": "Version of the C language standard to use for IntelliSense.",
+                        "description": "Version of the C language standard to use for IntelliSense. Note: GNU standards are only used to query the set compiler to get GNU defines and IntelliSense will emulate the equivalent C standard version.",
```
michelleangela: Is description clear? please advice if changes are needed.

sean-mcmanus: Yeah, it should be "defines, and" though. Also, "query the set compiler" might be more easy to understand as "query the compilerPath`.

sean-mcmanus (approved):

Issue #5119, opened 2020-03-14
Add check to ensure offline/insiders vsix is installed on correct platform


Colengms:
Installing a win32 cpptools vsix on Linux, for example, results in an error that the native binary is missing.  (Which is correct, as a different filename is used on each platform).  Perhaps existance of the wrong binaries could surface an indication of what is wrong, and fall back to downloading if possible.

sean-mcmanus:
For Mac binary detection we can check for LLVM/bin/clang-format.darwin, because we forgot to rename that binary.

The downloading fallback is not a good option for Insiders, because the binary is most likely out-of-date (in fact, we should probably make that impossible or detect a mismatched binary version, in case the install.lock is somehow deleted).

elahehrashedi:
This issue is fixed in 0.28.0.
0.28.0 release: https://github.com/microsoft/vscode-cpptools/releases/tag/0.28.0

PR #5332, opened 2020-04-20
check offline/insiders VSIX

elahehrashedi:
Add check to ensure offline/insiders VSIX is installed on the correct platform
add Github download link to the correct platform
bugfix: https://github.com/microsoft/vscode-cpptools/issues/5119

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -534,6 +534,33 @@ export function checkInstallLockFile(): Promise<boolean> {
     return checkFileExists(getInstallLockPath());
 }
 
+/** Get the platform that the installed binaries belong to.*/
+export function getInstalledBinaryPlatform(): Promise<string> {
+    return new Promise(async (resolve, reject) => {
+        // the debugAdapters folder is utilized to identify the platform
+        const path_to_check: string = getExtensionFilePath("debugAdapters");
+        let installedPlatform: string = "";
+        if (checkDirectoryExists(path_to_check)) {
+            const files: string[] = await readDir(path_to_check);
+            installedPlatform = "win32";
+            if (files !== undefined) {
+                (files).forEach((element: string) => {
+                    if (element.includes(".osx")) {
+                        installedPlatform = "darwin";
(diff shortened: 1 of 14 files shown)


Extension/src/main.ts
```diff
@@ -48,6 +48,24 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
     Telemetry.activate();
     util.setProgress(0);
 
+    // check if the correct offline/insiders vsix is installed on the correct platform
+    let installedPlatform: string = await util.getInstalledBinaryPlatform();
+    if (process.platform !== installedPlatform) {
+        let helpURL: string = "https://github.com/Microsoft/vscode-cpptools/releases";
+        errMsg = localize("native.binaries.not.supported", "The installed VSIX is incompatible with {0}. Download the {0} VSIX from: {1}.", process.platform, helpURL);
```
bobbrow: I would like to avoid using "VSIX" in the message since most non-MSFT folks don't refer to the extension that way.  I think a simple message that says something like: "The {0} version of the extension appears to be installed. Please download and install the {1} version of the extension." would be better.  Putting the URL in the message is unnecessary because the button will take you there.

Extension/src/common.ts
```diff
@@ -534,6 +534,33 @@ export function checkInstallLockFile(): Promise<boolean> {
     return checkFileExists(getInstallLockPath());
 }
 
+/** Get the platform that the installed binaries belong to.*/
+export function getInstalledBinaryPlatform(): Promise<string> {
+    return new Promise(async (resolve, reject) => {
```
bobbrow: Let's avoid creating explicit Promises.  If you mark this function as `async` you can use `await` directly and there's no need to call `resolve` or `reject`.  Just return the result if it succeeds and throw a `new Error` if it fails.

Extension/src/main.ts
bobbrow: Or "The {0} version of the extension is incompatible with your OS"

sean-mcmanus:
What are the loc string changes?

PR #5341, opened 2020-04-21
check offline/insiders VSIX

elahehrashedi:
Add check to ensure offline/insiders VSIX is installed on the correct platform
add Github download link to the correct platform
bugfix: https://github.com/microsoft/vscode-cpptools/issues/5119


Extension/src/common.ts
```diff
@@ -534,6 +534,29 @@ export function checkInstallLockFile(): Promise<boolean> {
...
+/** Get the platform that the installed binaries belong to.*/
+export async function getInstalledBinaryPlatform(): Promise<string | undefined> {
+    // the debugAdapters folder is utilized to identify the platform
+    const path_to_check: string = getExtensionFilePath("debugAdapters");
+    let installedPlatform: string = "";
+    if (checkDirectoryExists(path_to_check)) {
+        const files: string[] = await readDir(path_to_check);
+        installedPlatform = "win32";
+        if (files !== undefined) {
+            (files).forEach((element: string) => {
```
pieandcakes: it looks like you are relying on checking to see which `mono`  runtime we have installed to figure out the installed platform. Is that the best way to do this? Is the `uname` method not working? Also, once you have the installedPlatform set, you should exit instead of looping through all the files in the folder.

pieandcakes: Is the bug to do with the language service or with the debugger?

bobbrow: This is not related to debugger.  `uname` works, but we're trying to determine which "offline vsix" was installed and if it matches the OS is was installed on.  We need to look at something in the extension's files to determine this.  I think another alternative would be to embed a simple text file in the extension as part of our build process.  We already inject lock.txt when we build offline vsix'es.  We could potentially add content to lock.txt (currently it's empty) and just read it to see if it matches the OS.  If lock.txt is empty (like for online vsix), we wouldn't do the check.

Extension/src/githubAPI.ts
```diff
@@ -35,7 +35,7 @@ export interface Build {
 * @param vsixName The name of the VSIX to search for
 * @return The download URL of the VSIX
 */
-function getVsixDownloadUrl(build: Build, vsixName: string): string {
+export function getVsixDownloadUrl(build: Build, vsixName: string): string {
```
bobbrow: It didn't look like this was used anywhere else, so I don't think it needs to be `export`

Extension/src/common.ts
bobbrow: The only trouble with using `foreach` here is that there is no way to break out once you find what you are looking for.  Take a look at `some` which should allow you to return a `true` once you've set the `installedPlatform`.

WardenGnaw: This should be a bug on both. If a user decides to download the wrong offline installer it wont execute on their machine. 

`uname` will determine the current machine but does not determine the offline vsix.

I am not a fan of this solution since its assuming the debugger will always use a mono runtime with specific extensions to determine the offline package type.

Extension/src/platform.ts
```diff
@@ -14,6 +14,16 @@ import * as nls from 'vscode-nls';
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
+export function GetOsName(processPlatform: string | undefined): string | undefined {
```
bobbrow: minor: two letter acronyms stay capitalized in function names. e.g. `GetOSName`

Extension/src/githubAPI.ts
```diff
@@ -44,6 +44,9 @@ function getVsixDownloadUrl(build: Build, vsixName: string): string {
     return downloadUrl;
 }
 
+export function getOfflineDownloadUrl(): string {
```
bobbrow: This could probably just be a const value instead of a function.  I suspect it could be used elsewhere in this file too though I don't know off the top of my head.

WardenGnaw:
The original issue mentions:
> [This] results in an error that the native binary is missing. (Which is correct, as a different filename is used on each platform)

Can we not catch where the file not found exception happens and then show the user the correct download dialog?

Extension/src/common.ts
sean-mcmanus: You could just check the LLVM/clang-format or clang-format.exe or clang-format.darwin.

Extension/src/githubAPI.ts
elahehrashedi: I found it in main.ts being used in a log message (line 318). I can replace that one with the const value as well.

Extension/src/common.ts
elahehrashedi: A possible solution for the future is to add a dummy file in the VSIX folder that indicates the VSIX platform's info. For now, a unified solution for win/mac/linux32/linux64 is to check this folder. We can replace this with the name of the dummy file later on.

Extension/src/platform.ts
```diff
@@ -14,6 +14,16 @@ import * as nls from 'vscode-nls';
...
+    switch (processPlatform) {
+        case "win32": return "Windows";
+        case "darwin": return "MacOS";
+        case "linux32": return "Linux(32 bit)";
+        case "linux": return "Linux(64 bit)";
```
sean-mcmanus: Should be a space before (. Or you could remove the 32-bit Linux stuff and just use "Linux".

Extension/src/common.ts
bobbrow: We wouldn't need to add a new dummy file.  We could just add contents to install.lock and you can read the file.  But since we don't support 32-bit linux anymore, we can just look at clang-format as was discussed earlier today..

elahehrashedi: In the new approach, we are not reading the content of that folder anymore.

elahehrashedi: In the new approach, we check the LLVM/bin folder.
image

bobbrow (approved):

sean-mcmanus (approved):

PR #5350, opened 2020-04-22
Update for 3 assets (no more 32-bit Linux)

sean-mcmanus:
0.27.0 users won’t auto-update to 0.28.0-insiders without 4 assets, but we could fix that temporarily by adding a bogus 4th asset until we ship 0.28.0 with this change that checks for 3 or more assets.

32-bit Linux will get a failure when there's not cpptools-linux32.vsix in the assets, but when I tried it out the failure appears to be silent, so that seems okay.

### DIFF
--- a/Extension/src/githubAPI.ts
+++ b/Extension/src/githubAPI.ts
@@ -56,12 +56,12 @@ function isAsset(input: any): input is Asset {
 
 /**
  * Determine whether an object is of type Build. Note that earlier releases of the extension
- * do not have 4 or greater Assets (Mac, Win, Linux 32/64). Only call this on more recent Builds.
+ * do not have 3 or greater Assets (Mac, Win, Linux). Only call this on more recent Builds.
  * @param input Incoming object.
  * @return Whether input is of type Build.
  */
 function isBuild(input: any): input is Build {
-    return input && input.name && typeof(input.name) === "string" && isArrayOfAssets(input.assets) && input.assets.length >= 4;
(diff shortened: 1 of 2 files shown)


Extension/src/githubAPI.ts
```diff
@@ -56,12 +56,12 @@ function isAsset(input: any): input is Asset {
 
 /**
  * Determine whether an object is of type Build. Note that earlier releases of the extension
- * do not have 4 or greater Assets (Mac, Win, Linux 32/64). Only call this on more recent Builds.
+ * do not have 3 or greater Assets (Mac, Win, Linux). Only call this on more recent Builds.
```
bobbrow, line 59: I think there might actually be 6 assets currently (since the source is zipped with each release), and this will go to 5.  We might want to check that there are 3 assets with a ".vsix" extension.

sean-mcmanus: The code gives 4 when I debugged it, i.e. something is choosing the 4 vsix's as assets and ignoring the other stuff. I could look into why it's not 6...

sean-mcmanus: The raw JSON has 4 assets only. See https://api.github.com/repos/Microsoft/vscode-cpptools/releases . The other 2 that appear in the web interface are actually stored in different properties in the raw JSON, and the GitHub UI just presents them as "assets".

Colengms (approved):

Extension/src/githubAPI.ts
bobbrow: ok. It's been a while since we wrote this code.

PR #5351, opened 2020-04-23
adding CppStandard version to clang_format_style

### LINKED ISSUE #3578: Update default clang-format style to format "> >" correctly based on cppStandard
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 7
- VS Code Version:
Version: 1.33.1 (system setup)
Commit: 51b0b28134d51361cf996d2f0a1c698247aeabd8
Date: 2019-04-11T08:27:14.102Z
Electron: 3.1.6
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 6.1.7601
- C/C++ Extension Version:
Version 0.22.1
- A clear and concise description of what the bug is.
`std::vector<std::set<int> > my_vector;`
is automatically corrected to:
`std::vector<std::set<int>> my_vector;`

The latter is illegal (at least before C++ 11) and throws the error:
`error: '>>' should be '> >' within a nested template argument list`

**To Reproduce**

1. In a *.cpp file, type this line manually (not copy pasted):
`std::vector<std::set<int> > my_vector;`
2. Press enter.

**Expected behavior**

That the space between `> >` would not be removed in cases such as this.

**Screenshots**

**Additional context**

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/3578


Colengms (approved):

sean-mcmanus:
We shouldn't update the description. Most users won't be using C++03, so the fact that internally we'll switch to using that for the "Visual Studio" style doesn't need to be described here. The description is meant as a baseline for users to copy to modify if they wanted to change it slightly. The fact that users could add a CppStandard that matches their language (and their installed clang-format version) to their clang-format styles might be useful to describe, but probably somewhere else in the docs (not sure where exactly).

sean-mcmanus (changes requested):
.

bobbrow:
@sean-mcmanus should we just close this PR then?

sean-mcmanus:
> @sean-mcmanus should we just close this PR then?

Yes.

Issue #4873, opened 2020-01-21
For auto-update, check that the insiders version is compatible


bobbrow:
Type: General

We take a dependency on a minimum version of VS Code in our extension.  There is a concern that we might try to download/install an insiders version that has a dependency on a newer VS Code version than is currently installed.  In that case, we might be attempting multiple times to update the insiders version.

PR #5371, opened 2020-04-28
Disable Insiders updates for unsupported cases.

sean-mcmanus:
Partial fix for https://github.com/microsoft/vscode-cpptools/issues/4873 .

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -462,22 +462,22 @@ function realActivation(): void {
 
     vcpkgDbPromise = initVcpkgDatabase();
 
-	PlatformInformation.GetPlatformInformation().then(info => {
-		// Skip Insiders processing for 32-bit Linux.
-		if (info.platform !== "linux" || info.architecture === "x86_64") {
-			// Skip Insiders processing for VS Code newer than 1.42.1.
-			// TODO: Change this to not require the hardcoded version to be updated.
-			let vscodeVersion: PackageVersion = new PackageVersion(vscode.version);
-			let minimumSupportedVersionForInsidersUpgrades: PackageVersion = new PackageVersion("1.42.1");
-			if (vscodeVersion.isGreaterThan(minimumSupportedVersionForInsidersUpgrades, "insider")) {
-				if (settings.updateChannel === 'Default') {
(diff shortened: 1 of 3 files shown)


Colengms (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -462,12 +462,22 @@ function realActivation(): void {
 
     vcpkgDbPromise = initVcpkgDatabase();
 
-    if (settings.updateChannel === 'Default') {
-        suggestInsidersChannel();
-    } else if (settings.updateChannel === 'Insiders') {
-        insiderUpdateTimer = global.setInterval(checkAndApplyUpdate, insiderUpdateTimerInterval, settings.updateChannel);
```
bobbrow, line 468: Is the removal of the `insiderUpdateTimer` intentional?  I don't think it should be removed.

Extension/src/LanguageServer/extension.ts
```diff
@@ -462,12 +462,22 @@ function realActivation(): void {
...
-        checkAndApplyUpdate(settings.updateChannel);
-    }
+    PlatformInformation.GetPlatformInformation().then(info => {
+        // Skip Insiders processing for 32-bit Linux.
+        if (info.platform !== "linux" || info.architecture === "x86_64") {
```
bobbrow, line 468: This check should probably move to `checkAndApplyUpdate`.  Otherwise this logic needs to exist both here and in `onDidChangeSettings`.  I'd rather it be in just one place.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Ooops, not intentional.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Umm...but it needs to surround suggestInsidersChannel too. I'll look into onDidChangeSettings.

sean-mcmanus: I just added an enabled Boolean.

bobbrow (approved):
:shipit:

PR #5401, opened 2020-05-01
Switch to Semantic Tokens API

Colengms:
(Recreating PR from the correct account)
* Removes previous semantic colorization implementation.
* Adds a SemanticTokensProvider.
* Restores implementation of inactive regions from prior to being merged with semantic colorization.
* VS Code version dependency is increased to 1.44.0


Extension/src/LanguageServer/client.ts
```diff
@@ -1043,6 +1052,61 @@ export class DefaultClient implements Client {
...
+                    // Semantic token modifiers are bit indexes corresponding to the indexes in this list of modifiers in the legend.
+                    let tokenModifiersLegend: string[] = [];
+                    for (let e in SemanticTokenModifiers) {
+                        if (isNaN(Number(e))) {
+                            tokenModifiersLegend.push(e);
+                        }
+                    }
+                    const legend: vscode.SemanticTokensLegend = new vscode.SemanticTokensLegend(tokenTypesLegend, tokenModifiersLegend);
+
+                    class SemanticTokensProvider implements vscode.DocumentSemanticTokensProvider {
```
michelleangela: Is there a way to have `class SemanticTokensProvider` in a different file like the old colorization class? Or will this require some major refactoring in Client.ts?

Extension/src/LanguageServer/client.ts
```diff
@@ -1043,6 +1052,61 @@ export class DefaultClient implements Client {
...
+                                                    tokensResult.tokens.forEach((token) => {
+                                                        builder.push(token.line, token.character, token.length, token.type, token.modifiers);
+                                                    });
+                                                    resolve(builder.build());
+                                                }
+                                            }
+                                        });
+                                    token.onCancellationRequested(e => this.client.abortRequest(id));
+                                });
+                            });
```
michelleangela: Can this returned object be broken down into functions that can be called with `await` so it's easier to follow than the nested results?

Extension/src/LanguageServer/client.ts
```diff
@@ -363,6 +321,62 @@ interface AbortRequestParams {
...
+enum SemanticTokenModifiers {
+    // These are camelCase as the enum names are used directly as strings in our legend.
+    // eslint-disable-next-line no-bitwise
+    static = (1 << 0),
+    // eslint-disable-next-line no-bitwise
+    global = (1 << 1),
+    // eslint-disable-next-line no-bitwise
+    local = (1 << 2)
+}
+
```
michelleangela, line 380: Interfaces/enums related to colorization or semantic tokens can be placed in separate file and exported to be used in Client class.

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: So, I'm not sold on making these changes.  There is no colorization logic on the TypeScript side anymore, except for logic related to marshaling messages between VS Code and our native process.  All other such marshaling is in this file.  The implementation of the provider is mainly boilerplate (almost identical to code folding provider).  I would agree with a general refactor that moved all providers into separate files, but I don't think it makes sense to single out these changes to be moved out, as they are not the worst offenders by far.  I'd be open to making the changes in a subsequent PR, and for all of these providers.

michelleangela: Yes, I meant for refactoring the providers into separate files later would be nice.

michelleangela (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1067,6 +1133,9 @@ export class DefaultClient implements Client {
                             if (settings.codeFolding) {
                                 this.codeFoldingProviderDisposable = vscode.languages.registerFoldingRangeProvider(this.documentSelector, new FoldingRangeProvider(this));
                             }
+                            if (settings.enhancedColorization && this.semanticTokensLegend) {
```
sean-mcmanus, line 1139: Why is this check needed? Isn't it always true?

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
Colengms: I assume you're referring to `this.semanticTokensLegend`, as the setting is not always true.  This check was needed because the variable is nullable.  It should always be true here, but the check is necessary to silence the lint error.

Extension/src/LanguageServer/client.ts
Colengms: I'd like to do this in a subsequent change (when I move the providers into files), as this is currently a common pattern for multiple provides.

Themes/README.md
```diff
@@ -1,9 +1,11 @@
...
+[Semantic colorization was added to the C/C++ Extension in version 0.24.0](https://devblogs.microsoft.com/cppblog/visual-studio-code-c-c-extension-july-2019-update/).  At the time, colorization in VS Code was purely syntactic/lexical and leveraged TextMate grammar to associate named 'scopes' with syntactic elements.  Themes and settings can be used to associate colors with these scopes.  Our original implementation of semantic colorization leveraged the same system of associating colors with named scopes.  But, some tokens that can be colored by semantic colorization in C/C++ did not have existing analogs in VS Code's TextMate grammar.  So, new named scopes are required.  Because these scopes were new, existing themes did not include colors for them either.
 
 We created C/C++ Extension UI Themes to closely match Visual Studio themes, and include colors for many of the new scopes.
 
+VS Code has since provided an API for semantic colorization.  The C/C++ Extension has transitioned from it's own implementation to this new API.  These themes now include colors for some of the new semantic token scopes.
```
J-Fields: typo: `it's` -> `its`

sean-mcmanus: Thanks. Fixed.

PR #5406, opened 2020-05-04
add new error message for Custom config provider extension users

### LINKED ISSUE #3971: Custom config provider extension users should not see the message "Please update your includePath."
Currently when an include search path does not resolve, the message displayed to users is "#include error detected. Please update your includePath." This message is not helpful for custom configuration providers, since we don't want the users to update the includePath themselves, which would override the results the config provider provides.

In our scenario, the user would see this every time at start up since it takes a few seconds for the makefile parsing to complete. In this case, we especially don't want the users to be updating any configurations on their own.

elahehrashedi:
Custom config provider extension users should not see the message "Please update your includePath.
bug fix: https://github.com/microsoft/vscode-cpptools/issues/3971


elahehrashedi:
@sean-mcmanus do you think we should add more information to the message, for example:
"#include errors detected based on information provided by configurationProvider. IntelliSense features for this translation unit ({0}) will be provided by the Tag Parser."

sean-mcmanus:
> @sean-mcmanus do you think we should add more information to the message, for example:
> "#include errors detected based on information provided by configurationProvider. IntelliSense features for this translation unit ({0}) will be provided by the Tag Parser."

Yeah, that sounds good, but change it to "...provided by the configurationProvider setting." in case users don't know that "configurationProvider" is a setting, since it can get auto-set it from a popup dialog.

michelleangela (approved):

sean-mcmanus (approved):

Issue #3923, opened 2019-07-12
VSIX Files Adding To /tmp Each Hour VSCode is Open.


vindicatorr:
Type: General
**Describe the bug**
- OS and Version: Linux x64 5.1.15.a-1-hardened
- VS Code Version: Code - Insiders 1.36.0-insider (c089daa, 2019-06-07T05:15:26.964Z)
- C/C++ Extension Version: 0.23.1
- Other extensions you installed (and if the issue persists after disabling them): Unrelated
- A clear and concise description of what the bug is. https://github.com/microsoft/vscode/issues/76911
Everything has already been mentioned there. The short of it is that once a project is opened that uses this extension, a vsix file is created in /tmp every 1 hour, and the file is NEVER deleted.

**To Reproduce**
1. Open any project that uses this extension.
2. Monitor the /tmp folder for created vsix files over time.
3. Close VSCode
4. Note /tmp still has replicating vsix files.

**Expected behavior**
Don't automatically download vsix files when update is set to manual.
Clean up vsix files.

**Screenshots**
Not needed.

**Additional context**
I set the extension log verbosity to Debug and saw nothing related to vsix updates.
Make sure to look at the linked issue at vscode for details. We can continue from here if it is related to this extension and not a vscode issue.
Log (Main):
```
[2019-07-12 16:38:24.212] [main] [info] update#setState idle
[2019-07-12 16:38:24.213] [main] [info] update#ctor - manual checks only; automatic updates are disabled by user preference
```
$ ls -lat /tmp/*.vsix
-rw------- 1 \<uid> \<gid> 19283261 Jul 12 17:42 /tmp/tmp-10323wGcOCsJMxKxG.vsix
-rw------- 1 \<uid> \<gid> 19283261 Jul 12 16:41 /tmp/tmp-10323LYATa11ywpfA.vsix

sean-mcmanus:
You can set your C_Cpp.updateChannel Setting to "Default" as a workaround (or upgrade to 0.24.0). We have another workitem to check the extensions.autoUpdate. The "Insider" update channel is causing us to download the 0.24.0 vsix -- the mechanism was intended to only apply to "insider" vsix's and not released ones.

Do you know why the downloaded vsix is not installing? Repeatedly downloading seems buggy, as is the not deleting of tmp files.

vindicatorr:
Hmmm, yeah. I changed updateChannel to Default.
[...]

PR #5422, opened 2020-05-06
Delete temporary VSIX created during installation

michelleangela:
Fix for https://github.com/microsoft/vscode-cpptools/issues/3923

**Issue:**
The created temp file for the VSIX content does not get deleted after its use is no longer needed.

**Fix:**
Delete temporary VSIX file after downloading and installing operations completes or fails by calling `removeCallback()` on the temporary VSIX file `tmp.FileResult` object.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -823,7 +823,7 @@ async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
     }
 
     // Delete temp VSIX file
-    if (!tempVSIX) {
+    if (tempVSIX) {
         tempVSIX.removeCallback();
     }
 }
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -770,15 +770,10 @@ async function suggestInsidersChannel(): Promise<void> {
 }
 
 async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
-    const tempVSIX: any = await util.CreateTempFileWithPostfix('.vsix').catch(error => {
-        if (error.message.indexOf('/') !== -1 || error.message.indexOf('\\') !== -1) {
-            error.message = "Potential PII hidden";
-        }
-        telemetry.logLanguageServerEvent('installVsix', { 'error': error.message, 'success': 'false' });
-        return;
-    });
-
+    let tempVSIX: any;
     try {
+        tempVSIX = await util.createTempFileWithPostfix('.vsix');
+
         // Try to download VSIX
         let config: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
         let originalProxySupport: string | undefined = config.inspect<string>('http.proxySupport')?.globalValue;
@@ -828,7 +823,9 @@ async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
     }
 
     // Delete temp VSIX file
-    tempVSIX.removeCallback();
+    if (!tempVSIX) {
+        tempVSIX.removeCallback();
+    }
 }
 
 /**
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -21,7 +21,6 @@ import { getCustomConfigProviders } from './customProviders';
 import { PlatformInformation } from '../platform';
 import { Range } from 'vscode-languageclient';
 import { ChildProcess, spawn, execSync } from 'child_process';
-import * as tmp from 'tmp';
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -770,64 +770,64 @@ async function suggestInsidersChannel(): Promise<void> {
...
 
-function applyUpdate(buildInfo: BuildInfo): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
-        tmp.file({postfix: '.vsix'}, async (err, vsixPath, fd, cleanupCallback) => {
-            if (err) {
-                reject(new Error('Failed to create vsix file'));
-                return;
-            }
+async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
+    const tempVSIX: tmp.FileResult = await util.CreateTempFileWithPostfix('.vsix');
```
bobbrow: I believe you need a try/catch around this.  The helper function always returns a file when it succeeds, but throws an error when it fails (`reject` -> `throw`)

Extension/src/common.ts
```diff
@@ -776,6 +777,17 @@ export function promptReloadWindow(message: string): void {
     });
 }
 
+export async function CreateTempFileWithPostfix(postfix: string): Promise<tmp.FileResult> {
+    return new Promise<tmp.FileResult>((resolve, reject) => {
+        tmp.file({ postfix: postfix }, (err, path, fd, cleanupCallback) => {
+            if (err) {
+                return reject();
```
bobbrow: `reject` will result in a `throw` when awaited.  I think you should call `reject(err)` to pass the `Error` along.

Extension/src/common.ts
```diff
@@ -776,6 +777,17 @@ export function promptReloadWindow(message: string): void {
     });
 }
 
+export async function CreateTempFileWithPostfix(postfix: string): Promise<tmp.FileResult> {
```
bobbrow: This function returns a promise, but is not `async` because `await` is not used in it.

bobbrow (changes requested):

sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -776,6 +777,17 @@ export function promptReloadWindow(message: string): void {
     });
 }
 
+export function CreateTempFileWithPostfix(postfix: string): Promise<tmp.FileResult> {
```
bobbrow: Sorry I missed this the first time.  The function should start with a lower case letter.

Extension/src/LanguageServer/extension.ts
```diff
@@ -770,64 +769,66 @@ async function suggestInsidersChannel(): Promise<void> {
...
-                reject(new Error('Failed to create vsix file'));
-                return;
-            }
+async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
+    const tempVSIX: any = await util.CreateTempFileWithPostfix('.vsix').catch(error => {
+        if (error.message.indexOf('/') !== -1 || error.message.indexOf('\\') !== -1) {
+            error.message = "Potential PII hidden";
+        }
+        telemetry.logLanguageServerEvent('installVsix', { 'error': error.message, 'success': 'false' });
+        return;
```
bobbrow: I'm pretty sure that this just exits the `.catch`, not `applyUpdate`.  That's probably why you had to make `tempVSIX` an `any`.  If you use `try/catch` instead, you can exit within the `catch` block which is what I think you want to do.  If this was intentional, then remove the `return` because it is unnecessary.

bobbrow: Actually, I think you can move the `util.createTempFileWithPostfix` call inside the bigger try/catch below and piggy back on the error handling that sends the telemetry since it's the same code.

bobbrow (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -828,7 +823,9 @@ async function applyUpdate(buildInfo: BuildInfo): Promise<void> {
     }
 
     // Delete temp VSIX file
-    tempVSIX.removeCallback();
+    if (!tempVSIX) {
```
bobbrow: Remove the `!`

bobbrow (changes requested):
Looks like the boolean check was inverted. Fix that and this is ready to go.

Extension/src/LanguageServer/extension.ts
michelleangela: oops, was thinking backwards.. :(

bobbrow (approved):

PR #5423, opened 2020-05-06
Add ARM IntelliSenseMode options to default settings

michelleangela:
This change was missed when ARM IntelliSenseMode options were added to c_cpp_properties.json schema.


Extension/package.json
```diff
@@ -362,7 +362,13 @@
             "clang-x64",
             "msvc-x86",
             "gcc-x86",
-            "clang-x86"
+            "clang-x86",
```
sean-mcmanus: Where is this ordering from? Would it be easier for users to choose from the list if it were sorted alphabetically?

sean-mcmanus (approved):

Extension/package.json
michelleangela: I copied the ordering from the c_cpp_properties.json. I can change the ordering there too to make it alphabetical.

PR #5425, opened 2020-05-06
Add telemetry event for main process crashes

Colengms:


Extension/src/LanguageServer/client.ts
```diff
@@ -1240,6 +1240,7 @@ export class DefaultClient implements Client {
                 closed: () => {
                     languageClientCrashTimes.push(Date.now());
                     languageClientCrashedNeedsRestart = true;
+                    telemetry.logLanguageServerEvent("languageClientCrash", { "count": languageClientCrashTimes.length.toString() }, undefined);
```
bobbrow: I think the count is maxed at 5.  And since the count is lossy (we shift out old crashes), let's just not add a count to the payload at all. We can just count the number of events that fire.

sean-mcmanus: That seems fine, but the count would tell us if it's a rare crash that just occurred just 1 time in a session or a repeated/severe crash that crashed 5 times in 3 minutes or not...not sure if that's important enough to track or not.

sean-mcmanus (approved):

PR #5435, opened 2020-05-08
Add setting to enable/disable code folding provider

Colengms:
Adds a setting.

The provide is disposed when disabled, which causes VS Code to fall back to its default behavior.

I moved the provider class to outside of the constructor.  I will move it to it's own file (as well as others) in a subsequent PR, after 0.28.0 ships and semantic token changes are in.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -529,6 +529,16 @@
           "default": false,
           "description": "%c_cpp.configuration.debugger.useBacktickCommandSubstitution.description%",
           "scope": "window"
+        },
+        "C_Cpp.codeFolding": {
+          "type": "string",
+          "enum": [
+            "Enabled",
+            "Disabled"
+          ],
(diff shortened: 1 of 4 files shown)


Extension/package.json
```diff
@@ -529,6 +529,16 @@
...
+        },
+        "C_Cpp.codeFolding": {
+          "type": "string",
+          "enum": [
+            "Enabled",
+            "Disabled"
+          ],
+          "default": "Enabled",
+          "description": "%c_cpp.configuration.codeFolding.description%",
+          "scope": "resource"
```
sean-mcmanus, line 541: It seems like the scope should be "application"...I can't think of a reason a user would want otherwise.

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #5443, opened 2020-05-08
Fix online vsix error message.

sean-mcmanus:


bobbrow (approved):

sean-mcmanus:
I built a vsix from this branch and it fixes the issue (and the error message still correctly appears for the offline vsix on a mismatched OS).

PR #5446, opened 2020-05-09
Updates for 0.28.0.

sean-mcmanus:


Colengms (approved):

sean-mcmanus:
I'll keep this open in case there are any other quick updates we need for 0.28.0.

sean-mcmanus:
> I'll keep this open in case there are any other quick updates we need for 0.28.0.

Actually, I'll commit it now in case the string change is able to get translated in time (?).

PR #5465, opened 2020-05-12
updateChannel fixes.

sean-mcmanus:
I don't think these fixes need to go into 0.28.0. They're not regressions.


Extension/src/githubAPI.ts
```diff
@@ -181,8 +181,8 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
  * @param updateChannel The user's updateChannel setting.
  * @return The Build if the user should update to it, otherwise undefined.
  */
-export function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string): Build | undefined {
-    if (!vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
+export function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string, isFromSettingsChange: boolean): Build | undefined {
+    if (!isFromSettingsChange && !vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
```
Colengms, line 188: If I understand correctly, this checks the `autoUpdate` setting only if not reacting to a setting change.  So, if the user changes from "Default" to "Insiders" we will now download and install an insiders, even if `autoUpdate` is disabled. (?)  Is that desirable?  Or, is there somewhere else this is short-circuited when `autoUpdate` is disabled?  (In 0.25.1, we had a bug in which autoUpdate was not honored: https://github.com/microsoft/vscode-cpptools/issues/3632 )

bobbrow: I agree with @Colengms .  I don't think this is the desired behavior.  I think the only thing that should be allowed when autoUpdate is false would be downgrading when switching from "Insiders" to "Default".

sean-mcmanus: The "autoUpdate" is intended to be set to false for scenarios where the user doesn't want us automatically updating their versions without explicit consent/actions. When the user switches the channel to Insiders, they're essentially saying they want a one-time manual update to Insiders if one exists, but if autoUpdate is false, then they don't want future Insider updates unless they explicitly change the setting again. i.e. this allows users a mechanism to opt-in to updates on demand (pull updates) versus the current method of getting pushed updates whenever we happen to release. Is there some downside to this behavior I don't understand?

sean-mcmanus: In particular, autoUpdate is a VS Code setting that users may want to have set to false for other extensions to not auto-update, but would still like to have a mechanism to get the C/C++ Insiders.

michelleangela:
What was the issue of udpateChannel? What is the fix?

sean-mcmanus:
> What was the issue of udpateChannel? What is the fix?

There were multiple issues.
1. It doesn't check for a downgrade to non-insiders on startup, only when the setting changes.
2. There was a bug with the timer code that checks for a new insiders every hour.
3. It wasn't bypassing the autoUpdate check when the user manually switches the updateChannel (the thread discussion).

Extension/src/githubAPI.ts
```diff
@@ -182,8 +182,8 @@ export async function getTargetBuildInfo(updateChannel: string): Promise<BuildIn
  * @param updateChannel The user's updateChannel setting.
  * @return The Build if the user should update to it, otherwise undefined.
  */
-export function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string): Build | undefined {
-    if (!vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
+export function getTargetBuild(builds: Build[], userVersion: PackageVersion, updateChannel: string, isFromSettingsChange: boolean): Build | undefined {
```
bobbrow, line 187: If you're adding a new parameter to this function, please add it to the function comments as well.

Extension/src/githubAPI.ts
bobbrow: ok

Extension/src/LanguageServer/extension.ts
```diff
@@ -474,10 +474,15 @@ function realActivation(): void {
             if (vscodeVersion.isGreaterThan(minimumSupportedVersionForInsidersUpgrades, "insider")) {
                 insiderUpdateEnabled = true;
                 if (settings.updateChannel === 'Default') {
-                    suggestInsidersChannel();
+                    const userVersion: PackageVersion = new PackageVersion(util.packageJson.version);
+                    if (userVersion.suffix === "insiders") {
+                        checkAndApplyUpdate(settings.updateChannel, false);
```
bobbrow, line 479: Should this be `true` because there was a settings change since the last time we checked?

bobbrow (approved):
Just a few small things, then this is good to go.

Were we going to add the `minimumSupportedVersionForInsidersUpgrades` to cpptools.json as well so that we're not comparing against a hard coded version of VS Code? Is there an issue for that, or do we need to create one?

Extension/src/githubAPI.ts
sean-mcmanus: Okay, I didn't see those comments.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: No, this is the "onInit" case where we want the autoUpdate check to not be bypassed.

bobbrow: But in this case it would be a "downgrade", so autoUpdate doesn't apply (?)

sean-mcmanus: update could mean upgrade or downgrade, so it seems like we should observe autoUpdate still, which is the existing behavior.

Extension/src/githubAPI.ts
bobbrow: It looks like there are more function comments that should be updated too besides this one.

sean-mcmanus:
> Just a few small things, then this is good to go.
> 
> Were we going to add the `minimumSupportedVersionForInsidersUpgrades` to cpptools.json as well so that we're not comparing against a hard coded version of VS Code? Is there an issue for that, or do we need to create one?

You created an issue for that at https://github.com/microsoft/vscode-cpptools/issues/4873 . I was going to do that separately.

bobbrow (approved):

Issue #5464, opened 2020-05-11
Changing to "Default" updateChannel isn't downgrading from Insiders


sean-mcmanus:
Repros with 0.28.0-insiders3...but when I debug it, it works.

sean-mcmanus:
There's a related existing issue in that it doesn't do the downgrade check on activation, only after a settings change.

PR #5466, opened 2020-05-12
Fix a bug with downgrading

bobbrow:
Addresses: #5464


sean-mcmanus (approved):

bobbrow:
I committed this, but it doesn't have to go into 0.28.0 as it will have no effect until 0.29.0-insiders.

PR #5488, opened 2020-05-13
Fix code folding cancel error.

### LINKED ISSUE #5487: Adopt use of CancellationError when rejecting requests to providers cancelled using CancellationToken
[2020-05-13 14:04:53.007] [exthost] [error] [ms-vscode.cpptools] provider FAILED
[2020-05-13 14:04:53.007] [exthost] [error] undefined

I used the repro at https://github.com/microsoft/vscode-cpptools/issues/5481 and type "style::".

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/5487

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2,7 +2,7 @@
   "name": "cpptools",
   "displayName": "C/C++",
   "description": "C/C++ IntelliSense, debugging, and code browsing.",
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
Sounds like there's a better fix...

PR #5510, opened 2020-05-16
Join string arrays in `env` with the correct delimiter

### LINKED ISSUE #5509: String arrays in `env` are not joined properly
**Type: LanguageService**

**Describe the bug**
- OS and Version: macOS 10.15.4
- VS Code Version: 1.45.1
- C/C++ Extension Version: 0.28.0

In #5229 environment variables were changed to be split using a platform-specific delimiter (`;` on Windows and `:` for everything else, like Linux and macOS). However, joining arrays of strings in the `env` variable in `c_cpp_properties.json` was not properly updated to match this new behavior, causing such variables to be incorrectly parsed when substituted in configurations.

This causes bugs like lists of include directories to not resolve properly when used in multiple configurations on non-Windows machines.

**Steps to reproduce**
1. Add a variable containing an array of strings to the `env` section of `c_cpp_properties.json`.
1. Reference this variable in a configuration.

```json
{
    "env": {
        "commonIncludes": [
            "${workspaceFolder}/path/to/dir1",
            "${workspaceFolder}/path/to/dir2"
        ]
    },
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${commonIncludes}"
            ]
        }
    ],
    "version": 4
}
```

**Incorrect behavior**
```
-------- Diagnostics - 5/15/2020, 10:07:43 PM
Version: 0.28.0
Current Configuration:
{
    "name": "Mac",
    "includePath": [
        "${workspaceFolder}/path/to/dir1;${workspaceFolder}/path/to/dir2"
    ],
    "compilerArgs": [],
    "browse": {
        "path": [
            "${workspaceFolder}/path/to/dir1;${workspaceFolder}/path/to/dir2",
            "${workspaceFolder}"
        ],
    }
}
No active translation units.
```

**Expected behavior**

```
-------- Diagnostics - 5/15/2020, 10:09:10 PM
Version: 0.27.1
Current Configuration:
{
    "name": "Mac",
    "includePath": [
        "${workspaceFolder}/path/to/dir1",
        "${workspaceFolder}/path/to/dir2"
    ],
    "compilerArgs": [],
    "browse": {
        "path": [
            "${workspaceFolder}/path/to/dir1",
            "${workspaceFolder}/path/to/dir2",
            "${workspaceFolder}"
        ],
    }
}
No active translation units.
```

kersson:
Fixes #5509.


sean-mcmanus (approved):

Colengms (approved):

Extension/src/common.ts
```diff
@@ -339,6 +339,7 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
     let regexp: () => RegExp = () => /\$\{((env|config|workspaceFolder)(\.|:))?(.*?)\}/g;
     let ret: string = input;
     let cycleCache: Set<string> = new Set();
+    const delimiter: string = (process.platform === 'win32') ? ";" : ":";
```
michelleangela: For better maintainability and since this code is repeated elsewhere, let's create a function in `common.ts` that returns delimiter based on a string input of the platform. This way, only the function needs to be updated if delimiter ever changes again.

Example
```
export function getEnvDelimiter(platform: string): string {
    return (platform === 'win32') ? ";" : ":";
}
```

michelleangela (changes requested):

Extension/src/common.ts
kersson: Where's the best place to put this function? Sorry, not super familiar with the codebase. In `Extension/src/common.ts` or somewhere else?

sean-mcmanus: Yes, I think Extension/src/common.ts.

kersson: Since the delimiter won't change, how about this instead of a function?
```ts
// Platform-specific environment variable delimiter
export const envDelimiter: string = (process.platform === 'win32') ? ";" : ":";
```

sean-mcmanus: That seems good to me.

michelleangela (approved):

PR #5513, opened 2020-05-16
Handle quoted spaces when stripping args from compilerPath

Colengms:
Previously, args were being split from the compiler path based solely on spaces.

This change correctly handles compiler command lines such as:

C:\PROGRA~1\LLVM\bin\clang-cl.exe -DTEST1="blah 1" -DTEST2="\"blah 2\""  ....

With json escaping, i.e. :

"compilerPath": "C:\\PROGRA~1\\LLVM\\bin\\clang-cl.exe -DTEST1=\"blah 1\" -DTEST2=\"\\\"blah 2\\\"\"",


sean-mcmanus:
Is this for 0.28.1?

Colengms:
> Is this for 0.28.1?

It's pretty minor, but it doesn't need to be in 0.28.1.

sean-mcmanus (approved):

PR #5614, opened 2020-06-05
Update strings and fwlinks

sean-mcmanus:


Extension/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "cpptools",
   "displayName": "C/C++",
   "description": "C/C++ IntelliSense, debugging, and code browsing.",
-  "version": "0.28.2-master",
+  "version": "0.28.3-master",
```
Colengms: Should this be 0.29.0, or do we anticipate another patch?

sean-mcmanus: I anticipate it will be 0.29.0, but I didn't want to increase it too early. However, it should be fine to use 0.29.0, and then we could still decrease it later if there's a reason to.

Colengms (approved):

PR #5634, opened 2020-06-10
Add prefer-const to eslint. Fix some other lint errors.

sean-mcmanus:
Most of these changes were automatically made by running "eslint ** --fix". "const" variables in TypeScript are still mutable (e.g. arrays can be pushed to), they just can't be re-assigned. See https://eslint.org/docs/rules/prefer-const .


sean-mcmanus:
I wasn't planning to check this in until the other existing PRs have gone in, to avoid breaking the PRs.

michelleangela (approved):

PR #5639, opened 2020-06-12
Add 'quoteArgs' to launch.json schema

WardenGnaw:
With https://github.com/microsoft/MIEngine/pull/1006, the debug adapter
supports 'quoteArgs' in pipeTransports. If the user wishes to handle the
quoting for pipeTransport arguments, they should set this field to
'false'. It currently defaults to true.

For example, if quoteArgs is 'true', the pipe command will result in
`bash.exe -c '/usr/bin/gdb --interpreter=mi'` If false,
it will result to `bash.exe -c /usr/bin/gdb --interpreter=mi`.
In this example, the latter will not work, but will be needed for a pipeTransport command like `wsl.exe`

### DIFF
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -77,6 +77,7 @@
     "c_cpp.debuggers.pipeTransport.pipeProgram.description": "The fully qualified pipe command to execute.",
     "c_cpp.debuggers.pipeTransport.pipeArgs.description": "Command line arguments passed to the pipe program to configure the connection.",
     "c_cpp.debuggers.pipeTransport.pipeEnv.description": "Environment variables passed to the pipe program.",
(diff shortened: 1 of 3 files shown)


Extension/package.nls.json
```diff
@@ -77,6 +77,7 @@
     "c_cpp.debuggers.pipeTransport.pipeProgram.description": "The fully qualified pipe command to execute.",
     "c_cpp.debuggers.pipeTransport.pipeArgs.description": "Command line arguments passed to the pipe program to configure the connection.",
     "c_cpp.debuggers.pipeTransport.pipeEnv.description": "Environment variables passed to the pipe program.",
+    "c_cpp.debuggers.pipeTransport.quoteArgs.description": "If individual arguments contain characters (such as spaces or tabs), should it be quoted? If 'false', the debugger command will no longer be automatically quoted. \nDefault is 'true'.",
```
pieandcakes: I think we talked about rewording it to `If the pipeProgram's individual arguments...`

pieandcakes (approved):

sean-mcmanus (approved):

PR #5644, opened 2020-06-15
Print launch.json to C/C++ Output Log

WardenGnaw:
This PR adds a check in resolveDebugConfiguration so that if a user adds
logging.engineLogging = true in the launch.json, we will output the
resovled json in the C/C++ Output.

@bobbrow This will be useful for CMake extension if we have debug launch issues there.


sean-mcmanus (approved):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -188,85 +188,94 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
 	 */
     resolveDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration> {
-        if (config) {
-            if (config.type === 'cppvsdbg') {
-                // Fail if cppvsdbg type is running on non-Windows
-                if (os.platform() !== 'win32') {
-                    logger.getOutputChannelLogger().showWarningMessage(localize("debugger.not.available", "Debugger of type: '{0}' is only available on Windows. Use type: '{1}' on the current OS platform.", "cppvsdbg", "cppdbg"));
-                    return undefined;
-                }
+        // if config or type is not specified, return null to trigger VS Code to open a configuration file https://github.com/Microsoft/vscode/issues/54213
```
pieandcakes: use github bug notation and put bug at the front in brackets might make it easierl
`[Microsoft/vscode#54213] If config or type...`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -188,85 +188,94 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
...
-                }
+                return undefined;
             }
         }
-        // if config or type is not specified, return null to trigger VS Code to open a configuration file https://github.com/Microsoft/vscode/issues/54213
-        return config && config.type ? config : null;
+
+        if (config.logging?.engineLogging) {
+            const outputChannel: logger.Logger = logger.getOutputChannelLogger();
+            outputChannel.appendLine(JSON.stringify(config, undefined, 2));
```
pieandcakes, line 275: do you want to preface this with` Launch configuration` or something so it doesn't seem like you are just dumping a blob of json into the output window. Similar to what we do with the XML in MIEngine.

pieandcakes (approved):

Issue #5370, opened 2020-04-27
Localize some phrases/words in autocomplete info


michelleangela:
Localize following words/phrases:

- "preprocessor keyword"
![image](https://user-images.githubusercontent.com/38734287/80421599-946c9180-8891-11ea-93cb-145a72003771.png)

- "overloads"
![image](https://user-images.githubusercontent.com/38734287/80421648-a9e1bb80-8891-11ea-925d-f7f9de8d9a78.png)

- "specialization"

Colengms:
Since these aren't log messages, we're either going to need a new request/response so the native code can ask the TypeScript side to provide a localized string, or we'll need to reroute completion through the equivalent TypeScript provider class to allow the TypeScript side an opportunity to do localized string replacement before returning the results.

Colengms:
Since it's going to require the same solution, merging this issue with https://github.com/microsoft/vscode-cpptools/issues/5340.

PR #5651, opened 2020-06-17
Additional localized strings for hover and auto-complete

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/5370

 * Adds new stings that needed to be localized
 * Sends all native localized strings to native process during initialization
   * Note that strings requiring substition will include placeholders (i.e. "{0}").
 * Adds support for hints in nativeStrings.json


bobbrow:
Does this mean we have to add support for `{#}` substitution in the native portion of the extension?  Could we change it to `%s` or `%d` in the appropriate cases to save us a step?

Extension/src/nativeStrings.json
```diff
@@ -13,7 +13,6 @@
     "could_not_find_compile_commands": "\"{0}\" could not be found. 'includePath' from c_cpp_properties.json in folder '{1}' will be used instead.",
     "file_not_found_in_path": "\"{0}\" not found in \"{1}\". 'includePath' from c_cpp_properties.json in folder '{2}' will be used for this file instead.",
     "database_reset": "The IntelliSense database was successfully reset.",
-    "global_scope": "",
```
sean-mcmanus: Would it be worth keeping this around to avoid breaking strings for any users who use the master branch and have the 0.28.1 binaries still?

Extension/gulpfile.js
```diff
@@ -444,36 +444,49 @@ gulp.task("generate-native-strings", (done) => {
...
         
         // Add to native string table
-        nativeStringTableContent += `    ${JSON.stringify(stringTable[property])},\n`;
+        nativeStringTableContent += `    ${JSON.stringify(stringValue)},\n`;
 
         // Add to TypeScript switch
         // Skip empty strings, which can be used to prevent enum/index reordering
-        if (stringTable[property] != "") {
+        if (stringValue != "") {
             // It's possible that a translation may skip "{#}" entries, so check for up to 20 of them.
```
sean-mcmanus: Change comment to 50?

sean-mcmanus (approved):

PR #5652, opened 2020-06-17
[cppvsdbg] Add logging ThreadExit and ProcessExit

### LINKED ISSUE #1017: [cppvsdbg] Option to disable thread exit logging in launch.json
_From @akbyrd on September 7, 2017 21:20_

Desired Behavior:

I'd like the ability to disable thread exit messages in the Debug Console. Options already exist for other types of messages such as module loads (under logging.moduleLoad), so this would simply be another one added to the list.

vscode thread exit messages

- VSCode Version: Code - Insiders 1.17.0-insider (ac167c41bababe6c47180012723fa2bc52610ff4, 2017-09-07T05:24:04.020Z)
- OS Version: Windows_NT x64 10.0.15063
- Extensions: 

Extension|Author (truncated)|Version
---|---|---
python|don|0.7.0
Lua|key|0.0.9
cpptools|ms-|0.12.3
code-settings-sync|Sha|2.8.3

---

Steps to Reproduce:

1. Launch a C++ program under the debugger
2. Notice "The thread #### has exited with code 0 (0x0)" messages in the Debug Console

_Copied from original issue: Microsoft/vscode#33965_

WardenGnaw:
This PR adds two log flags for ThreadExit and ProcessExit events.
This will remove the "The thread 10564 has exited with code 0 (0x0)."
and "The program '[24104] ConsoleApplication1.exe' has exited with code
0 (0x0)." messages.


Extension/package.json
```diff
@@ -1353,6 +1353,16 @@
                     "type": "boolean",
                     "description": "%c_cpp.debuggers.logging.engineLogging.description%",
                     "default": false
+                  },
+                  "threadExit": {
+                    "type": "boolean",
+                    "description": "%c_cpp.debuggers.logging.threadExit.description%",
+                    "default": false
+                  },
+                  "processExit": {
```
sean-mcmanus, line 1362: Can the name be more descriptive like threadExitLogging, processExitLogging?

WardenGnaw: I don't think that would be good since it would be redundant, because it would look like:
```
"logging": {
   "threadExitLogging": true
   "processExitLogging": false
}
```

sean-mcmanus: Ah, okay. I was looking at engineLogging and forgot about the surrounding "logging".

sean-mcmanus (approved):

Extension/package.json
```diff
@@ -1353,6 +1353,16 @@
                     "type": "boolean",
                     "description": "%c_cpp.debuggers.logging.engineLogging.description%",
                     "default": false
+                  },
+                  "threadExit": {
+                    "type": "boolean",
+                    "description": "%c_cpp.debuggers.logging.threadExit.description%",
+                    "default": false
```
pieandcakes, line 1360: why is this default `false` but the processExit `true` ? by default we should be printing all these messages and allow the user to override with `false`. In the past the `default` has been what we have it set internally as `default` if the value isn't set.

Extension/package.nls.json
```diff
@@ -86,6 +86,8 @@
     "c_cpp.debuggers.logging.engineLogging.description": "Optional flag to determine whether diagnostic debug engine messages should be logged to the Debug Console. Defaults to false.",
     "c_cpp.debuggers.logging.trace.description": "Optional flag to determine whether diagnostic adapter command tracing should be logged to the Debug Console. Defaults to false.",
     "c_cpp.debuggers.logging.traceResponse.description": "Optional flag to determine whether diagnostic adapter command and response tracing should be logged to the Debug Console. Defaults to false.",
+    "c_cpp.debuggers.logging.threadExit.description": "Controls if a message is logged when a thread in the target process exits. Default: `false`.",
```
pieandcakes: copy the same text format from ModuleLoad or Exceptions (above)

pieandcakes (approved):

Extension/package.json
sean-mcmanus: I assumed that an app could have tons of threads exiting, so that could be spammy to show by default, while the process exiting is only supposed to occur one time?

Issue #5285, opened 2020-04-12
Default 'preLaunchTask' in launch.json is not the same as 'label' in tasks.json


ghost:
Type: Debugger

**Environment**
- OS and Version: Windows 10
- VS Code Version: 1.44.0
- C/C++ Extension Version:  0.27.0: March 30, 2020

**Bug Describe**

Default 'preLaunchTask' in launch.json is not the same as 'label' in tasks.json
- "preLaunchTask": "g++.exe build active file"
- "label": "C/C++: g++.exe build active file"

Users have to change one of them by themselves. It will be good if them are default the same, like both with "C/C++:" start.

jmarmstrong1207:
I have this issue too with WSL Ubuntu & VSCode. I was just about to make a bug report about this until I found this one

bobbrow:
@elahehrashedi did you fix this with your build tasks change?

sean-mcmanus:
I'm not reproing this with 1.0.1 (prelaunch tasks has "C/C++: ...".

elahehrashedi:
This is fixed in the new build task changes. Will be available in C/C++ Extension Version: 1.1.0

ghost:
Thanks for your work!

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/1.1.0-insiders .

Issue #5270, opened 2020-04-09
Do not create tasks.json when creating debug configuration for gcc build active file


NaLaDcP:
Type: Debugger

**Describe the bug**
- OS and Version:Ubuntu 18.04.4
- VS Code Version:  1.43.2
- C/C++ Extension Version: 0.27.0
- Other extensions you installed (and if the issue persists after disabling them): Python and LiveShare, issue persistts after disabling the extensions
- A clear and concise description of what the bug is.

*Please include a code sample and `launch.json` configuration.*
`
[launch.json.txt](https://github.com/microsoft/vscode-cpptools/files/4454797/launch.json.txt)

`
Steps to Reproduce:

1. Install C/C++ extension ms-vscode.cpptools
2. Start debugging
3. Select environment: C++(GDB/LLDB)
4. Select gcc run and debug active file

A launch.json file is created but no tasks.json. So it does not compile before starting the debugger. I tried to do it by hand by going to Menu -> Terminal -> Configure tasks -> Create tasks.json file from template I do not have such option for creating tasks.json for 'gcc build active file'

**Additional context**
*If applicable, please include logging by adding "logging": { "engineLogging": true, "trace": true, "traceResponse": true } in your `launch.json`*
Add any other context about the problem here including log or error messages in your Debug Console or Output windows.

bobbrow:
@elahehrashedi did you fix this with your build tasks change?

elahehrashedi:
This is fixed in the new build task changes. Will be available in C/C++ Extension Version: 1.1.0.

sean-mcmanus:
With https://github.com/microsoft/vscode-cpptools/releases/tag/1.1.0-insiders this is fixed for tasks.json, but not launch.json -- seems like VS Code doesn't yet have a way to fix the launch.json case, other than us adding text to the end of the launch.json name/lable.

Issue #4278, opened 2019-09-19
"Run and Debug" and "F5" doesn't show the compilers detail in the API


ranolfi:
Type: General

- OS and Version: **Arch Linux**
- VS Code Version: **Code-OSS 1.37.1**
- C/C++ Extension Version: **0.25.1**

When we have a workspace folder open, in the Debug panel when clicking the Add Configuration option and then select _C++ (GDB/LLDB)_, some options with the same name are offered:

![image](https://user-images.githubusercontent.com/8507649/65284259-69f27300-db0f-11e9-9db6-8664905b02fa.png)

I tried all three "_g++ build and debug active file_". Options 1 and 2 generates the same `launch.json` and `tasks.json`; option 3 creates a slightly different `launch.json` file (`"miDebuggerPath": "/bin/gdb"` instead of `"/usr/bin/gdb"`).

In any case, this is confusing.

elahehrashedi:
this issue is addressed in this PR: https://github.com/microsoft/vscode-cpptools/pull/5659

elahehrashedi:
This is fixed in the new build task changes. Will be available in C/C++ Extension Version: 1.1.0

sean-mcmanus:
@elahehrashedi I'm still reproing this with 1.1.0-insiders. What's supposed to happen?
![image](https://user-images.githubusercontent.com/19859882/95529692-7357b300-0990-11eb-9fa7-703da258c9fe.png)

sean-mcmanus:
That screenshot is from F5 with no tasks.json generated.

sean-mcmanus:
I see it fixed for Configure Default Build Task.

elahehrashedi:
@sean-mcmanus can you please use the vscode-insiders so that I can see the details of each of those compilers?

sean-mcmanus:
Yeah, it doesn't show the details with vscode-insiders. Use F5 (or Build and Debug Active File).

elahehrashedi:
The user is getting this screen after running "Run and debug" or pressing "F5". I fixed this issue and sent a PR.
![image](https://user-images.githubusercontent.com/16143955/96952776-7bd9de80-14a4-11eb-89ed-58c38593ab91.png)
the paths are not visible:
![image](https://user-images.githubusercontent.com/16143955/96953165-4da8ce80-14a5-11eb-9c8f-1eac785fe0ea.png)
[...]

Issue #3674, opened 2019-05-21
CTRL+SHIFT+B no longer gives the option to set up a build task when a cpp file is open


dunhor:
**Type: LanguageService**
When I do the key combo 'CTRL+SHIFT+B' on a new directory (not configured for any build task), it would previously prompt me to set up a new task. This was wonderful since I likely wanted to invoke cmake/ninja/etc. It is now, however, only presenting me the option to invoke cl.exe on the active file. While I can still go in and manually set up a new build task, it was very convenient to do everything the old way. This could also be an editor change unrelated to the extension, but it's still annoying nonetheless. FWIW if the active file is not a C++ file, then it presents me with the old experience, so I suppose at least I still have that

**Describe the bug**
- OS and Version: Windows Insiders (but hit this on 1803 machine at home)
- VS Code Version: 1.35.0-insider
- C/C++ Extension Version: 0.23.1

sean-mcmanus:
Yeah, we recently changed it to add those default build tasks -- maybe we can add the "default" that was previously shown to the list as well, so the previous behavior is still obtainable.

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/1.1.0-insiders ,

PR #5659, opened 2020-06-18
Task Provider Feature

elahehrashedi:
https://github.com/microsoft/vscode-cpptools/issues/3674 : Add support to run cpp build tasks (Tasks: Run Build Task)

Will also fix these bugs: 
https://github.com/microsoft/vscode-cpptools/issues/4278
https://github.com/microsoft/vscode-cpptools/issues/5285
https://github.com/microsoft/vscode-cpptools/issues/4278
https://github.com/microsoft/vscode-cpptools/issues/5270

Edited:
Refactoring the Task tools: **_Task: Configure Task_** and **_Task: Run Build Task_**, and **_C/C++: Build and Debug Active File_**
The multi-root is not implemented in this version yet. 
You have to use the insider Vs Code to test the proposed API for this feature (1.49.0-insider).

These are the behavior of the current version:

when **_Task: Run Build Task_** is called:
1. If there is a tasks.json file exists:
- >> if there is a default build task: the default build task will be run (no task will be shown to the user to choose from).
image
- >> if there is NO default task: The list of all defined tasks in tasks.json will be shown to the user to choose from.
image

2. If tasks.json file doesn't exist:
- >> the detected build tasks will be shown to the user to choose from. no tasks.json will be created.
image

P.S. right now only **_Task: Configure Task_** and **_C/C++: Build and Debug Active File_** can create a tasks.json file for the first time.


Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        const fileExt: string = path.extname(editor.document.fileName);
+        if (!fileExt) {
+            return [];
+        }
+
+        // Don't offer tasks for header files.
+        const fileExtLower: string = fileExt.toLowerCase();
+        const isHeader: boolean = !fileExt || [".hpp", ".hh", ".hxx", ".h", ".inl", ""].some(ext => fileExtLower === ext);
```
Colengms: I know you didn't write this part originally, but it looks like the native code (settings.h, line ~80-88) is also tracking ".h++", ".hp", ".i", ".ii", ".ino" and ".ipp" header extensions, and ".c++", ".cp", and ".tcc" source file extensions.  Could you add those here?   (It also associates ".idl" with C++, but idl isn't actually C++, so I'm not sure about that one.)

Extension/package.json
```diff
@@ -40,9 +40,35 @@
   ],
   "activationEvents": [
     "*"
-  ],
+  ],
```
sean-mcmanus: extra space

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+        task.group = vscode.TaskGroup.Build;
+
+        return task;
+    };
+}
+
+class CustomBuildTaskTerminal implements vscode.Pseudoterminal {
+    private writeEmitter  = new vscode.EventEmitter<string>();
+    // eslint-disable-next-line no-invalid-this
+    onDidWrite: vscode.Event<string> = this.writeEmitter.event;
```
Colengms: Perhaps implement these are properties, so we don't have to disable lint errors for them?  i.e. public get onDidWrite(): vscode.Event<string> { return this.writeEmitter.event; }

Extension/package.json
```diff
@@ -40,9 +40,35 @@
...
+  ], 
   "main": "./dist/main",
   "contributes": {
+    "taskDefinitions": [
+			{
+				"type": "cppbuild",
+				"required": [
+					"command"
+				],
+				"properties": {
```
sean-mcmanus: Can you format the doc? Seems like tabs are used.

Extension/package.json
```diff
@@ -40,9 +40,35 @@
...
+						"type": "string",
+						"description": "The compiler path."
+					},
+					"args": {
+						"type": "array",
+						"description": "Additional build args."
+          },
+          "options": {
+            "type": "object",
+						"description": "Options {\"cwd\": cwd}."
```
sean-mcmanus: Does this need to go in an options object or can a "cwd" property be added directly?

Extension/package.nls.json
```diff
@@ -18,7 +18,7 @@
...
-    "c_cpp.configuration.clang_format_style.description": "Coding style, currently supports: Visual Studio, LLVM, Google, Chromium, Mozilla, WebKit. Use \"file\" to load the style from a .clang-format file in the current or parent directory. Use {key: value, ...} to set specific parameters. For example, the \"Visual Studio\" style is similar to: { BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Allman, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All }",
+    "c_cpp.configuration.clang_format_style.description": "Coding style, currently supports: Visual Studio, LLVM, Google, Chromium, Mozilla, WebKit. Use \"file\" to load the style from a .clang-format file in the current or parent directory. Use {key: value, ...} to set specific parameters. For example, the \"Visual Studio\" style is similar to: { BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Allman, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All, Standard: Cpp03 }",
```
sean-mcmanus: I don't think this change should be here. The Cpp03 is a special case and not the default.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -7,7 +7,7 @@ import * as debugUtils from './utils';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { getBuildTasks, BuildTaskDefinition } from '../LanguageServer/extension';
+import { CppBuildTaskProvider, CppBuildTaskDefinition} from '../LanguageServer/cppbuildTaskProvider';
```
sean-mcmanus: cppBuildTaskProvider instead? Or is the cppbuildTaskProvider  casing intentional to match the cppbuild type?

Extension/package.json
```diff
@@ -40,9 +40,35 @@
   ],
   "activationEvents": [
     "*"
-  ],
+  ], 
   "main": "./dist/main",
   "contributes": {
+    "taskDefinitions": [
+			{
+				"type": "cppbuild",
```
bobbrow: Can you format this file?  It should only have spaces in it, but I see TABs.

Extension/package.nls.json
bobbrow: This looks like it was meant to be part of one of your previous clang-format changes.  I think that commit was incorrectly added to this branch.

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+            return resolve({});
+        }
+        fs.exists(path, exists => {
+            if (!exists) {
+                return resolve({});
+            }
+            let fileContents: string = fs.readFileSync(path).toString();
+            fileContents = fileContents.replace(/^\s*\/\/.*$/gm, ""); // Remove start of line // comments.
+            let rawTasks: any = {};
+            try {
```
sean-mcmanus: Would it be easy to switch to jsonc.parse instead? colorization.ts uses that. That would enable the comment removal code to be omitted. We might have another issue tracking that.

sean-mcmanus (changes requested):
I'm not finished reviewing...

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+
+export class CppBuildTaskProvider implements vscode.TaskProvider {
+    static CppBuildScriptType: string = 'cppbuild';
+    static CppBuildSourceStr: string = "C/C++";
+    private tasks: vscode.Task[] | undefined;
+
+    constructor() {}
+
+    public async provideTasks(): Promise<vscode.Task[]> {
+        if (this.tasks !== undefined) {
```
michelleangela: Can `if (this.tasks !== undefined)` be `if (this.tasks)` ? #Resolved

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    // Resolves a task that has no [`execution`](#Task.execution) set.
+    public resolveTask(_task: vscode.Task): vscode.Task | undefined {
+        const execution: vscode.ProcessExecution | vscode.ShellExecution | vscode.CustomExecution | undefined = _task.execution;
+        if (execution === undefined) {
+            const definition: CppBuildTaskDefinition = <any>_task.definition;
+            return this.getTask(definition.command, false, false, definition.args ? definition.args : [], definition);
+        }
+        return undefined;
```
michelleangela: Just trying to understand what function does. Why does it return undefined if `(execution === undefined)` is false?

Extension/src/LanguageServer/extension.ts
```diff
@@ -47,12 +47,10 @@ let realActivationOccurred: boolean = false;
 let tempCommands: vscode.Disposable[] = [];
 let activatedPreviously: PersistentWorkspaceState<boolean>;
 let buildInfoCache: BuildInfo | undefined;
-const taskTypeStr: string = "shell";
-const taskSourceStr: string = "C/C++";
 const cppInstallVsixStr: string = 'C/C++: Install vsix -- ';
 let taskProvider: vscode.Disposable;
 let codeActionProvider: vscode.Disposable;
-const intelliSenseDisabledError: string = "Do not activate the extension when IntelliSense is disabled.";
+export const intelliSenseDisabledError: string = "Do not activate the extension when IntelliSense is disabled.";
```
michelleangela, line 56: Is this a user facing error that needs to be localized?

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+        }
+        const options: cp.ProcessEnvOptions | undefined = {"cwd": cwd};
+
+        // Double-quote the command if it is not already double-quoted.
+        let resolvedcompilerPath: string = isCl ? compilerPathBase : compilerPath;
+        if (resolvedcompilerPath && !resolvedcompilerPath.startsWith("\"") && resolvedcompilerPath.includes(" ")) {
+            resolvedcompilerPath = "\"" + resolvedcompilerPath + "\"";
+        }
+
+        if (definition === undefined) {
```
michelleangela: This can be `'if (!definition)`

sean-mcmanus:
I'm seeing an issue when there are spaces in the ${file} path. I'm not able to get mingw or cl.exe to work. When the workspace folder is named "New folder", I get `cl : Command line warning D9024 : unrecognized source file type 'folder', object file assumed` . That scenario works with 0.28.3.

Extension/src/LanguageServer/cppbuildTaskProvider.ts
```diff
@@ -0,0 +1,320 @@
...
+            this.tasks.push(this.getTask(userCompilerPath, returnCompilerPath, appendSourceToName, userCompilerPathAndArgs?.additionalArgs));
+        }
+
+        return this.tasks;
+    }
+
+    private getTask: (compilerPath: string, returnCompilerPath: boolean, appendSourceToName: boolean, compilerArgs?: string [], definition?: CppBuildTaskDefinition) => vscode.Task = (compilerPath: string, returnCompilerPath: boolean, appendSourceToName: boolean, compilerArgs?: string [], definition?: CppBuildTaskDefinition) => {
+
+        const filePath: string = path.join('${fileDirname}', '${fileBasenameNoExtension}');
+        const compilerPathBase: string = path.basename(compilerPath);
```
sean-mcmanus: extra blank line?

sean-mcmanus:
Build errors are not reported:
```
Build finished with error:
        Command failed: 
```
I think we would need to implement our own problem matcher to report the errors and/or forward the results to the terminal?

sean-mcmanus:
Also, I'm seeing unusual tab and newline behavior in the terminal output with cl.exe.

image

This is the 0.28.3 result:
Microsoft (R) C/C++ Optimizing Compiler Version 19.27.28919.3 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

test2.c
...\Test1\test2.c(5): error C2143: syntax error: missing ';' before '}'

Extension/src/LanguageServer/cppbuildTaskProvider.ts
elahehrashedi: I modify the header and source file set as below:
["*.hpp", "*.hh", "*.hxx", "*.h++", "*.hp", "*.h", "*.ii", "*.ino", "*.inl", "*.idl", ""]
["*.cpp", "*.cc", "*.cxx", "*.c++", "*.cp", "*.ipp", "*.tcc"]

Extension/package.json
elahehrashedi: there might be other properties added such as cwd, env, and shell.

Extension/src/LanguageServer/cppbuildTaskProvider.ts
elahehrashedi: As far as I know, we can implement this function to return only undefined and the behaviour of the task feature won't change for now.
The tasks defined in tasks.json file pass-through this function to be resolved (the blanks are filled-out I guess).

Extension/src/LanguageServer/extension.ts
elahehrashedi: I am not sure. I had to use this error in cppBuildTaskProvider, so I exported it.

sean-mcmanus (changes requested):
https://github.com/microsoft/vscode-cpptools/pull/5659#issuecomment-646374058
https://github.com/microsoft/vscode-cpptools/pull/5659#issuecomment-646385834
https://github.com/microsoft/vscode-cpptools/pull/5659#issuecomment-646387500

elahehrashedi:
- A sample results of a compiler error (the file name includes Space as well):
image
image

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,357 @@
...
+    private fileWatcher: vscode.FileSystemWatcher | undefined;
+
+
+    constructor(private command: string, private args: string[], private options: cp.ExecOptions | undefined, private workspaceRoot: string) {
+    }
+
+
+    open(_initialDimensions: vscode.TerminalDimensions | undefined): void {
+        telemetry.logLanguageServerEvent("cppBuildTaskStarted");
+        const pattern: string = path.join(this.workspaceRoot, 'cppBuild');
```
bobbrow: This file watcher stuff appears to be an artifact of the extension sample.  I think it can be safely removed.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,357 @@
...
+                    temp = "\"" + temp + "\"";
+                }
+                activeCommand = activeCommand + " " + temp;
+            });
+            if (this.options?.cwd) {
+                this.options.cwd = util.resolveVariables(this.options.cwd, this.AdditionalEnvironment);
+            }
+            cp.exec(activeCommand, this.options, (_error, stdout, _stderr) => {
+                if (_error) {
+                    telemetry.logLanguageServerEvent("cppBuildTaskError", { "error": _error.message });
```
bobbrow: I don't think we can safely log the "_error.message". This might have PII in it.  Let's just log the event name by itself instead.

Extension/package.json
```diff
@@ -43,6 +43,33 @@
...
+            "type": "string",
+            "description": "Task's name."
+          },
+          "command": {
+            "type": "string",
+            "description": "The compiler path."
+          },
+          "args": {
+            "type": "array",
+            "description": "Additional build args."
```
michelleangela: >build [](start = 39, length = 5)

Are these args used for compiling? If so, describe as "Additional compiler args." ?

Extension/src/LanguageServer/extension.ts
michelleangela: I see that this error is thrown in function `realActivation()` and was never localized. May not be user facing then.

michelleangela:
Are there any associated issues to this PR?

sean-mcmanus (approved):

Extension/package.json
```diff
@@ -43,6 +43,33 @@
...
+      {
+        "type": "cppbuild",
+        "required": [
+          "command",
+          "label"
+        ],
+        "properties": {
+          "label": {
+            "type": "string",
+            "description": "Task's name."
```
bobbrow: We need to localize these

Extension/package.json
michelleangela: What was the resolution to this question?

Extension/package.json
bobbrow: Description texts:
`name` - "The name of the task."
`command` - "The path to either a compiler or script that performs compilation."
`args` - "Additional arguments to pass to the compiler or compilation script."

Extension/package.json
```diff
@@ -43,6 +43,33 @@
...
+          },
+          "command": {
+            "type": "string",
+            "description": "The compiler path."
+          },
+          "args": {
+            "type": "array",
+            "description": "Additional compiler args."
+          },
+          "options": {
```
bobbrow, line 65: What is `options`?

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+                    resolve(0);
+                }
+            });
+        }).finally (() => {
+            this.closeEmitter.fire(0);
+        });
+    }
+
+    private get AdditionalEnvironment(): { [key: string]: string | string[] } | undefined {
+
```
bobbrow: remove empty newline

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+class CustomBuildTaskTerminal implements vscode.Pseudoterminal {
+    private writeEmitter  = new vscode.EventEmitter<string>();
+    private closeEmitter = new vscode.EventEmitter<number>();
+    public get onDidWrite(): vscode.Event<string> { return this.writeEmitter.event; }
+    public get onDidClose(): vscode.Event<number> { return this.closeEmitter.event; }
+
+    constructor(private command: string, private args: string[], private options: cp.ExecOptions | undefined) {
+    }
+
```
bobbrow, line 303: remove 1 empty newline here

Extension/package.json
elahehrashedi: some build options like "cwd" and "env" goes to this part.

Extension/package.json
```diff
@@ -43,6 +43,33 @@
...
+          "label"
+        ],
+        "properties": {
+          "label": {
+            "type": "string",
+            "description": "Task's name."
+          },
+          "command": {
+            "type": "string",
+            "description": "The compiler path."
```
Colengms: Elsewhere in pacakge.json, you will see settings, etc., that have `description` fields containing `%`'s.  i.e. `"description": "%c_cpp.configuration.clang_format_style.description%"`.  These refer to localized strings, with the English version of the string specified in package.nls.json.  I believe you can localize these descriptions using the same approach.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+        // At this point we can start using the terminal.
+        this.writeEmitter.fire("Starting build...\r\n");
+        this.doBuild();
+    }
+
+    close(): void {
+        // The terminal has been closed. Shutdown the build.
+    }
+
+    private async doBuild(): Promise<number> {
```
Colengms: Since this function does not 'await' any other functions, I don't believe it needs to be declared as `async`.  There is also a `execChildProcess` in common.ts which converts `exec` into an await-able function.  You could either a) remove the `await`, or b) await `execChildProcess` instead of invoking `exec` directly.  B seems better, as you could avoid creating a new Promise object directly, and use `try`/`catch`.

Also, where `doBuild` is called, the Promise you are returning is not awaited.  Is the intention to complete the `exec` before `open` returns?  If so, you'll want to await the result of doBuild.  Or, if not, doBuild() doesn't need to return a promise at all or be async.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+        const file: string = editor.document.fileName;
+        return {
+            "file": file,
+            "fileDirname": fileDir.uri.fsPath,
+            "fileBasenameNoExtension": path.parse(file).name,
+            "workspaceFolder": fileDir.uri.fsPath};
+    }
+}
+
+export function getRawTasksJson(): Promise<any> {
```
Colengms: I know you didn't write this function, but...  There is a `checkFileExists` function in common.ts that is await-able.  If that was used instead, this function could be made await-able and simplified.

Colengms (changes requested):

Extension/package.json
elahehrashedi: It's updated in the new branch.

Extension/package.json
elahehrashedi: what would be a good description for it? 
"c_cpp.taskDefinitions.options.description": "Additional options including cwd and env."

bobbrow: If there are additional properties supported in here, you need to list them out.  I believe you use `"properties": { }`

Reply: in shell tasks, it was called "options", and in vscode, the interface is named execoptions,  that's why I kept that same name.

Extension/CHANGELOG.md
```diff
@@ -1,35 +1,24 @@
 # C/C++ for Visual Studio Code Change Log
 
-## Version 0.29.0-insiders2: July 8, 2020
+## Version 0.29.0: July 15, 2020
 ### New Features
-* Added support for LogMessage Breakpoints for debug type `cppdbg`. [#1013](https://github.com/microsoft/MIEngine/pull/1013)
-* Switch to using the VS Code Semantic Tokens API for semantic colorization. [PR #5401](https://github.com/microsoft/vscode-cpptools/pull/5401)
+* Add Doxygen comment support (to tooltip display of hover, completion, and signature help). [#658](https://github.com/microsoft/vscode-cpptools/issues/658)
+  * Add `C_Cpp.simplifyStructuredComments` setting. [#5706](https://github.com/microsoft/vscode-cpptools/issues/5706)
```
bobbrow: The link to the issue on this sub bullet is not needed because it's part of the parent feature and was addressed as part of the insiders.
You can just mention "The way comments are formatted is controlled by the `C_Cpp.simplifyStructuredComments` setting."

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+        const fileDir: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(editor.document.uri);
+        if (!fileDir) {
+            return undefined;
+        }
+        const file: string = editor.document.fileName;
+        return {
+            "file": file,
+            "fileDirname": fileDir.uri.fsPath,
+            "fileBasenameNoExtension": path.parse(file).name,
+            "workspaceFolder": fileDir.uri.fsPath};
```
bobbrow: nit: the closing curly brace should be on a new line

elahehrashedi: I format the whole file.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: I added an await on the `exec` directly. It was a good idea to re-use `execChildProcess` instead, but it had a `vscode.OutputChannel` as input, so I couldn't re-use it. 
yes, we want to wait for `exec` to complete before `open` returns. I added `await` to the result of `dobuild` as well.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: `getRawTasksJson` is modified now.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -0,0 +1,342 @@
...
+            let temp: string = util.resolveVariables(value, this.AdditionalEnvironment);
+            if (temp && temp.includes(" ")) {
+                temp = "\"" + temp + "\"";
+            }
+            activeCommand = activeCommand + " " + temp;
+        });
+        if (this.options?.cwd) {
+            this.options.cwd = util.resolveVariables(this.options.cwd, this.AdditionalEnvironment);
+        }
+        await new Promise<number>((resolve, reject) => {
```
bobbrow: If you await something that will potentially `reject` the promise, you should wrap it with `try/finally`.  Though it probably makes more sense to resolve to -1 than to reject the promise.  Give this a try:
```
        try {
            const result = await new Promise<number>((resolve, reject) => {
                cp.exec(activeCommand, this.options, (_error, stdout, _stderr) => {
                    if (_error) {
                        telemetry.logLanguageServerEvent("cppBuildTaskError");
                        this.writeEmitter.fire("Build finished with error:\r\n");
                        this.writeEmitter.fire(stdout.toString());
                        resolve(-1);
                    } else {
                        this.writeEmitter.fire(stdout.toString());
                        this.writeEmitter.fire("\r\nBuild finished successfully.\r\n");
                        resolve(0);
                    }
                });
            });
        } catch {
            this.closeEmitter.fire(-1);
        }
```

Extension/src/LanguageServer/cppBuildTaskProvider.ts
Colengms: The question mark in `channel?: vscode.OutputChannel` indicates the channel param is optional.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
Colengms: I was hoping we could remove creation of a new promise here, as it's not necessary if this function is async, and calls await on checkFileExists (instead of then()).  This function could be very simple:
```
export async function getRawTasksJson(): Promise<any> {
    const path: string | undefined = getTasksJsonPath();
    if (!path) {
        return {};
    }
    const fileExists: boolean = await util.checkFileExists(path);
    if (!fileExists) {
        return {};
    }

    const fileContents: string = await util.readFileText(path);
    let rawTasks: any = {};
    try {
        rawTasks = jsonc.parse(fileContents);
    } catch (error) {
        throw new Error(failedToParseTasksJson);
    }
    return rawTasks;
}
```

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi:
```
        try {
            await new Promise<number>((resolve, reject) => {
                cp.exec(activeCommand, this.options, (_error, stdout, _stderr) => {
                    if (_error) {
                        telemetry.logLanguageServerEvent("cppBuildTaskError");
                        const endOfLine: string = stdout ? ":" : ".";
                        this.writeEmitter.fire("Build finished with error" + endOfLine + "\r\n");
                        this.writeEmitter.fire(stdout.toString());
                        resolve(-1);
                    } else {
                        this.writeEmitter.fire(stdout.toString());
                        this.writeEmitter.fire("\r\nBuild finished successfully.\r\n");
                        resolve(0);
                    }
                });
            });
            this.closeEmitter.fire(0);
        } catch {
            this.closeEmitter.fire(-1);
        }
```

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: I need to print the stdout (instead of error message) in the case that exec finishes with error.

Extension/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * Add `quoteArgs` to `launch.json` schema. [PR #5639](https://github.com/microsoft/vscode-cpptools/pull/5639)
 * Add logs for a resolved `launch.json` if "engineLogging" is enabled. [PR #5644](https://github.com/microsoft/vscode-cpptools/pull/5644)
 * Add threadExit and processExit logging flags for 'cppvsdbg'. [PR #5652](https://github.com/microsoft/vscode-cpptools/pull/5652)
+* Add support to run cpp build tasks  (Tasks: Run Build Task) [#3674](https://github.com/microsoft/vscode-cpptools/issues/3674)
```
sean-mcmanus: This is moved to a 0.30.0-insiders changelog...we could update the changelog in a separate PR, since it's still in flux.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
Colengms: It could be updated to return that also.  :)  Generally, I'm suggesting that creation of Promises are pushed down into helper functions, as async/await syntax is so much easier to read and understand.  Not critical, though.

Extension/package.json
```diff
@@ -7,6 +7,7 @@
   "preview": true,
   "icon": "LanguageCCPP_color_128x.png",
   "readme": "README.md",
+  "enableProposedApi": true,
```
bobbrow: We can't check in this line

michelleangela (approved):

Colengms (approved):

Extension/package.nls.json
```diff
@@ -220,10 +220,10 @@
     "c_cpp.debuggers.symbolLoadInfo.loadAll.description": "If true, symbols for all libs will be loaded, otherwise no solib symbols will be loaded. Default value is true.",
     "c_cpp.debuggers.symbolLoadInfo.exceptionList.description": "List of filenames (wildcards allowed) separated by semicolons ';'. Modifies behavior of LoadAll. If LoadAll is true then don't load symbols for libs that match any name in the list. Otherwise only load symbols for libs that match. Example: \"foo.so;bar.so\"",
     "c_cpp.debuggers.requireExactSource.description": "Optional flag to require current source code to match the pdb.",
-    "c_cpp.taskDefinitions.name.description": "The name of the task.",
-    "c_cpp.taskDefinitions.command.description": "The path to either a compiler or script that performs compilation.",
-    "c_cpp.taskDefinitions.args.description": "Additional arguments to pass to the compiler or compilation script.",
-    "c_cpp.taskDefinitions.options.description": "Additional command options.",
+    "c_cpp.taskDefinitions.name.description": "The name of the task",
```
bobbrow, line 223: Should these periods have remained?

PR #5669, opened 2020-06-20
ignore 'screen size is bogus' error; fixes microsoft/vscode/issues/100563 (and microsoft/vscode/issues/75932)

nukoyluoglu:
Corresponding issue: microsoft/vscode#100563
The error below appears when using the Debugger with "type": "cppdbg" and "request": "attach", while picking and attaching to a process through WSL. The root of the error is in the process of fetching remote diagnostics for 'WSL: Ubuntu' since the same error occurs when viewing the Process Explorer, as well. The proposed fix ignores this error silently. The fix imitates a similar commit implemented for the Node - Debug (Legacy) Extension (https://github.com/microsoft/vscode-node-debug/commit/5298920), which also fixes the same issue (microsoft/vscode#75932). The issue was closed, but the fix was only implemented for that particular extension.
'screen size is bogus' error

### DIFF
--- a/Extension/src/Debugger/nativeAttach.ts
+++ b/Extension/src/Debugger/nativeAttach.ts
@@ -175,6 +175,10 @@ function execChildProcess(process: string, workingDirectory?: string): Promise<s
             }
 
             if (stderr && stderr.length > 0) {
+                if (stderr.indexOf('screen size is bogus') >= 0) {
+                    // ignore this error silently; see https://github.com/microsoft/vscode/issues/75932
+                    // see similar fix for the Node - Debug (Legacy) Extension at https://github.com/microsoft/vscode-node-debug/commit/5298920
+                }
                 reject(new Error(stderr));
                 return;
             }
--- a/Extension/src/Debugger/nativeAttach.ts
+++ b/Extension/src/Debugger/nativeAttach.ts
@@ -178,8 +178,9 @@ function execChildProcess(process: string, workingDirectory?: string): Promise<s
                 if (stderr.indexOf('screen size is bogus') >= 0) {
                     // ignore this error silently; see https://github.com/microsoft/vscode/issues/75932
                     // see similar fix for the Node - Debug (Legacy) Extension at https://github.com/microsoft/vscode-node-debug/commit/5298920
+                } else {
+                    reject(new Error(stderr));
                 }
-                reject(new Error(stderr));
                 return;
             }
 


Extension/src/Debugger/nativeAttach.ts
```diff
@@ -175,6 +175,10 @@ function execChildProcess(process: string, workingDirectory?: string): Promise<s
             }
 
             if (stderr && stderr.length > 0) {
+                if (stderr.indexOf('screen size is bogus') >= 0) {
```
sean-mcmanus, line 178: I don't understand how this fixes it. Should there some code in the "if" statement?

sean-mcmanus (changes requested):

Extension/src/Debugger/nativeAttach.ts
sean-mcmanus: Or put the "reject" line in an "else".

nukoyluoglu:
putting the reject statement inside the “else” statement was the intention, i believe i must have committed a wrong version. will fix it.

Extension/src/Debugger/nativeAttach.ts
nukoyluoglu: putting the reject statement inside the “else” statement was the intention, i believe i must have committed a wrong version. will fix it.

sean-mcmanus (approved):

sean-mcmanus:
Thanks

sean-mcmanus:
Your fix should be available with https://github.com/microsoft/vscode-cpptools/releases/tag/0.29.0-insiders .

nukoyluoglu:
@sean-mcmanus 
> Your fix should be available with https://github.com/microsoft/vscode-cpptools/releases/tag/0.29.0-insiders .

I could not find my fix in this version. See: https://github.com/microsoft/vscode-cpptools/compare/0.28.3...0.29.0-insiders and https://github.com/microsoft/vscode-cpptools/blob/0.29.0-insiders/Extension/src/Debugger/nativeAttach.ts - where you can see that the fix has not been merged.

Also, currently the fix does not allow process picking. Please request changes to move the "return" statement on line 184 inside the else statement above.

Extension/src/Debugger/nativeAttach.ts
```diff
@@ -175,7 +175,12 @@ function execChildProcess(process: string, workingDirectory?: string): Promise<s
...
 
             if (stderr && stderr.length > 0) {
-                reject(new Error(stderr));
+                if (stderr.indexOf('screen size is bogus') >= 0) {
+                    // ignore this error silently; see https://github.com/microsoft/vscode/issues/75932
+                    // see similar fix for the Node - Debug (Legacy) Extension at https://github.com/microsoft/vscode-node-debug/commit/5298920
+                } else {
+                    reject(new Error(stderr));
+                }
                 return;
```
nukoyluoglu, line 184: This `return `statement must be inside the `else `statement above it.

sean-mcmanus:
@nukoyluoglu Oops. Sorry about that -- I didn't realize the PR was to our release branch and not master.

Extension/src/Debugger/nativeAttach.ts
sean-mcmanus: Okay, thanks. I'll submit a new PR to master for our next release.

sean-mcmanus: https://github.com/microsoft/vscode-cpptools/pull/5705

sean-mcmanus:
This should be available (for real) this time with https://github.com/microsoft/vscode-cpptools/releases/tag/0.29.0-insiders2

Issue #5687, opened 2020-06-24
compile_commands.json is only updated on reload


bpbradley:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10 (OS Build 19041.329)
- VS Code Version: 1.46.1
- C/C++ Extension Version: 0.28.3
- Other extensions you installed (and if the issue persists after disabling them):
C/C++ Advanced Lint
CMake
CMake Tools
CMake Integration
cmake-format
DeviceTree
Doxygen Documentation Generator
Markdown PDF
markdownlint
Python
Remote - WSL
reStructredText

I have disabled all of these and the problem persists. 

- Does this issue involve using SSH remote to run the extension on a remote machine?: No.
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).

Fairly large project with a number of external dependencies. Numerous cmake projects are added with add_subdirectory, and compile_commands.json is generated specifically for the intellisense.

**Steps to reproduce**

1. Create cmake project. 
2. Build with -DCMAKE_EXPORT_COMPILE_COMMANDS=1
3. Point extension to the generated compile_commands.json
4. Verify intellisense works as expected
5. Add new header file to project
6. Repeat step 2
7. Without reloading, try to include header. Intellisense will complain that it cant locate the header. 
8. Reload window, verify that it works as expected again

**Expected behavior**

The intellisense should be updated automatically, either periodically or when a file is opened/changed. 

<details>
  <summary><strong>Logs</strong></summary>

```
Insert logs here.
```
-------- Diagnostics - 6/24/2020, 10:27:20 AM
Version: 0.28.3
Current Configuration:
{
    "name": "removed",
    "compileCommands": "${workspaceFolder}/build/compile_commands.json",
    "compilerArgs": [],
    "configurationProvider": "go2sh.cmake-integration",
    "browse": {
        "limitSymbolsToIncludedHeaders": true
    }
}
Translation Unit Mappings:

   <removed/path>

    <removed/path>
Translation Unit Configurations:

    Process ID: 8680
    Memory Usage: 33 MB
    Compiler Path: C:\gnuarmemb\bin\arm-none-eabi-gcc.exe
    Includes:
        <removed/paths>
    Defines: <removed>

    Forced Includes:
        <removed/paths>
    Standard Version: c11
    IntelliSense Mode: gcc-x64
    Other Flags:
        --gcc
        --gnu_version=70301
[...]

PR #5688, opened 2020-06-24
Fix compileCommand updating.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/5687


michelleangela:
How does removing the "rename" event check fix the issue?

sean-mcmanus:
> How does removing the "rename" event check fix the issue?

It's only firing rename events (in some cases), which is a known problem with the fs.watch API. The check to exclude rename is unnecessary.

michelleangela (approved):

PR #5730, opened 2020-07-06
Add settings to optionally simplify structured comments (Doxygen)

### LINKED ISSUE #5706: Leave the non-recognized doxygen commands in comments
It is possible to leave the commands that are not being converted? We use \note, \pre, and \post. The current behavior is that they don't show up in the tool tip at all.

_Originally posted by @jhortenstine in https://github.com/microsoft/vscode-cpptools/issues/658#issuecomment-650328227_

michelleangela:
Add option to simplify structured comments. When the option is set to true, tooltips will display a simplified form of structured comments. For example, for Doxygen, it will only display certain tags. If this option is disabled, it will display complete structured comments.

### Screenshot of setting:
image

### Example displays for Doxygen comment:
image

### Simplified display
image

### Complete display
image


sean-mcmanus (approved):

Extension/src/nativeStrings.json
```diff
@@ -189,5 +189,6 @@
         "hint": "This label is for the return value description for a function. Usage example: 'Returns: Area of a shape.'"
     },
     "deprecated_label": "Deprecated:",
-    "exceptions_label": "Exceptions:"
+    "exceptions_label": "Exceptions:",
```
sean-mcmanus, line 192: Are these still used?

sean-mcmanus: Oh, it is.

michelleangela: Yes, these labels are used when the comment is simplified.

sean-mcmanus (changes requested):
It seems like the scope should be "application"?

michelleangela:
> It seems like the scope should be "application"?

Since this is UI, then it should only be application? Similar to the experience of changing color theme of VS Code?

sean-mcmanus:
> > It seems like the scope should be "application"?
> 
> Since this is UI, then it should only be application? Similar to the experience of changing color theme of VS Code?

Yeah. If it doesn't make sense for the setting (UI) to change after the user switches workspace folders, switches workspaces, or switches to a remote machine, then the setting should be "application" scope (i.e. User scope).

sean-mcmanus (approved):

PR #5787, opened 2020-07-15
Use GitHub created MIT License.

sean-mcmanus:
Previous change at https://github.com/microsoft/vscode-cpptools/pull/5786 was insufficient.


sean-mcmanus:
I'm still following up with our legal team on if we can find some way to make our licensing more clear in regards to what exactly the MIT versus restrictive license applies to.

Colengms (approved):

sean-mcmanus:
FYI, we may have a follow up next week that adds some more license readme files to explain what the licenses apply to.

PR #5823, opened 2020-07-22
Extension installation/download failures

### LINKED ISSUE #4829: Implement a solution to deal with extension installation/download failures
I can't get the intellie server to start after upgraded to the 0.26.3-insiders3 because I missed Microsoft.VSCode.CPP.Extension.exe in bin folder. So I have to copy one from the 0.26.2/bin folder, then it'll work fine.

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/4829

Edit: The messages are error messages now.

- Check for the core binary files in the extension folder and show this warning message if any of the binaries are missing:

image

- Check for the important JSON files in the extension folder and show this error message if any of the binaries are missing (the message is shown if the user has installed a VSIX that doesn't match their OS).

image


bobbrow:
>                                         const fileMode: number = (pkg.binaries && pkg.binaries.indexOf(absoluteEntryPath) !== -1) ? 0o755 : 0o664;

This is the line of code that expects binaries to be empty on Win32.  Can you just exclude win32 from this check?  (though maybe it doesn't matter on Windows?)

---
Refers to: Extension/src/packageManager.ts:390 in f7cb016. [](commit_id = f7cb016771fc477255199d84e99080a66e6c0ca8, deletion_comment = False)

Extension/src/common.ts
```diff
@@ -557,6 +574,67 @@ export function getInstalledBinaryPlatform(): string | undefined {
...
+                    console.log(`Extension file ${file} is missing.`);
+                }
+            }));
+        }
+    }
+    return installBinariesExist;
+}
+
+/* Check if the core Json files exists in extension's installation folder */
+export async function checkInstallJasonsExist(): Promise<boolean> {
```
bobbrow: "Json", not Jason :)

Extension/src/common.ts
```diff
@@ -214,6 +215,22 @@ export async function isExtensionReady(): Promise<boolean> {
...
+            fs.rename(sourcePath, getDebugAdaptersPath("bin/OpenDebugAD7.exe.config.unused"), (err: NodeJS.ErrnoException | null) => {
+                if (err) {
+                    getOutputChannelLogger().appendLine(localize("rename.failed.delete.manually",
+                        'ERROR: fs.rename failed with "{0}". Delete {1} manually to enable debugging.', err.message, sourcePath));
+                }
+            });
+        }
+    }
+
+    return Promise.resolve();
```
Colengms: It doesn't look like this function returns a Promise other than this call to `Promise.resolve()`.  Maybe this function doesn't need to return a Promise?  There is a helper function that wraps `fs.rename()` in common.ts, called `renamePromise`.  It's async.  I'd suggest calling `await renamePromise()` instead of using the callback, making `renameDebugAdaptersPath` `async`, and omitting the call to `Promose.resolve()` (as an async function implicitly returns a Promise).

Extension/src/common.ts
```diff
@@ -557,6 +574,67 @@ export function getInstalledBinaryPlatform(): string | undefined {
...
+            installJasonsExist = false;
+            console.log(`Extension file ${file} is missing.`);
+        }
+    }));
+    return installJasonsExist;
+}
+
+export function removeInstallLockFile(): boolean {
+    try {
+        fs.unlinkSync(path.join(extensionPath, "install.lock"));
```
Colengms: There is a `unlinkPromise` helper function in common.ts, which is awaitable.  I'd suggest making this function awaitable and calling `await unlinkPromise()` instead of calling `fs.unlinkSync`.

Colengms (changes requested):

Extension/src/common.ts
bobbrow: It probably had a promise because of the callback for rename. If that is the case, then this code needs to be wrapped in a new Promise and you call resolve when the rename callback is invoked.

Colengms: Wouldn't it be better to `await renamePromise()` instead, and make this function async?  I like the idea of avoiding creation of new Promises, pushing them down into async helpers.  (Also, IMHO, I'd rename `renamePromise` to `renameAsync`(?))

elahehrashedi: I made the change. 
I can't see the function `renameDebugAdaptersPath `is being called anywhere.

Extension/src/main.ts
```diff
@@ -72,9 +72,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
 
-    // check if the correct offline/insiders vsix is installed on the correct platform
     const installedPlatform: string | undefined = util.getInstalledBinaryPlatform();
-    if (!installedPlatform || (process.platform !== installedPlatform)) {
+
+    // Check the main binaries files to declare if the extension has been installed successfully.
+    if (!installedPlatform || !(await util.checkInstallBinariesExist())) {
+        errMsg = localize("extension.installation.failed", "The C/C++ extension failed to install successfully. You will need to repair or reinstall the extension for C/C++ language features to function properly.");
+        const reload: string = localize("remove.extension", "Attempt to Repair");
+        vscode.window.showInformationMessage(errMsg, reload).then(async (value?: string) => {
```
sean-mcmanus: Why is this an information message and not an error message?

Extension/src/common.ts
Colengms: I don't see where `renameDebugAdaptersPath` is being called either.  Isn't it added in this PR?  Can it be removed?

elahehrashedi: @Colengc should I remove it? It was not added in this PR (it was an existing code).

Extension/src/main.ts
elahehrashedi: @sean-mcmanus  Do you think the error message will be better?

When the **VSIX** belongs to another platform => **error** message (go to the download page)
When some **Json** files are missing                  => **warning** message (go to the download page)
When some **Binaries** are missing                    => **warning** message (attempt to repair)

Extension/src/common.ts
Colengms: I'm confused.  In the diff it shows as green (newly added), which means you are adding this function.  I don't see any red section in the diff indicating it was moved, or any removed files.  I assumed you had written it.  Maybe this is left over from a bad merge?  I do see a function `removeUnnecessaryFile()` in main.ts, which looks to be the same function, and that is called in main.ts in 2 places.  If this one is not used, I'd suggest removing it.

Extension/src/main.ts
elahehrashedi: I changed them all to error messages

Extension/package.json
```diff
@@ -1826,7 +1826,10 @@
       "platforms": [
         "win32"
       ],
-      "binaries": []
+      "binaries": [
+        "./bin/cpptools",
```
sean-mcmanus: it seems like these should include the extension, i.e. cpptools.exe, since the entry is platform-specific. Was there a reason not to?

Extension/src/common.ts
```diff
@@ -557,6 +558,61 @@ export function getInstalledBinaryPlatform(): string | undefined {
...
+    }
+    let installBinariesExist: boolean = true;
+    const info: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+    const packageManager: PackageManager = new PackageManager(info);
+    const packages: Promise<IPackage[]> = packageManager.GetPackages();
+    for (const pkg of await packages) {
+        if (pkg.binaries) {
+            await Promise.all(pkg.binaries.map(async (file: string) => {
+                if (!await checkFileExists(file) &&
+                    !(info.platform === "win32" && await checkFileExists(file + ".exe"))) {
```
sean-mcmanus: This additional check seems unnecessary if the entry in the package.json is changed.

sean-mcmanus (changes requested):

Extension/src/common.ts
elahehrashedi: I removed it

sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -557,6 +558,60 @@ export function getInstalledBinaryPlatform(): string | undefined {
...
+                }
+            }));
+        }
+    }
+    return installBinariesExist;
+}
+
+/* Check if the core Json files exists in extension's installation folder */
+export async function checkInstallJsonsExist(): Promise<boolean> {
+    let installJasonsExist: boolean = true;
```
bobbrow: Jsons, not "Jasons"

Colengms (approved):

Extension/src/common.ts
sean-mcmanus: I did the rename.

bobbrow: thanks!

Issue #5776, opened 2020-07-13
compilerArg isn't used if compilerPath isn't explicitly set


sean-mcmanus:
I wanted to add a compilerArg (using the existing default CompilerPath), but it didn't get used until I explicitly added the compilerPath.

sean-mcmanus:
Or it should give some sort of warning that the compilerPath is required to be explicitly added.

bobbrow:
@Colengms can you take a look at this?  I was thinking this was supposed to work because the Windows team has a custom config provider extension that depends on this behavior.  It could be that this only is a problem for standard `c_cpp_properties.json` configs.

sean-mcmanus:
I hit this issue again today, when I added compilerArgs with "/clr" etc. with the intent of it applying to the default cl.exe compiler -- it didn't work until I manually added the compilerPath to be "cl.exe".

Colengms:
I have tried a number of things, with no repro.

Colengms:
Only repro's if the default compiler is not explicitly set.

PR #5838, opened 2020-07-27
Use detected default compiler

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/5776

The detected default compiler and associated values were used to populate newly created configurations, but were not used as a fallback if the compiler was unspecified.

This change handles the situation where no compiler was specified in the c_cpp_properties, or explicitly as a default setting.  Instead of using no/blank compiler, the detected compiler and associated values are used.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -628,7 +628,7 @@ export class CppProperties {
             configuration.cStandard = this.updateConfigurationString(configuration.cStandard, settings.defaultCStandard, env);
             configuration.cppStandard = this.updateConfigurationString(configuration.cppStandard, settings.defaultCppStandard, env);
             configuration.intelliSenseMode = this.updateConfigurationString(configuration.intelliSenseMode, settings.defaultIntelliSenseMode, env);
-            if (!configuration.compilerPath && !!this.defaultCompilerPath) {
+            if ((configuration.compilerPath === undefined) && !!this.defaultCompilerPath) {
                 configuration.compilerPath = this.defaultCompilerPath;
                 if (!configuration.cStandard && !!this.defaultCStandard) {
                     configuration.cStandard = this.defaultCStandard;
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -628,6 +628,27 @@ export class CppProperties {
             configuration.cStandard = this.updateConfigurationString(configuration.cStandard, settings.defaultCStandard, env);
             configuration.cppStandard = this.updateConfigurationString(configuration.cppStandard, settings.defaultCppStandard, env);
             configuration.intelliSenseMode = this.updateConfigurationString(configuration.intelliSenseMode, settings.defaultIntelliSenseMode, env);
+            if ((configuration.compilerPath === undefined || configuration.compilerPath === "") && !!this.defaultCompilerPath) {
```
sean-mcmanus: It doesn't seem like "" should use the default, right?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -628,6 +628,27 @@ export class CppProperties {
...
+                configuration.compilerPath = this.defaultCompilerPath;
+                if ((!configuration.cStandard || configuration.cStandard === "") && !!this.defaultCStandard) {
```
sean-mcmanus: Is it intentional to condition these on the compilerPath?

Colengms: Yes.  If we're not actually using the detected default compiler, we should not apply the detected default compiler's standards, intelliSenseMode, etc.  Right?

Colengms: But, I'm not sure about defaultWindowsSdkVersion, defaultIncludes, defaultFrameworks .

Extension/src/LanguageServer/configurations.ts
```diff
@@ -628,6 +628,27 @@ export class CppProperties {
             configuration.cStandard = this.updateConfigurationString(configuration.cStandard, settings.defaultCStandard, env);
             configuration.cppStandard = this.updateConfigurationString(configuration.cppStandard, settings.defaultCppStandard, env);
             configuration.intelliSenseMode = this.updateConfigurationString(configuration.intelliSenseMode, settings.defaultIntelliSenseMode, env);
+            if (!configuration.compilerPath && !!this.defaultCompilerPath) {
+                configuration.compilerPath = this.defaultCompilerPath;
```
bobbrow: `!configuration.compilerPath` will be true when compilerPath is set to "".  We should not be overwriting `""` with `this.defaultCompilerPath` in that case. That will break folks who depend on the `""` behavior.

bobbrow (approved):

PR #5839, opened 2020-07-27
Check the insiders version to be compatible with vscode

elahehrashedi:
For Auto-update, Check the insider's version to be compatible with vscode's version.
Github issue: https://github.com/microsoft/vscode-cpptools/issues/4873

Edit (2): 
"minimumVSCodeVersion" is the minimum VS Code's version that is supported by the latest insiders' C/C++ extension.
It has only one value. 
for example, if the latest extension insiders version is 0.30.0-insiders, the minimum supported VS Code version is "1.44.0".
If an insider user is using a lower version of VsCode on a 0.29.0 extension, we won't upgrade his version to an insider version. 

    "minimumVSCodeVersion": "1.44.0"

Edit (1):
the current compatible versions are:
    "minimumVSCodeVersion": {
        "0.28.0": "1.44.0"
    }

The following can be used for testing:
    "minimumVSCodeVersion": {
        "0.31.0": "1.48.0",
        "0.30.0": "1.47.0",
        "0.29.0": "1.46.0",
        "0.28.0": "1.44.0"
    }
The new entries should be added at the beginning of this map.


Extension/package.json
```diff
@@ -981,6 +981,12 @@
           "default": "Enabled",
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
+        },
+        "C_Cpp.supportExtensionVscodeVersion": {
```
bobbrow: We do not want a setting for this.  It should just be read from the cpptools.json file.

Extension/cpptools.json
```diff
@@ -3,5 +3,10 @@
     "defaultIntelliSenseEngine": 100,
     "recursiveIncludes": 100,
     "gotoDefIntelliSense": 100,
-    "enhancedColorization": 100
+    "enhancedColorization": 100,
+    "supportExtensionVscodeVersion": {
+        "0.28.0": "1.44.0",
```
bobbrow: I don't think we should add each extension version in here.  We only need to add an entry when the vscode dependency changes.  We could end up forgetting to update this when there is no dependency change.

Extension/package.nls.json
```diff
@@ -187,5 +187,6 @@
     "c_cpp.debuggers.symbolLoadInfo.description": "Explicit control of symbol loading.",
     "c_cpp.debuggers.symbolLoadInfo.loadAll.description": "If true, symbols for all libs will be loaded, otherwise no solib symbols will be loaded. Default value is true.",
     "c_cpp.debuggers.symbolLoadInfo.exceptionList.description": "List of filenames (wildcards allowed) separated by semicolons ';'. Modifies behavior of LoadAll. If LoadAll is true then don't load symbols for libs that match any name in the list. Otherwise only load symbols for libs that match. Example: \"foo.so;bar.so\"",
-    "c_cpp.debuggers.requireExactSource.description": "Optional flag to require current source code to match the pdb."
+    "c_cpp.debuggers.requireExactSource.description": "Optional flag to require current source code to match the pdb.",
+    "c_cpp.configuration.supportExtensionVscodeVersion.description": "The minimum version of VS Code to support C/C++ extension"
```
bobbrow: This can be removed when you remove the setting

Extension/src/abTesting.ts
```diff
@@ -12,12 +12,15 @@ import * as fs from 'fs';
...
 const localConfigFile: string = "cpptools.json";
+const minimumVscodeVersion: string = "1.43.2";
 
 interface Settings {
     defaultIntelliSenseEngine?: number;
     recursiveIncludes?: number;
     gotoDefIntelliSense?: number;
     enhancedColorization?: number;
+    // a record of <extensionVersion, vscodeVersion>
+    supportExtensionVscodeVersion?: Record<string, string> | null;
```
bobbrow: If the property has a `?`, don't add `null` to the type declaration.

Extension/src/packageVersion.ts
```diff
@@ -77,4 +77,8 @@ export class PackageVersion {
             }
         }
     }
+
+    public get noPatchVersion(): string {
+        return this.major.toString() + "." + this.minor.toString() + ".0";
```
bobbrow: use this instead:
```ts
return `${this.major}.${this.minor}.0`;
```

Extension/src/abTesting.ts
bobbrow: And if we don't add an entry for every version, we shouldn't use `Record<,>` here.  `Map<,>` would be better, I think.

Extension/cpptools.json
```diff
@@ -3,5 +3,10 @@
     "defaultIntelliSenseEngine": 100,
     "recursiveIncludes": 100,
     "gotoDefIntelliSense": 100,
-    "enhancedColorization": 100
+    "enhancedColorization": 100,
+    "supportExtensionVscodeVersion": {
```
bobbrow: I'm kind of confused by the name here.  I would expect something about "minimum" versions in the name.  Maybe `minimumVSCodeVersions`?

Extension/src/abTesting.ts
```diff
@@ -61,6 +67,12 @@ export class ABTestSettings {
         return util.isNumber(this.settings.gotoDefIntelliSense) ? this.settings.gotoDefIntelliSense >= this.bucket.Value : true;
     }
 
+    public minimumVscodeSupportedVersion(extensionVersion: string): string {
+        return (this.settings.supportExtensionVscodeVersion && this.settings.supportExtensionVscodeVersion[extensionVersion]) ?
```
bobbrow: With only entries for the dependency changes, this will need to be reworked a bit.
Iterate through the keys in the map and for each one, if it is less than the `extensionVersion`, make sure that the current vscode version is greater than the value for that key.  You'll want to create PackageVersion objects for these comparisons since we already have a comparison function in that class.

bobbrow (changes requested):

Extension/package.json
```diff
@@ -567,10 +567,10 @@
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
         },
-        "C_Cpp.supportExtensionVscodeVersion": {
+        "C_Cpp.minimumVSCodeVersions": {
```
bobbrow: This setting should be deleted

Extension/src/LanguageServer/extension.ts
```diff
@@ -621,7 +623,7 @@ function installVsix(vsixLocation: string): Thenable<void> {
 
     // 1.33.0 introduces workbench.extensions.installExtension.  1.32.3 was immediately prior.
     const lastVersionWithoutInstallExtensionCommand: PackageVersion = new PackageVersion('1.32.3');
-    if (userVersion.isGreaterThan(lastVersionWithoutInstallExtensionCommand, "insider")) {
+    if (userVersion.isGreaterThan(lastVersionWithoutInstallExtensionCommand, "insiders")) {
```
sean-mcmanus: This needs to be "insider", because VS Code uses "insider".

sean-mcmanus (dismissed):

Extension/src/abTesting.ts
```diff
@@ -73,11 +95,19 @@ export class ABTestSettings {
                 this.recursiveIncludesDefault.Value = util.isNumber(newSettings.recursiveIncludes) ? newSettings.recursiveIncludes : this.recursiveIncludesDefault.DefaultValue;
                 this.gotoDefIntelliSenseDefault.Value = util.isNumber(newSettings.gotoDefIntelliSense) ? newSettings.gotoDefIntelliSense : this.gotoDefIntelliSenseDefault.DefaultValue;
                 this.enhancedColorizationDefault.Value = util.isNumber(newSettings.enhancedColorization) ? newSettings.enhancedColorization : this.enhancedColorizationDefault.DefaultValue;
+                const newMap: Map<string, string> = new Map<string, string>();
+                if (newSettings.minimumVSCodeVersion) {
+                    for (const [extensionVersion, vscodeVersion] of Object.entries(newSettings.minimumVSCodeVersion)) {
+                        newMap.set(extensionVersion, <string> vscodeVersion);
+                    }
+                }
+                this.minimumVSCodeVersionDefault = newSettings.minimumVSCodeVersion ? newMap : this.minimumVSCodeVersionDefault;
```
michelleangela: Set `this.minimumVSCodeVersionDefault` to newMap within if statement block of line 99 since it checks for  `newSettings.minimumVSCodeVersion`. So no need to do another check on line 104.

Extension/src/abTesting.ts
```diff
@@ -8,16 +8,21 @@ import * as util from './common';
 import * as Telemetry from './telemetry';
 import { PersistentState } from './LanguageServer/persistentState';
 import * as fs from 'fs';
+import { PackageVersion } from './packageVersion';
 
 const userBucketMax: number = 100;
 const userBucketString: string = "CPP.UserBucket";
 const localConfigFile: string = "cpptools.json";
+const minimumVSCodeVersionDefault: PackageVersion = new PackageVersion("1.43.2");
```
michelleangela: Will `minimumVSCodeVersionDefault` need to be updated whenever extension updates its VSCode version dependency? Is there a way to get this version number from the package.json instead of hardcoding it here?

michelleangela (dismissed):

Extension/src/abTesting.ts
```diff
@@ -61,6 +69,20 @@ export class ABTestSettings {
...
     }
 
+    public getMinimumVSCodeVersion(curExtension: PackageVersion): PackageVersion {
+        if (this.minimumVSCodeVersionDefault.size === 0) {
+            return minimumVSCodeVersionDefault;
+        }
+        for (const [extensionString, vscodeString] of this.minimumVSCodeVersionDefault) {
+            const extensionPackage: PackageVersion = new PackageVersion(extensionString);
+            const vscodePackage: PackageVersion = new PackageVersion(vscodeString);
+            if (!extensionPackage.isGreaterThan(curExtension, undefined)) {
```
bobbrow: Something doesn't seem right here.  It looks like you return the vscode requirement of an older version of the extension (which we can assume has already been satisfied).  We would actually want the requirement of a newer version in the event that a new requirement was published.

I apologize now for recommending doing this as a Map.  I don't think the insiders update code even works in a way that is compatible with what we were trying to design.  I think it just tries to get the most recent insiders build.  In that case, minimumVSCodeVersion could just be a string with the current VS code version requirement.  That would simplify all of this logic down to just converting the string from cpptools.json into a PackageVersion object.

Extension/src/abTesting.ts
```diff
@@ -26,18 +29,21 @@ export class ABTestSettings {
     private recursiveIncludesDefault: PersistentState<number>;
     private gotoDefIntelliSenseDefault: PersistentState<number>;
     private enhancedColorizationDefault: PersistentState<number>;
+    private minimumVSCodeVersionDefault: string;
```
bobbrow: Please make this a `PersistentState<string>` type.

Extension/src/packageVersion.ts
```diff
@@ -50,8 +50,8 @@ export class PackageVersion {
             this.suffix === other.suffix && this.suffixVersion === other.suffixVersion;
     }
 
-    public isGreaterThan(other: PackageVersion, suffixStr: string = 'insiders'): boolean {
-        if ((this.suffix && !this.suffix.startsWith(suffixStr)) || (other.suffix && !other.suffix.startsWith(suffixStr))) {
+    public isGreaterThan(other: PackageVersion, suffixStr: string | undefined): boolean {
```
bobbrow: Can you explain what is wrong with this function that requires a re-write?  I would rather we not change this functionality.

bobbrow (changes requested):

Extension/src/abTesting.ts
elahehrashedi: This is a default in case the package.json doesn't contain the version number. This version (i.e 1.43.2) was hardcoded before, so I used it as a default. The version extracted from package.json will replace this.

bobbrow: This won't be needed if you use PersistentState for the class field.  You put this value as the default for the PersistentState instead.

Extension/src/packageVersion.ts
elahehrashedi: I changed this for when I didn't want to consider the suffix when comparing the VsCode versions, 
because adding the input "insider" here didn't make sense to me, So I changed the first line. 
For example, I want this:
"1.44.0" equal to "1.44.0.master"
"1.44.0" equal to "1.44.0.insdier"
"1.44.0"            equal to "1.44.0"

in the current code
"1.44.0".isgreaterthan("1.44.0.insider") is false
"1.44.0.insider".isgreaterthan("1.44.0.") is true

bobbrow: There are a few things to think about to achieve this behavior.
1. `undefined` and `""` are both "falsy" values, and the comparision you've used would allow you to get the same behavior without adding `| undefined` to the type definition in the signature.
2. Changing the behavior of a public function based on magic values like `undefined` and `""` can result in confusion in the future and should be avoided when possible. Consider adding a new versions of the function that encode what you want in the function name and then have the forward to a private function that understands the magic values. For example:
```ts
public isGreaterThan(other: PackageVersion, suffixStr: string = 'insiders'): boolean {
    return isGreaterThanInternal(other, suffixStr);
}

public isMajorMinorPatchGreaterThan(other: PackageVersion): boolean {
    return isGreaterThanInternal(other, "");
}

private isGreaterThanInternal(other: PackageVersion, suffixStr: string): boolean {
    // The real work happens here and suffixStr === "" is documented as a special thing that changes the behavior.
}
```
If you don't do it that way, at the very least, suffixStr === "" needs to be clearly documented as a special value that changes the behavior of the function.

elahehrashedi: Thank you for the detailed explanation. 
I created different functions for different behaviors of `isGreaterThan` for vscode package and extension package.

Extension/src/abTesting.ts
```diff
@@ -61,6 +67,12 @@ export class ABTestSettings {
         return util.isNumber(this.settings.gotoDefIntelliSense) ? this.settings.gotoDefIntelliSense >= this.bucket.Value : true;
     }
 
+    public getMinimumVSCodeVersion(): PackageVersion {
+        // Get minimum VS Code's supported version for latest insiders upgrades.
+        return new PackageVersion(
+            this.settings.minimumVSCodeVersion ? this.settings.minimumVSCodeVersion : this.minimumVSCodeVersionDefault.Value);
```
bobbrow: `this.settings.minimumVSCodeVersion` should always be valid, so there's no need for the `?` anymore.

Extension/src/packageVersion.ts
```diff
@@ -66,13 +78,18 @@ export class PackageVersion {
...
                     return diff > 0;
-                } else if (this.suffix) {
-                    if (!other.suffix) {
-                        return false;
-                    }
-                    return (this.suffixVersion > other.suffixVersion);
                 } else {
-                    return other.suffix ? true : false;
+                    if (!suffixStr) {
+                        return false; // isEqual = true
```
bobbrow: Please update the comment to explain that when `suffixStr` is empty that we only compare the major/minor/patch components of the PackageVersion.

Extension/src/LanguageServer/extension.ts
```diff
@@ -468,10 +469,11 @@ function realActivation(): void {
         // Skip Insiders processing for 32-bit Linux.
         if (info.platform !== "linux" || info.architecture === "x86_64" || info.architecture === "arm" || info.architecture === "arm64") {
             // Skip Insiders processing for unsupported VS Code versions.
-            // TODO: Change this to not require the hardcoded version to be updated.
             const vscodeVersion: PackageVersion = new PackageVersion(vscode.version);
-            const minimumSupportedVersionForInsidersUpgrades: PackageVersion = new PackageVersion("1.43.2");
-            if (vscodeVersion.isGreaterThan(minimumSupportedVersionForInsidersUpgrades, "insider")) {
+            const abTestSettings: ABTestSettings = getABTestSettings();
+            const minimumSupportedVersionForInsidersUpgrades: PackageVersion =
+                abTestSettings.getMinimumVSCodeVersion();
```
bobbrow: nit: please move this up to the previous line.

Extension/src/abTesting.ts
elahehrashedi: Yes, but the same logic applies to other members of `Settings`, like `this.settings.gotoDefIntelliSense`. 
I needed to define `Settings `like this: 
```
interface Settings {
    defaultIntelliSenseEngine?: number;
    recursiveIncludes?: number;
    gotoDefIntelliSense?: number;
    enhancedColorization?: number;
    minimumVSCodeVersion**?**: string;
}
```
because I use this interface to read the `Settings `from the JSON file, and the `minimumVSCodeVersion `might not exists in the JSON file.

bobbrow (approved):

PR #5855, opened 2020-07-29
Fix another issue with resolving compilerPath when compileCommands are used.

Colengms:
Address a detail missed in https://github.com/microsoft/vscode-cpptools/pull/5838

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -652,6 +652,14 @@ export class CppProperties {
                         configuration.macFrameworkPath = this.defaultFrameworks;
                     }
                 }
+            } else {
+                // However, if compileCommands are used and compilerPath is explicitly set, it's still necessary to resolve variables in it.
+                if (configuration.compilerPath === "${default}") {
+                    configuration.compilerPath = settings.defaultCompilerPath;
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -652,6 +652,16 @@ export class CppProperties {
                         configuration.macFrameworkPath = this.defaultFrameworks;
                     }
                 }
+            } else {
+                // However, if compileCommands are used and compilerPath is explicitly set, it's still necessary to resolve variables in it.
+                if (configuration.compilerPath === null) {
+                    configuration.compilerPath = undefined;
+                } else {
+                    if (configuration.compilerPath === "${default}") {
+                        configuration.compilerPath = settings.defaultCompilerPath;
```
sean-mcmanus: I think we may need to handle the case where settings.defaultCompilerPath is null/undefined?

sean-mcmanus: Should this check be moved to be the 1st checked?

sean-mcmanus (approved):

sean-mcmanus (approved):

michelleangela (approved):

PR #5876, opened 2020-07-31
Fix C Standard selector in UI

majibow:
Fix C Standard Selector in UI:  gnu17 -> gnu18.


bobbrow (approved):
Thank you for the fix!

Issue #5885, opened 2020-08-03
Use jsonc parser for c_cpp_properties.json


heartacker:
settings is using json with comment format,
c_cpp_p.json does not have the same bahavier as setting.json
hopes that Commas  `,` is supported in c_cpp_p json file to make it easier for users to write configurations
otherwise, some time it is very very annoying
```json
            "defines": [
                "_DEBUG",
                "UNICODE",
                "_UNICODE",
                "WIN32"
            ],
```
```json
            "defines": [
                "_DEBUG",
                "UNICODE",
                "WIN32"       //using alt + uparrow/downarrow to change position, it doesn't make sense, which is very bothered
                "_UNICODE",

            ],
```
![image](https://user-images.githubusercontent.com/1876302/89160957-7b683080-d5a4-11ea-8745-a6be7d6f66e3.png)

sean-mcmanus:
This is fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/0.30.0-insiders. However, comments will give squiggles unless the default JSON file type is changed to "JSON with comments", which will be fixed in our next release (see https://github.com/microsoft/vscode-cpptools/pull/5911 ).

Issue #5822, opened 2020-07-22
Request: Option to Always Show C/C++ Configuration in Status Bar


benjaminc247:
As of the latest release I am able to tie my build task and launch command back to the c/c++ configuration, which is awesome!

I have one problem though, the c/c++ configuration does not show up in the status bar unless I have c/c++ files open. It would be nice to be able to quickly change configurations, build, and run without necessarily opening a file. I realize I can select the configuration from the command palette, but I'm after convenience here.

Would it be possible to add a setting (which I would set in my workspace) to always show the c/c++ configuration in the status bar?

sean-mcmanus:
Yeah, I've hit this issue too, mostly when my cursor is the Output logging.

The code that filters by cpp is at https://github.com/microsoft/vscode-cpptools/blob/release/Extension/src/LanguageServer/ui.ts#L164 .

So it would be pretty trivial to add/remove some filter at that spot (if the file types are general enough to be added without a setting), but more work to add a setting to always show it (if anyone wanted to make PR for the change).

PR #5911, opened 2020-08-06
Set jsonc for c_cpp_properties.json.

sean-mcmanus:
Fix c_cpp_properties.json not opening as JSON with comments. Follow up to https://github.com/microsoft/vscode-cpptools/issues/5885
Fix the configuration UI disappearing when selection moves to the C/C++ or Log Diagnostics window, mentioned at https://github.com/microsoft/vscode-cpptools/issues/5822 .
Fix an unreported bug where the configuration UI wouldn't appear for non-cpp files that are active on startup (until the selection changed).

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -452,6 +452,7 @@ function realActivation(): void {
 
     disposables.push(vscode.workspace.onDidChangeConfiguration(onDidChangeSettings));
     disposables.push(vscode.window.onDidChangeActiveTextEditor(onDidChangeActiveTextEditor));
+    ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
     disposables.push(vscode.window.onDidChangeTextEditorSelection(onDidChangeTextEditorSelection));
     disposables.push(vscode.window.onDidChangeVisibleTextEditors(onDidChangeVisibleTextEditors));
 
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -452,6 +452,7 @@ function realActivation(): void {
 
     disposables.push(vscode.workspace.onDidChangeConfiguration(onDidChangeSettings));
     disposables.push(vscode.window.onDidChangeActiveTextEditor(onDidChangeActiveTextEditor));
+    ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
```
sean-mcmanus, line 455: This was also causing a bug where the config wasn't being shown initially for open non-cpp documents until a cursor change was done.

Extension/src/LanguageServer/ui.ts
```diff
@@ -164,9 +164,13 @@ export class UI {
...
             // It's sometimes desirable to see the config and icons when making settings changes.
-            const isSettingsJson: boolean = ((activeEditor.document.fileName.endsWith("c_cpp_properties.json") || activeEditor.document.fileName.endsWith("settings.json")));
-
-            this.ShowConfiguration = isCpp || isSettingsJson;
+            const isCppPropertiesJson: boolean = activeEditor.document.fileName.endsWith("c_cpp_properties.json");
+            if (isCppPropertiesJson) {
+                vscode.languages.setTextDocumentLanguage(activeEditor.document, "jsonc");
+            }
+            this.ShowConfiguration = isCpp || isCppPropertiesJson ||
+                activeEditor.document.uri.scheme === "output" ||
```
sean-mcmanus, line 174: This is an unrelated change -- so the configuration doesn't disappear when in C/C++ or Log Diagnostics output.

Colengms (approved):

PR #5944, opened 2020-08-10
add telemetry for missing files

elahehrashedi:
add telemetry to report the missing Binary/JSON files from the extension folder.


Extension/src/common.ts
```diff
@@ -573,6 +576,7 @@ export async function checkInstallBinariesExist(): Promise<boolean> {
                 if (!await checkFileExists(file)) {
                     installBinariesExist = false;
                     console.log(`Extension file ${file} is missing.`);
+                    Telemetry.logLanguageServerEvent("missingBinary", { "source": `${file}` });
```
bobbrow: You should use `path.basename(file)` instead. We can't log absolute paths in telemetry.

Extension/src/common.ts
```diff
@@ -603,6 +607,7 @@ export async function checkInstallJsonsExist(): Promise<boolean> {
         if (!await checkFileExists(path.join(extensionPath, file))) {
             installJsonsExist = false;
             console.log(`Extension file ${file} is missing.`);
+            Telemetry.logLanguageServerEvent("missingJson", { "source": `${file}` });
```
bobbrow, line 611: Same here. Just log the name of the file instead of the full path.

bobbrow (changes requested):

Extension/src/common.ts
elahehrashedi: fixed.

Extension/src/common.ts
elahehrashedi: the file name used here already a base name (not a full path)

bobbrow (approved):

sean-mcmanus (approved):

Issue #5896, opened 2020-08-05
Don't downgrade from an unreleased insiders vsix to the released vsix if updateChannel is "Default"


sean-mcmanus:
...should only affects internal users, so not sure if it's high enough priority, since the workaround is to use "Insiders" updateChannel or autoUpdate "false".

PR #5946, opened 2020-08-10
don't downgrade pre-release

elahehrashedi:
Don't downgrade from an unreleased insiders vsix to the released vsix if updateChannel is "Default"
bugfix: https://github.com/microsoft/vscode-cpptools/issues/5896

Edit: new test added for "internal testing":
image


Extension/src/githubAPI.ts
```diff
@@ -191,7 +191,7 @@ export function getTargetBuild(builds: Build[], userVersion: PackageVersion, upd
     }
     const latestVersion: PackageVersion = new PackageVersion(builds[0].name);
```
bobbrow: Can you rename this variable to `latestVersionOnline` so that it's clear what we're checking here?  I think it would be good to add a test for this case as well.

sean-mcmanus (approved):

Extension/src/githubAPI.ts
elahehrashedi: new test added:
image

PR #5954, opened 2020-08-11
Adds a JSON parsing library for tasks.json

dan-shaw:
Currently, comments are removed when parsing tasks.json. This PR adds a new dependency that keeps user comments.


Extension/package.json
```diff
@@ -39,7 +39,31 @@
     "Snippets"
   ],
   "activationEvents": [
-    "*"
+    "onLanguage:cpp",
```
sean-mcmanus: This activationEvent change should be reverted (manually is fine). It's an artifact of the fact that we modify the package.json on runtime, i.e. the normal procedure to update our package.json would be to undo this change first, apply the change, stage/commit the change, then Launch extension (which would change the activationEvents again). The other alternative is to copy the Extension folder to a different location to use for debugging.

sean-mcmanus:
@Colengms Was there a GitHub issue tracking the comment removal issue?

Colengms:
> @Colengms Was there a GitHub issue tracking the comment removal issue?

@sean-mcmanus Are you referring to the issue when editing `c_cpp_properties.json` via the Configuration UI?  I dont think so.

@dan-shaw Can this also be extended to replace existing uses of `jsonc-parser`?  It would be nice to depend on only 1 jsonc parser.

sean-mcmanus (approved):

sean-mcmanus (changes requested):
I don't see anything fixed by this PR. When I have a tasks.json or c_cpp_properties.json and then trigger an action that causes one of those to be editted by our code, then the comments get overwritten still. Maybe some extra code needs to be added preserver the comments?

sean-mcmanus:
> I don't see anything fixed by this PR. When I have a tasks.json or c_cpp_properties.json and then trigger an action that causes one of those to be editted by our code, then the comments get overwritten still. Maybe some extra code needs to be added preserver the comments?

Oh, I had the wrong branch checked out.

sean-mcmanus:
Closing this PR was accidental, not sure how that happened.

Extension/src/common.ts
```diff
@@ -123,7 +123,7 @@ export async function ensureBuildTaskExists(taskName: string): Promise<void> {
         throw new Error("Failed to get tasksJsonPath in ensureBuildTaskExists()");
     }
 
-    await writeFileText(tasksJsonPath, JSON.stringify(rawTasksJson, null, settings.editorTabSize));
```
sean-mcmanus, line 126: Text "// TODO: It's dangerous to overwrite this file. We could be wiping out comments." above should be removed.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -16,7 +16,7 @@ import { getCustomConfigProviders } from './customProviders';
 import { SettingsPanel } from './settingsPanel';
 import * as os from 'os';
 import escapeStringRegExp = require('escape-string-regexp');
-import * as jsonc from 'jsonc-parser';
+import * as jsonc from 'comment-json';
```
sean-mcmanus, line 19: This file has to 2 references to JSON.stringify that should be changed to jsonc.stringify.

sean-mcmanus (changes requested):

elahehrashedi:
> @Colengms Was there a GitHub issue tracking the comment removal issue?
This issue happens in Vs Code as well: https://github.com/microsoft/vscode/issues/37138
that's good if we fix it here to preserve comments. However, if the user runs some commands such as "Tasks -> Configure Default Build Task" the comments will be removed by Vs Code.

sean-mcmanus:
@dan-shaw This PR seems close to getting checked in and seems to just need 3 minor changes I mentioned in my previous comments. Did you plan to make those changes or did you want us to do that?

dan-shaw:
@sean-mcmanus Sorry, just fixed it.

sean-mcmanus (approved):

elahehrashedi:
C/C++ extension will keep the comments while modifying tasks.json file. However, Vs Code doesn't. 
I opened an issue on Vs Code to follow up on this issue: https://github.com/microsoft/vscode/issues/105885

PR #5955, opened 2020-08-11
Updates for aarch64

sean-mcmanus:
This is for 0.30.0-insiders3 probably (or maybe and an addition to 0.30.0-insiders2 after it's released for the other platforms). Still working on build definition changes and more testing.


Extension/README.md
```diff
@@ -53,6 +53,7 @@ The extension has platform-specific binary dependencies, therefore installation
 :--- | :---
 `cpptools-linux.vsix` | Linux 64-bit
 `cpptools-linux-armhf.vsix` | Linux ARM
+`cpptools-linux-aarch64.vsix` | Linux ARM 64-bit
```
sean-mcmanus, line 56: I went with using aarch64 instead of arm64 since that's what the architecture is reported as on the 64-bit Raspberry Pi OS (same reason I used armhf instead of just arm)...let me know if you if you think I should change that to arm64.

Extension/package.json
```diff
@@ -2300,6 +2300,20 @@
         "./bin/cpptools-srv"
       ]
     },
+    {
+      "description": "C/C++ language components (Linux / aarch64)",
+      "url": "https://go.microsoft.com/fwlink/?linkid=2135628",
```
sean-mcmanus, line 2305: These links will be updated for the 64-bit versions later, although the 32-bit binaries still work on the 64-bit OS Raspberry Pi OS (not sure if other OS's would have issues).

michelleangela (approved):

PR #5959, opened 2020-08-12
insiders upgrade/downgrade after we recall an insider release

### LINKED ISSUE #5947: insiders upgrade/downgrade is broken if we recall an insider release
While reviewing PR #5946, I was looking at the code that determines a valid array of builds.  It requires 3 or more assets to be marked as "valid".  But we also did a feature where if the assets are removed, then we downgrade a user to an older vsix.  It looks like our validity check will block the feature from being invoked (and actually break all downloads) if we ever remove the assets online, so instead of check `isArrayOfBuilds`, we need to `getArrayOfBuilds` which automatically filters out the recalled releases.  Am I reading the code right?

elahehrashedi:
insiders upgrade/downgrade is broken if we recall an insider release

bug fix: https://github.com/microsoft/vscode-cpptools/issues/5947

I changed the functionality of the isBuild function, to recognize if an input is a "build" type, or if it is a "build" type with 3 assets (i.e. "valid" build type)


Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -35,6 +35,16 @@ suite("UpgradeDowngrade", () => {
                 assert.equal(targetBuild.name, release0);
             });
         });
+        suite("Internal Testing, no Downgrade", () => {
```
elahehrashedi: this new test came here from another branch that is already merged. Please ignore.

Extension/src/githubAPI.ts
```diff
@@ -60,8 +60,11 @@ function isAsset(input: any): input is Asset {
...
  * @return Whether input is of type Build.
  */
-function isBuild(input: any): input is Build {
-    return input && input.name && typeof(input.name) === "string" && isArrayOfAssets(input.assets) && input.assets.length >= 3;
+function isBuild(input: any, isValid: boolean = true): input is Build {
+    // if isValid = true, check if the input is a valid build type with 3 assets,
+    // otherwise, check if the input is a build type.
+    return input && input.name && typeof (input.name) === "string" && isArrayOfAssets(input.assets) &&
+        (!isValid || input.assets.length >= 3);
 }
```
elahehrashedi, line 75: I changed the functionality of the isBuild function, to recognize if an input is a "build" type, or if it is a "build" type with 3 assets (i.e. "valid" build type)

Extension/src/githubAPI.ts
```diff
@@ -84,7 +87,7 @@ function isArrayOfBuilds(input: any): input is Build[] {
     }
     // Only check the five most recent builds for validity -- no need to check all of them
     for (let i: number = 0; i < 5 && i < input.length; i++) {
-        if (!isBuild(input[i])) {
+        if (!isBuild(input[i], false)) {
```
elahehrashedi: Here, I check if the entries are type "build" with any number of assets. During upgrading or downgrading, the "invalid" assets will be filtered out.

Extension/src/githubAPI.ts
bobbrow: I don't think we want to change this.  What we want is for `getReleaseJson` to return a filtered array of 5 or so `Build`s that are all valid.  So we just need to add a new function that gets 5 valid builds out of the json.  I don't think anything else has to change.

The other problem with doing it this way is that `getTargetBuildInfo` is going to have to do more checks to work correctly.  It currently exits early if the first build in the array matches the installed build, but doesn't check that the build is valid.  I think there is an expectation by all the logic involved that `getReleaseJson` should return a valid array of `Build`s.

bobbrow: We talked about this offline. We do still want to include the invalid builds in the array, but we can stop adding entries once we see a "release" version.

Extension/test/unitTests/updowngrade.test.ts
```diff
@@ -60,8 +64,8 @@ suite("UpgradeDowngrade", () => {
                     name: insider2, assets: three_assets}, {
                     name: insider1, assets: three_assets}, {
                     name: release0, assets: three_assets}];
-
-                const userVersion: PackageVersion = new PackageVersion(release1);
+                // In internal testing, the name of the release has a "-master" at the end of it.
+                const userVersion: PackageVersion = new PackageVersion(release1 + "-master");
```
elahehrashedi, line 68: For internal developers, the name of the branch has a "-master" at the end of it. I added this to the tests, and solved it by this check:
`(!testingInsidersVsixInstall && userVersion.suffix && userVersion.suffix !== 'insiders')`

Extension/src/githubAPI.ts
```diff
@@ -192,7 +194,8 @@ export function getTargetBuild(builds: Build[], userVersion: PackageVersion, upd
     }
     const latestVersionOnline: PackageVersion = new PackageVersion(builds[0].name);
     // Allows testing pre-releases without accidentally downgrading to the latest version
-    if (userVersion.isExtensionVersionGreaterThan(latestVersionOnline)) {
+    if ((!testingInsidersVsixInstall && userVersion.suffix && userVersion.suffix !== 'insiders') ||
```
elahehrashedi, line 203: In this version, all the checks are placed in the `getTargetBuild `function. Previously, some of the checked were placed in `getTargetBuildInfo `function.

Extension/src/githubAPI.ts
elahehrashedi: we are including the invalid builds. 
The behaviour of isBuild is as below:
IsValid === false, when the input has a name and the name is a type string, it is a **build**
IsValid === True, when the input has a name and it has 3 assets, it is a **Valid build**

elahehrashedi: So first, I extract all builds (isValid=false) until I reach a valid release build,
and later on, I choose only the valid builds (isValid=true).

Extension/src/githubAPI.ts
```diff
@@ -88,11 +95,11 @@ function getArrayOfBuilds(input: any): Build[] {
     }
     // Only return the the most recent release and insider builds.
     for (let i: number = 0, j: number = 0; i < input.length; i++) {
-        if (isBuild(input[i], false)) {
+        if (isBuild(input[i])) {
             builds[j] = input[i];
             j++;
```
bobbrow: Instead of tracking the index manually, usually you just to a [`.push`](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/array/push) into the array in TypeScript.

bobbrow (approved):

Issue #5751, opened 2020-07-08
Improve the description of the browse.path setting.


sean-mcmanus:
browse.path doesn't control where headers are searched for in regards to IntelliSense (that is includePath) -- it controls where the workspace indexing occurs for the database for source file and when dependent `#include` are found it happens to search for headers at that location in order to tag parse them as well.

The description of browse.path is incorrect/misleading -- I'll create an issue to track improving that.

_Originally posted by @sean-mcmanus in https://github.com/microsoft/vscode-cpptools/issues/5735#issuecomment-655712830_

sean-mcmanus:
The current description is `"A list of paths for the tag parser to use while searching for included headers. Searching on these paths is recursive by default. Specify '*' to indicate non-recursive search. For example: '/usr/include' will search through all subdirectories while '/usr/include/*' will not."`

It should probably say something like "A list of paths to use for indexing and parsing of workspace symbols (for use by Go to Definition, Find All References, etc.)..." And the example path of `'/usr/include'` should probably be changed to `'${workspaceFolder}'`, since that is a more practical example (since system header paths are already added by default).

PR #5974, opened 2020-08-17
Improve the description of the browse.path setting

elahehrashedi:
Improve the description of the browse.path setting
issue: https://github.com/microsoft/vscode-cpptools/issues/5751


michelleangela:
The description will also need to be updated at https://code.visualstudio.com/docs/cpp/c-cpp-properties-schema-reference

michelleangela (approved):

PR #5978, opened 2020-08-17
Generate and validate integrity of downloaded packages

### LINKED ISSUE #5268: Validate crypto signatures of binaries we download
Type: Compliance

michelleangela:
https://github.com/microsoft/vscode-cpptools/issues/5268

- Add function to generate hash keys per runtime dependency package listed in package.json. This can be invoked by using the gulp script "generate-package-hashes". Run this script whenever URL of package changes.
- Add ability to validate hash of runtime dependency package at time of download.
- Added hash or integrity keys to package.json.

### DIFF
--- a/Extension/gulpfile.js
+++ b/Extension/gulpfile.js
@@ -298,16 +298,10 @@ gulp.task("translations-import", (done) => {
 // ****************************
 
 async function DownloadFile(urlString) {
-    const url = new URL(urlString);
-    const options = {
-        host: url.host,
-        path: url.path,
-    };
-
     const buffers = [];
     return new Promise((resolve, reject) => {
         const req = https.request(urlString, (response) => {
-            if (response.statusCode === 301 || response.statusCode === 302 && !response.headers.location) {
+            if (response.statusCode === 301 || response.statusCode === 302) {
                 // Redirect - download from new location
                 let redirectUrl;
                 if (typeof response.headers.location === "string") {
--- a/Extension/gulpfile.js
+++ b/Extension/gulpfile.js
@@ -21,6 +21,8 @@ const vinyl = require('vinyl');
 const parse5 = require('parse5');
 const traverse = require('parse5-traverse');
 const jsonc = require('jsonc-parser'); // Used to allow comments in nativeStrings.json
+const crypto = require('crypto');
+const https = require('https');
 
 
 // Patterns to find HTML files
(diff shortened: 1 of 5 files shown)


Extension/package.json
```diff
@@ -2194,6 +2194,7 @@
     "translations-import": "node ./tools/prepublish.js && gulp translations-import",
     "prepublishjs": "node ./tools/prepublish.js",
     "pretest": "tsc -p test.tsconfig.json",
+    "generatePackageHashes": "gulp generate-package-hashes",
```
Colengms, line 2211: Should this run as part of the build process, so `integrity` will always get automatically updated before webpacked?  That might protect us from forgetting to update it manually.

Extension/src/packageManager.ts
```diff
@@ -300,7 +317,14 @@ export class PackageManager {
                             }
                         });
 
-                        response.on('end', resolve);
+                        response.on('end', () => {
+                            const packageBuffer: Buffer = Buffer.concat(buffers);
+                            if (isValidPackage(packageBuffer, pkg.integrity)) {
+                                resolve();
+                            } else {
+                                reject(new PackageManagerError('Invalid content received. Hash is incorrect.', localize("invalid.content.received", 'Invalid content received. Hash is incorrect.'), 'DownloadFile', pkg));
```
Colengms, line 325: Although this whole scenario is unlikely, should this error be more geared at advising the user what to do about the issue?

Extension/gulpfile.js
```diff
@@ -290,6 +292,108 @@ gulp.task("translations-import", (done) => {
...
+            console.log(`No hash generated for package '${dependency.description}`);
+        }
+        console.log(`\n`);
+    }
+
+    let content = JSON.stringify(packageJson, null, 2);
+    return content;
+}
+
+gulp.task('generate-package-hashes', async (done) => {
```
Colengms, line 389: I don't think anything else we do in gulpfile.js is 'async'.  Does this imply that when this script is run, it will return before the operation is completed?

michelleangela: It would error without the async. I found that it needs there because there is an await, and without the await, it would not wait for the async operation to complete.

Extension/package.json
michelleangela: Yes, I also thought to run "generate-package-hashes" during the build process. But I don't know yet where in the build process to put it. Would this be in the yaml file?

Extension/src/packageManager.ts
michelleangela: If the hash incorrect, then the whole install would fail. I'll see what messages is further given to the user when the whole install is incomplete or if the entire install is undone.

Extension/package.json
Colengms: build.yml (and build.windows.yml), I think.  Sean could confirm.  It would go before `yarn run compile`, so it's already available when that command webpacks.

Extension/gulpfile.js
Colengms: It looks like the await is required because DownloadFile was implemented using a Promise rather than sync.  Here, it would be fine if that were sync.  My concern is that being async, maybe the command being executed might not be finished before it returns, and the next command executed.  Maybe a race in which running generatePackageHashes returns before they are actually generated, and the next command (yarn run compile) picks up the file before it's ready.  Unlikely, but would be good to ensure there isn't a possible issue here.  The gulp process probably waits for 'done' to be signalled, I'm just not sure.

sean-mcmanus (approved):

Extension/src/packageManager.ts
michelleangela: I tested that if it rejects the downloaded package then, the install fails. Failed installs results in message to reinstall again.

Extension/package.json
michelleangela: GenerateHashes script is set to run before compiling in yml build files.

Extension/gulpfile.js
michelleangela: Verified that the gulp task waits before running other tasks.

PR #5989, opened 2020-08-19
Adding the defaultFolder to the configuration.includePath

### LINKED ISSUE #5621: ${workspaceFolder}/** is auto-added when default.includePath is set and no c_cpp_properties.json exists
**Type: LanguageService**

**Describe the bug**
- OS and Version:
MacOS 10.15.4

- VS Code Version:
1.45.1

- C/C++ Extension Version:
0.28.2

Remote -- SSH 

- Does this issue involve using SSH remote to run the extension on a remote machine?:
Yes

After VSCode is being used a while (say, couple of hours) Goto Symbol in Editor could take tens of seconds to show the list of symbols in the current file. Closing VSCode and killing all vscode-related processes on the remote machine, and then restarting everything back solves the issue... for some time.

Is there anything I could do to help diagnosing this? E.g. should I collect the logs when this feature becomes laggy?

Thanks in advance.

**Steps to reproduce**

**Expected behavior**

<details>
  <summary><strong>Logs</strong></summary>

```
Insert logs here.
```
</details>

**Screenshots**

**Additional context**

elahehrashedi:
Adding only the `defaultFolder` to the `configuration.includePath`, when the default `includePath `is set.
Improve the description of `inludePath`.
bug fix: https://github.com/microsoft/vscode-cpptools/issues/5621


Extension/c_cpp_properties.schema.json
```diff
@@ -64,8 +64,7 @@
                         "type": "string"
                     },
                     "includePath": {
-                        "description": "A list of paths for the IntelliSense engine to use while searching for included headers. Searching on these paths is not recursive.",
-                        "type": "array",
+                        "description": "A list of paths for the IntelliSense engine to use while searching for included headers. Searching on these paths is not recursive. Specify '**' to indicate recursive search. For example, '${workspaceFolder}/**' will search through all subdirectories while '${workspaceFolder}' will not.",                        "type": "array",
```
bobbrow: nit: reformat this line.  It looks like `"type": "array"` got merged into this line.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -308,7 +308,7 @@ export class CppProperties {
         if (isUnset(settings.defaultIncludePath)) {
             configuration.includePath = [rootFolder].concat(this.vcpkgIncludes);
         } else {
-            configuration.includePath = [rootFolder].concat(this.vcpkgIncludes).concat([defaultFolder]);
+            configuration.includePath = [defaultFolder].concat(this.vcpkgIncludes);
```
bobbrow: I don't think we want to add the `vcpkgIncludes` here either because that's not the behavior they get when they don't have `c_cpp_properties.json`.  Just copy the `defaultFolder` into the includePath (`slice(0)`, I think)

michelleangela:
The documentation https://code.visualstudio.com/docs/cpp/c-cpp-properties-schema-reference also needs an update.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Yes

elahehrashedi:
@michelleangela I sent a PR to update the vscode document:
https://github.com/microsoft/vscode-docs/pull/3952

Extension/src/LanguageServer/configurations.ts
elahehrashedi: the `defaultFolder `is a string:  `const defaultFolder: string = "${default}";`

Will this be good?  `configuration.includePath = [defaultFolder]`, 

or should I find the array of default folders defined in `c_cpp.default.includePath` and add them here (using` slice(0)`)?

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: defaultFolder seems good

Extension/c_cpp_properties.schema.json
```diff
@@ -64,7 +64,7 @@
                         "type": "string"
                     },
                     "includePath": {
-                        "description": "A list of paths for the IntelliSense engine to use while searching for included headers. Searching on these paths is not recursive.",
+                        "description": "A list of paths for the IntelliSense engine to use while searching for included headers. Searching on these paths is not recursive. Specify '**' to indicate recursive search. For example, '${workspaceFolder}/**' will search through all subdirectories while '${workspaceFolder}' will not.",
```
michelleangela, line 67: I forgot about the html of the configuration UI will also need an update. See file https://github.com/microsoft/vscode-cpptools/blob/master/Extension/ui/settings.html . Or maybe the description on that UI is good?

michelleangela (changes requested):

Extension/c_cpp_properties.schema.json
sean-mcmanus: The description in the settings.html seems good to me. Having 2 different descriptions seems okay as well...might help users understand better when the same thing is described in slightly different ways.

elahehrashedi: The new description:
image

It matches the new description in the Docs (the new description is checked in).
image

michelleangela (approved):

Issue #6012, opened 2020-08-23
C++ wrong syntax check


yihuajack:
Issue Type: <b>Bug</b>

I use C/C++ extension in Windows environment. Using CMake to build is successful and the executable file gives the correct answer, but it wrongly shows me four syntax errors. I have included <iostream> and using namespace std but it still tells me that cin is unrecognizable identifer, and the operator << also not matched. Besides, it tells me the class vector has no proper copy constructor and no matched operator for != of vector<int>. The program runs right in Visual Studio. Such error does not exist in WSL remote mode. I have tested the problem only occurs if C/C++ extension is enabled and not relevant to other extensions.

Extension version: 0.30.0-insiders3
VS Code version: Code 1.48.1 (3dd905126b34dcd4de81fa624eb3a8cbe7485f13, 2020-08-19T17:12:13.244Z)
OS version: Windows_NT x64 10.0.19041

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 x 2208)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>protected_video_decode: enabled<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>viz_display_compositor: enabled_on<br>viz_hit_test_surface_layer: disabled_off_ok<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.79GB (4.33GB free)|
|Process Argv||
|Screen Reader|no|
|VM|50%|
</details>

Part of code:
```
int main() {
    int i, num;
    string output;
    vector<int> nums;
    for (i = 0; i < 3; i++) {
        cin >> num;
        nums.push_back(num);
    }
    Solution s;
    output =  s.judgePoint24(nums) ? "True"  : "False";
    cout << output << endl;
}
```

sean-mcmanus:
We need more info. Most likely something configured incorrectly in regards to your system includes. Can you provide the output of running the C/C++: Log Diagnostics command?

yihuajack:
[...]

PR #6023, opened 2020-08-25
Remove unintended loc.

sean-mcmanus:
One text in the C/C++ Output is being localized. The intent was that it was not to be localized. If we later choose to localize the text, it should all be localized. See https://github.com/microsoft/vscode-cpptools/issues/6012#issuecomment-679966603 for an example.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2982,7 +2982,7 @@ export class DefaultClient implements Client {
             break;
         }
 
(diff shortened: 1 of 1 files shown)


elahehrashedi (approved):

elahehrashedi:
why do we localize some messages and not localize some others?

sean-mcmanus:
> why do we localize some messages and not localize some others?

We didn't localize the C/C++ Log Diagnostics messages because it was intended to be used by us...but we could change that later, since "Debug" strings are localized. This change just makes it consistent within the logging.

Issue #2595, opened 2018-10-02
__cplusplus macro is not properly evaluated


graph:
**Type: LanguageService**

c_cpp_properties.json
```
{
    "configurations": [
        {
            "name": "Win32",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "defines": [
                "_DEBUG",
                "UNICODE",
                "_UNICODE"
            ],
            "windowsSdkVersion": "10.0.17134.0",
            "compilerPath": "C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.15.26726/bin/Hostx64/x64/cl.exe",
            "cStandard": "c11",
            "cppStandard": "c++17",
            "intelliSenseMode": "msvc-x64",
            "compileCommands": "${workspaceFolder}/build/release/compile_commands.json"
        }
    ],
    "version": 4
}
```

**Describe the bug**
- OS and Version: Windows 10
- VS Code Version:
Version: 1.29.0-exploration (user setup)
Commit: f28f62c944c3f72b8c03deaff6f98ea3741ca677
Date: 2018-10-02T05:46:41.453Z
Electron: 3.0.2
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
Architecture: x64

- C/C++ Extension Version: 0.19.0

**To Reproduce**
have a c++ file with this code and above c_cpp_properties.json:

```
#pragma once

#ifdef __cplusplus
#if __cplusplus < 201402L
#error "C++14 or higher is required"
#endif
#endif
```

The plugin reports an error that C++14 or higher is required

bobbrow:
Issue #2248 and https://blogs.msdn.microsoft.com/vcblog/2018/04/09/msvc-now-correctly-reports-__cplusplus/ have some more background on this issue.

Short story is that this was actually the expected behavior until earlier this year when `/Zc:__cplusplus` support was added to cl.exe.  I think it was a mistake to close #2248 and we should use this issue to track adding support for this in IntelliSense.

sean-mcmanus:
Support for `/Zc:__cplusplus` in compilerArgs was added with https://github.com/microsoft/vscode-cpptools/releases/tag/0.30.0-insiders4 .

PR #6028, opened 2020-08-25
reports the correct __cplusplus macro for MSVC IntelliSense Mode

elahehrashedi:
feature-request: https://github.com/microsoft/vscode-cpptools/issues/2595
the language-server side is also altered to accommodate this feature. 
The reason I am adding a default value for _Cplusplus is that if the IntelliSense mode is MSVC and the std version is c++17, the default value for _cplusplus is  201703L.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -105,6 +105,7 @@ export interface CompilerDefaults {
     windowsSdkVersion: string;
     intelliSenseMode: string;
     rootfs: string;
+    cplusplusMacro: string;
 }
 
 export class CppProperties {
@@ -126,6 +127,7 @@ export class CppProperties {
     private vcpkgIncludes: string[] = [];
     private vcpkgPathReady: boolean = false;
     private defaultIntelliSenseMode?: string;
+    private defaultCplusplusMacro?: string;
     private defaultCustomConfigurationVariables?: { [key: string]: string };
     private readonly configurationGlobPattern: string = "c_cpp_properties.json";
     private disposables: vscode.Disposable[] = [];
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
How does this work? I wouldn't think "/Zc" would be a valid define. It would think it would have be passed as a compilerArg, although then cpptools would have to be updated to process that.

bobbrow:
~~Does cl.exe set this by default in newer versions?~~  You have to opt in.  I didn't expect to see any change to the TypeScript for this issue.

EDIT: From: https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
You need to compile with the `/Zc:__cplusplus` switch to see the updated value of the `__cplusplus` macro. We tried updating the macro by default and discovered that a lot of code doesn’t compile correctly when we change the value of `__cplusplus`. **We’ll continue to require use of the `/Zc:__cplusplus` switch for all minor versions of MSVC in the 19.xx family.**

sean-mcmanus:
> ~Does cl.exe set this by default in newer versions?~ You have to opt in. I didn't expect to see any change to the TypeScript for this issue.
> 
> EDIT: From: https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
> You need to compile with the `/Zc:__cplusplus` switch to see the updated value of the `__cplusplus` macro. We tried updating the macro by default and discovered that a lot of code doesn’t compile correctly when we change the value of `__cplusplus`. **We’ll continue to require use of the `/Zc:__cplusplus` switch for all minor versions of MSVC in the 19.xx family.**

That seems to imply we shouldn't change the defaults?

bobbrow:
We shouldn't change the defaults for now.  When the compiler updates the version to 20.xx we may have to make some changes, but we're still on 19.xx now.

Issue #414, opened 2016-12-19
File encoding specified in "files.encoding" setting is not honored


strrchr:
In my workspace's settings.json 
{
 "files.encoding": "gbk"
}
![default](https://cloud.githubusercontent.com/assets/8603093/21301997/91e80c2e-c5ed-11e6-9153-0b744338344d.png)
These is still encoding problem in the auto completetion.

sean-mcmanus:
I'm not able to repro the problem. From your screenshot, it looks like you may be using another C++ extension for auto-completion, because our extensions doesn't provide test like that. Can you provide a sample that repros this? I created a file with GBK encoding and I didn't see a problem.

strrchr:
- VSCode Version: Code 1.8.1 (ee428b0eead68bf0fb99ab5fdc4439be227b6281, 2016-12-19T14:49:23.350Z)
- OS Version: Windows_NT ia32 6.1.7601
- Extensions:

|Extension|Author|Version|
|---|---|---|
|xml|DotJoshJohnson|1.6.0|
|auto-close-tag|formulahendry|0.3.6|
|auto-rename-tag|formulahendry|0.0.8|
|green-theme|vscode|0.1.0|
|beautify|HookyQR|0.6.2|
|theme-material-theme|jprestidge|1.0.1|
|cpptools|ms-vscode|0.9.3|
|vscode-icons|robertohuertasm|4.2.0|

---

![default](https://cloud.githubusercontent.com/assets/8603093/21335841/7d07c83a-c69c-11e6-87d0-e84d1241562a.png)

The gbk file should contain chinese characters.
Such as: 中文字符中文字符中文字符

sean-mcmanus:
Okay, I am able to repro the bug now. Thanks. I don't know how you got the text write(sprint ("%s..."... to appear, but I was able to get a repro via int func2(string s = "中文字符中文字符中文字符") {...

sean-mcmanus:
This bug was fixed a while ago.

sean-mcmanus:
Oops, it's not fixed. Looks like we don't handle GBK encoding correctly.

sean-mcmanus:
It looks like "most" functionality is broken when it interacts with 2-byte GBK encoding Chinese characters. Our code assumes UTF-8 characters are being used and we never check the encoding.

sean-mcmanus:
[...]

PR #6040, opened 2020-08-27
Plumbing for sending 'files.encoding' to native process

Colengms:
* Added plumbing for sending 'files.encoding' to native process. See https://github.com/microsoft/vscode-cpptools/issues/414
* Cleaned up some uses of null in settings.ts.


sean-mcmanus:
You should add libiconv's LGPL license to the ThirdPartyNotices.txt in the Additional Third Party Notices section at the bottom. I'm planning to refactor the license later, but it'll be easier if it's not missing stuff that I might forget to add.

Colengms:
> You should add libiconv's LGPL license to the ThirdPartyNotices.txt in the Additional Third Party Notices section at the bottom. I'm planning to refactor the license later, but it'll be easier if it's not missing stuff that I might forget to add.

Do you know if that is necessary?  We are not installing the library on the system.  The man pages for the API show up on Linux and Mac without any reference to libiconv or a license.  It would seem that the OSes have already addressed the licensing.  We may need to ask legal.

Extension/src/LanguageServer/client.ts
```diff
@@ -1772,6 +1778,9 @@ export class DefaultClient implements Client {
             cppSettingsScoped["default"] = { systemIncludePath: cppSettingsResourceScoped.get("default.systemIncludePath") };
         }
 
+        const otherSettingsWorkspace: OtherSettings = new OtherSettings(this.RootUri);
```
sean-mcmanus: This shouldn't be named "workspace", because existing code like workspaceOtherSettings is being used for that. It could be WorkspaceFolder and the Global one could be Workspace.

sean-mcmanus (changes requested):

sean-mcmanus:
> Do you know if that is necessary? We are not installing the library on the system. The man pages for the API show up on Linux and Mac without any reference to libiconv or a license. It would seem that the OSes have already addressed the licensing. We may need to ask legal.

See https://www.gnu.org/software/libiconv/ . The iconv.h is covered by the LGPL license.

Colengms:
> > Do you know if that is necessary? We are not installing the library on the system. The man pages for the API show up on Linux and Mac without any reference to libiconv or a license. It would seem that the OSes have already addressed the licensing. We may need to ask legal.
> 
> See https://www.gnu.org/software/libiconv/ . The iconv.h is covered by the LGPL license.

The page you are referecing mentions: "This library provides an iconv() implementation, for use on systems which don't have one, or whose implementation cannot convert from/to Unicode."

The OSes we are targetting have iconv() implementations.

sean-mcmanus:
> The OSes we are targetting have iconv() implementations.

Oh...I see, so we're not using the GNU libiconv library. So, yeah, I'm not sure if we need a TPN or not.

Colengms:
> > The OSes we are targetting have iconv() implementations.
> 
> Oh...I see, so we're not using the GNU libiconv library. So, yeah, I'm not sure if we need a TPN or not.

The wikipedia page states: "Most Linux distributions provide an implementation, either from the GNU Standard C Library (included since version 2.1, February 1999), or the more traditional GNU libiconv, for systems based on other Standard C Libraries.

The iconv function[6] on both is licensed as LGPL, so it is linkable with closed source applications."

So, it seems like a good idea to include the license.  I added something simple.  Let me know if I need to do something different.

sean-mcmanus:
> So, it seems like a good idea to include the license. I added something simple. Let me know if I need to do something different.

Yeah, this is what the header shows with gcc 10.2. The dates are probably not correct because we build with gcc 9 currently (might be 2019).

/* Copyright (C) 1997-2020 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <https://www.gnu.org/licenses/>.  */

Colengms:
> This file is part of the GNU C Library.

That doesn't seem appropriate in a file that is outside of the GNU C Library.  :)  Also, now that I consider the comment from the Wikipedia page, I'm back to thinking it's the OS's responsibility to address this license, as the library is provided with the OS.  The page implies the library "is licensed" by the OS.

sean-mcmanus:
> > This file is part of the GNU C Library.
> 
> That doesn't seem appropriate in a file that is outside of the GNU C Library. :) Also, now that I consider the comment from the Wikipedia page, I'm back to thinking it's the OS's responsibility to address this license, as the library is provided with the OS. The page implies the library "is licensed" by the OS.

But our code uses the header to build and links to the library.

sean-mcmanus (approved):

michelleangela (approved):

PR #6098, opened 2020-09-09
Fix casing of some loc strings

Colengms:
The loc team opened some bugs due to confusion about these strings, why they were not-capitalized, and if they should do the same.


sean-mcmanus (approved):

bobbrow:
They are the strings for enum values and were not meant to be capitalized.

EDIT: (i.e. they are not complete sentences)

Extension/package.nls.json
```diff
@@ -24,25 +24,25 @@
...
     "c_cpp.configuration.vcFormat.indent.withinParentheses.alignToParenthesis.description": "New line is aligned under the opening parenthesis",
     "c_cpp.configuration.vcFormat.indent.withinParentheses.indent.description": "New line is indented based on `C_Cpp.vcFormat.indent.multiLineRelativeTo`",
     "c_cpp.configuration.vcFormat.indent.preserveWithinParentheses.description": "In existing code, preserve the existing indent alignment of new lines within parentheses",
     "c_cpp.configuration.vcFormat.indent.caseLabels.description": "Labels are indented relative to switch statements by the amount specified in the Editor: Tab Size setting",
-    "c_cpp.configuration.vcFormat.indent.caseContents.description": "Code inside case is indented relative to its label by the amount specified in the Editor: Tab Size setting",
+    "c_cpp.configuration.vcFormat.indent.caseContents.description": "Code inside case block is indented relative to its label by the amount specified in the Editor: Tab Size setting",
```
bobbrow, line 35: Keep this change, but I think the capitalizations should be reverted.

bobbrow (changes requested):

Extension/package.nls.json
sean-mcmanus: Maybe they should be changed to be complete sentences? Do you know of another example of an enum value not having a complete sentence description?

Colengms: I don't think it makes sense for these to be non-capitalized partial sentences.  This does not consider other languages in which it may not be possible to split these sentences in the same way.  It's also difficult to provide guidance to the Loc team on how to maintain these strings if not self-contained (as there isn't a way to add hints to package.nls.json).

Colengms: The only prior use of `enumDescriptions` was `C_Cpp.formatting`, which are complete sentences, were not continuations of a sentence, and were capitalized (except when starting with a 'clang-format', as the binary name is not capitalized).  Note that these enum descriptions are also not continuations of a sentence.  So, I made them complete sentences.  Also, I omitted periods, as we generally do not appear to be using them in this file, even for complete sentences.

sean-mcmanus: That seems good with the sentences and capitalization, but I'm seeing the VS Code setting descriptions end with periods? Can we update them all to end in periods?

sean-mcmanus (approved):

PR #6109, opened 2020-09-10
Update changelog for 1.0 and update aarch64 fwlinks

sean-mcmanus:


elahehrashedi (approved):

Extension/CHANGELOG.md
```diff
@@ -1,91 +1,61 @@
...
+* Validate crypto signatures of binaries we download. [#5268](https://github.com/microsoft/vscode-cpptools/issues/5268)
 * Add link to the documentation in the configuration UI. [#5875](https://github.com/microsoft/vscode-cpptools/issues/5875)
   * Abhishek Pal (@devabhishekpal) [PR #5991](https://github.com/microsoft/vscode-cpptools/pull/5991)
+* Allow comments, trailing commas, etc. in `c_cpp_properties.json` [#5885](https://github.com/microsoft/vscode-cpptools/issues/5885)
+* Prevent comments from being removed from json files when the extension modifies them.
+  * @dan-shaw [PR #5954](https://github.com/microsoft/vscode-cpptools/pull/5954)
+* Add diagnostics on potentially conflicting recursive includes to `C/C++: Log Diagnostics`, i.e. if a workspace uses files with the same name as system headers. [#6009](https://github.com/microsoft/vscode-cpptools/issues/6009)
+* Add workspace parsing diagnostics. [#6048](https://github.com/microsoft/vscode-cpptools/issues/6048)
+* Add `wmain` snippet on Windows. [#6064](https://github.com/microsoft/vscode-cpptools/issues/6064)
+* More C++20 support (with more on the way in 1.1.0).
```
elahehrashedi: Do we want to keep this? (with more on the way in 1.1.0). We always have more on the way.

sean-mcmanus: ...not sure. It's not guaranteed (to users) that 1.1.0 would have new C++20 support, so I added the "more on the way", because I know there are bunch more changes that are going into 1.1.0.

michelleangela: I suggest to remove "more on the way in 1.1.0" and just leave the message as "More C++ support" to let users know that some C++ 20 support has been added. The change log does not have to let users know what may be coming in the future. Also, nothing is ever guaranteed, so better to leave info about potential future stuff off the change log.

michelleangela (changes requested):

Extension/CHANGELOG.md
sean-mcmanus: Done.

Colengms (approved):

michelleangela (approved):

Issue #6132, opened 2020-09-13
"Comments are not permitted in JSON" appears when c_cpp_properties.json is open but hasn't been made active yet


sean-mcmanus:
It goes away after the c_cpp_properties.json is made active. Not sure if there's a way to call setTextDocumentLanguage on the files open in the workspace beforehand.

sean-mcmanus:
vscode.workspace.textDocuments doesn't contain first that haven't been activated yet.

sean-mcmanus:
There's a thread at https://github.com/microsoft/vscode/issues/15178

sean-mcmanus:
Oh...we can just check if the file exists, even if it's not open.

PR #6133, opened 2020-09-13
Fix jsonc handling for non-active c_cpp_properties.json.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/6132


sean-mcmanus:
This doesn't seem high priority enough to rebuild the vsix.

michelleangela (approved):

PR #6141, opened 2020-09-15
Fix issue with LSP process not getting shut down gracefully

Colengms:
This addresses an issue with how we were shutting down the LSP process.  We were not calling stop() due to an incorrect check, resulting in the shutdown message never being delivered to the LSP process.  This actually only impacted Windows users (likely causing frequent database rebuilds, as safety checks failed), due to Windows killing all child processes as VS Code terminates.  On Mac and Linux, the native process continued to execute, and gracefully shut down (and closed the DB properly) due to closing of std::cin.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -3103,7 +3103,7 @@ export class DefaultClient implements Client {
     }
 
     public dispose(): Thenable<void> {
-        const promise: Thenable<void> = (this.languageClient && clientCollection.Count === 0) ? this.languageClient.stop() : Promise.resolve();
+        const promise: Thenable<void> = this.languageClient ? this.languageClient.stop() : Promise.resolve();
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
I think I added that for the multi-root scenario. Can you check if the multi-root scenario still works? Or I could double-check later. The problem is that in the multi-root scenario, there are multiple clients but we only want to shutdown the 1 languageClient when all the Clients are gone (multiple Clients share a single languageClient).

Colengms:
> I think I added that for the multi-root scenario. Can you check if the multi-root scenario still works? Or I could double-check later. The problem is that in the multi-root scenario, there are multiple clients but we only want to shutdown the 1 languageClient when all the Clients are gone (multiple Clients share a single languageClient).

Aha.  The check should probably be === 1 then.

sean-mcmanus (approved):

Issue #6166, opened 2020-09-18
IntelliSense gives "No IL" (or crashes) in a certain scenario on macOS


Colengms:
With 1.0.1-insiders, I noticed cpptools become unresponsive, though so far I only repro on macOS (Catalina 10.15.6).  Stacks are below.  cpptools-srv was at 100%.  An IntelliSense update appears to have been stuck.  I waited >10 minutes.  I have a project that repro's consistently, but reducing it may be difficult.

<details><summary>cpptools</summary>
<p>

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff6c41f69e libsystem_kernel.dylib`__read_nocancel + 10
    frame #1: 0x00007fff6c36c2cb libsystem_c.dylib`_sread + 16
    frame #2: 0x00007fff6c36b946 libsystem_c.dylib`__srefill1 + 24
    frame #3: 0x00007fff6c36ba68 libsystem_c.dylib`__srget + 14
    frame #4: 0x00007fff6c367d1c libsystem_c.dylib`getc + 52
    frame #5: 0x00007fff695bbbc9 libc++.1.dylib`std::__1::__stdinbuf<char>::__getchar(bool) + 109
    frame #6: 0x0000000109113656 cpptools`std::__1::basic_istream<char, std::__1::char_traits<char> >& std::__1::getline<char, std::__1::char_traits<char>, std::__1::allocator<char> >(std::__1::basic_istream<char, std::__1::char_traits<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&, char) + 150
    frame #7: 0x00000001090d474c cpptools`vscode::message_handler::main_loop() + 1724
    frame #8: 0x00000001090d121e cpptools`main + 206
    frame #9: 0x00007fff6c2dccc9 libdyld.dylib`start + 1
  thread #2
    frame #0: 0x00007fff6c41e81e libsystem_kernel.dylib`read + 10
    frame #1: 0x00000001091be48e cpptools`comm_pipe_nix::read(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) + 94
[...]

PR #6177, opened 2020-09-21
1.0.1

bobbrow:


sean-mcmanus:
I'm waiting to see if we want to take the fix for https://github.com/microsoft/vscode-cpptools/issues/6166 or not.

Also, I'll push changes to the integrity hashes.

sean-mcmanus:
> I'm waiting to see if we want to take the fix for #6166 or not.
> 
> Also, I'll push changes to the integrity hashes.

Looks like fixing #6166 will have to be for later.

sean-mcmanus:
> Looks like fixing #6166 will have to be for later.

Actually, Colen says the it fixes the infinite loop but gives the "No IL", but that's intentional until we can get a better fix.

sean-mcmanus (approved):

PR #6240, opened 2020-10-01
Add memory usage limit of an IntelliSense process

### LINKED ISSUE #6230: Shutdown IntelliSense cpptools-srv process when it reaches N amount of memory usage
There has been issues on memory usage around IntelliSense engine process on macOS and Linux platforms. On Windows, the IntelliSense engine process will shutdown when it reaches a certain threshold of memory usage, therefore there has been no feedback issues on memory usage on Windows as the process will terminates before it uses an undesirable amount of memory.

To mitigate the memory usage issue on macOS and Linux, add a mechanism to terminate the IntelliSense engine process when it reaches N amount of memory usage. 

Value of memory usage limit:

- The limit is configurable
- default is 4 GB
- minimum is 256 MB
- maximum is 6 GB

michelleangela:
This is to address issue #6230 

- Add setting to set memory usage limit of an IntelliSense process.
- Default limit is 4096 MB
- Minimum limit is 256 MB
- Maximum limit is 16384 MB or 16 GB


Extension/package.json
```diff
@@ -848,6 +848,13 @@
           "scope": "machine-overridable",
           "minimum": 0
         },
+        "C_Cpp.intelliSenseMemoryLimit": {
+          "type": "number",
+          "default": 4096,
+          "description": "%c_cpp.configuration.intelliSenseMemoryLimit.description%",
+          "scope": "machine-overridable",
+          "minimum": 2048
```
sean-mcmanus: Should we enable this to be lower, e.g. for Raspberry PI machines with low memory?

sean-mcmanus (changes requested):

Extension/package.nls.json
```diff
@@ -133,6 +133,7 @@
...
     "c_cpp.configuration.intelliSenseCachePath.description": "Defines the folder path for cached precompiled headers used by IntelliSense. The default cache path is \"%LocalAppData%/Microsoft/vscode-cpptools\" on Windows, \"$XDG_CACHE_HOME/vscode-cpptools/\" on Linux (or \"$HOME/.cache/vscode-cpptools/\" if XDG_CACHE_HOME is not defined), and \"$HOME/Library/Caches/vscode-cpptools/\" on Mac. The default path will be used if no path is specified or if a specified path is invalid.",
     "c_cpp.configuration.intelliSenseCacheSize.description": "Maximum size of the per-workspace hard drive space in megabytes for cached precompiled headers; the actual usage may fluctuate around this value. The default size is 5120 MB. Precompiled header caching is disabled when the size is 0.",
+    "c_cpp.configuration.intelliSenseMemoryLimit.description": "Memory usage limit of an IntelliSense process. The default is 4096 MB. The extension will shutdown an IntelliSense process when it reaches the limit. Set this to a high value (e.g. 99999 MB) to stop automatic shutdown of an IntelliSense process.",
```
bobbrow: Maybe call out in the first sentence that the units are "in megabytes"

Extension/package.json
michelleangela: I'll make it 1024 MB (1 GB). The smallest Raspberry Pi memory is 2 GB.

sean-mcmanus: Raspberry Pi 3 is 1 GB. Can the limit be 256 MB? The OS seems to start freezing when processes start using > 256 MB. Unfortunately, VS Code itself easily uses > 256 MB, so the user is likely to hit that issue before they hit that issue with cpptools-srv....well, with VS Code running locally, remote is probably better.

sean-mcmanus (approved):

bobbrow (approved):

PR #6241, opened 2020-10-02
Add custom actions to close and lock stale issues

Colengms:
Custom actions, based on the locker and needs-more-info-closer actions here: https://github.com/microsoft/vscode-github-triage-actions

Much of this code was copied from vscode's actions, and does not align with the coding style of cpptools.  Since this is not part of cpptools itself, changing a lot of code to adjust the style doesn't seem particularly important to me, unless others disagree.

There are 2 actions:

* Locker - Locks a closed issue after it has been closed for a period of time, and there has been no activity for a period of time.

* StaleCloser - If an issue had no activity for more than a specific period of time, it is either closed (with a message to the user), or, optionally, if there is an assignee and the last commenter was not someone with write or admin permissions to the repo (or a member of 'additionalTeam'), a comment is added which pings both the author and assignee.  If there is no assignee, no one is pinged.  (There is not a good way to know who to ping, since the last commenter was not a team member.  Alternatively, I could scan all comments for all team members, and ping them all, or pick one?).  Optionally, a specific label can be applied to the issue when closed.

I've added a common base class which generically handles filtering candidate issues based on labels (to include/exclude, or 'no:label'), milestone (singular required milestone, a set of milestones to exclude, or 'no:milestone'), minimum number of votes, or maximum number of votes.

The following workflows are based on the StaleCloser action: question-closer, more-info-needed-closer, feature-request-closer

The feature-request-closer is based on our initial idea for how it would work: It will close an inactive Feature Request with less than a minimum number of votes.  It will not be locked, so users may continue to up-vote it.

Feel free to suggest policy changes, changes to the comment messages, etc..

I've manually tested these in a test repo, using delays of 0 days (immediate) and running them directly via the Actions tab.  But, we should still look very closely at these, as they will be making bulk changes to our issues.

... It occurs to me that I may be able to combine these 2 actions into a single action, by adding params to indicates whether candidate issues should be filtered by open/closed/locked and whether the result of the operation should be to open/close/lock the issue.  That would make a singular generic data-driven action...


.github/actions/.gitignore
```diff
@@ -0,0 +1,3 @@
+# ignore dependency packages
```
bobbrow, line 1: speaking of ignore files, you should update `.vscodeignore` to ignore the `.github` folder so it doesn't end up in our VSIX.

.github/workflows/feature-request-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v2
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          labels: Feature Request
+          ignoredLabels: debugger
+          addLabels: more votes needed
+          closeDays: 14
```
bobbrow: 14 days seems short to close feature requests.  I was thinking we'd keep it at 60.

.github/workflows/more-info-needed-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          labels: more info needed
+          ignoredLabels: debugger
+          closeDays: 14
+          closeComment: "This issue has been closed automatically because it needs more information and has not had recent activity."
+          pingDays: 80
```
bobbrow, line 22: I don't understand the 80 for ping vs 14 for close here.  Won't the issue be closed and locked already?

.github/workflows/question-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          labels: question
+          ignoredLabels: debugger
+          closeDays: 14
+          closeComment: "This issue has been closed automatically because it's labeled as a 'question' and has not had recent activity."
+          pingDays: 80
```
bobbrow, line 22: same comment as above about the 80 days for ping.

.github/workflows/question-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v2
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          labels: question
+          ignoredLabels: debugger
```
sean-mcmanus, line 19: Why would the debugger issues be ignored?

.github/workflows/more-info-needed-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Checkout Actions
+        uses: actions/checkout@v2
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          labels: more info needed
+          ignoredLabels: debugger
+          closeDays: 14
```
sean-mcmanus: Is this 14 day value enough time? -- it took around 4 weeks to get enough info on the recent "No IL" regression...although in that case there happened to be a continuous stream of "activity".

.github/workflows/feature-request-closer.yml
```diff
@@ -0,0 +1,23 @@
...
+          maximumVotes: 1
+          closeComment: "This feature request is being closed due to insufficient upvotes.  When at least 2 upvotes are received, this issue will be elidgable for our backlog."
```
sean-mcmanus: Spelling

sean-mcmanus (changes requested):

.github/actions/.gitignore
Colengms: This is not valid, correct?  `.github. is at the root of the repo.  The extension is built from the Extension subdirectory, which is also where the .vscodeignore file is.

.github/workflows/more-info-needed-closer.yml
Colengms: StaleCloser.ts is fairly brief, so should be easy to decipher.  The issue is not closed if pingDays is set and if the last commenter is not specified in additionalTeam and does not have write or admin access to the repo (or is a bot).  If there is an assignee, a comment will be posted to ping them instead (once per pingDays).

.github/workflows/question-closer.yml
Colengms: Bob asked for debugger issues to be ignored.  But, we can ask Andrew if he wants to opt in to the policies.

.github/workflows/more-info-needed-closer.yml
Colengms: These timeouts are based on the last update, so latest activity.  The issue would need to have no activity for that number of days, (which I am increasing to 60 days, based on Bob's feedback).  It also is not automatically closed if the last commenter is not on the team (if not in additionalTeam field, does not have write or admin access to the repo, and is not a bot).

.github/workflows/feature-request-closer.yml
```diff
@@ -18,6 +18,6 @@ jobs:
           labels: Feature Request
           ignoredLabels: debugger
           addLabels: more votes needed
-          closeDays: 14
+          closeDays: 60
           maximumVotes: 1
-          closeComment: "This feature request is being closed due to insufficient upvotes.  When at least 2 upvotes are received, this issue will be elidgable for our backlog."
+          closeComment: "This feature request is being closed due to insufficient upvotes.  When at least 2 upvotes are received, this issue will be elidgible for our backlog."
```
sean-mcmanus: The spelling is still incorrect. I think it's supposed to be `eligible`

sean-mcmanus:
@WardenGnaw Which of the actions should use the debugger label filter?

WardenGnaw:
None for now. I can modify the YAML later to add support for the debugger ones.

.github/workflows/more-info-needed-closer.yml
bobbrow: But we'd be pinging after the issue is closed?  80 > 60

.github/actions/.gitignore
bobbrow: You're right. I've been working in 2 other repos where .github/ and src/ are both in the root.  I forgot about how this repo doesn't do that.

.github/workflows/more-info-needed-closer.yml
Colengms: No, the issue would not be closed if the last comment was made by a user, and not us (as we may not have seen their comment, and the issue may be on our plate, not theirs).  Instead, after pingDays, the assignee is pinged.  After an issue is closed, no-one will be pinged.  The StaleCloser action only queries for issues that are: "is:open is:unlocked"...  Though, the code as written does appear to assume pingDays will be greater than closeDays, as the initial query is based on the closeDays value.

.github/actions/common/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+	;(['core', 'graphql', 'search'] as const).forEach(async (category) => {
+		usage[category] = 1 - usageData[category].remaining / usageData[category].limit
+	})
+	return usage
+}
+
+export const errorLoggingIssue = (() => {
+	try {
+		const repo = context.repo.owner.toLowerCase() + '/' + context.repo.repo.toLowerCase()
+		if (repo === 'microsoft/vscode' || repo === 'microsoft/vscode-remote-release') {
```
JacksonKearl: this can be modified to log to your own repo

.github/workflows/more-info-needed-closer.yml
bobbrow: Got it. So "closeDays" is for when it's assigned to us and "pingDays" is for when it's assigned to them.

Colengms: If you are using the term 'assigned' loosely, then yes.  If you are using it to refer to assigning the issue using the 'assignees' field, then no.  closeDays is for when we were the last ones to comment on the issue.  Otherwise, if a user was the last to comment, it doesn't get closed, and IF there is an assignee they are pinged.

PR #6255, opened 2020-10-05
Work around issue with linux tests

Colengms:
(Not yet ready for review.  Creating PR to see if the fix works)


sean-mcmanus:
Did this work? I assume not? Is there a bug tracking the GitHub test issue?

Colengms:
> Did this work? I assume not? Is there a bug tracking the GitHub test issue?

I don't know if this would work.  This is blocked due to tests not launching at all.  I don't have any insights on that.  I assume it's a GitHub issue.  Who would we report that to?

michelleangela (approved):

Issue #5759, opened 2020-07-10
telemetry updates for openfile


bobbrow:
Add warm start/cold start telemetry for openfile

PR #6257, opened 2020-10-06
add telemetry updates for openfile

elahehrashedi:
telemetry request: https://github.com/microsoft/vscode-cpptools/issues/5759

These are the timestamps:
 
- `firstFile`: when the extension is activated by `realActivation` and the active text editor is a c/cpp file. Defined only for "cold" start cases.
- `didOpen`: when the file appears in the editor (all files including the first file). Defined for "warm" start cases.
- `setup`: when the Intellisense_client constructor is completed and we give the control back to the user at the end of `textDocument_didOpen` in language server side
- `updateRange`: when `publishDiagnostics `is completed

We defined an extra `startTime` timestamp: when the extension activates, if the active text document is a c/cpp file, the startTime for this file is set to `firstFile` timestamp, otherwise it is set to `didOpen` timestamp.

These are the durations that are being calculated based on timestamps and logged:

- `activationTime = firstFile - didOpen`  logged only for the first file (cold start)
- `setupTime = setup - didOpen`
- `updateRangeTime = updateRange - setup`
- `totalTime = updateRange - startTime`

there is a related pull request on the native side.


Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -215,3 +220,69 @@ export class ClientCollection {
...
 }
+
+interface TimeStampSequence {
+    activationTime: number; // when the file appears in the editor. Defined for both "cold/warm" start cases.
+    setupTime: number; // when the Intellisense_client constructor is completed
+    updateRangeTime: number; // when publishDiagnostics & provideSemanticTokens is completed
+    totalTime: number;
+}
+
+export class TimeTelemetryCollector {
```
bobbrow: Please move the collector class and dependent interfaces to a new file.  This doesn't belong in clientCollection.ts.

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -215,3 +220,69 @@ export class ClientCollection {
...
+    constructor() {
+        this.extensionStartTime = new Date().getTime();
+        this.firstFile = 0;
+    }
+
+    public setFirstFile() {
+        if (!this.firstFile){
+            this.firstFile = new Date().getTime();
+        }
+        Telemetry.logLanguageServerEvent("firstFile", { "firstFile": (this.firstFile - this.extensionStartTime).toString() });
```
bobbrow: I don't think there should be a separate event for "firstFile".  Instead, this should be a bool on the openFile event.

Extension/src/LanguageServer/client.ts
```diff
@@ -2149,6 +2160,11 @@ export class DefaultClient implements Client {
         }
     }
 
+    public LogIntellisenseSetup(notification: SetUpTimeStamp): void {
+        clientCollection.timeTelemetryCollector.setSetupTime(vscode.Uri.parse(notification.uri).fsPath);
+        return;
```
bobbrow: `return` is not needed here.

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -215,3 +220,69 @@ export class ClientCollection {
...
+    public setUpdateRangeTime(uri: string) {
+        let curTimeStamps: TimeStampSequence = this.getTimeStamp(uri);
+        if (!curTimeStamps.updateRangeTime) {
+            curTimeStamps.updateRangeTime = new Date().getTime();
+            this.cachedTimeStamps.set(uri, curTimeStamps);
+        }
+        if (!curTimeStamps.totalTime && curTimeStamps.activationTime && curTimeStamps.setupTime){
+            curTimeStamps.totalTime = curTimeStamps.updateRangeTime - curTimeStamps.activationTime;
+            Telemetry.logLanguageServerEvent("timeStamps", {
+                "activationTime": (curTimeStamps.activationTime).toString(),
```
bobbrow: These should not be string "Properties".  We want them to be numeric "Measures".  And I think we should report the deltas in milliseconds instead of the raw timestamps (e.g. setupTime: (setupTime - activationTime), updateRangeTime: (updateRangeTime - setupTime), totalTime).

Extension/src/LanguageServer/extension.ts
```diff
@@ -331,6 +331,9 @@ function realActivation(): void {
     clients.ActiveClient.notifyWhenReady(() => {
         intervalTimer = global.setInterval(onInterval, 2500);
     });
+
+    // Log cold start.
+    clients.timeTelemetryCollector.setFirstFile();
```
bobbrow: We only mark an openfile event as "cold start" when there is a C/C++ file open in the editor at activation time.  I didn't see any code that checked this.  I think you could check this in the timestamp collector constructor and save the state then.  Doing it here is not measuring the right thing.

Extension/src/LanguageServer/clientCollection.ts
bobbrow: The one field that would be in the "Properties" would be the "coldStart" boolean.  But we could also say that that property is unnecessary and we just log an "activationTime" Measure which is (activationTime - extensionStartTime).  The existence of that measure could mean that this is a "cold start" event.

Extension/src/LanguageServer/client.ts
```diff
@@ -1392,6 +1400,8 @@ export class DefaultClient implements Client {
         if (document.uri.scheme === "file") {
             openFileVersions.set(document.uri.toString(), document.version);
         }
+        // Log warm start.
+        clientCollection.timeTelemetryCollector.setActivationTime(document.uri.fsPath);
```
bobbrow: I'm going to diverge from the table we created and suggest that we call this "didOpen" time.  Activation means something else and only happens once, so I think it is confusing to refer to this as "activation" time.

bobbrow: `.fsPath` is not the URI.  We should use URIs everywhere to avoid messing with the capitalization.  I believe it should be just `.path` here.

Extension/src/LanguageServer/client.ts
```diff
@@ -156,6 +156,8 @@ function publishDiagnostics(params: PublishDiagnosticsParams): void {
 
     const realUri: vscode.Uri = vscode.Uri.parse(params.uri);
     diagnosticsCollection.set(realUri, diagnostics);
+
+    clientCollection.timeTelemetryCollector.setUpdateRangeTime(realUri.fsPath);
```
bobbrow: `.path` instead of `.fsPath`

Extension/src/LanguageServer/client.ts
```diff
@@ -2149,6 +2160,11 @@ export class DefaultClient implements Client {
         }
     }
 
+    public LogIntellisenseSetup(notification: SetUpTimeStamp): void {
+        clientCollection.timeTelemetryCollector.setSetupTime(vscode.Uri.parse(notification.uri).fsPath);
```
bobbrow: `.path` instead of `.fsPath`

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -32,6 +34,7 @@ export class ClientCollection {
         return result;
     }
     public get Count(): number { return this.languageClients.size; }
+    public get ActiveDocument(): vscode.TextDocument | undefined { return this.activeDocument; }
```
bobbrow: I didn't see where this is used.  Is it needed?

Extension/src/LanguageServer/client.ts
```diff
@@ -419,6 +421,10 @@ enum SemanticTokenModifiers {
     local = (1 << 2)
 }
 
+interface IntellisenseSetup {
```
bobbrow: nit: IntelliSenseSetup (capital 'S' in the middle)

Extension/src/LanguageServer/client.ts
```diff
@@ -477,6 +483,7 @@ const ShowMessageWindowNotification: NotificationType<ShowMessageWindowParams, v
 const ShowWarningNotification: NotificationType<ShowWarningParams, void> = new NotificationType<ShowWarningParams, void>('cpptools/showWarning');
 const ReportTextDocumentLanguage: NotificationType<string, void> = new NotificationType<string, void>('cpptools/reportTextDocumentLanguage');
 const SemanticTokensChanged: NotificationType<string, void> = new NotificationType<string, void>('cpptools/semanticTokensChanged');
+const IntellisenseSetupNotification:  NotificationType<IntellisenseSetup, void> = new NotificationType<IntellisenseSetup, void>('cpptools/intellisenseSetup');
```
bobbrow: nit: IntelliSenseSetupNotification (capital 'S' in the middle)

Extension/src/LanguageServer/client.ts
```diff
@@ -2149,6 +2159,10 @@ export class DefaultClient implements Client {
         }
     }
 
+    public LogIntellisenseSetupTime(notification: IntellisenseSetup): void {
```
bobbrow: style: logIntelliSenseSetupTime (functions start with lower case letters & capital 'S' in IntelliSense)

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    public setDidOpenTime(uri: vscode.Uri): void {
+        const curTimeStamps: TimeStampSequence = this.getTimeStamp(uri.path);
+        curTimeStamps.didOpen = new Date().getTime();
+        this.cachedTimeStamps.set(uri.path, curTimeStamps);
+    }
+
+    public setSetupTime(uri: string): void {
```
bobbrow: I think `setSetupTime` should accept `vscode.Uri` as a parameter instead of `string` to match all the other functions.

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,79 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import * as telemetry from '../telemetry';
+import * as util from '../common';
+import * as vscode from 'vscode';
+
+interface TimeStampSequence {
+    firstFile: number | undefined; // when the extension is activated by realActivation. Defined only for "cold" start cases.
```
bobbrow: Use `firstFile?: number` instead of `firstFile: number | undefined`

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    private removeTimeStamp(uri: string): void {
+        this.cachedTimeStamps.delete(uri);
+    }
+
+    private logTelemetry(uri: string, timeStamps: TimeStampSequence): void {
+        const startTime: number = timeStamps.firstFile ? timeStamps.firstFile : timeStamps.didOpen;
+        telemetry.logLanguageServerEvent("timeStamps",
+            timeStamps.firstFile ? { "coldstart": "true" } : {}, {
+                "activationTime": (timeStamps.didOpen - startTime),
```
bobbrow: activationTime should only be logged for the cold start case

bobbrow (changes requested):

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import * as telemetry from '../telemetry';
+import * as util from '../common';
+import * as vscode from 'vscode';
+
+interface TimeStampSequence {
+    firstFile?: number; // when the extension is activated by realActivation. Defined only for "cold" start cases.
+    didOpen: number; // when the file appears in the editor. Defined for both "warm" start cases.
```
michelleangela: > when the file appears in the editor. Defined for both "warm" start cases.

What are the two warm start cases? Or does this comment mean to say "for both warm and cold start cases"?

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    private removeTimeStamp(uri: string): void {
+        this.cachedTimeStamps.delete(uri);
+    }
+
+    private logTelemetry(uri: string, timeStamps: TimeStampSequence): void {
+        const startTime: number = timeStamps.firstFile ? timeStamps.firstFile : timeStamps.didOpen;
+        let properties: any = {};
+        let metrics: any = {
+            "setupTime": (timeStamps.setup - startTime),
```
bobbrow, line 72: I think this should be (setup - didOpen)

Extension/src/LanguageServer/timeTelemetryCollector.ts
```diff
@@ -0,0 +1,84 @@
...
+            "updateRangeTime": (timeStamps.updateRange - timeStamps.setup),
+            "totalTime": (timeStamps.updateRange - startTime)
```
bobbrow, line 74: I think this should be (updateRange - (firstFile ? startTime : didOpen))

bobbrow (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1161,6 +1169,8 @@ function handleCrashFileRead(err: NodeJS.ErrnoException | undefined | null, data
 }
 
 export function deactivate(): Thenable<void> {
+    // Log timestamps telemetry.
```
michelleangela: >// Log timestamps telemetry. 
Is this comment for the line `clients.timeTelemetryCollector.clear();`? It doesn't match what the function `clear()` is doing.

michelleangela (changes requested):

Extension/src/LanguageServer/timeTelemetryCollector.ts
elahehrashedi: I defined `const startTime: number = timeStamps.firstFile ? timeStamps.firstFile : timeStamps.didOpen;`
because the starting point is different for first file and other files.

Extension/src/LanguageServer/timeTelemetryCollector.ts
elahehrashedi: I defined `const startTime: number = timeStamps.firstFile ? timeStamps.firstFile : timeStamps.didOpen;`
because the starting point is different for first file and other files.

Extension/src/LanguageServer/timeTelemetryCollector.ts
elahehrashedi: "both" needs to be deleted

Extension/src/LanguageServer/extension.ts
elahehrashedi: I changed the functionality but forgot to change the comment part.

michelleangela (approved):

Extension/src/LanguageServer/timeTelemetryCollector.ts
bobbrow: I missed that.  Thanks.

Extension/src/LanguageServer/timeTelemetryCollector.ts
bobbrow: I think that is correct for total time, but not for this one.  This one is supposed to measure from didOpen to setup.  For firstFile, you add activationTime so you don't want to double count it here.

bobbrow:
I will clear my review since I'm about to leave for the day. If you fix the last issue about the "setupTime" property, I approve.

bobbrow:
@elahehrashedi you didn't fix the last bug. I cleared my review so that you could check in after fixing that. Please fix the value of setup time for the warm start case.

PR #6304, opened 2020-10-13
Fix 'Build and Debug Active File' race condition with EngineLogs

WardenGnaw:
If users use 'Build and Debug Active File' with "engineLogging", the resolveDebugConfigurations will quickly switch to the output window and `${fileBasenameNoExtension}` will resolve to something similar to `./extension-output-#5`.

This may be a bug with VS Code determining the output window to be a file, but this will unblock users who need to enable engine logging for this scenario.

Example broken task.json
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "g++ - Build and debug active file",
            "type": "cppdbg",
            "request": "launch",
            "program": "${fileDirname}/${fileBasenameNoExtension}",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${workspaceFolder}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ],
            "preLaunchTask": "C/C++: g++ build active file",
            "miDebuggerPath": "/usr/bin/gdb",
            "logging": {
                "engineLogging": true
            }
        }
    ]
}
```

Will result in:
```
g++: error: extension-output-#5: No such file or directory
g++: fatal error: no input files
compilation terminated.
The terminal process "/bin/bash '-c', '/usr/bin/g++ -g extension-output-#5 -o ./extension-output-#5'" failed to launch (exit code: 1).
```

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -274,6 +274,8 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
             const outputChannel: logger.Logger = logger.getOutputChannelLogger();
             outputChannel.appendLine(localize("debugger.launchConfig", "Launch configuration:"));
             outputChannel.appendLine(JSON.stringify(config, undefined, 2));
+	    // TODO: Enable when https://github.com/microsoft/vscode/issues/108619 is resolved.
+	    // logger.showOutputChannel();
         }
 
         return config;
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -274,7 +274,6 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
             const outputChannel: logger.Logger = logger.getOutputChannelLogger();
             outputChannel.appendLine(localize("debugger.launchConfig", "Launch configuration:"));
             outputChannel.appendLine(JSON.stringify(config, undefined, 2));
-            logger.showOutputChannel();
         }
 
         return config;


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -274,7 +274,6 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
             const outputChannel: logger.Logger = logger.getOutputChannelLogger();
             outputChannel.appendLine(localize("debugger.launchConfig", "Launch configuration:"));
             outputChannel.appendLine(JSON.stringify(config, undefined, 2));
-            logger.showOutputChannel();
```
pieandcakes, line 277: So if you remove that, the user won't see the log and We want to show the logging window.  I think this is a hack so you should mark it with a bug number and comment it out and file a bug with VS Code to fix the problem.

WardenGnaw:
VS Code Issue: https://github.com/microsoft/vscode/issues/108619

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: There a weird fight between the terminal window and output window in the "Build and Debug Active File" scenario. It flashes the output window then goes to terminal for compiling.

sean-mcmanus (approved):

PR #6305, opened 2020-10-13
Add action to reopen feature requests that get enough votes

Colengms:
* Add an action/workflow to re-open closed Feature Requests that receive enough votes
* Change vote threshold to 5 votes (open for debate.  We should stick with what we choose, as it adds a comment that refers to the number of votes, which is not easy to change later).
* Added support for removing labels and setting milestone, to StaleCloser action
* Update Feature Request closer to move thing to Triage milestone when closing.


michelleangela (approved):

.github/workflows/feature-request-closer-triage.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           ignoredLabels: debugger
           addLabels: more votes needed
           closeDays: 60
-          maximumVotes: 1
+          maximumVotes: 4
           closeComment: "This feature request is being closed due to insufficient upvotes.  When at least 2 upvotes are received, this issue will be eligible for our backlog."
```
bobbrow: Update comment

PR #6306, opened 2020-10-13
Regenerate action JS files

Colengms:
The actions I copied from VS Code's action repo don't compiler TS on the fly.  Instead, JS files are checked in.  Those JS files were not regenerated after a couple of previous fixes.


WardenGnaw:
Future TODO: I would create a simple pipeline to trigger if your `ts` files has changed, force it to compile it and check for `git diff --exit-code`

How to trigger on specific paths; https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#paths

WardenGnaw (approved):

PR #6319, opened 2020-10-15
[cppdbg] Support new sourceFileMap schema

WardenGnaw:
For cppdbg, sourceFileMap can be used for binding breakpoints.
If keeping with the old format `"<source-path>":"<target-path>"`, it will
always be used for breakpoints.

If you use the new:
```
"<compile-path>" : {
   "editorPath": "<target-path>",
   "useForBreakpoints": true
}
```

The sourceMap will only be used for frame enumeration.

Descriptions are from:
https://github.com/microsoft/MIEngine/blob/cc361a883dbb7e61de78054f838f2ea9d9f5e065/src/MICore/LaunchOptions.xsd#L603-L626


Extension/tools/GenerateOptionsSchema.ts
```diff
@@ -80,6 +80,17 @@ function replaceReferences(definitions: any, objects: any): any {
             objects[key] = refReplace(definitions, objects[key]);
         }
 
+        // Handle 'anyOf' with references
+        if (objects[key].hasOwnProperty('anyOf')) {
+            for (const index in objects[key].anyOf)
+            {
+                if (objects[key].anyOf[index].hasOwnProperty('$ref'))
+                {
```
sean-mcmanus: The brace styling doesn't match...is TSLint not running on this file?

Extension/package.nls.json
```diff
@@ -226,5 +226,8 @@
...
     "c_cpp.taskDefinitions.options.description": "Additional command options",
     "c_cpp.taskDefinitions.options.cwd.description": "The current working directory of the executed program or script. If omitted Code's current workspace root is used.",
-    "c_cpp.taskDefinitions.detail.description": "Additional details of the task"
+    "c_cpp.taskDefinitions.detail.description": "Additional details of the task",
+    "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.description": "Current and compile-time paths to the same source trees. Files found under the EditorPath are mapped to the CompileTimePath path for breakpoint matching and mapped from CompileTimePath to EditorPath when displaying stacktrace locations.",
+    "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.editorPath.description": "The path to the source tree the editor will use.",
+    "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.useForBreakpoints.description": "False if entry is only used for stack frame location mapping. True if this entry should also be used when specifying breakpoint locations."
```
sean-mcmanus: Seems like the first "entry" should be "this entry" (like is used in the 2nd sentence)...or "the entry".

sean-mcmanus (changes requested):

Extension/tools/GenerateOptionsSchema.ts
WardenGnaw: We skip the `tools` directory for tslint.

sean-mcmanus: Was there a reason for that?

WardenGnaw: Some configuration from `tsconfig` or another config file skips it, or else we would compile and webpack the files in tools into `main.js`.

sean-mcmanus (approved):

sean-mcmanus:
@WardenGnaw Is this supposed to be checked in for 1.1.0-insiders3? It's seemed to be mentioned in the changelog.

WardenGnaw:
Yep. This will need to be checked in or the new schema type will not launch the debugger since it currently assumes sourceFileMap's value is always a string

PR #6337, opened 2020-10-17
Fixes for Windows ARM64

sean-mcmanus:
Still testing this...


Extension/README.md
```diff
@@ -56,6 +56,7 @@ The extension has platform-specific binary dependencies, therefore installation
 `cpptools-linux-aarch64.vsix` | Linux ARM 64-bit
 `cpptools-osx.vsix` | macOS
 `cpptools-win32.vsix` | Windows 64-bit & 32-bit
+`cpptools-win-arm64.vsix` | Windows ARM64
```
michelleangela, line 59: Also update the other README.md in the root folder

sean-mcmanus: Good catch.

Colengms (approved):

sean-mcmanus:
The TypeScript changes seem okay so far, but the ARM64 binary isn't running and I haven't tested on 32-bit Windows yet.

sean-mcmanus:
I'm checking this in, but tag parsing isn't working yet...

Extension/src/githubAPI.ts
```diff
@@ -114,7 +114,15 @@ function getArrayOfBuilds(input: any): Build[] {
 export function vsixNameForPlatform(info: PlatformInformation): string {
     const vsixName: string | undefined = function(platformInfo): string | undefined {
         switch (platformInfo.platform) {
-            case 'win32': return 'cpptools-win32.vsix';
+            case 'win32':
+                switch (platformInfo.architecture) {
+                    case 'x86_64': return 'cpptool-win32.vsix'; // TODO: Change to cpptools-win64?
```
bobbrow, line 119: It looks like this is misspelled

bobbrow: This needs to be fixed or else insiders will be broken for x64.

sean-mcmanus: yikes, good catch.

Issue #6366, opened 2020-10-21
[BUG]: Compiler ARGS not taken over/considered when running build with g++ on Win10


hasenradball:
- VSCode Version: 150.1
- OS Version: Win 10.0.18363
- c/c++: v1.1.0-insiders2

### General Information
When compiling a simple c++ programm like "hello world" or just want to compile 2 or 3 files and link them together by adapting the `tasks.json` file, every change in the file works until the **`args`** part.
I you change the **`args`**like change the output filename or using `*.cpp` instead of `${file}`.
The compiler only uses the active file and nothing else!

### Steps to Reproduce:
1. create the "hello world" example from Documentation
2. change the compiler "args" in tasks.json like an other output name, or change the args to be able to compile multiple files
3. run Task
4. the changes in "args" are not considered!

### Hint:
On MAC and Linux there is no problem, it works fine.
On Win10 it does not work!

{ "version": "2.0.0", "tasks": [ { "type": "cppbuild", "label": "C/C++: g++.exe build active file", "command": "C:\\Program Files (x86)\\mingw-w64\\i686-8.1.0-posix-dwarf-rt_v6-rev0\\mingw32\\bin\\g++.exe", "args": ["-g", "${workspaceFolder}\\*.cpp", "-o", "${fileDirname}\\${fileBasenameNoExtension}.exe"], "options": { "cwd": "${workspaceFolder}" }, "problemMatcher": [ "$gcc" ], "group": { "kind": "build", "isDefault": true }, "detail": "compiler: \"C:\\Program Files (x86)\\mingw-w64\\i686-8.1.0-posix-dwarf-rt_v6-rev0\\mingw32\\bin\\g++.exe\"" } ] }

Does this issue occur when all extensions are disabled?:
actually not possible to tes

sean-mcmanus:
I repro the issue. Looks like a regression from 1.1.0-insiders/2. Downgrading to 1.0.1 and changing the task type to "shell" should be a workaround.

sean-mcmanus:
@elahehrashedi The problem is that F5 (Build and Debug Active File) is causing the wrong build task to be run, i.e. the default build task is not run (it's using the pre-generated task). I see it working correctly when Run Build Task is used.

sean-mcmanus:
@hasenradball A workaround is to change the label in tasks.json and the preLaunchTasks to match that new name in launch.json.

hasenradball:
Ok thanks for the reply.
[...]

PR #6374, opened 2020-10-22
don't overwrite options and args

elahehrashedi:
don't overwrite options and args when building from a custom defined task
bugfix: https://github.com/microsoft/vscode-cpptools/issues/6366


sean-mcmanus (approved):

sean-mcmanus:
This fixes the args being used, but I have to change the label/name still. Are you able to repro the label/name issue? Should I create a new issue for that?

Also I'm seeing debugging not working with mingw on Windows, but that seems to repro without this change.

sean-mcmanus:
If you want to get this into 1.1.0-insiders3 can you update the changelog.md?

elahehrashedi:
@sean-mcmanus I changes the changelog. the label name issue is in another reported issue that I have to address.

PR #6377, opened 2020-10-23
show details of the compiler path

### LINKED ISSUE #4278: "Run and Debug" and "F5" doesn't show the compilers detail in the API

elahehrashedi:
Edit: detail is removed from Launch.json.
the paths are visible as before. 

the paths are visible now:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/4278

image

image

image

Same result with pressing F5


michelleangela (approved):

Extension/package.nls.json
```diff
@@ -230,5 +230,6 @@
     "c_cpp.taskDefinitions.detail.description": "Additional details of the task",
     "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.description": "Current and compile-time paths to the same source trees. Files found under the EditorPath are mapped to the CompileTimePath path for breakpoint matching and mapped from CompileTimePath to EditorPath when displaying stacktrace locations.",
     "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.editorPath.description": "The path to the source tree the editor will use.",
-    "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.useForBreakpoints.description": "False if this entry is only used for stack frame location mapping. True if this entry should also be used when specifying breakpoint locations."
+    "c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.useForBreakpoints.description": "False if this entry is only used for stack frame location mapping. True if this entry should also be used when specifying breakpoint locations.",
+    "c_cpp.debuggers.configuration.detail.description": "Additional details of the preLaunch task"
```
sean-mcmanus: Should this be preLaunchTask?

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -2003,6 +2007,10 @@
                 "type": "boolean",
                 "description": "%c_cpp.debuggers.requireExactSource.description%",
                 "default": true
+              },
+              "detail": {
```
WardenGnaw: Why are we adding `detail` to the launch configuration? I thought this was just a change so the UI has additional information. 

Any specific detail  to tasks should be in tasks.json, not launch.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -144,6 +144,7 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
             newConfig.externalConsole = false;
             const exeName: string = path.join("${fileDirname}", "${fileBasenameNoExtension}");
             newConfig.program = platform === "win32" ? exeName + ".exe" : exeName;
+            newConfig.detail = definition.command;
```
WardenGnaw, line 156: This just seems to duplicate a task's `command`. What if the user decides to change the `command` in their `tasks.json`, this `detail` will now be out of date and the user has to update this comment themselves. 

I would only just add this detail to the MenuItem and not actually return it to the user's `launch.json` configuration.

WardenGnaw (changes requested):

Extension/package.json
sean-mcmanus: This is for the F5 (Build and Debug Active File) scenario in which the launch.json needs the details (see the screenshots). tasks.json already behaves correctly. Is there a problem with this?

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: @elahehrashedi Does just adding it to the MenuItem work?

Extension/package.json
WardenGnaw: Yes. 

This will add `details` into `launch.json` and it will be outdated if the user changes the command in `tasks.json`. If the user wants to know the exact command when it has ran, the user can see what is executed in the terminal window or just go to the `tasks.json` and look at the `command`.
Here is an image of it being populated in the `launch.json`.
image

Adding descriptions for it in the debugger schema in `package.json` will create an autocompletion for `detail`. However, this is not used for the debug adapter at all and will be sent as part of the launch configuration.
image

I highly prefer it to not be written to the `launch.json` or show up as an item you can add in `launch.json`. It should only be used to for the MenuItem since I doubt it will be needed once it has been generated.

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: I removed the "detail" item after showing the menu Items. line 62

Extension/package.json
WardenGnaw: You need to undo all changes in `package.json`. This will still create an autocomplete prompt for `cppvsdbg`.

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: Please add a comment saying this is not used in `launch.json` and it is just used to add a description for the MenuItem in `QuickPickConfigurationProvider.provideDebugConfigurations` and where `noDetailConfig.detail = undefined;` is set, add a comment saying this deletes the key so it does not get written to the `launch.json`

WardenGnaw (changes requested):

WardenGnaw (approved):

PR #6384, opened 2020-10-23
fix terminal issues (windows/linux/linux-on-wsl verified)

elahehrashedi:
updated on 10/27:
wsl:
image

windows (g++):
image

windows (cl.exe):
image


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -325,23 +326,26 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
             this.options.cwd = util.resolveVariables(this.options.cwd, this.AdditionalEnvironment);
         }
 
+        const splitWriteEmitter = (lines: string | Buffer) => {
+            for (const line of lines.toString().replace("\r\n", "\n").split("\n")) {
```
bobbrow: I think you can split on a regex so that you don't have to substitute before splitting the lines.  Something like this (not tested): `lines.toString().split(/\r?\n/g)`

sean-mcmanus (approved):

PR #6396, opened 2020-10-26
Use x86 vsdbg for Windows ARM64.

sean-mcmanus:


WardenGnaw:
@gregg-miskelly Do you see any issues running the C++ VsDbg on all Windows platforms?

gregg-miskelly:
The current x86 VsDbg will only work for debugging x86 processes on Windows ARM64, so at the very least we would want to include the ARM64 remote debugger (and maybe ARM also?) and fix the code to code to launch the correct one.

This will also mean doing most of the "heavy lifting" in a relatively short-lived x86 process (it will exit when debugging stops). So I don't know if performance will be fast enough to be acceptable.

sean-mcmanus:
Yeah, I think for 1.1.0 we'll only have support for debugging x86 executables. It sounds like support for making the ARM64 remote debugger easier to get setup would need to be tracked a new issue.

I'm still testing x86 vsdbg on Windows ARM64, so I might find some other issues...

gregg-miskelly:
Make sure at least the error message is reasonable when attaching/launching ARM/ARM64 processes, and that we have reasonable telemetry on how often it is happening...

sean-mcmanus:
@WardenGnaw Vsdbg isn't working on Windows ARM64 with an x86 binary. It just shows `Unknown error: 0x80040028 at t.RawDebugSession.handleErrorResponse.

gregg-miskelly:
@WardenGnaw In case this saves you some time, that error (E_AUTO_LAUNCH_EXEC_FAILURE) likely means we are trying to launch msvsmon.exe because we think we should be doing pseudo-remote debugging. Though obviously we shouldn't be.

PR #6401, opened 2020-10-27
Don't register cppvsdbg for Windows ARM.

sean-mcmanus:
This causes a no-op when the C++ (Windows) debug option is selected. The behavior matches what occurs on Linux/Mac, although ideally it seems like the "C++ (Windows)" option should be disabled in the UI or a message would be shown.


WardenGnaw:
Does gdb scenario work on Windows ARM64?

sean-mcmanus:
> Does gdb scenario work on Windows ARM64?

Yeah, mingw/gdb debugging works fine. Did you expect some issue in that scenario?

WardenGnaw:
It might be using mono x86 so it may be slow. It would be nice to have a native arm64 runtime.

But if that scenario works, that is good to hear.

sean-mcmanus:
> It might be using mono x86 so it may be slow. It would be nice to have a native arm64 runtime.
> 
> But if that scenario works, that is good to hear.

Oh, well, debugging of a simple program works (e.g. math operations from cmath), but when I use a std::string or std::vector it fails. The logging shows "During startup program exited with code 0xc0000135". Any ideas how to fix that?

WardenGnaw:
`0xc0000135` is STATUS_DLL_NOT_FOUND.  <del>They may be an issue with mono framework x86, it is trying to load a dll and failing. </del> Edit: We dont use mono for windows, we have .net framework. 
You might be able to see which dll it is loading with `Fuslogvw.exe`

sean-mcmanus:
> `0xc0000135` is STATUS_DLL_NOT_FOUND. They may be an issue with mono framework x86, it is trying to load a dll and failing.
> You might be able to see which dll it is loading with `Fuslogvw.exe`

It says
```
Bind result: hr = 0x80004005. Unspecified error.
...
Rejecting native image because native image dependency C:\WINDOWS\Microsoft.Net\assembly\GAC_32\mscorlib\...\mscorlib.dll had a different identity than expected.
```

WardenGnaw:
No idea, I would have to investigate why OpenDebugAD7.exe does not like the mscorlib shipped in Windows ARM64.

sean-mcmanus:
> No idea, I would have to investigate why OpenDebugAD7.exe does not like the mscorlib shipped in Windows ARM64.

So is this change good enough for 1.1.0? Or do you have another recommendation?

bobbrow:
We can merge this as-is.  I think our message for 1.1 is going to be "debugging is unsupported on Windows ARM64"

bobbrow (approved):

WardenGnaw (approved):

PR #6415, opened 2020-10-30
Command to generate .editorconfig file

Colengms:
If a .editorconfig file exists, it's updated using WorkspaceEdits, which allows the user to potentially undo or edit changes before saving.

If a .editorconfig file does not exist, one will be created at the correct location and opened for edit with new contents.  Because the file creation itself is done using a WorkspaceEdit, the user can undo creation of the file itself, if they choose to.

If existing settings are found, they are replaced with the new values.  Any settings that were not found, are added to a new section at the end of the file.  Ideally, this should either update the existing C/C++ section(s) to use the new values, and/or create a new section.  If a user has a more complex setup, such as different VC Format settings for different file extensions, that we not be discernable, they should probably not be 'generating' a new .editorconfig file.

An effort is made not to output a section ("[*]") unnecessarily, if already in that section at the end of the file.  Tries to be smart about line spacing before adding to the end of the file.

Updated to new version of @types/vscode, in order to pick up the definition of Uri.joinPath().


Extension/src/LanguageServer/settings.ts
```diff
@@ -458,3 +488,210 @@ export class OtherSettings {
...
+        if (value === "oneLiners") {
+            return "one_liners";
+        }
+    }
+    return "never";
+}
+
+function populateEditorConfig(rootUri: vscode.Uri | undefined, document: vscode.TextDocument): void {
+    // Set up a map of setting names and values. Parse through the document line-by-line, looking for
+    // existing occurances to replace. Replaced occurances are removed from the map. If any remain when
```
sean-mcmanus: occurrences

sean-mcmanus: occurrences

Extension/src/LanguageServer/settings.ts
```diff
@@ -458,3 +488,210 @@ export class OtherSettings {
...
+        // If a folder is open and '.editorconfig' exists at the root, use that.
+        const uri: vscode.Uri = vscode.Uri.joinPath(rootUri, ".editorconfig");
+        const edits: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
+        edits.createFile(uri, { ignoreIfExists: true, overwrite: false });
+        vscode.workspace.applyEdit(edits).then(() => {
+            vscode.workspace.openTextDocument(uri).then(
+                (document) => populateEditorConfig(rootUri, document),
+                () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
+        },
+        () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
```
sean-mcmanus: The formatter seems to want this on the previous line? i.e. 693 added to 692.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/settings.ts
```diff
@@ -458,3 +488,209 @@ export class OtherSettings {
...
+        if (value === "oneLiners") {
+            return "one_liners";
+        }
+    }
+    return "never";
+}
+
+function populateEditorConfig(rootUri: vscode.Uri | undefined, document: vscode.TextDocument): void {
+    // Set up a map of setting names and values. Parse through the document line-by-line, looking for
+    // existing occurences to replace. Replaced occurences are removed from the map. If any remain when
```
sean-mcmanus: occurences->occurrences

sean-mcmanus: occurences->occurrences

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #6419, opened 2020-10-30
Add telemetry for machineId changes and remoteName

Colengms:
Bob requested logging of the machineId and if it changes, as well as the remoteName, per session.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -812,6 +812,20 @@ export class DefaultClient implements Client {
                     this.semanticTokensLegend = new vscode.SemanticTokensLegend(tokenTypesLegend, tokenModifiersLegend);
 
                     if (firstClient) {
+                        // Don't log telemetry for machineId if it's a special value used by the dev host: someValue.machineid
+                        if (vscode.env.machineId !== "someValue.machineId") {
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -812,6 +812,20 @@ export class DefaultClient implements Client {
                     this.semanticTokensLegend = new vscode.SemanticTokensLegend(tokenTypesLegend, tokenModifiersLegend);
 
                     if (firstClient) {
+                        // Don't log telemetry for machineId if it's a special value used by the dev host: someValue.machineid
+                        if (vscode.env.machineId !== "someValue.machineId") {
```
bobbrow: I was thinking we'd do this during activation instead of during client creation.  Were you thinking it would be better to ignore people who aren't using C++ anymore?  I would prefer to have a single "activate" event with Properties:
* remoteName
* oldMachineId - optional (machineIdPersistentState.Value)
* newMachineId - optional (vscode.env.machineId)

The two last properties are only logged when the value changes.  Could you check and see whether this triggers when using Remote-WSL or one of the other Remote extensions?

bobbrow (changes requested):

bobbrow (approved):
Did you check whether this fires an id change on Remote-WSL?  (you might have to build/install the VSIX to do it though)

PR #6420, opened 2020-10-31
"F5" using the correct build task

### LINKED ISSUE #6375: "C/C++ Build and Debug Active File" and "F5" are causing the wrong build task to be run, i.e. the default build task is not run (it's using the pre-generated task)
@elahehrashedi The problem is that F5 (Build and Debug Active File) is causing the wrong build task to be run, i.e. the default build task is not run (it's using the pre-generated task). I see it working correctly when Run Build Task is used.

_Originally posted by @sean-mcmanus in https://github.com/microsoft/vscode-cpptools/issues/6366#issuecomment-714684171_

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/6375
renaming is as below:
`"label": "C/C++: cl.exe build active file",`
`"label": "C/C++: cl.exe build active file ver(1)",`
`"label": "C/C++: cl.exe build active file ver(2)",`


sean-mcmanus:
Is it possible to change the name of the default task by adding a space at the end and then removing the space when the duplicate is detected?

If this is only an issue the 1st time the launch.json is written, then having a number that increments to append to the name doesn't seem necessary.

sean-mcmanus:
There's some issue with F5 debugging not reacting to changes to the tasks.json -- the issue doesn't repro with 1.0.1 so it seems important to fix for 1.1.0, although it repros with 1.1.0-insiders3 too. I'm not sure yet if it's a VS Code bug or what area of our code could be changed to fix that.

Jianchaoliang:
unsubscribe


| |
菅朝梁
|
|
18205601716@163.com
|
签名由网易邮箱大师定制
On 11/3/2020 13:34，Sean McManus<notifications@github.com> wrote：

There's some issue with F5 debugging not reacting to changes to the tasks.json -- the issue doesn't repro with 1.0.1 so it seems important to fix for 1.1.0, although it repros with 1.1.0-insiders3 too. I'm not sure yet if it's a VS Code bug or what area of our code could be changed to fix that.

sean-mcmanus (approved):

PR #6421, opened 2020-11-01
Fix whitespace in imports

dsanders11:


sean-mcmanus (approved):

sean-mcmanus:
FYI, we could add `"object-curly-spacing": ["error", "always"],` to catch the other cases of this (see https://eslint.org/docs/rules/object-curly-spacing and https://github.com/microsoft/vscode-cpptools/blob/master/Extension/.eslintrc.js). However, that generates a bunch of other linter errors for other cases where we don't have consistent spacing.

Jianchaoliang:
unsubscribe


| |
菅朝梁
|
|
18205601716@163.com
|
签名由网易邮箱大师定制
On 11/3/2020 06:05，Sean McManus<notifications@github.com> wrote：

FYI, we could add "object-curly-spacing": ["error", "always"], to catch the other cases of this (see https://eslint.org/docs/rules/object-curly-spacing and https://github.com/microsoft/vscode-cpptools/blob/master/Extension/.eslintrc.js). However, that generates a bunch of other linter errors for other cases where we don't have consistent spacing.

PR #6493, opened 2020-11-10
Remove symbol_add_failed.

sean-mcmanus:


michelleangela (approved):

Extension/src/nativeStrings.json
```diff
@@ -31,8 +31,8 @@
     "formatting_failed": "Formatting failed:",
     "database_add_file_failed": "Unable to add file to database, error = {0}: {1}",
     "reset_timestamp_failed": "Failed to reset timestamp during abort, error = {0}: {1}",
+    "symbol_add_failed": "",
```
Colengms: Can't we just remove this line entirely?

Colengms: As long as the native file is regenerated, the indexes should line up.

sean-mcmanus: It causes changes to the indexes, so retranslation.

Colengms: It would not cause retranslation.  We export an xlf file that is used for translations, and it does not include these indexes, just the label names.

sean-mcmanus: Okay, I removed the empty strings.

Issue #6150, opened 2020-09-15
Automatically configure to use custom config provider if available and no other config exists


Colengms:
Opening this issue to track automatic use of a custom config provider, if present/registered.  This would only occur if `c_cpp_properties.json` has not been created, or if the user specified any `C_Cpp` related settings at the workspace level, or any `C_Cpp.default.*` settings at the global level.  This might involve generating a `c_cpp_properties.json`, or perhaps configuring to use the provider without generating that file.

Since we don't currently know if a custom config provider is actually being used for the current workspace, or which is being used if there are multiple, we would either need to isolate this to only if there is just 1 registered (and assume it's used), or we would need to add something to the cpptools API to allow a provider to inform us that it's being used for the current workspace.

PR #6504, opened 2020-11-12
Automatically fall back to config provider

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/6150

Triggers a config refresh when a config provider is registered.
If `c_cpp_properties.json` does not exist, there are no other configuration setting specified in `C_Cpp.defaults.*`, and there is only one custom configuration provider registered, use that provider.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -715,17 +715,18 @@ export class CppProperties {
             if (providers.size === 1
                 && !this.propertiesFile
                 && !settings.defaultCompilerPath
+                && settings.defaultCompilerPath !== ""
                 && !settings.defaultIncludePath
                 && !settings.defaultDefines
                 && !settings.defaultMacFrameworkPath
-                && settings.defaultWindowsSdkVersion === ''
+                && settings.defaultWindowsSdkVersion === ""
                 && !settings.defaultForcedInclude
-                && settings.defaultCompileCommands === ''
+                && settings.defaultCompileCommands === ""
                 && !settings.defaultCompilerArgs
-                && settings.defaultCStandard === ''
-                && settings.defaultCppStandard === ''
-                && settings.defaultIntelliSenseMode === ''
-                && settings.defaultConfigurationProvider === '') {
+                && settings.defaultCStandard === ""
+                && settings.defaultCppStandard === ""
+                && settings.defaultIntelliSenseMode === ""
+                && settings.defaultConfigurationProvider === "") {
                 providers.forEach(provider => { configuration.configurationProvider = provider.extensionId; });
             }
         }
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -708,6 +708,26 @@ export class CppProperties {
 
             configuration.browse.limitSymbolsToIncludedHeaders = this.updateConfigurationStringOrBoolean(configuration.browse.limitSymbolsToIncludedHeaders, settings.defaultLimitSymbolsToIncludedHeaders, env);
             configuration.browse.databaseFilename = this.updateConfigurationString(configuration.browse.databaseFilename, settings.defaultDatabaseFilename, env);
+
+            // If there is no c_cpp_properties.json, there are no relevant C_Cpp.default.* settings set,
+            // and there is only 1 registered custom config provider, default to using that provider.
+            const providers: CustomConfigurationProviderCollection = getCustomConfigProviders();
+            if (providers.size === 1
+                && !this.propertiesFile
+                && !settings.defaultCompilerPath
```
bobbrow, line 717: If the customer sets `C_Cpp.default.compilerPath` to empty string (to disable compiler querying), I think this will do the wrong thing.  And I'm wondering if it's safer to use something like `settingsTracker.getUserModifiedSettings` and check that against the settings names in case we ever change defaults.

Colengms: This seems simple to fix by checking for an empty compilerPath here.  getUserModifiedSettings would also gather settings that don't apply to the configuration.  There are various other checks in this function of these individual settings.  We could move some or all of that into new SettingsTracker functions, but that seems like a can-of-worms, and may not wind up being simpler.

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
bobbrow: I think it's safest to do explicit comparison (`!==` or `===`) with the default settings values for all the settings instead of using `!` then.

Colengms: It's already comparing all string based settings with `===` against `''`, instead of using `!`.  VS Code does not represent string settings using null or undefined, it actually returns an empty string if those settings are not set.  This is related to the settings UI, in which an empty text field and an unset setting are equivalent.  compilerPath is our only current exception (which results in this setting not being editable using a text field in the settings UI, which is unfortunate), and for that I'm now treating `""` differently from unset.  For ALL other strings, it is not using `!`, it's checking against empty string, which is correct.  The remaining uses of `!` are for arrays, and that seems preferable over explicitly checking against both null and undefined.  Unlike strings, when an array is defined but empty, `!` will return false.  So, if there is something that should be changed here to be more explicit, I don't see it.  Let me know.  :)

Extension/src/LanguageServer/configurations.ts
```diff
@@ -483,12 +483,6 @@ export class CppProperties {
                     resolve();
                 }, () => {});
             } else {
-                const settings: CppSettings = new CppSettings(this.rootUri);
-                if (providerId) {
-                    settings.update("default.configurationProvider", providerId);
```
Colengms, line 488: I removed this arbitrarily add/erasing of settings.  It doesn't seem necessary, and reading/writing the user's settings seems like a terrible idea to me.  If they see it set a particular way, and we arbitrarily add/remove it out from under them, that's extremely confusing.

PR #6529, opened 2020-11-16
Update TPN.

### LINKED ISSUE #6517: Replace msgpack with libuv in ThirdPartyNotices.txt
https://github.com/libuv/libuv/blob/v1.x/LICENSE

sean-mcmanus:


sean-mcmanus:
Will probably need to add https://github.com/USCiLab/cereal/blob/master/LICENSE soon.

bobbrow:
This won't go in until 1.2.0-insiders, correct?  I see msgpack removed and replaced with libuv, which is a 1.2.0 change.

sean-mcmanus:
> This won't go in until 1.2.0-insiders, correct? I see msgpack removed and replaced with libuv, which is a 1.2.0 change.

Yeah.

michelleangela (approved):

Issue #6538, opened 2020-11-18
Run-> add configuration ->C++ (GDB/LLDB) "Cannot read property 'includes' of undefined


Kanstarry9T:
Type: Debugger

**Describe the bug**
- OS and Version:
- VS Code Version:
- C/C++ Extension Version:
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is.

**To Reproduce**
*Please include a code sample and `launch.json` configuration.*

**Additional context**
*If applicable, please include logging by adding "logging": { "engineLogging": true, "trace": true, "traceResponse": true } in your `launch.json`*
Add any other context about the problem here including log or error messages in your Debug Console or Output windows.
![image](https://user-images.githubusercontent.com/43159305/99477659-f0157e00-298d-11eb-99ae-6f904ec6981d.png)
![image](https://user-images.githubusercontent.com/43159305/99477677-f6a3f580-298d-11eb-857c-bea84a188fe6.png)

sean-mcmanus:
If you access Help->Toggle Developer Tools->Console do you see a call stack for that error?

You're using 1.1.2 of the extension, right?

Also, any other repro details?

gegogi:
I second this.
Using 1.1.2.

![image](https://user-images.githubusercontent.com/639708/99484441-c5cebb00-29a3-11eb-84dc-f24e00322d47.png)

gegogi:
Looks like this commit might cause the problem.

https://github.com/microsoft/vscode-cpptools/commit/df8523e5509bf8cb2cdde0805c8d9cc87117c8d4

sean-mcmanus:
@gegogi Thanks for the info -- you are correct.

sean-mcmanus:
It appears to be line https://github.com/microsoft/vscode-cpptools/blob/release/Extension/src/Debugger/configurationProvider.ts#L158 (or the similar one). I repro the bug when I delete the "command" field in my tasks.json. Does your tasks.json not have a "command" field for some reason? Do you have a C/C++ compiler installed?

gegogi:
This is my tasks.json and it has command field.
Compiler (VC Toolset) is also installed and I can compile.
Is there any way that I can debug against my local installation of cpptools?

```
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "shell",
			"label": "Build",
			"command": "cl.exe",
			"args": [
				"/I",
[...]

PR #6539, opened 2020-11-18
Fix build task bugs.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/6538
Also, switch to using VS Code's taks variable resolving.
Also, avoids an error in the (invalid) case where there's no label for task.

### DIFF
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -4,7 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 import * as path from 'path';
 import {
-    TaskDefinition, Task, TaskGroup, WorkspaceFolder, ShellExecution, Uri, workspace,
+    TaskDefinition, Task, TaskGroup, ShellExecution, Uri, workspace,
     TaskProvider, TaskScope, CustomExecution, ProcessExecution, TextEditor, Pseudoterminal, EventEmitter, Event, TerminalDimensions, window
 } from 'vscode';
 import * as os from 'os';
@@ -185,9 +185,9 @@ export class CppBuildTaskProvider implements TaskProvider {
 
         const scope: TaskScope = TaskScope.Workspace;
         const task: CppBuildTask = new Task(definition, scope, definition.label, CppBuildTaskProvider.CppBuildSourceStr,
-            new CustomExecution(async (): Promise<Pseudoterminal> =>
+            new CustomExecution(async (resolvedDefinition: TaskDefinition): Promise<Pseudoterminal> =>
                 // When the task is executed, this callback will run. Here, we setup for running the task.
-                new CustomBuildTaskTerminal(resolvedcompilerPath, definition ? definition.args : [], definition ? definition.options : undefined)
+                new CustomBuildTaskTerminal(resolvedcompilerPath, resolvedDefinition ? resolvedDefinition.args : [], resolvedDefinition ? resolvedDefinition.options : undefined)
             ), isCl ? '$msCompile' : '$gcc');
 
         task.group = TaskGroup.Build;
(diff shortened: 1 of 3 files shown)


sean-mcmanus:
Might be a 1.1.3 candidate, not sure yet. Only 2 user reports and there's a workaround.

sean-mcmanus:
> Might be a 1.1.3 candidate, not sure yet. Only 2 user reports and there's a workaround.

...workaround doesn't work if the unsupported task variables (that got added) are required.

sean-mcmanus:
@elahehrashedi It sounds like VS Code might have a way to resolve the tasks variables for us but I haven't found the API yet, see https://github.com/microsoft/vscode/issues/81007 . Looks like it's described at https://github.com/microsoft/vscode/issues/81007#issuecomment-675921123 , but I'm not sure if it's worth the effort to try to hook that up...maybe we could revisit that in a later PR.

alexr00:
The API to have VS Code resolve variables in your task definition properties is here: 

https://github.com/microsoft/vscode/blob/0286c4f793a508837176e1ee3b76a0431d22bb6a/src/vs/vscode.d.ts#L6198-L6198

VS Code will pass your task definition back to you with all variables resolved.

sean-mcmanus:
> The API to have VS Code resolve variables in your task definition properties is here:
> 
> https://github.com/microsoft/vscode/blob/0286c4f793a508837176e1ee3b76a0431d22bb6a/src/vs/vscode.d.ts#L6198-L6198
> 
> VS Code will pass your task definition back to you with all variables resolved.

Okay, thanks -- I figured out how to get it to work -- doing that also fixes other variable resolutions like ${lineNumber} that we weren't handling.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -185,9 +185,9 @@ export class CppBuildTaskProvider implements TaskProvider {
 
         const scope: TaskScope = TaskScope.Workspace;
         const task: CppBuildTask = new Task(definition, scope, definition.label, CppBuildTaskProvider.CppBuildSourceStr,
-            new CustomExecution(async (): Promise<Pseudoterminal> =>
+            new CustomExecution(async (resolvedDefinition: TaskDefinition): Promise<Pseudoterminal> =>
```
elahehrashedi, line 188: @sean-mcmanus I couldn't get this to work. 
Argument of type '(resolvedDefinition: TaskDefinition) => Promise<Pseudoterminal>' is not assignable to parameter of type '() => Thenable<Pseudoterminal>'.

elahehrashedi: can you add a comment similar to this:
`// VS Code will pass task definition back with all variables resolved.`

elahehrashedi (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
sean-mcmanus: Ok

sean-mcmanus: Actually, I'm hesitant to add that comment, because it's described already in the comment for CustomExecution:

image

and the name of the resolvedDefinition implies variables are resolved.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -185,9 +185,9 @@ export class CppBuildTaskProvider implements TaskProvider {
...
                 // When the task is executed, this callback will run. Here, we setup for running the task.
-                new CustomBuildTaskTerminal(resolvedcompilerPath, definition ? definition.args : [], definition ? definition.options : undefined)
+                new CustomBuildTaskTerminal(resolvedcompilerPath, resolvedDefinition ? resolvedDefinition.args : [], resolvedDefinition ? resolvedDefinition.options : undefined)
```
bobbrow: Can it be `resolvedDefinition?.options` instead of `resolvedDefinition ? resolvedDefinition.options : undefined`?

sean-mcmanus: Sure

sean-mcmanus: The ? checks were actually unnecessary.

PR #6544, opened 2020-11-18
Add localization to strings in LanguageServer\cppBuildTaskProvider.ts

### LINKED ISSUE #6436: Add localization to strings in LanguageServer\cppBuildTaskProvider.ts
I see several user facing strings that aren't localized, for example:
1. "compiler: "
2. "Generated task by Debugger".
3. "Starting build..."
4. "Build finished with error".
5. "Build finished successfully"

Not sure if " build active file" can be localized (it's possible it may intentionally be in English due to technical reasons, in which case a comment should be added).

elahehrashedi:
fix : https://github.com/microsoft/vscode-cpptools/issues/6436


Extension/src/Debugger/extension.ts
```diff
@@ -50,7 +50,7 @@ export function initialize(context: vscode.ExtensionContext): void {
             // Not enabled because we do not react to single-file mode correctly yet.
             // We get an ENOENT when the user's c_cpp_properties.json is attempted to be parsed.
             // The DefaultClient will also have its configuration accessed, but since it doesn't exist it errors out.
-            vscode.window.showErrorMessage('This command is not yet available for single-file mode.');
+            vscode.window.showErrorMessage(localize("single_file_mode_not_available", "This command is not yet available for single-file mode."));
```
Colengms: I'd omit the "yet", so it doesn't set an expectation that the functionality will eventually be supported.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -333,7 +337,7 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
     async open(_initialDimensions: TerminalDimensions | undefined): Promise<void> {
         telemetry.logLanguageServerEvent("cppBuildTaskStarted");
         // At this point we can start using the terminal.
-        this.writeEmitter.fire(`Starting build...${this.endOfLine}`);
+        this.writeEmitter.fire(localize("starting_build", "Starting build...{0}", this.endOfLine));
```
Colengms: Maybe add the endOfLine outside of the call to localize?  Otherwise, translators are going to ask us what the {0} is for, as they might wonder if it's relevant to the translation.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -366,13 +370,13 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
                     if (_error) {
                         telemetry.logLanguageServerEvent("cppBuildTaskError");
                         const dot: string = (stdout || _stderr) ? ":" : ".";
-                        this.writeEmitter.fire(`Build finished with error${dot}${this.endOfLine}`);
+                        this.writeEmitter.fire(localize("build_finished_with_error", "Build finished with error{0}{1}", dot, this.endOfLine));
```
Colengms: Similar feedback as above.  Avoid including vars in calls to localize() unless they are significant to how the line is translated.  Alternatively, you could use the other prototype for localize(), which allows a 'comment' field to be passed in, in which you can provide guidance to the loc team about how to translate the string.

Colengms (changes requested):

Colengms (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -156,7 +160,7 @@ export class CppBuildTaskProvider implements TaskProvider {
 
         if (!definition) {
             const taskLabel: string = ((appendSourceToName && !compilerPathBase.startsWith(CppBuildTaskProvider.CppBuildSourceStr)) ?
-                CppBuildTaskProvider.CppBuildSourceStr + ": " : "") + compilerPathBase + " build active file";
+                CppBuildTaskProvider.CppBuildSourceStr + ": " : "") + compilerPathBase + localize("build_active_file", " build active file");
             const filePath: string = path.join('${fileDirname}', '${fileBasenameNoExtension}');
```
sean-mcmanus, line 164: Do you think the translators will understand this? It seems safer to use "build active file" and add the " " before the translated string; otherwise, the translators might not understand and remove the space.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: You are right

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -366,13 +370,13 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
                     if (_error) {
                         telemetry.logLanguageServerEvent("cppBuildTaskError");
                         const dot: string = (stdout || _stderr) ? ":" : ".";
-                        this.writeEmitter.fire(`Build finished with error${dot}${this.endOfLine}`);
```
sean-mcmanus, line 372: Is this ever false? How can _error be non-null but _stderr not?

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -366,13 +370,13 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
...
+                        this.writeEmitter.fire(localize("build_finished_with_error", "Build finished with error") + dot + this.endOfLine);
```
sean-mcmanus: "Build finished with error." is not grammatically correct...if that case is possible it should be "Build finished with an error" or "Build finished with errors", so correct translation of "Build finished with error" might not have enough info to translate correctly...something like "Build finished with errors:" might work (not sure if "error" versus "errors"  matters if there's one error).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: the `dot `variable is either `:` or `.` that is why it is not included in the sentence.
I replaced it with a more common sentence `Build finished with error(s)`

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: `_stderr `could be an empty string. although the build is finished with an error.

sean-mcmanus (approved):

Issue #6595, opened 2020-11-30
Support Apple Silicon ARM64 architecture natively


michelleangela:
VS Code now works on Apple Silicon (https://github.com/microsoft/vscode/issues/101662).

This is a feature request to enable the language server binaries to also run on Apple Silicon natively.

The feature request to have the debugger binaries run on Apple Silicon natively is tracked by https://github.com/microsoft/vscode-cpptools/issues/7035.

**Note:** the cpptools, cpptools-srv, and debugger binaries built for macOS x64 may work on Apple Silicon's x64 emulator, however this may use more resources than having binaries run natively as ARM64 on Apple Silicon.

sean-mcmanus:
I think we'll need Azure support for Big Sur first: https://github.com/actions/virtual-environments/issues/2072 ... unless we can get a "hosted" Big Sur machine added (I don't see one currently).

michelleangela:
This now available on our 1.3.0-insiders4 version as VSIX package **cpptools-osx-arm64.vsix** from
 https://github.com/microsoft/vscode-cpptools/releases/tag/1.3.0-insiders4

FischLu:
> This is now available on our 1.3.0-insiders4 version as VSIX package **cpptools-osx-arm64.vsix** from
> https://github.com/microsoft/vscode-cpptools/releases/tag/1.3.0-insiders4

hello, I have tried this version, but still cannot debug any cpp code, the error message is "ERROR: Unable to start debugging. Unexpected LLDB output from command "-exec-run". process exited with status -1 (attach failed ((os/kern) invalid argument))"

bobbrow:
Thank you for trying out the insiders build!  We are tracking support for IntelliSense and the debugger separately.  The work for native debugger support on Apple Silicon is #7035 and has not been completed yet.

michelleangela:
This is now available in the 1.3.0 release version https://github.com/microsoft/vscode-cpptools/releases/tag/1.3.0

PR #6601, opened 2020-12-01
Allow extension's x64 binaries to run on Apple Silicon x64 emulator

xymeng16:
Since arm64 and x86 macOS should have identical C++ support and features, I simply modified the if-condition in the activation() function at main.ts to allow the extension to work on my M1-chip Macbook. Now it has worked for 1 week with no error.


sean-mcmanus (approved):

sean-mcmanus:
FYI, the cpptools and cpptools-srv binaries are built for x64, so they may run slower and/or use more battery via running in an emulated mode compared to native arm64 binaries would.

WardenGnaw:
Same with the debugger binaries.

The mono runtime shipped is built for macOS x64 and same with the lldb-mi that is built.

michelleangela:
Also note that some default IntelliSense configuration may need to be modified, as these settings are based on the host architecture, platform, and available compiler.

xymeng16:
It seems that cpptools and cpptools-srv are close-sourced (or I don't know where to fetch their code). Hope that native arm64 version binaries will be online soon. Thanks for all of your work!

sean-mcmanus:
@xymeng16 We're tracking that with https://github.com/microsoft/vscode-cpptools/issues/6595 .

PR #6604, opened 2020-12-01
Fix infinite loop after cpptools crashes.

### LINKED ISSUE #6603: VS Code uses non-stop CPU and memory after the cpptools process is killed
Reproed on Windows with VS Code and VS Code Insiders, 1.1.2 and 1.2.0-insiders.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/6603


Colengms (approved):

sean-mcmanus:
This change is required as a result of a "regression" in behavior starting with 1.51.1 (doesn't repro with 1.50.1).

PR #6606, opened 2020-12-01
check for valid user compiler path

### LINKED ISSUE #3760: Default build tasks (Ctrl+Shift+B) should not use the compilerPath if it's invalid
It should use the "knownCompilers" instead. The Build and Debug Active File command behaves correctly.

This is when an old c_cpp_properties.json has an old compilerPath set to a non-existent file (e.g. old cl.exe path).

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/3760


Colengms (changes requested):
Could you add the missing space that the linter is complaining about?  Otherwise, it looks good.

sean-mcmanus (approved):

PR #6637, opened 2020-12-08
add GitHub actions for debugger issues

michelleangela:
Added following GitHub actions to manage debugger issues:
- Close issues with "more info needed" after 14 days of inactivity and involves "andrew" or "pierson"
- Close issues with "questions" after 14 days of inactivity and involves "andrew" or "pierson"
- Close issues with "external" or "by design" immediately
- Close issues with "investigate" or "investigate: costing" after 180 days of inactivity

These actions are defined in separate yml workflows.


Colengms (approved):

Colengms (changes requested):
If added a param to StaleCloser.ts, I think you'll need to also add that param to action.yml and update index.ts to pass the value through.

michelleangela:
@WardenGnaw when these are checked in, I will run these new actions manually to test but without committing to any issues.

Colengms (approved):

.github/workflows/question-closer-debugger.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Checkout Actions
+        uses: actions/checkout@v2
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          readonly: ${{ github.event.inputs.readonly }}
+          labels: question, debugger
+          ignoreLabels: language service, internal
```
michelleangela, line 24: @WardenGnaw this query returns an issue with labels "question" and "debugger" plus other labels the issue may additionally have like "help wanted". Would you like to ignore such issues or auto close them too?

WardenGnaw: This query should be good enough for now. I can update it if we run into issues in the future.

.github/workflows/more-info-needed-closer-debugger.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v2
+      - name: Install Actions
+        run: cd ./.github/actions && npm install --production && cd ../..
+      - name: Stale Closer
+        uses: ./.github/actions/StaleCloser
+        with:
+          readonly: ${{ github.event.inputs.readonly }}
+          labels: more info needed, debugger
+          ignoreLabels: language service, internal
+          involves: wardengnaw, pieandcakes
```
WardenGnaw: `+ calgagi`

Calvin also needs to be added to this list.

WardenGnaw (approved):

.github/workflows/more-info-needed-closer-debugger.yml
michelleangela: I added `calgagi` to the "question" query issues too.

PR #6644, opened 2020-12-08
Fix yaml syntax for GitHub actions

michelleangela:
add option to put quotes `"` around parameter values.

### DIFF
--- a/.github/actions/common/ActionBase.ts
+++ b/.github/actions/common/ActionBase.ts
@@ -54,6 +54,9 @@ export class ActionBase {
 
 		// All indicated labels must be present
 		if (this.labels) {
+			if (this.labels?.length > 2  && this.labels?.startsWith('"') && this.labels?.endsWith('"')) {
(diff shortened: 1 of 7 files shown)


.github/actions/common/ActionBase.ts
```diff
@@ -54,10 +54,13 @@ export class ActionBase {
 
 		// All indicated labels must be present
 		if (this.labels) {
+			if (this.labels?.length > 2  && this.labels?.startsWith('"') && this.labels?.endsWith('"')) {
```
Colengms, line 57: What if; instead of supporting quoting of the entire string, the individual labels could be quoted?  And, if there are already quotes, not adding additional quoting to the query (on line 63 here)?

Colengms (changes requested):

.github/actions/common/ActionBase.ts
michelleangela: individual quotes doesn't work with yaml syntax. Those were the first things I tried.

Colengms (approved):

Issue #6569, opened 2020-11-23
semantic tokens: use 'method' instead of 'member'


aeschli:
After requests in microsoft/language-server-protocol#1087 we decided to deprecate the 'member' token type in favor of a new 'method' token type.

The 'method' token type stands for all types of member functions (also if the language calls it differently)

VSCode issue: microsoft/vscode#111172

@Colengms FYI

Colengms:
I gather this change will need to wait until VS Code 1.52 is released.

aeschli:
Yes, you better wait until 1.52 is out (or restrict the extension to 1.52)

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/1.2.0-insiders2

Issue #5429, opened 2020-05-07
Code folding incorrectly matches to an inactive }


sean-mcmanus:
Originally reported at https://github.com/microsoft/vscode-cpptools/issues/407.

Use code
```
int main()
{
    if (false)
    {
    #if false
    }
    #else
    }
    #endif
    return 0;
}
```

Bug: Code folding incorrectly matches to the inactive } instead of the active one that comes afterwards. The bug doesn't repro with VS.

Colengms:
Code folding is lexer based.  It doesn't know the values of preprocessor conditions.  It looks like VS is initially folding this incorrectly as well, but doing another folding pass after inactive regions are available.  We might be able to do something similar, but need a way to trigger VS Code to request ranges from us again (without having to edit the document).  I've sent them a question on this, and am waiting for a reply.

Colengms:
This is blocked on: https://github.com/microsoft/vscode/issues/99914

Colengms:
Taking this out of tracking, as the blocking VS Code issue has been fixed.

Colengms:
Although the blocking issue has been addressed by VS Code, it's still in insiders and not yet available in a general release.  This will need to be delayed until the next VS Code release.  This will also involve a dependency on that new VS Code version.

Colengms:
The type we need is not yet present in @types/vscode 1.51.0.  It's expected to be in 1.52.0.

PR #6667, opened 2020-12-14
Handle inactive regions when determining folding ranges

Colengms:
When combined with the associated native PR, addresses: https://github.com/microsoft/vscode-cpptools/issues/5429
Also addresses: https://github.com/microsoft/vscode-cpptools/issues/6569 (which is just an enum value change from "member" -> "method")

Also fixed an issue that would cause semantic tokens not to refresh properly after a manual change of the `C_Cpp.enhancedColorization` setting.

### DIFF
--- a/Extension/cpptools.json
+++ b/Extension/cpptools.json
@@ -4,5 +4,5 @@
     "recursiveIncludes": 100,
     "gotoDefIntelliSense": 100,
     "enhancedColorization": 100,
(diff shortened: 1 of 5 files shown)


sean-mcmanus (changes requested):
"minimumVSCodeVersion": "1.49.0"

this needs to be updated too -- and the new cpptools.json published.

sean-mcmanus (approved):

PR #6679, opened 2020-12-17
change master to main

elahehrashedi:


sean-mcmanus (changes requested):
There are still more references that should be updated.

sean-mcmanus (approved):

Issue #4854, opened 2020-01-16
feature request: add node-gyp include paths


bmacnaughton:
When writing node-gyp-based C/C++ extensions to nodejs, node-gyp adds the currently active node include directory to the include path. the binding.gyp file can specify additional paths. These are not configured in system environment variables (and when using nvm the location of the node include directory is dynamic which is the real issue). When using node-addon-api, the `"<!@(node -p \"require('node-addon-api').include\")"` is used in binding.gyp to add node-addon-api's include directory to the include path.

While node-gyp automatically adds these paths for compilation, cpp-tools does not know about them. It would be useful to have an option to automatically include the node include directory as well as those defined in binding.gyp so that vscode can find those files. Currently it squiggles them and isn't able to find declarations and definitions.

Overall it would be a nice addition to have an option of something like `C_Cpp.addNodeGypIncludePaths` or something similar. The `nvm` setting seems most important, because hardcoding a possibly non-existent node-addon-api location is not a big deal. So maybe it's more limited and something like `C_Cpp.addNvmNodeIncludePath`.

Thanks for your consideration.

As a workaround, I created a simple extension that implements a command to return the node include directory:
```
let disposable = vscode.commands.registerCommand('extension.nodeIncludeNvm', function () {
  // just return a best guess as to the include directory for this version
  // of node. presume NVM_BIN var points to bin if it exists. if not, presume
  // /usr/local/include/node. no, it's not sophisticated.
  let path;
  if (process.env.NVM_BIN) {
    path = process.env.NVM_BIN.replace(/\/bin$/, '/include');
  } else {
    path = '/usr/local/include/node';
  }
  return Promise.resolve(path);
});
```

It works within the "echo nodeIncludeNvm path" task:
```
    "tasks": [
        {
            "label": "echo nodeIncludeNvm path",
            "type": "shell",
            "command": "echo ${command:extension.nodeIncludeNvm}",
            "problemMatcher": []
        },
        {
          "label": "echo C_Cpp.default.includePath",
          "type": "shell",
          "command": "echo ${config:C_Cpp.default.includePath}",
          "problemMatcher": []
        }
    ]
```

but I am unable to get it to "stick" in `C_Cpp.default.includePath`. My project settings.json:
```
  "C_Cpp.default.includePath": [
    "${workspaceFolder}",
    "${command:extension.nodeIncludeNvm}",
    "${workspaceFolder}/node_modules/node-addon-api"
  ],
```

[...]

PR #6731, opened 2021-01-01
implement addNodeAddonIncludePaths

bmacnaughton:
This PR addresses the common cases for #4854 by adding the include paths specified by `nan` and `node-addon-api` when they are listed as dependencies in package.json. This was the first time I've looked at vscode-cpptools code and it's also the first time I've written any typescript, so suggestions about how I can improve in either area are welcome.

I haven't changed the changelog nor even looked at adding a specific test pending feedback. Afaict, it passes the unit tests and integration tests but I don't see any positive feedback to be sure.

I modified the insiders branch, so that is the target for this PR. Not sure whether that's the regular practice or not.

Questions:

- Is "machine-overridable" the correct scope [package.json 1109](https://github.com/microsoft/vscode-cpptools/compare/insiders...bmacnaughton:node-addon-includes?expand=1#diff-62bdf835c36931e2307d4d55bb2f11bcfef5e8dc5714ef35b7d5ae81bc5a10bcR1109)
- should it check `settings.addNodeAddonIncludePaths` or just concat an empty array [configuration.ts 327](https://github.com/microsoft/vscode-cpptools/compare/insiders...bmacnaughton:node-addon-includes?expand=1#diff-9b32cba7ca4094f68784eb207728f8c4dd5aa6f0a3aa601c01c55fec735b589eR327)
- mapping the dependencies to commands uses an inline definition. not sure the best practices for this [configuration.ts 398](https://github.com/microsoft/vscode-cpptools/compare/insiders...bmacnaughton:node-addon-includes?expand=1#diff-9b32cba7ca4094f68784eb207728f8c4dd5aa6f0a3aa601c01c55fec735b589eR327).
- `pfs` (fs/promises) is defined at the top of the file but only used one place [configuration.ts 408](https://github.com/microsoft/vscode-cpptools/compare/insiders...bmacnaughton:node-addon-includes?expand=1#diff-9b32cba7ca4094f68784eb207728f8c4dd5aa6f0a3aa601c01c55fec735b589eR327). prefer it locally scoped?
- the include path loop uses a labeled loop. I didn't see those anywhere else; it could be replaced with an if-level. [configuration.ts 414](https://github.com/microsoft/vscode-cpptools/compare/insiders...bmacnaughton:node-addon-includes?expand=1#diff-9b32cba7ca4094f68784eb207728f8c4dd5aa6f0a3aa601c01c55fec735b589eR327)

Obviously, you're welcome to change names, wording, etc. or ask me to do so.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1106,7 +1106,7 @@
           "type": "boolean",
           "default": false,
           "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.description%",
-          "scope": "machine-overridable"
+          "scope": "resource"
         },
         "C_Cpp.renameRequiresIdentifier": {
           "type": "boolean",
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1106,7 +1106,7 @@
           "type": "boolean",
           "default": false,
           "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.description%",
-          "scope": "resource"
+          "scope": "application"
         },
         "C_Cpp.renameRequiresIdentifier": {
           "type": "boolean",
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1102,6 +1102,12 @@
           "markdownDescription": "%c_cpp.configuration.vcpkg.enabled.markdownDescription%",
           "scope": "resource"
         },
+        "C_Cpp.addNodeAddonIncludePaths": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.description%",
+          "scope": "machine-overridable"
+        },
         "C_Cpp.renameRequiresIdentifier": {
           "type": "boolean",
           "default": true,
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1367,6 +1367,11 @@ export class DefaultClient implements Client {
                             this.semanticTokensProvider = undefined;
                         }
                     }
+                    // if addNodeAddonIncludePaths was turned on but no includes have been found yet then 1) presume that nan
+                    // or node-addon-api was installed so prompt for reload.
+                    if (changedSettings["addNodeAddonIncludePaths"] && settings.addNodeAddonIncludePaths && this.configuration.nodeAddonIncludesFound() === 0) {
+                        util.promptForReloadWindowDueToSettingsChange();
+                    }
                 }
                 this.configuration.onDidChangeSettings();
                 telemetry.logLanguageServerEvent("CppSettingsChange", changedSettings, undefined);
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -404,8 +404,8 @@ export class CppProperties {
 
         const pathToNode: string = which.sync("node");
         const nodeAddonMap: { [dependency: string]: string } = {
-            "nan": `${pathToNode} --no-warnings -e "require('nan')"`,
-            "node-addon-api": `${pathToNode} --no-warnings -p "require('node-addon-api').include"`
+            "nan": `"${pathToNode}" --no-warnings -e "require('nan')"`,
+            "node-addon-api": `"${pathToNode}" --no-warnings -p "require('node-addon-api').include"`
         };
 
         if (!error) {
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -322,10 +322,6 @@ export class CppProperties {
             configuration.includePath = [defaultFolder];
         }
 
-        if (settings.addNodeAddonIncludePaths) {
-            configuration.includePath = configuration.includePath.concat(this.nodeAddonIncludes);
-        }
-
         // browse.path is not set by default anymore. When it is not set, the includePath will be used instead.
         if (isUnset(settings.defaultDefines)) {
             configuration.defines = (process.platform === 'win32') ? ["_DEBUG", "UNICODE", "_UNICODE"] : [];
@@ -701,7 +697,8 @@ export class CppProperties {
 
             configuration.includePath = this.updateConfigurationStringArray(configuration.includePath, settings.defaultIncludePath, env);
             if (settings.addNodeAddonIncludePaths) {
-                configuration.includePath = this.updateConfigurationStringArray(configuration.includePath, this.nodeAddonIncludes, env);
+                const includePath: string[] = configuration.includePath || [];
+                configuration.includePath = includePath.concat(this.nodeAddonIncludes.filter(i => includePath.indexOf(i) < 0));
             }
             configuration.defines = this.updateConfigurationStringArray(configuration.defines, settings.defaultDefines, env);
             configuration.macFrameworkPath = this.updateConfigurationStringArray(configuration.macFrameworkPath, settings.defaultMacFrameworkPath, env);
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -21,6 +21,8 @@ import * as nls from 'vscode-nls';
 import { setTimeout } from 'timers';
 import * as which from 'which';
 
+const pfs = fs.promises;
+
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
@@ -130,6 +132,8 @@ export class CppProperties {
     private defaultWindowsSdkVersion: string | null = null;
     private vcpkgIncludes: string[] = [];
     private vcpkgPathReady: boolean = false;
+    private nodeAddonIncludes: string[] = [];
+    private nodeAddonIncludePathsReady: boolean = false;
     private defaultIntelliSenseMode?: string;
     private defaultCustomConfigurationVariables?: { [key: string]: string };
     private readonly configurationGlobPattern: string = "c_cpp_properties.json";
@@ -155,6 +159,7 @@ export class CppProperties {
         this.configFolder = path.join(rootPath, ".vscode");
         this.diagnosticCollection = vscode.languages.createDiagnosticCollection(rootPath);
         this.buildVcpkgIncludePath();
+        this.addNodeAddonIncludeLocations(rootPath);
         this.disposables.push(vscode.Disposable.from(this.configurationsChanged, this.selectionChanged, this.compileCommandsChanged));
     }
 
@@ -288,7 +293,8 @@ export class CppProperties {
     }
 
     private applyDefaultIncludePathsAndFrameworks(): void {
-        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady) {
+        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady
+            && this.nodeAddonIncludePathsReady) {
             const configuration: Configuration | undefined = this.CurrentConfiguration;
             if (configuration) {
                 this.applyDefaultConfigurationValues(configuration);
@@ -317,6 +323,9 @@ export class CppProperties {
         } else {
             configuration.includePath = [defaultFolder];
         }
+        // check settings.addNodeAddonIncludePaths?
+        configuration.includePath = configuration.includePath.concat(this.nodeAddonIncludes);
+
         // browse.path is not set by default anymore. When it is not set, the includePath will be used instead.
         if (isUnset(settings.defaultDefines)) {
             configuration.defines = (process.platform === 'win32') ? ["_DEBUG", "UNICODE", "_UNICODE"] : [];
@@ -382,6 +391,46 @@ export class CppProperties {
             }
         } catch (error) {} finally {
             this.vcpkgPathReady = true;
+            this.handleConfigurationChange();
+        }
+    }
+
+    private nodeAddonMap: {[dependency: string]: string} = {
+      "nan": "node --no-warnings -e \"require('nan')\"",
+      "node-addon-api": "node --no-warnings -p \"require('node-addon-api').include\"",
+    };
+
+    private async addNodeAddonIncludeLocations(rootPath: string) : Promise<void> {
+        try {
+
+            const settings: CppSettings = new CppSettings(this.rootUri);
+            if (settings.addNodeAddonIncludePaths) {
+                const package_json = JSON.parse(await pfs.readFile(path.join(rootPath, "package.json"), "utf8"));
+                dependencyLoop:
+                for (const dep in this.nodeAddonMap) {
+                  if (dep in package_json.dependencies) {
+                      const execCmd: string = this.nodeAddonMap[dep];
+                      let stdout = await util.execChildProcess(execCmd, rootPath);
+                      if (!stdout) continue dependencyLoop;
+
+                      // cleanup newlines
+                      if (stdout[stdout.length - 1] === "\n") {
+                          stdout = stdout.slice(0, -1);
+                      }
+                      // node-addon-api returns a quoted string, e.g., '"/home/user/dir/node_modules/node-addon-api"'.
+                      if (stdout[0] === "\"" && stdout[stdout.length - 1] === "\"") {
+                          stdout = stdout.slice(1, -1);
+                      }
+                      if (stdout) {
+                          this.nodeAddonIncludes.push(stdout);
+                      }
+                    }
+                }
+            }
+        } catch (error) {
+          console.log(error);
+        } finally {
+            this.nodeAddonIncludePathsReady = true;
             this.handleConfigurationChange();
         }
     }
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -406,9 +406,10 @@ export class CppProperties {
             .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
             .catch(e => (error = e));
 
+        const pathToNode: string = which.sync("node");
         const nodeAddonMap: { [dependency: string]: string } = {
-            "nan": "node --no-warnings -e \"require('nan')\"",
-            "node-addon-api": "node --no-warnings -p \"require('node-addon-api').include\""
+            "nan": `${pathToNode} --no-warnings -e "require('nan')"`,
+            "node-addon-api": `${pathToNode} --no-warnings -p "require('node-addon-api').include"`
         };
 
         if (!error) {
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -131,7 +131,6 @@ export class CppProperties {
     private vcpkgIncludes: string[] = [];
     private vcpkgPathReady: boolean = false;
     private nodeAddonIncludes: string[] = [];
-    private nodeAddonIncludePathsReady: boolean = false;
     private defaultIntelliSenseMode?: string;
     private defaultCustomConfigurationVariables?: { [key: string]: string };
     private readonly configurationGlobPattern: string = "c_cpp_properties.json";
@@ -291,8 +290,7 @@ export class CppProperties {
     }
 
     private applyDefaultIncludePathsAndFrameworks(): void {
-        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady
-            && this.nodeAddonIncludePathsReady) {
+        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady) {
             const configuration: Configuration | undefined = this.CurrentConfiguration;
             if (configuration) {
                 this.applyDefaultConfigurationValues(configuration);
@@ -402,14 +400,14 @@ export class CppProperties {
             .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
             .catch(e => (error = e));
 
-        const pathToNode: string = which.sync("node");
-        const nodeAddonMap: { [dependency: string]: string } = {
-            "nan": `"${pathToNode}" --no-warnings -e "require('nan')"`,
-            "node-addon-api": `"${pathToNode}" --no-warnings -p "require('node-addon-api').include"`
-        };
-
         if (!error) {
             try {
+                const pathToNode: string = which.sync("node");
+                const nodeAddonMap: { [dependency: string]: string } = {
+                    "nan": `"${pathToNode}" --no-warnings -e "require('nan')"`,
+                    "node-addon-api": `"${pathToNode}" --no-warnings -p "require('node-addon-api').include"`
+                };
+
                 for (const dep in nodeAddonMap) {
                     if (dep in package_json.dependencies) {
                         const execCmd: string = nodeAddonMap[dep];
@@ -447,12 +445,14 @@ export class CppProperties {
                 error = e;
             }
         }
-        if (error && pdjFound) {
-            // only log an error if package.json exists.
-            console.log('readNodeAddonIncludeLocations', error.message);
+        if (error) {
+            if (pdjFound) {
+                // only log an error if package.json exists.
+                console.log('readNodeAddonIncludeLocations', error.message);
+            }
+        } else {
+            this.handleConfigurationChange();
         }
-        this.nodeAddonIncludePathsReady = true;
-        this.handleConfigurationChange();
     }
 
     private getConfigIndexForPlatform(config: any): number | undefined {
@@ -696,8 +696,10 @@ export class CppProperties {
             const configuration: Configuration = this.configurationJson.configurations[i];
 
             configuration.includePath = this.updateConfigurationStringArray(configuration.includePath, settings.defaultIncludePath, env);
+            // in case includePath is reset below
+            const origIncludePath: string[] | undefined = configuration.includePath;
             if (settings.addNodeAddonIncludePaths) {
-                const includePath: string[] = configuration.includePath || [];
+                const includePath: string[] = origIncludePath || [];
                 configuration.includePath = includePath.concat(this.nodeAddonIncludes.filter(i => includePath.indexOf(i) < 0));
             }
             configuration.defines = this.updateConfigurationStringArray(configuration.defines, settings.defaultDefines, env);
@@ -736,8 +738,9 @@ export class CppProperties {
                         if (!configuration.windowsSdkVersion && !!this.defaultWindowsSdkVersion) {
                             configuration.windowsSdkVersion = this.defaultWindowsSdkVersion;
                         }
-                        if (!configuration.includePath && !!this.defaultIncludes) {
-                            configuration.includePath = this.defaultIncludes;
+                        if (!origIncludePath && !!this.defaultIncludes) {
+                            const includePath: string[] = configuration.includePath || [];
+                            configuration.includePath = includePath.concat(this.defaultIncludes);
                         }
                         if (!configuration.macFrameworkPath && !!this.defaultFrameworks) {
                             configuration.macFrameworkPath = this.defaultFrameworks;
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -401,36 +401,57 @@ export class CppProperties {
     };
 
     private async addNodeAddonIncludeLocations(rootPath: string): Promise<void> {
-        try {
+        let error: Error | undefined;
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
+            .then(pdj => JSON.parse(pdj))
+            .catch(e => (error = e));
 
-            const package_json: any = JSON.parse(await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8"));
-            for (const dep in this.nodeAddonMap) {
-                if (dep in package_json.dependencies) {
-                    const execCmd: string = this.nodeAddonMap[dep];
-                    let stdout: string = await util.execChildProcess(execCmd, rootPath);
-                    if (!stdout) {
-                        continue;
-                    }
+        if (!error) {
+            try {
+                for (const dep in this.nodeAddonMap) {
+                    if (dep in package_json.dependencies) {
+                        const execCmd: string = this.nodeAddonMap[dep];
+                        let stdout: string = await util.execChildProcess(execCmd, rootPath);
+                        if (!stdout) {
+                            continue;
+                        }
 
-                    // cleanup newlines
-                    if (stdout[stdout.length - 1] === "\n") {
-                        stdout = stdout.slice(0, -1);
-                    }
-                    // node-addon-api returns a quoted string, e.g., '"/home/user/dir/node_modules/node-addon-api"'.
-                    if (stdout[0] === "\"" && stdout[stdout.length - 1] === "\"") {
-                        stdout = stdout.slice(1, -1);
-                    }
-                    if (stdout) {
-                        this.nodeAddonIncludes.push(stdout);
+                        // cleanup newlines
+                        if (stdout[stdout.length - 1] === "\n") {
+                            stdout = stdout.slice(0, -1);
+                        }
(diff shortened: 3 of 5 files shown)


sean-mcmanus:
FYI, we might not get around to reviewing this till Wednesday.

sean-mcmanus:
> I haven't changed the changelog

That's fine. We generally update the changelog ourselves before a release, e.g. https://github.com/microsoft/vscode-cpptools/pull/6674/files -- the reason is that modifying the changelog with every PR can cause a merge conflict with every commit, which is annoying.

> I modified the insiders branch

Targeting the insiders branch seems fine and we can merge to master (insiders is almost identical to master currently).

> Is "machine-overridable" the correct scope package.json 1109

No, I think it should be "resource". "machine" and "machine-overridable" scopes are for scopes that are not expected to be valid across different machines/OS's, such as file paths.

> mapping the dependencies to commands uses an inline definition

Seems okay to me. It could also be made local to the function or file scope, but I'm not familiar with the recommended TypeScript practice either.

More comments on the way...

sean-mcmanus (changes requested):
You should install the ESLint extension, and run the the "lint" npm script or "gulp lint". The linter enforces are our whitespace/style rules.

configurations.ts
   24:7   error  expected variable-declaration: 'pfs' to have a typedef (tslint:typedef)           @typescript-eslint/tslint/config
  399:1   error  Expected indentation of 8 spaces but found 6                                      @typescript-eslint/indent
  400:1   error  Expected indentation of 8 spaces but found 6                                      @typescript-eslint/indent
  400:86  error  Unexpected trailing comma                                                         comma-dangle
  403:66  error  Unexpected space before the ':'                                                   @typescript-eslint/type-annotation-spacing       
  408:23  error  expected variable-declaration: 'package_json' to have a typedef (tslint:typedef)  @typescript-eslint/tslint/config
  411:1   error  Expected indentation of 20 spaces but found 18                                    @typescript-eslint/indent
  412:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  413:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  413:27  error  expected variable-declaration: 'stdout' to have a typedef (tslint:typedef)        @typescript-eslint/tslint/config
  414:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  414:23  error  Expected { after 'if' condition                                                   curly
  416:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  417:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  418:1   error  Expected indentation of 28 spaces but found 26                                    @typescript-eslint/indent
  419:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  420:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  421:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  422:1   error  Expected indentation of 28 spaces but found 26                                    @typescript-eslint/indent
  423:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  424:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  425:1   error  Expected indentation of 28 spaces but found 26                                    @typescript-eslint/indent
  426:1   error  Expected indentation of 24 spaces but found 22                                    @typescript-eslint/indent
  431:1   error  Expected indentation of 12 spaces but found 10                                    @typescript-eslint/indent

✖ 24 problems (24 errors, 0 warnings)
  21 errors and 0 warnings potentially fixable with the `--fix` option.

bobbrow:
> Targeting the insiders branch seems fine and we can merge to master (insiders is almost identical to master currently).

Process-wise, I think it's better if we always target `main` because we won't regularly merge `insiders` back to `main` (I was thinking we really only did it one direction `main` -> `insiders` since it's a "release" branch.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -317,6 +323,9 @@ export class CppProperties {
         } else {
             configuration.includePath = [defaultFolder];
         }
+        // check settings.addNodeAddonIncludePaths?
```
sean-mcmanus: Yeah, can you check settings.addNodeAddonIncludePaths and skip the concat if it's false? -- that would enable dynamic disabling without a reload. For the enabling case, in onDidChangeSettings you can just check for the addNodeAddonIncludePaths settings being enabled and then call promptForReloadWindowDueToSettingsChange ...assuming you don't want to make dynamic enabling possible.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +395,46 @@ export class CppProperties {
...
+                      // cleanup newlines
+                      if (stdout[stdout.length - 1] === "\n") {
+                          stdout = stdout.slice(0, -1);
+                      }
+                      // node-addon-api returns a quoted string, e.g., '"/home/user/dir/node_modules/node-addon-api"'.
+                      if (stdout[0] === "\"" && stdout[stdout.length - 1] === "\"") {
+                          stdout = stdout.slice(1, -1);
+                      }
+                      if (stdout) {
+                          this.nodeAddonIncludes.push(stdout);
```
sean-mcmanus: I think you should check if the stdout is a valid directory, maybe calling util.checkDirectoryExists. If it doesn't exist, you could check if path.join("${workspaceFolder}", stdout) exists, since I'm getting back a relative path `node_modules\\nan` from nan.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +395,46 @@ export class CppProperties {
...
+
+            const settings: CppSettings = new CppSettings(this.rootUri);
+            if (settings.addNodeAddonIncludePaths) {
+                const package_json = JSON.parse(await pfs.readFile(path.join(rootPath, "package.json"), "utf8"));
+                dependencyLoop:
+                for (const dep in this.nodeAddonMap) {
+                  if (dep in package_json.dependencies) {
+                      const execCmd: string = this.nodeAddonMap[dep];
+                      let stdout = await util.execChildProcess(execCmd, rootPath);
+                      if (!stdout) continue dependencyLoop;
```
sean-mcmanus: This can just be "continue", right? I don't see why the label is needed.

sean-mcmanus:
> > Targeting the insiders branch seems fine and we can merge to master (insiders is almost identical to master currently).
> 
> Process-wise, I think it's better if PR's always target `main` because we won't regularly merge `insiders` back to `main` (I was thinking we really only did it one direction `main` -> `insiders` since it's a "release" branch).

The potential problem is that main might have breaking changes that don't work without unreleased binaries, while the insider branch is guaranteed to work with the 1.2.0-preview binaries.

There shouldn't be any issue with merging from insiders to main, right?

Extension/src/LanguageServer/configurations.ts
bmacnaughton: i use labels so that if the code changes there is no need to worry about the scope of the continue. but i'm happy to change it.

bmacnaughton:
@sean-mcmanus @bobbrow - thank you for the feedback. i will make these changes this weekend (unless you have a need for them before that or i find a little spare time) as this is a side project of mine.

i'm not sure how the lint errors came about; i have eslint and the extension installed. i didn't run the lint script because i presumed that the extension would use the same rules as the script. i'll run the script and take care of the errors too.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Yeah, I think the label should be removed. It is expected that in the rare a case an sub-loop is added inside another loop then any "continue" would need to be addressed.

sean-mcmanus:
> @sean-mcmanus @bobbrow - thank you for the feedback. i will make these changes this weekend (unless you have a need for them before that or i find a little spare time) as this is a side project of mine.
> 
> i'm not sure how the lint errors came about; i have eslint and the extension installed. i didn't run the lint script because i presumed that the extension would use the same rules as the script. i'll run the script and take care of the errors too.

That's fine. Our goal is to release 1.2.0-insiders some time next week, but I'm not sure what day yet, but this could go into the next release if it doesn't make that.

Yeah, ESLint is supposed to squiggle the lines with issues without having to run the specific command.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: i used `isUnset()` because all the other checks do but am wondering why `!settings.addNodeAddonIncludePaths` might not be valid.

i'm now looking at *which* `onDidChangeSettings()` would be appropriate to modify. any hints appreciated as i don't see any of them that (directly) call `util.promptForReloadWindowDueToSettingsChange()`.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: nan always returns the path relative to the working directory, so i'm calling `path.resolve()` on the results and checking that. i think the only time `util.checkDirectoryExists()` will fail is if the node command executed by `addNodeAddonIncludeLocations()` fails.

bmacnaughton:
addressed with the exception of dynamic enabling of the setting. i'm having to dig around a bit to understand that.

questions:
- i used `: any` for `pfs` and `package_json`. i saw multiple uses of `: any` for reading package.json so that seems ok. but is there a better way to get access to `require('fs').promises`?
- i'm digging into how to dynamically enable the setting but if you can point me at an example that would be great.

btw, my problem not catching the eslint problems was that i didn't notice that eslint wanted to ask me permission to use the locally installed version.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: IsUnset is used because in TypeScript values like "" also implicitly convert to false in "if" statements. So using `isUnset` is not what you want, since it just checks if the setting doesn't exist.

Looking more closely at the code, `applyDefaultConfigurationValues` maybe be the wrong place for setting this. The reason the vcpkgIncludes is set here is because we wanted it to only appear when users create a new default c_cpp_properties.json, but in your case it sounds like you want it to always be auto-added?

The correct spot may be `updateServerOnFolderSettingsChange` (I'm not 100% sure).

Yeah, we currently call util.promtForReloadWindowDueToSettingsChange after we get a response from cpptools from the setting change. I was referring to `DefaultClient.onDidChangeSettings(event: vscode.ConfigurationChangeEvent, isFirstClient: boolean): { [key: string]: string }`, which has some existing code that checks if certain settings were changed.

bmacnaughton:
thanks - this weekend.

sean-mcmanus:
> * i used `: any` for `pfs` and `package_json`. i saw multiple uses of `: any` for reading package.json so that seems ok. but is there a better way to get access to `require('fs').promises`?
> * i'm digging into how to dynamically enable the setting but if you can point me at an example that would be great.

Yeah, dynamic enabling seemed tricky because it would involve having to somehow delay the processing until after addNodeAddonIncludeLocations has generated the nodeAddonIncludes , which I don't think we have any similar case/example that does that, so that's why I was thinking just letting the user know a reload was required was sufficient, particular since it's not expected for that setting to change frequently. Enabling should be simple if nodeAddonIncludes has the data already though (assuming you don't want it to try to regenerate that dynamically).

"any" are used in cases where we don't have type info or if the object is a dynamic type. I think it would be better to just use fs.promises directly instead of defining 'pfs'.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: path.resolve doesn't work because it uses a working path of VS Code, such as `C:\\Users\\<user>\\AppData\\Local\\Programs\\Microsoft VS Code\`, and not the user's workspace. Was there a reason `path.join("${workspaceFolder}", stdout)` or something similar wouldn't work?

bmacnaughton:
my bad. i'll switch it. it seemed more general.

bmacnaughton:
i appreciate your help; i will have more time to devote on the weekend; i'm sorry i haven't really given this the attention it deserves this week.

Extension/package.json
```diff
@@ -1102,6 +1102,12 @@
           "markdownDescription": "%c_cpp.configuration.vcpkg.enabled.markdownDescription%",
           "scope": "resource"
         },
+        "C_Cpp.addNodeAddonIncludePaths": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.description%",
+          "scope": "resource"
```
sean-mcmanus: On second thought, I think this should be "application" scope (i.e. User scope in the UI), because otherwise it would be impossible for a user to download/open/browse a workspace from the internet securely, because execChildProcess is called. Also, it seems rare to want to be enable/disable the setting on a per-project basis anyway.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +396,52 @@ export class CppProperties {
...
+        "nan": "node --no-warnings -e \"require('nan')\"",
+        "node-addon-api": "node --no-warnings -p \"require('node-addon-api').include\""
+    };
+
+    private async addNodeAddonIncludeLocations(rootPath: string): Promise<void> {
+        try {
+
+            const settings: CppSettings = new CppSettings(this.rootUri);
+            if (settings.addNodeAddonIncludePaths) {
+                const package_json: any = JSON.parse(await pfs.readFile(path.join(rootPath, "package.json"), "utf8"));
```
sean-mcmanus: It seems like it would be better to check for this file's existence first?

sean-mcmanus (changes requested):

sean-mcmanus:
> i appreciate your help; i will have more time to devote on the weekend; i'm sorry i haven't really given this the attention it deserves this week.

Yeah, that's fine...we usually don't work on weekends so you'll have to wait till Monday for a response.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: > but in your case it sounds like you want it to always be auto-added

correct. those paths should always be auto-added. this is where i am working now. it doesn't appear that `updateServerOnFolderSettingsChange()` would work if the function `addNodeAddonIncludeLocations` always reads package.json and sets `configuration.nodeAddonIncludes`. right now that is only set at startup when the setting 'addNodeAddonIncludePaths` is set, so removing that check results in those not being read at startup. but it looks like i might need a call to `addNodeAddonIncludeLocations` in both places to handle startup and a settings change.

i'll keep debugging and trying to determine how to get the logic working. i ran into the "unbound breakpoint" problem but was able to recover by checking out a previous commit. it took me a bit though.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: no reason that doesn't work. i was trying to get it resolved with one shot as opposed to looking around but obviously hadn't checked my assumption about the working directory. i will address this in the next push.

Extension/package.json
bmacnaughton: ok. the scope definitions are still a bit of a mystery to me but next push will have this change.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: yes, i should or the user will get a message about an unhandled promise rejection.

i will change this logic to better capture errors and issue better messages. should i use `vscode.window.showErrorMessage()` to display a message if there is a problem? it doesn't seem like using console.log is optimal.

bmacnaughton:
ok, just pushed changes. biggest changes:

- the code now reads any node addon include paths at start up. this change allows the setting to be changed on the fly (mostly see below)
- `addNodeAddonIncludePaths()` uses `vscode.window.showErrorMessage()` when it detects errors.
- i'm thinking of renaming `addNodeAddonIncludePaths` to `readNodeAddonIncludePaths` because it doesn't add them.


(mostly allows changing on the fly) if the user adds `nan` or `node-addon-api` to the project on the fly the code doesn't know that and doesn't know to re-execute `addNodeAddonIncludePaths()`. is there a way to get notified when a specific file (the project's package.json) is modified?

feedback is welcome.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: For serious errors that users should always see, showErrorMessage should be used, the console.log may only get seen if the user looks at the logs or does a Toggle Developer Console. For this case, it seems like it should just silently skip the processing, since the user would just be opening a non-nodejs project.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: > * i'm thinking of renaming `addNodeAddonIncludePaths` to `readNodeAddonIncludePaths` because it doesn't add them.

That sounds good.

> (mostly allows changing on the fly) if the user adds `nan` or `node-addon-api` to the project on the fly the code doesn't know that and doesn't know to re-execute `addNodeAddonIncludePaths()`. is there a way to get notified when a specific file (the project's package.json) is modified?

Yes, you can add a file watcher on the package.json and react to changes. We do the same thing for c_cpp_properties.json at https://github.com/microsoft/vscode-cpptools/blob/insiders/Extension/src/LanguageServer/configurations.ts#L207 .

In some cases, the file watchers may file, so you could fallback to checking the time stamp in our idle loop processing like was done in https://github.com/microsoft/vscode-cpptools/pull/6748/files for compile_commands.json .
 
> and i need to clean up the `showErrorMessage('error.adding.node.addon.includes')` but wanted to get your feedback on the preferred way to showing errors like this first.
> 
> feedback is appreciated.

I didn't get a chance to finish reviewing the changes today.

bmacnaughton:
ok, i think it's pretty close at this point.
- it now fails silently. the only times it really can fail now are
    - package.json doesn't exist or is malformed (it can't be a node-addon project)
    - requiring node-addon-api/nan fails (they haven't been added yet or are malformed. could keep track of what is present).
    - the directory node-addon-api/nan returns is not valid (can't happen now but maybe could in the future).
    - a coding error (consider reporting this one?)
- prompts for reload if `addNodeAddonIncludePaths` is turned on but no addon includes were found by `readNodeAddonIncludeLocations`. this doesn't catch the case where `nan` was found but they added `node-addon-api` before turning on the setting. i looked at creating a filewatcher but chose not to because i don't believe this is a major use of vscode for c/c++, so it should avoid using resources unnecessarily. (it also would require async activity in order to handle it and i didn't see how that could be handled in the file watcher, but that was secondary. if there is a filewatcher that already watches package.json or files added in node_modules and that could be piggy-backed on it might be worth revisiting.)
- in theory `DefaultClient.onDidChangeSettings()` could initiate the process by kicking off `readNodeAddonIncludeLocations`
- maybe the docs could note that adding node-addon-api or nan require reloading and the check for the length of node addon include paths found could be removed.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +395,71 @@ export class CppProperties {
...
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
+            .then(pdj => JSON.parse(pdj))
+            .catch(e => (error = e));
+
+        const nodeAddonMap: { [dependency: string]: string } = {
+            "nan": "node --no-warnings -e \"require('nan')\"",
+            "node-addon-api": "node --no-warnings -p \"require('node-addon-api').include\""
+        };
+
+        let added: number = 0;
```
sean-mcmanus: This isn't being used currently. Should it be removed?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +395,71 @@ export class CppProperties {
...
+                    }
+                }
+            } catch (e) {
+                error = e;
+            }
+        }
+        if (error) {
+            // if it's possible for the directories returned by nan or node-addon-api to fail then
+            // this should be uncommented.
+            // console.log('readNodeAddonIncludeLocations', error.message);
```
sean-mcmanus: The logging seems like it would be okay as long as the missing package.json case is excluded (currently it's not).

sean-mcmanus:
The "unbound breakpoint" issue you were hitting may have been caused by the extension not being activated...opening a .cpp should trigger the activation.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -317,6 +321,11 @@ export class CppProperties {
         } else {
             configuration.includePath = [defaultFolder];
         }
+
+        if (settings.addNodeAddonIncludePaths) {
+            configuration.includePath = configuration.includePath.concat(this.nodeAddonIncludes);
```
sean-mcmanus: It seems like the changes to applyDefaultIncludePathsAndFrameworks should be removed, because it sounds like you don't want the default c_cpp_properties.json to have the nodeAddonIncludes added...otherwise they would be added twice by updateServerOnFolderSettingsChange. The reason we added vcpkg paths to the default was because we expected that to be user modified.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: yes, nice eye - thanks. it was an experiment to avoid `handleConfigurationChange()` but it didn't work correctly.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: done

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +395,71 @@ export class CppProperties {
...
+            "node-addon-api": "node --no-warnings -p \"require('node-addon-api').include\""
+        };
+
+        let added: number = 0;
+        if (!error) {
+            try {
+                for (const dep in nodeAddonMap) {
+                    if (dep in package_json.dependencies) {
+                        const execCmd: string = nodeAddonMap[dep];
+                        let stdout: string = await util.execChildProcess(execCmd, rootPath);
```
sean-mcmanus, line 414: This will automatically execute any node or node.exe process in the user's root workspace if they have a package.json, which is potentially unwanted. Could you use something like
```
        const pathToNode: string = which.sync("node");
        const nodeAddonMap: { [dependency: string]: string } = {
            "nan": `"${pathToNode}" --no-warnings -e \"require('nan')\"`,
            "node-addon-api": `"${pathToNode}" --no-warnings -p \"require('node-addon-api').include\"`
        };
```
instead? That will cause the "node" on the user's path to execute instead of potentially one in their current working directory.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
bmacnaughton: no problem and good catch. thank you - i had wondered about how to execute "which" cross-platform. that wouldn't execute a local node in linux which is where i spend my time. i hadn't realized that it would do so in windows - a good learning for me.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: i can remove but if i do it creates this problem (which maybe i should be handling in some other way):

- start with "Add Node Addon Include Paths" checked - no squiggles
- uncheck "Add Node Addon Include Paths" - get squiggles
- check "Add Node Addon Include Paths" - squiggles do not go away

with the lines as they are the squiggles go away when the setting is checked (3rd bullet).

bmacnaughton: just need a simple project depending on `node-addon-api` with an `include <napi.h>` in a c/c++ file.

bmacnaughton: could change it to `configuration.includePath.concat(this.nodeAddonIncludes.filter(i => configuration.includePath.indexOf(i) < 0)` to avoid duplicating if there's not a better way.

bmacnaughton: i will figure it out. it's not clear why that makes a difference but it's not working the way i thought it was.

bmacnaughton: i believe it's fixed now. the way it was working was accidental - not what i thought. thanks for the note.

not sure i handled the possibly undefined `configuration.includePath` in a standard way. let me know if you'd like it changed. i appreciate your time on this - i have learned a lot stumbling around with your help.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +391,70 @@ export class CppProperties {
...
+    private async readNodeAddonIncludeLocations(rootPath: string): Promise<void> {
+        let error: Error | undefined;
+        let pdjFound: boolean = false;
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
+            .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
+            .catch(e => (error = e));
+
+        const pathToNode: string = which.sync("node");
+        const nodeAddonMap: { [dependency: string]: string } = {
+            "nan": `${pathToNode} --no-warnings -e "require('nan')"`,
```
sean-mcmanus: ${pathToNode} needs to be `"${pathToNode}"` in order to handle spaces in the path on Windows (I haven't checked Linux).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
bmacnaughton: just pushed. it would need them on linux too (though most people don't use spaces in paths). i've been a bit sloppy.

bmacnaughton: would you like me to resolve these when i think i have or would you like to?

sean-mcmanus: Yeah, you can resolve them.

bmacnaughton:
i believe that i have made all the changes you requested. let me know if you have any additional.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -627,6 +696,10 @@ export class CppProperties {
             const configuration: Configuration = this.configurationJson.configurations[i];
 
             configuration.includePath = this.updateConfigurationStringArray(configuration.includePath, settings.defaultIncludePath, env);
+            if (settings.addNodeAddonIncludePaths) {
```
sean-mcmanus, line 701: configuration.includePath is checked/modified in the code below under certain settings. The "unmodified" configuration.includePath should be saved to some variable here to be used later.

sean-mcmanus:
```
if (!configuration.includePath && !!this.defaultIncludes) {
                            configuration.includePath = this.defaultIncludes;
                        }
```
should check the unmodified/original variable and then do something like
```
                const includePath: string[] = configuration.includePath || [];
                configuration.includePath = includePath.concat(this.defaultIncludes);
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -288,7 +291,8 @@ export class CppProperties {
     }
 
     private applyDefaultIncludePathsAndFrameworks(): void {
-        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady) {
+        if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady
+            && this.nodeAddonIncludePathsReady) {
```
sean-mcmanus: This variable is no longer needed. It was only needed when the applyDefaultConfigurationValues was being used.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +391,70 @@ export class CppProperties {
...
+                        if (stdout) {
+                            this.nodeAddonIncludes.push(stdout);
+                        }
+                    }
+                }
+            } catch (e) {
+                error = e;
+            }
+        }
+        if (error && pdjFound) {
```
sean-mcmanus: I think this should be something like
```
if (error) {
if (pdjFound) {
...
}
} else {
this.handleConfigurationChange();
}
```
i.e. to skip this.handleConfigurationChange when it's not needed.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -386,6 +391,70 @@ export class CppProperties {
...
+    }
+
+    private async readNodeAddonIncludeLocations(rootPath: string): Promise<void> {
+        let error: Error | undefined;
+        let pdjFound: boolean = false;
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
+            .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
+            .catch(e => (error = e));
+
+        const pathToNode: string = which.sync("node");
```
sean-mcmanus: This should be inside the `try {` below (avoids calling which.sync when it's not needed).

Extension/src/LanguageServer/client.ts
```diff
@@ -1367,6 +1367,11 @@ export class DefaultClient implements Client {
                             this.semanticTokensProvider = undefined;
                         }
                     }
+                    // if addNodeAddonIncludePaths was turned on but no includes have been found yet then 1) presume that nan
+                    // or node-addon-api was installed so prompt for reload.
+                    if (changedSettings["addNodeAddonIncludePaths"] && settings.addNodeAddonIncludePaths && this.configuration.nodeAddonIncludesFound() === 0) {
+                        util.promptForReloadWindowDueToSettingsChange();
```
sean-mcmanus, line 1373: This will only get shown if a user adds the nan/node-addon-api to their package.json and then enables the setting -- is that intentional? It seems fine to me, but it just covers a subset of the cases.

sean-mcmanus (changes requested):

sean-mcmanus:
> i believe that i have made all the changes you requested. let me know if you have any additional.

Yeah, I think it's close. I noticed some additional minor issues.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Or checking this.nodeAddonIncludes.length > 0 (i.e. in case one of the cases errors?).

Extension/src/LanguageServer/client.ts
bmacnaughton: > prompts for reload if addNodeAddonIncludePaths is turned on but no addon includes were found by readNodeAddonIncludeLocations. this doesn't catch the case where nan was found but they added node-addon-api before turning on the setting.

above is my original comment on this. it's definitely a subset but seemed like one that is likely to be hit with a typical workflow - create a project, start editing, add node-addon-api or nan, notice the squiggles, and then fix it.

if your experience is different i'm happy to take it out. but it's kind of the best that could be done without creating a file-watcher or changing `readNodeAddonIncludeLocations` to be synchronous.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: good point.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: i think it is (conceivably) possible that both node-addon-api and nan are referenced in package.json, both requires succeed, but one of two util.checkDirectoryExists() fail.

but i am changing it as i think it's 1) not possible now given what node-addon-api and nan return and 2) a really tiny edge case going forward because it means the package has to return an invalid directory.

thank you.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: yes, obvious when i see it. you have a much better handle on all the moving pieces than i do - thanks for spending time on this.

Extension/src/LanguageServer/configurations.ts
bmacnaughton: thanks for this; i feel like a dog watching tv when i work in this codebase. everything is new and my focus is on a very tiny slice of what is going on.

bmacnaughton:
@sean-mcmanus - thanks for the review. i think i've addressed the issues; i don't think the first one was minor and i appreciate your continuing help.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, seems fine to me.

sean-mcmanus (approved):

sean-mcmanus:
> @sean-mcmanus - thanks for the review. i think i've addressed the issues; i don't think the first one was minor and i appreciate your continuing help.

Yeah, thanks for your contribution. Looks like it'll make 1.2.0-insiders -- hopefully it'll be available by end of day Thursday unless we find any last minute blocking issues.

Issue #1083, opened 2017-09-30
Support cross-compilation configurations for IntelliSense


jfroy:
I am sometimes working on my project using a cross-compilation toolchain (for example, working on a Mac laptop; the project targets Linux). I tried to configure IntelliSense to only search the project and the cross-compilation sysroot. But I got this error:

```
file: 'file:///<redacted>.cc'
severity: 'Info'
message: 'cannot open source file "Availability.h" (dependency of "<redacted>.h")'
at: '1,1'
source: ''
```

This is of course a macOS header. I am guessing the indexer might be scanning hard-coded host paths. It seems likely that to properly support cross-compilation, there would need to be a sysroot parameter to act as a final catch-all.

sean-mcmanus:
What is the configuration you're using in c_cpp_properties.json? It sounds like it may be using the macFrameworkPath. We don't scan any hard-coded host paths that aren't in the c_cpp_properties.json.

jfroy:
This is my c_cpp_properties.json, with minor edits for confidentiality.

```json
{
    "configurations": [
        {
            "name": "Foo",
            "includePath": [
                "${workspaceRoot}",
                "${workspaceRoot}/build/vscode/gen",
                "${workspaceRoot}/build/vscode/GLog-prefix/install/include",
                "${workspaceRoot}/third_party/gflags/include/linux",
                "${workspaceRoot}/third_party/googletest/googletest/include",
                "${workspaceRoot}/third_party/protobuf/src",
                "${workspaceRoot}/tools/sdk/toolchain/include/c++/v1",
                "${workspaceRoot}/tools/sdk/toolchain/lib/clang/4.0.1/include",
                "${workspaceRoot}/tools/sdk/sysroot/usr/include",
                "${workspaceRoot}/tools/sdk/sysroot/usr/include/x86_64-linux-gnu"
            ],
            "defines": [
            ],
            "intelliSenseMode": "clang-x64",
            "browse": {
                "path": [
                    "${workspaceRoot}",
                    "${workspaceRoot}/build/vscode/gen",
                    "${workspaceRoot}/build/vscode/GLog-prefix/install/include",
                    "${workspaceRoot}/third_party/gflags/include/linux",
                    "${workspaceRoot}/third_party/googletest/googletest/include",
                    "${workspaceRoot}/third_party/protobuf/src",
                    "${workspaceRoot}/tools/sdk/toolchain/include/c++/v1",
[...]

Issue #6075, opened 2020-09-03
Mixed up default intellisense options


edwood-grant:
**Type: LanguageService**

**Describe the bug**
When I change intellisense default mode, it seems that some settings are mixed up and show different values when setting some values. I happened to fond this when it was showing VLA intellisense errors on a C18 standard when it sohuldn't in clang on windows.

- OS and Version:  Windows 10
- VS Code Version: 1.48.2
- C/C++ Extension Version:  0.30.0-insiders4
- Other extensions you installed (and if the issue persists after disabling them): I do have **CodeLLDB 1.5.3**, but the issue seems to be not related to this extension.

**Steps to reproduce**
1. Go to **settings.json**
2. Set the **C_Cpp.default.intelliSenseMode** value to anything you want to.
3. Call the C/C++ diagnostics. Look for the **IntelliSense Mode** log file.
4. See mixed up versions. This will also trigger errors given the intelligence (for example VLAs)
    a. **clang-x86** and **clang-x64** results in **msvc-x86** and **msvc-x64**
    b. **gcc-x86** and **gcc-x64** results in **clang-x86** and **clang-x64**
    c. **gcc-arm** results in **clang-arm**

The rest of the settings work fine.

This also happens if you set this values via the UI  method. When I reload the project, the feature seems to correct itself, up until I close the settings window or file.

**Expected behavior**
Have correct relations to the settings.

sean-mcmanus:
We'll generally override the intelliSenseMode based on the compilerPath that is found. What compilerPath is being used? You could see via running C/C++: Log Diagnostics.

edwood-grant:
The compiler path points to clang, version 10.0. This is my setup in **c_cpp_properties.json**: 
`"compilerPath": "C:/Program Files/LLVM/bin/clang.exe",`

The compiler path showing in the diagnostics is the same in both the **Current Configuration** and the **Translation Unit Configurations**. 

It does seem that its showing wrong in the translation configurations section only. I may be doing something wrong, maybe the extra inclusions for windows might be a problem, and force it to msvc-64 (is there a way for me to remove those additional includes if I want to?). Or maybe there is something in clang that connects to msvc-64 and I cannot change it on Windows.

I'll attach my **C++ Diagnostic log file** here and my **c_cpp_properties.json** file as well.

Thank you.

[C_CPP_Diagnostics.log](https://github.com/microsoft/vscode-cpptools/files/5213382/C_CPP_Diagnostics.log)
[...]

PR #6745, opened 2021-01-06
Add triplet IntelliSense modes to support cross-compilation

michelleangela:
Issue https://github.com/microsoft/vscode-cpptools/issues/1083

**New IntelliSenseMode options**
These new IntelliSenseMode options are added to support a target by "platform", "compiler" and "architecture".

IntelliSenseMode for targeting Linux platform
IntelliSenseMode | platform | compiler | architecture
-- | -- | -- | --
`Linux-gcc-x86` | linux | gcc | X86
`Linux-gcc-x64` | linux | gcc | X64
`Linux-gcc-arm` | linux | gcc | Arm
`Linux-gcc-arm64` | linux | gcc | Arm64
`Linux-clang-x86` | linux | clang | X86
`Linux-clang-x64` | linux | clang | X64
`Linux-clang-arm` | linux | clang | Arm
`Linux-clang-arm64` | linux | clang | Arm64

IntelliSense modes for targeting macOS platform
IntelliSenseMode | platform | compiler | architecture
-- | -- | -- | --
`macos-clang-x86` | macOS | clang | X86
`macos-clang-x64` | macOS | clang | X64
`macos-clang-arm` | macOS | clang | Arm
`macos-clang-arm64` | macOS | clang | Arm64
`macos-gcc-x86` | macOS | gcc | X86
`macos-gcc-x64` | macOS | gcc | X64
`macos-gcc-arm` | macOS | gcc | Arm
`macos-gcc-arm64` | macOS | gcc | Arm64

IntelliSense modes for targeting Windows platform
IntelliSenseMode | platform | compiler | architecture
-- | -- | -- | --
`windows-msvc-x86` | windows | msvc | X86
`windows-msvc-x64` | windows | msvc | X64
`windows-msvc-arm` | windows | msvc | Arm
`windows-msvc-arm64` | windows | msvc | Arm64
`windows-clang-x86` | windows | clang | X86
`windows-clang-x64` | windows | clang | X64
`windows-clang-arm` | windows | clang | Arm
`windows-clang-arm64` | windows | clang | Arm64
`windows-gcc-x86` | windows | gcc | X86
`windows-gcc-x64` | windows | gcc | X64
`windows-gcc-arm` | windows | gcc | Arm
`windows-gcc-arm64` | windows | gcc | Arm64

Legacy or old IntelliSenseMode options will still be available. However, they will be converted automatically as follows when IntelliSenseMode is used for the IntelliSense engine:

Legacy   IntelliSenseMode | Corresponding new  IntelliSenseMode
-- | --
`msvc-<architecture>` | `windows-msvc-<architecture>`
`clang-<architecture>` | `<host   platform>-clang-<architecture>`
`gcc-<architecture>` | `<host   platform>-gcc-<architecture>`

**Code changes:**
The predefined macros set based on IntelliSenseMode option are now defined in the IntelliSenseMode *.JSON files. The naming of a *.JSON mirrors the naming pattern of IntelliSenseMode options, which is `<platform>-<compiler>-<architecture>`.

TODO:
#6768


Extension/src/LanguageServer/configurations.ts
```diff
@@ -441,7 +441,10 @@ export class CppProperties {
         // Valid compiler + IntelliSenseMode combinations:
         // 1. compiler is cl/clang-cl and IntelliSenseMode is MSVC
         // 2. compiler is not cl/clang-cl and IntelliSenseMode is not MSVC
-        const isValid: boolean = compilerPathAndArgs.compilerName.endsWith("cl.exe") === configuration.intelliSenseMode.startsWith("msvc");
+        const isValid: boolean = compilerPathAndArgs.compilerName.endsWith("cl.exe") ===
+            (configuration.intelliSenseMode.startsWith("msvc")
+            || configuration.intelliSenseMode.startsWith("windows-msvc")
+            || configuration.intelliSenseMode.startsWith("windows-clang"));
```
Colengms: I think the addition of "windows-clang" here is wrong.  1. On Windows, it's possible to use either clang or clang-cl.  2. We moved away from using MSVC intellisense mode for clang/clang-cl, and are instead using the clang intellisense mode with the appropriate ms extensions enabled.  This was addressed with https://github.com/microsoft/vscode-cpptools/issues/6075 (And changed by native PR: 274593)  I don't believe we should be matching up clang/clang-cl with MSVC intellisense mode anymore.  If you see any checks for this, they may be erroneous.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -441,7 +441,10 @@ export class CppProperties {
         // Valid compiler + IntelliSenseMode combinations:
         // 1. compiler is cl/clang-cl and IntelliSenseMode is MSVC
         // 2. compiler is not cl/clang-cl and IntelliSenseMode is not MSVC
-        const isValid: boolean = compilerPathAndArgs.compilerName.endsWith("cl.exe") === configuration.intelliSenseMode.startsWith("msvc");
+        const isValid: boolean = compilerPathAndArgs.compilerName.endsWith("cl.exe") ===
```
Colengms: Not your change, but could you also update this to do a case-insensitive check against "cl.exe"?  i.e. Right now, if "CL.exe" is specified, it works, but we show a squiggle.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
Colengms: Doh.  I misunderstood.  Disregard my last comment.

Was the intention to no longer squiggle use of clang intellisense mode when using "clang-cl.exe"?  If so, maybe this could be slightly refactored to compare the whole filename part == "cl.exe" rather than using endsWith(), and only check against MSVC modes?

michelleangela: >Was the intention to no longer squiggle use of clang intellisense mode when using "clang-cl.exe"? 

Yeah, the use of `windows-clang*` mode indicates the use of "clang-cl.exe" for windows platform.

Extension/src/common.ts
```diff
@@ -951,7 +966,7 @@ function extractArgs(argsString: string): string[] {
 }
 
 export function extractCompilerPathAndArgs(inputCompilerPath?: string, inputCompilerArgs?: string[]): CompilerPathAndArgs {
-    let compilerPath: string | undefined = inputCompilerPath;
+    let compilerPath: string | undefined = inputCompilerPath?.toLowerCase();
```
Colengms: I think converting the path to lower case here would cause some issues.  The compiler path is only case insensitive on Windows.  And, at this point, it could contain compiler args as well.  This function parses out those args, and adds them to `additionalArgs`.  The casing of compiler args matter on all platforms.

Colengms (changes requested):

sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -972,9 +987,9 @@ export function extractCompilerPathAndArgs(inputCompilerPath?: string, inputComp
             // Input has no quotes but can have a compiler path with spaces and args.
             // Go from right to left checking if a valid path is to the left of a space.
             let spaceStart: number = compilerPath.lastIndexOf(" ");
-            if (spaceStart !== -1 && (!isWindows || !compilerPath.endsWith("cl.exe")) && !checkFileExistsSync(compilerPath)) {
+            if (spaceStart !== -1 && (!isWindows || !compilerPath.toLowerCase().endsWith("cl.exe")) && !checkFileExistsSync(compilerPath)) {
```
Colengms: What is `compilerPath.toLowerCase().endsWith("cl.exe")` accomplishing on line 990?  If `compilerPath` ended with a filename of "cl.exe", it would have been caught on line 974.  So, this would only detect clang-cl.exe, which I don't think we should be do anything for here, anymore.  I think this IF statement should be reduced to just `if (spaceStart !== -1 && !checkFileExistsSync(compilerPath))`.  Also, Nit: Could you use a new var to avoid calling toLowerCase() multiple times on `compilerPath`, on line 974?  :)

Colengms (changes requested):

Extension/src/common.ts
```diff
@@ -972,9 +987,9 @@ export function extractCompilerPathAndArgs(inputCompilerPath?: string, inputComp
...
                 let potentialCompilerPath: string = compilerPath.substr(0, spaceStart);
-                while ((!isWindows || !potentialCompilerPath.endsWith("cl.exe")) && !checkFileExistsSync(potentialCompilerPath)) {
+                while ((!isWindows || !potentialCompilerPath.toLowerCase().endsWith("cl.exe")) && !checkFileExistsSync(potentialCompilerPath)) {
```
Colengms: Also, this check might match `clang-cl.exe` as well.  I think it needs to grab the base filename from potentialCompilerPath and check the whole name.

Colengms (changes requested):

Extension/src/common.ts
sean-mcmanus: Or something like compilerPath.toLowerCase().endsWith("\\cl.exe") || compilerPath.toLowerCase().endsWith("/cl.exe") || compilerPath.toLowerCase() === "cl.exe" which is done earlier on.

michelleangela: I updated the function and removed the logic for checking `clang-cl.exe` as that's not needed anymore.

Extension/src/common.ts
```diff
@@ -1003,6 +1018,9 @@ export function extractCompilerPathAndArgs(inputCompilerPath?: string, inputComp
                 return inputCompilerArgs.indexOf(item) < 0;
             }));
     }
+    if (compilerName) {
+        compilerName = compilerName.toLowerCase();
```
Colengms: This makes compilerName lower case on all platforms.  But, it looks like it's being compared against a path in getTasks() in cppBuildTaskProvider.ts, which would seem to cause a bug.  It may be safer to keep as the original case, and convert to lower case when comparing against "cl.exe".

Colengms (changes requested):

Colengms (approved):

Extension/src/common.ts
```diff
@@ -897,6 +913,7 @@ export function downloadFileToStr(urlStr: string, headers?: OutgoingHttpHeaders)
     });
 }
 
+/** CompilerPathAndArgs retains original casing of text input for compiler path and args */
```
sean-mcmanus, line 916: Is this supposed to be `/*`?

sean-mcmanus:
Should the package.json be changed now to reference version 5.0.0 of vscode-cpptools-api or do we need to wait till that's published, i.e. in a 2nd PR?

michelleangela:
> Should the package.json be changed now to reference version 5.0.0 of vscode-cpptools-api or do we need to wait till that's published, i.e. in a 2nd PR?

We need to wait until "5.0.0 of vscode-cpptools-api" is published. Otherwise there's no way for the extension to get it.

Extension/src/common.ts
michelleangela: The comment doesn't show on hover for `/*`, it needs to be `/**`.

sean-mcmanus: Do you know why? I didn't see `/**` being used in other cases.

sean-mcmanus: Ah, I see now -- they're all like /** -- I was seeing a /* before for some reason.

Extension/src/common.ts
```diff
@@ -544,19 +544,35 @@ export async function checkInstallBinariesExist(): Promise<boolean> {
 export async function checkInstallJsonsExist(): Promise<boolean> {
```
sean-mcmanus, line 544: I was looking at this case -- looks like it should also be changed to /**, but that's unrelated to your changes so you don't need to fix that -- we probably have other cases of `/*` that should be fixed. I never realized it was an issue.

PR #6748, opened 2021-01-06
edit to compile-commands take effect immediately 

### LINKED ISSUE #6285: Edits to compile_commands.json don't take effect immediately on Linux (fs.watch fails to work)
Using vscode.workspace.createFileSystemWatcher doesn't seem to work either.

Checking the time stamp as a fallback on an interval timer like we do for c_cpp_properties.json might be a workaround.

elahehrashedi:
`Creating `a new Compile_commands file, `deleting`, and `editing `is taking effect sooner. 
Some code that was not being hit before is removed as well.

bug fix: https://github.com/microsoft/vscode-cpptools/issues/6285


Extension/src/LanguageServer/configurations.ts
```diff
@@ -120,7 +119,9 @@ export class CppProperties {
     private currentConfigurationIndex: PersistentFolderState<number> | undefined;
     private configFileWatcher: vscode.FileSystemWatcher | null = null;
     private configFileWatcherFallbackTime: Date = new Date(); // Used when file watching fails.
-    private compileCommandFileWatchers: fs.FSWatcher[] = [];
+    private compileCommandsFile: vscode.Uri | undefined | null = undefined;
+    private compileCommandsFileWatchers: fs.FSWatcher[] = [];
```
Colengms, line 124: Is this used?  I see it get cleared, but I don't see where it gets populated.  Does that mean that file watchers are no longer used for compile_commands.json files at all, and it's always checked on an interval instead (every 2.5 seconds, like c_cpp_properties.json) ?  If the list of watchers is not leveraged, the variable could be removed.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
elahehrashedi: Yes actually the array of filewatchers is not being used and I have to remove it. In the interval, we check for any updates.

sean-mcmanus: Should we keep the file watcher code in case it works for some users? I'm not sure if it was always broken or only in certain cases/OS's.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1669,13 +1623,29 @@ export class CppProperties {
...
+            return;
+        }
+        fs.stat(compileCommandsFile, (err, stats) => {
+            if (err) {
+                if (err.code === "ENOENT" && this.compileCommandsFile) {
+                    this.onCompileCommandsChanged(compileCommandsFile);
+                    this.compileCommandsFile = null; // File deleted
+                }
+            } else if (stats.mtime > this.compileCommandsFileWatcherFallbackTime) {
+                this.compileCommandsFileWatcherFallbackTime = new Date();
```
sean-mcmanus, line 1688: I think this should be moved into this.onCompileCommandsChanged, which matches what is done with configFileWatcherFallbackTime .

Extension/src/LanguageServer/configurations.ts
Colengms: It looks like we don't currently use FileWatchers at all for detecting changes to `c_cpp_properties.json`.  Since that is so similar a scenario, we should probably handle both the same way.

Do we know why FileWatchers sometimes fail?  Perhaps they only fail for files outside of the workspace, or if too many are created?  I think we either need to understand in what circumstances they fail and work around them, or we should avoid them entirely.  The only reason I can think of to keep FileWatchers as well as onInternal polling, would be if 2.5 seconds wasn't responsive enough, and we wanted to handle something ASAP if we happened to get a notification from a FileWatcher.

We still have a rootPathFileWatcher, which watches the entire workspace.  That looks to be where our file create/change/delete events are coming from for files that aren't currently open.  (We use those and seem to get them reliably, but I'm not sure if there would be a noticeable issue if they went missing).  ".editorconfig" files are checked there, instead of creating their own file watcher.  Any file within the workspace should already be reported via the root watcher, so we would only need additional file watchers for files outside of the workspace.  c_cpp_propeties.json is within the workspace.  Perhaps we should (also) be checking for changes to it via the rootPathFileWatcher?

sean-mcmanus: What? No, I think we do use file watchers for detecting changes to c_cpp_properties.json -- the configFileWatcher-related code.

There are some known cases where file watchers can fail (such as hitting a Linux limit), but I don't know why they were failing for my original repro. I could try again and see if they still fail or not.

The polling case is supposed to be a fallback (which is why the variables have fallback in the names).

The didcreate/change/delete cases are for updating IntelliSense stuff and not configurations...the callbacks are different.

Colengms: Ah, I was looking only at checkCppProperties().  It looks like handleConfigurationChange() is also called by configFileWatcher.

What I was saying was that configFileWatcher is not technically needed.  If we need a FileWatcher for the whole workspace anyway (rootPathFileWatcher), everything that needs to monitor files in the workspace could potentially leverage a shared instance of vscode.FileWatcher instead of creating new ones.  (In case creating new watchers is an issue).

It looks like in order to watch files outside of the workspace, we can't use vscode.FileWatcher.  We used fs.FSWatcher, only for compile_commands files.  Do we expect people are using compile_commands files outside of their workspaces?  Perhaps there are issues specific to that FSWatcher API?

I would also be in favor of keeping the FSWatcher's for compile commands, to avoid a potential lag of up to 2.5 seconds.  The ideal would be a reliable API, or at least a way to detect if the API is failing in order to only incur FS overhead every 2.5 seconds when absolutely necessary.

elahehrashedi: I kept the file watcher code.

sean-mcmanus (approved):

Issue #2667, opened 2018-10-14
"Switch header/source" won't switch to an existing file in another group if it's not visible


akb825:
**Type: LanguageService**

**Describe the bug**
When you use the "switch header/source" feature, and the file it would switch to is open in a separate group (but not visible), it will open a new tab instead of switching to the existing one.

- OS and Version: Arch Linux
- VS Code Version: 1.28.1
- C/C++ Extension Version: 0.19
- Other extensions you installed (and if the issue persists after disabling them): CMake

**To Reproduce**

1. Open a .cpp file.
2. Open the .h file corresponding to the .cpp file in a separate group.
3. Open a separate file in the second group and make sure that's the currently visible file.
4. Go back to the .cpp file and select "Switch header/source"

**Expected behavior**

The existing .h file open in the second group is displayed instead of opening a new tab.

**Screenshots**

![separate tabs](https://user-images.githubusercontent.com/19942638/46922616-6af65880-cfc0-11e8-915f-26175f1b11ca.png)

**Additional context**

It was mentioned [in a separate bug](https://github.com/Microsoft/vscode-cpptools/issues/444) that this is dependent on an additional API for VS code. However, this was closed almost 2 years ago without any indication of when (or if) this will be resolved. Some bug for this issue should remain open as long as it's present.

Colengms:
This is currently blocked on a VS Code issue: https://github.com/microsoft/vscode/issues/110537

I looked for an existing issue against VS Code that was tracking this and could not find one, so created a new one.

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/1.2.0-insiders2 , but you need to change the setting workbench.editor.revealIfOpen to true.

PR #6751, opened 2021-01-07
Fix SwitchHeaderSource with RevealIfOpen

### LINKED ISSUE #6749: "workbench.editor.revealIfOpen" option does not work for command "C/C++: Switch Header/Source"
VSCode Version: 1.52.1
C++ extension version: v1.1.3
Commit: ea3859d
Date: 2020-12-16T16:30:02.420Z (2 wks ago)
Electron: 9.3.5
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Darwin x64 19.6.0

Steps to Reproduce:

Use the C++ extension
HAve two different editor panels
Open a header on one panel, and the implementation (cc) on the other
Set the above option in the user settings.json to true (workbench.editor.revealIfOpen) (possibly restart?)
Use the command "C/C++: Switch Header/Source"
Expected:

The focus should shift to the other editor panel

Actual:

A new tab will open in the active panel, with the corresponding header / implementation file, duplicating the one already open in the other panel.

Does this issue occur when all extensions are disabled?: Yes/No

Unknown

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/6749 .

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -806,14 +806,9 @@ function onSwitchHeaderSource(): void {
                     vscode.window.showTextDocument(document, editor.viewColumn);
                 }
             });
-            // TODO: Handle non-visibleTextEditor...not sure how yet.
+
             if (!foundEditor) {
-                if (vscode.window.activeTextEditor !== undefined) {
-                    // TODO: Change to show it in a different column?
-                    vscode.window.showTextDocument(document, vscode.window.activeTextEditor.viewColumn);
-                } else {
-                    vscode.window.showTextDocument(document);
-                }
+                vscode.window.showTextDocument(document);
             }
         });
     });


Extension/src/LanguageServer/extension.ts
```diff
@@ -806,14 +806,9 @@ function onSwitchHeaderSource(): void {
                     vscode.window.showTextDocument(document, editor.viewColumn);
                 }
             });
-            // TODO: Handle non-visibleTextEditor...not sure how yet.
+
             if (!foundEditor) {
-                if (vscode.window.activeTextEditor !== undefined) {
-                    // TODO: Change to show it in a different column?
-                    vscode.window.showTextDocument(document, vscode.window.activeTextEditor.viewColumn);
```
Colengms, line 813: Wasn't this logic intended to address a previous bug here?
Does this fix also address? :  https://github.com/microsoft/vscode-cpptools/issues/2667

sean-mcmanus: Yeah -- it was intended to fix the bug that VS Code fixed via adding the revealIfOpen setting. This method leaves the behavior up to VS and the user setting instead of forcing it to use the activeTextEditor.viewColumn.

Yes, this fixes https://github.com/microsoft/vscode-cpptools/issues/2667, i.e. they're duplicate issues. I couldn't locate that.

sean-mcmanus: It looks like VS Code added revealIfOpen as a mitigation for https://github.com/microsoft/vscode/issues/15178, which is the request to add a more powerful "open editors" API.

Colengms (approved):

PR #6752, opened 2021-01-07
show a Squiggle if there is a duplicate configuration name

### LINKED ISSUE #3412: Show a c_cpp_properties.json squiggle when a config is used with the same name as another config
If you have 2 configs with the same name you select the wrong one it can be really confusing, particularly if the duplicate one isn't visible, and you select the non-visible one in the drop down by mistake.

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/3412
update:
image
and:
image


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1343,9 +1356,28 @@ export class CppProperties {
             }
             curTextStartOffset = configStart + 1;
             curText = curText.substr(curTextStartOffset); // Remove earlier configs.
-            const nameEnd: number = curText.indexOf(":");
+            let nameEnd: number = curText.indexOf(":");
             curTextStartOffset += nameEnd + 1;
             curText = curText.substr(nameEnd + 1);
+            // Check if the conig name is unique, before removing later configs
```
Colengms: spelling: "config"

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1343,9 +1356,28 @@ export class CppProperties {
...
+            let tailText: string = curText;
+            let tailTextStartOffset: number = curTextStartOffset;
+            let dupliacteNameStart: number = tailText.search(new RegExp(`{\\s*"name"\\s*:\\s*"${escapeStringRegExp(currentConfiguration.name)}"`));
```
Colengms: spelling: duplicateNameStart

Colengms: It looks like this RegExp is created twice.  Maybe create just one to reference from both places?

Colengms (changes requested):

Colengms:
This approach will only show squiggles for other configurations that have the same name as the currently selected configuration.  It would be possible to flag all duplicate names.  The `configurations` variable in `handleSquiggles()` contains the entire set of configurations as a JSON object.  You could use that to determine all duplicates.  You could then scan through the text looking for `name` fields, similar to what you are doing now, but just to find the right offsets to apply the squiggles at.

sean-mcmanus:
> This approach will only show squiggles for other configurations that have the same name as the currently selected configuration. It would be possible to flag all duplicate names. The `configurations` variable in `handleSquiggles()` contains the entire set of configurations as a JSON object. You could use that to determine all duplicates. You could then scan through the text looking for `name` fields, similar to what you are doing now, but just to find the right offsets to apply the squiggles at.

The current squiggle design is that we only squiggle stuff that is applicable to the currently active configuration, so I don't think we should add squiggles for duplicates config names for inactive configurations. The design philosophy is to not over-squiggle for stuff that is not currently relevant to the user.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1294,6 +1298,15 @@ export class CppProperties {
         return errorMsg;
     }
 
+    private isConfigNameUnique(configName: string): string | undefined {
+        let errorMsg: string | undefined;
+        const occurances: number | undefined = this.ConfigurationNames?.filter(function (name): boolean { return name === configName; }).length;
```
sean-mcmanus: occurances -> occurrences

sean-mcmanus (approved):

Colengms:
> The current squiggle design is that we only squiggle stuff that is applicable to the currently active configuration, so I don't think we should add squiggles for duplicates config names for inactive configurations. The design philosophy is to not over-squiggle for stuff that is not currently relevant to the user.

That makes sense for most values in the configuration, as we shouldn't be squiggling configs that may be intended for a different system (due to a checked-in `c_cpp_properties.json`), for example.  However, I think that does not apply to squiggles due to duplicate config names.  The issues encountered due to duplicate names transcend the context of the current configuration.  i.e. The list of configs to select between.  This feature is already introducing squiggles in config sections other than the active one.  I think it would make more sense (to users) to squiggle all duplicate names.

sean-mcmanus:
> That makes sense for most values in the configuration, as we shouldn't be squiggling configs that may be intended for a different system (due to a checked-in `c_cpp_properties.json`), for example. However, I think that does not apply to squiggles due to duplicate config names. The issues encountered due to duplicate names transcend the context of the current configuration. i.e. The list of configs to select between. This feature is already introducing squiggles in config sections other than the active one. I think it would make more sense (to users) to squiggle all duplicate names.

Ah, okay, that makes sense -- I forgot about the duplicate names in the config list.

sean-mcmanus (changes requested):
Can you change it to squiggle all the duplicates as Colen suggested?

sean-mcmanus (approved):

sean-mcmanus:
> Ah, okay, that makes sense -- I forgot about the duplicate names in the config list.

Actually, after thinking about it more, I changed my mind. The duplicates of inactive configs don't really matter because they're inactive, the squiggle of the duplicate is important for the active config because it warns the user they might have the wrong config selected, but if I'm on Windows with a "Win32" config then I don't really care to be squiggled from having 2 "Mac" configs, and the fact that the switch configuration dropdown will have 2 Mac configs doesn't matter, because the user may never switch configs, and as soon as they switch configs it to the duplicate one it should squiggle.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1146,6 +1147,9 @@ export class CppProperties {
         const isWindows: boolean = os.platform() === 'win32';
         const config: Configuration = this.configurationJson.configurations[configIndex];
 
+        // Check if config name is unique.
+        errors.name = this.isConfigNameUnique(config.name);
```
elahehrashedi, line 1150: @sean-mcmanus @Colengms I am checking for all configs that are duplicated here, active ones and inactive ones.

elahehrashedi: I realized something:
image
I send an error regarding duplicate names here, but the web view doesn't show the error.

michelleangela: Note: the errors on the UI config page are for the current _selected_ configuration. The errors are **not** from the _active_ configuration like on the JSON config page.

A new html element will need to be added to show/hide errors for configuration name. I think the error would be shown when "add configuration" is done or when the current selected configuration (not active one) has a duplicate name.

elahehrashedi: image
fixed now

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1343,9 +1356,29 @@ export class CppProperties {
             }
             curTextStartOffset = configStart + 1;
             curText = curText.substr(curTextStartOffset); // Remove earlier configs.
-            const nameEnd: number = curText.indexOf(":");
+            let nameEnd: number = curText.indexOf(":");
             curTextStartOffset += nameEnd + 1;
             curText = curText.substr(nameEnd + 1);
+            // Check if the config name is unique, before removing later configs
+            const nameRegex: RegExp = new RegExp(`{\\s*"name"\\s*:\\s*"${escapeStringRegExp(currentConfiguration.name)}"`);
```
Colengms: I had commented without starting a 'review', so I can't check if my topic was considered 'resolved' or not.  Opening this to track, just in case.  :)

This is where the `currentConfiguration` was used to detect duplicates of only the current config's name.  The local `configurations` var contains the text as parsed JSON object(s), and could be used to detect all duplicate names.  Something similar to the code below would still be needed to determine the exact positions to put squiggles.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
elahehrashedi: image
this is the latest update

Colengms: Shouldn't we also squiggle the first duplicate?  If the other duplicates happen to be scrolled out of view when the user changes the name of an earlier config, they may not notice.

Colengms: For example, when there are duplicate field names in a JSON object, in a .json file, VS Code will squiggle all duplicates names.

Extension/ui/settings.html
```diff
@@ -455,7 +455,7 @@
             <tr>
             <!-- select configuration name -->
             <td>
-                <div class="select-editable" style="width: 300px; margin-right: 10px">
+                <div class="select-editable" style="width: 300px; margin-left: -3px; margin-right: 7px">
```
michelleangela, line 458: `margin-left: -3px; margin-right: 7px` is this still aligned with the other selection boxes on the UI page?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1337,8 +1350,39 @@ export class CppProperties {
             envText = curText.substr(envStart, envEnd);
             const envTextStartOffSet: number = envStart + 1;
 
+            // Check if all config names are unique.
```
michelleangela, line 1350: Since `handleSquiggles()` is getting long, could the code that does `// Check if all config names are unique.` be factored out to a different function just for checking config names?

Extension/ui/settings.html
elahehrashedi: yes, there is a screenshot. Let me know

Extension/src/LanguageServer/configurations.ts
michelleangela: Don't need to refactor for insiders. if you have time, you can refactor later

michelleangela (approved):

Colengms (approved):

PR #6777, opened 2021-01-14
Merge latest changes for 1.2.0-insiders

sean-mcmanus:


sean-mcmanus:
My previous PR inadvertently merged main into insiders already, which is why this PR only has the last few changes.

PR #6781, opened 2021-01-14
Fix for downgrade from 1.2.0-insiders.

sean-mcmanus:


Extension/src/githubAPI.ts
```diff
@@ -96,10 +96,12 @@ function getArrayOfBuilds(input: any): Build[] {
     // Only return the the most recent release and insider builds.
     for (let i: number = 0; i < input.length; i++) {
         if (isBuild(input[i])) {
-            builds.push(input[i]);
-            // the latest "valid" released build
-            if (input[i].name.indexOf('-') === -1 && isValidBuild(input[i])) {
-                break;
+            // Only return "valid" builds, i.e. ignore builds without assets.
+            if (isValidBuild(input[i])) {
```
bobbrow, line 100: I don't think this is correct.  We need all the builds, including the invalid ones if our rollback feature is going to work.  Let's just not do this part of the change right now.  We'll delete the 1.2.0-preview release soon enough.

bobbrow (changes requested):

Extension/src/githubAPI.ts
sean-mcmanus: Ah, okay, so the releases without assets are treated as a rollback, that makes sense now.

PR #6784, opened 2021-01-15
Fix issue with multiple didOpen

Colengms:
Added a check to didOpen in the protocolFilter to avoid handling didOpen for a file that is already being tracked, since it's possible that we may have received onDidChangeVisibleTextEditors and considered this a 'delayedOpen' prior to receiving a didOpen.

Fixed placement of telemetry tracking didOpen.


sean-mcmanus:
Can you explain this change? It looks like " if (!me.TrackedDocuments.has(document)) {" was added -- when is that false?

sean-mcmanus (approved):

Issue #6787, opened 2021-01-15
1.2.0-insiders crashes with macOS 10.13 (High Sierra) or older (in "uv_run")


sean-mcmanus:
If you're hitting that, change the C_Cpp.updateChannel setting to "Default" until we're able to release a fix, although it's possible we could drop support for those older OS's for 1.2.0? (preferably not since 10.13 is still supported by Apple).

My guess is that it doesn't work at all due to some missing dynamic library dependency with the libuv library we started using. Let us know if you know of a workaround.

NOTE: I changed references to 17.7 (the Darwin/Platform version) to 10.13 (the OS version number).

sean-mcmanus:
I've removed the Mac binaries from 1.2.0-insiders until we've disable Insiders for those versions (https://github.com/microsoft/vscode-cpptools/pull/6788) or have a workaround.

PR #6788, opened 2021-01-15
Disable Insiders update for Mac versions older than 18

sean-mcmanus:
Due to issue https://github.com/microsoft/vscode-cpptools/issues/6787

If 1.1.4 is published, it looks like we'd also need to publish a 1.2.0-insiders2 (or republish 1.2.0-insiders?) afterwards in order for the Insider update mechanism to work correctly since it appears to stop once it encounters the latest non-Insiders.

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -1,5 +1,9 @@
 # C/C++ for Visual Studio Code Change Log
 
+## Version 1.1.4: Janurary 15, 2021
+### Bug Fix
+* Disable the Insiders update channel for Mac versions less than 18 due to bug [#6787](https://github.com/microsoft/vscode-cpptools/issues/6787).
+
 ## Version 1.1.3: December 3, 2020
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


michelleangela (approved):

Extension/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # C/C++ for Visual Studio Code Change Log
 
+## Version 1.1.4: Janurary 15, 2021
+### Bug Fix
+* Disable the Insiders update channel for Mac versions less than 18 due to bug [#6787](https://github.com/microsoft/vscode-cpptools/issues/6787).
```
bobbrow, line 5: We would probably use the 10.x version string here, right?  But I'm not sure that we need to ship this yet.

bobbrow (changes requested):

Extension/CHANGELOG.md
sean-mcmanus: Yeah, good catch. Keep getting the darwin versus regular versions confused.

sean-mcmanus: ...maybe turn this disabling on/off via a cpptools.json option...

PR #6794, opened 2021-01-15
Add new "console" launch config for cppvsdbg

WardenGnaw:
This PR adds new console support for the cppvsdbg debug type.

This removes the legacy 'externalConsole' flag with 'console'.
Console has four enums:
- internalConsole: Output to the VS Code Debug Console. This doesn't
  support reading console input (ex:Console.ReadLine)
- integratedTerminal: VS Code's integrated terminal
- externalTerminal: External terminal that can be configured via user
  settings
- newExternalWindow: Console applications will be launched in their own
  external console window which will end when the application stops.
  Non-console applications will run without a terminal, and
  stdout/stderr will be ignored

@gregg-miskelly


Extension/package.json
```diff
@@ -1984,10 +1984,22 @@
                 "description": "%c_cpp.debuggers.cppvsdbg.visualizerFile.description%",
                 "default": ""
               },
-              "externalConsole": {
```
gregg-miskelly: >externalConsole [](start = 15, length = 15)

Do you want to leave this in and just update the description to indicate that it is deprecated? Otherwise VS Code will yell at any user that has the previous default template.

gregg-miskelly (approved):
Otherwise LGTM

Extension/package.json
WardenGnaw: Yep. Looks like we leave in deprecated keys. I'll bring this back. 
E.g. `targetArchitecture`

Extension/package.json
```diff
@@ -1984,10 +1984,22 @@
...
+                  "newExternalWindow"
+                ],
+                "enumDescriptions": [
+                  "%c_cpp.debuggers.cppvsdbg.console.internalConsole.description%",
+                  "%c_cpp.debuggers.cppvsdbg.console.integratedTerminal.description%",
+                  "%c_cpp.debuggers.cppvsdbg.console.externalTerminal.description%",
+                  "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
+                ],
+                "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
+                "default": "internalConsole"
```
sean-mcmanus: Is this the best default? It seems like a lot of users hit the cin not working issue. Would another default generate less issues?

Extension/package.nls.json
```diff
@@ -213,7 +213,11 @@
     "c_cpp.debuggers.serverLaunchTimeout.description": "Optional time, in milliseconds, for the debugger to wait for the debugServer to start up. Default is 10000.",
     "c_cpp.debuggers.coreDumpPath.description": "Optional full path to a core dump file for the specified program. Defaults to null.",
     "c_cpp.debuggers.cppdbg.externalConsole.description": "If true, a console is launched for the debuggee. If false, on Linux and Windows, it will appear in the Integrated Console.",
-    "c_cpp.debuggers.cppvsdbg.externalConsole.description": "If true, a console is launched for the debuggee. If false, no console is launched.",
+    "c_cpp.debuggers.cppvsdbg.console.description": "Where to launch the debug target.",
+    "c_cpp.debuggers.cppvsdbg.console.internalConsole.description": "Output to the VS Code Debug Console. This doesn't support reading console input (ex:Console.ReadLine)",
```
sean-mcmanus: Isn't Console.Readline a .NET API? Should we reference something like std::cin or scanf?

sean-mcmanus (approved):

Extension/package.json
WardenGnaw: If not defined, this is the behavior vsdbg uses.  

"default" here is just what is autopropogated if the users tab in `launch.json`

Extension/package.nls.json
WardenGnaw: Good catch. Thanks!

sean-mcmanus (approved):

Extension/package.json
gregg-miskelly: To be clear, this is vsdbg's default, but the configuration provider could pick a different default for templates if we think that is better in these scenarios. If I recall correctly, we do use an external console by default in GDB cases? If yes, it might make sense to do the same here. The only downside is that VS Code doesn't switch focus like VS does when a breakpoint is hit. So it can be a bit confusing.

Extension/package.nls.json
```diff
@@ -213,7 +213,11 @@
...
+    "c_cpp.debuggers.cppvsdbg.console.integratedTerminal.description": "VS Code's integrated terminal",
+    "c_cpp.debuggers.cppvsdbg.console.externalTerminal.description": "External terminal that can be configured via user settings",
```
gregg-miskelly: The description for externalTerminal is wrong - we will always just launch VsDebugConsole.exe as a new top level window using CreateProcess.

gregg-miskelly: I would suggest: Console applications will be launched in an external terminal window. The window will be reused in relaunch scenarios, and will not automatically disappear when the application exits.

---
In reply to: [558714356](https://github.com/microsoft/vscode-cpptools/pull/6794#discussion_r558714356) [](ancestors = 558714356)

gregg-miskelly (approved):

PR #6823, opened 2021-01-20
Use cpptools API v5

michelleangela:


Colengms (changes requested):
I had made some changes to the API, to these fields (to make them optional. https://github.com/microsoft/vscode-cpptools-api/pull/28 ) :

	SourceFileConfiguration 

		intelliSenseMode
		standard

	WorkspaceBrowseConfiguration 

		standard

They now accept undefined.  But, we were waiting until we ingested the new API to make the corresponding changes to cpptools.  Could you update `isSourceFileConfigurationItem()` to use `isOptionalString()` instead of `isString()` when validating?  Perhaps consider it an error if either of these are undefined when the `compilerPath` is also undefined.  We need `compilePath` to be present to auto-detect these.

Similarly for `WorkspaceBrowseConfiguration` in `sendCustomBrowseConfiguration().` That is already using `isOptionalString` for `standard`, but should probably flag an error if it's undefined and `compilerPath` is also undefined.

Extension/src/LanguageServer/client.ts
```diff
@@ -2364,19 +2364,23 @@ export class DefaultClient implements Client {
         this.notifyWhenReady(() => this.languageClient.sendNotification(ChangeCompileCommandsNotification, params));
     }
 
-    private isSourceFileConfigurationItem(input: any): input is SourceFileConfigurationItem {
+    private isSourceFileConfigurationItem(input: any, providerVersion: Version): input is SourceFileConfigurationItem {
+        const intelliSenseModeValid = providerVersion > Version.v4 ?
+            util.isOptionalString(input.configuration.intelliSenseMode) : util.isString(input.configuration.intelliSenseMode)
```
Colengms: It would probably be sufficient to use isOptionalString in both cases.  The prior API did not allow undefined, but it wouldn't hurt to handle it anyway.

michelleangela:
> Is there an issue tracking getting CMake-Tools to use the v5 API?

I made an issue for it https://github.com/microsoft/vscode-cmake-tools/issues/1624

Colengms (approved):

Issue #6724, opened 2020-12-29
1.2.0-preview repeated crash messages can occur


sean-mcmanus:
I reproed it with https://github.com/microsoft/vscode-cpptools/issues/6698, but then I wasn't able to get it to repro again to investigate a cause. It might be some race condition with the update retry code. It didn't seem like it was actually crashing/retrying repeated, so it was just a bug with the output and updating code.

sean-mcmanus:
This appears to be happening with 1.2.0-insiders2 -- does anyone have any repro info?

PR #6851, opened 2021-01-26
Add crash count telemetry.

sean-mcmanus:
To differentiate between 1 time crashes (on shutdown, temporary code) and repeated crashes that don't go away. Also, enables us to tell fix what the max crash count is, i.e. a fast repeated crash should hit the 10 limit in 3 minutes, while periodic crashing could generate > 10 over a longer session, or if there's a bug (like https://github.com/microsoft/vscode-cpptools/issues/6724).

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1106,10 +1106,10 @@ function reportMacCrashes(): void {
                         fs.readFile(path.resolve(crashFolder, filename), 'utf8', (err, data) => {
                             if (err) {
                                 // Try again?
-                                fs.readFile(path.resolve(crashFolder, filename), 'utf8', handleCrashFileRead);
+                                fs.readFile(path.resolve(crashFolder, filename), 'utf8', handleMacCrashFileRead);
                                 return;
                             }
-                            handleCrashFileRead(err, data);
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -1120,10 +1120,16 @@ function reportMacCrashes(): void {
     }
 }
 
+let previousMacCrashData: string;
+let previousMacCrashCount: number = 0;
 function logCrashTelemetry(data: string): void {
```
michelleangela: is `logCrashTelemetry` only for Mac or will it also get used on other platforms? The new variables `previousMacCrashData` and `previousMacCrashCount` and telemetry name "MacCrash" seem to suggest that `logCrashTelemetry` is only being executed on Mac?

sean-mcmanus: Yes, it's just for Mac -- logCrashTelemetry method should probably be renamed logMacCrashTelemetry.

sean-mcmanus: I renamed the methods -- the code should probably be moved to it's own file, but that could be left for later.

michelleangela (approved):

PR #6929, opened 2021-02-08
Clarify legacy IntelliSense modes in description

michelleangela:
Clarify that the `<compiler>-<architecture>` IntelliSense mode variants (e.g. `gcc-x64`) are legacy modes.


Extension/ui/settings.html
```diff
@@ -538,18 +538,18 @@
...
-            <option value="clang-arm64">clang-arm64</option>
-            <option value="gcc-x86">gcc-x86</option>
-            <option value="gcc-x64">gcc-x64</option>
-            <option value="gcc-arm">gcc-arm</option>
-            <option value="gcc-arm64">gcc-arm64</option>
-            <option value="msvc-x86">msvc-x86</option>
-            <option value="msvc-x64">msvc-x64</option>
-            <option value="msvc-arm">msvc-arm</option>
-            <option value="msvc-arm64">msvc-arm64</option>
+            <option value="clang-x86">clang-x86 (legacy)</option>
```
sean-mcmanus, line 541: Would "deprecated" be more understandable or is that not the correct term to use?

sean-mcmanus (approved):

Extension/ui/settings.html
michelleangela: "deprecated" usually means that something will be removed in the future. "legacy" means something is old and is being kept for backward compatibility.

PR #6952, opened 2021-02-11
Update codespaces check to also include VS Code remoting codespaces

Colengms:
Using an expected environment variable to check for CODESPACES.  Alternatively, we could use a combination of UIKind.Web, and the remote name == "codespaces" when remoted.  But, this approach would allow us to set the env variable to test locally.


sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -1273,3 +1273,7 @@ export function getUniqueWorkspaceName(workspaceFolder: vscode.WorkspaceFolder):
 export function getUniqueWorkspaceStorageName(workspaceFolder: vscode.WorkspaceFolder): string {
     return getUniqueWorkspaceNameHelper(workspaceFolder, true);
 }
+
+export function is_codespaces(): boolean {
```
bobbrow: Can you remove the underscore? (naming conventions)

Issue #6981, opened 2021-02-17
Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: TypeError: Cannot read property 'setDidOpenTime' of undefined


gunarkroeger:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10
- VS Code Version: 1.53.2
- C/C++ Extension Version:
- Other extensions you installed (and if the issue persists after disabling them):
- Does this issue involve using SSH remote to run the extension on a remote machine?:
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).

**Steps to reproduce**

**Expected behavior**

<details>
  <summary><strong>Logs</strong></summary>

```
![Screenshot 2021-02-17 162812](https://user-images.githubusercontent.com/22892870/108235343-a4fe3900-713d-11eb-8c3a-3931857483d3.png)

[Error - 16:25:25] Server initialization failed.
TypeError: Cannot read property 'setDidOpenTime' of undefined
	at Object.didOpen (c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:60328:52)
	at c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:8696:32
	at Array.forEach (<anonymous>)
	at DidOpenTextDocumentFeature.register (c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:8685:42)
	at DidOpenTextDocumentFeature.initialize (c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:8676:18)
	at LanguageClient.initializeFeatures (c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:10684:21)
	at c:\Users\gunar\.vscode\extensions\ms-vscode.cpptools-1.2.1\dist\main.js:10389:18
```
</details>
I opened a cpp project today and was greeted with these errors. Intellisense not working now. I don't know what caused it.
**Screenshots**

![Screenshot 2021-02-17 162812](https://user-images.githubusercontent.com/22892870/108235674-f5759680-713d-11eb-8113-755a50f8b224.png)

**Additional context**

gunarkroeger:
Also, building and uploading is not working with the Platform.io extension anymore...

[...]

PR #6984, opened 2021-02-17
Move initialization of TimeTelemetryCollector

Colengms:
Might address: https://github.com/microsoft/vscode-cpptools/issues/6981

Since the repro is dereferencing 'clients', it would have to be after the ClientCollection was constructed (or else that would be undefined).  In that case, timeTelemetryCollector would already have been initialized.  So, I have no confidence that moving the initialization of timeTelemetryCollector will make any difference.


sean-mcmanus:
Can you update the changelog for 1.2.2-insiders in this PR? Or do you want me to create a new PR for that?

sean-mcmanus (approved):

PR #7049, opened 2021-02-25
Fix launch.json squiggle for "console".

### LINKED ISSUE #7048: "console" property in launch.json for cppdbg squiggles with "Property console is not allowed."
...when generating a cppbuild task.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/7048

### DIFF
--- a/Extension/tools/OptionsSchema.json
+++ b/Extension/tools/OptionsSchema.json
@@ -329,6 +329,23 @@
           "description": "%c_cpp.debuggers.cppdbg.externalConsole.description%",
           "default": false
         },
+        "console": {
+          "type": "string",
+          "enum": [
+            "internalConsole",
+            "integratedTerminal",
+            "externalTerminal",
+            "newExternalWindow"
+          ],
+          "enumDescriptions": [
+            "%c_cpp.debuggers.cppvsdbg.console.internalConsole.description%",
+            "%c_cpp.debuggers.cppvsdbg.console.integratedTerminal.description%",
+            "%c_cpp.debuggers.cppvsdbg.console.externalTerminal.description%",
+            "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
+          ],
+          "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
+          "default": "externalTerminal"
+        },
         "avoidWindowsConsoleRedirection": {
           "type": "boolean",
           "description": "%c_cpp.debuggers.avoidWindowsConsoleRedirection.description%",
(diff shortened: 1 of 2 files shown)


WardenGnaw (changes requested):
`cppdbg` does not use console.

sean-mcmanus:
> `cppdbg` does not use console.

Ah, okay, then there's a different bug.

Extension/tools/OptionsSchema.json
```diff
@@ -561,7 +578,7 @@
             "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
           ],
           "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
-          "default": "internalConsole"
+          "default": "externalTerminal"
```
WardenGnaw, line 581: This is going to mismatch when the user does not provide `console`. vsdbg will default to `internalConsole`.

sean-mcmanus: @WardenGnaw Why is package.json "default" set to "externalTerminal" then? Isn't that a mismatch?

sean-mcmanus: When I run generateOptionsSchema this gets updated.

WardenGnaw:
The error is:
https://github.com/microsoft/vscode-cpptools/blob/2199ec324d0074253522695863314d52621d83ec/Extension/src/Debugger/configurations.ts#L47

it should be
```
type === "cppdbg ? "externalConsole" : false : "console" : "externalTerminal"
```

sean-mcmanus:
> The error is:
> https://github.com/microsoft/vscode-cpptools/blob/2199ec324d0074253522695863314d52621d83ec/Extension/src/Debugger/configurations.ts#L47
> 
> it should be
> 
> ```
> type === "cppdbg ? "externalConsole" : false : "console" : "externalTerminal"
> ```

Yeah, but something is adding `"console": "externalTerminal"` to the end of the configuration...not sure where that's coming from.

WardenGnaw:
The issue is that `createLaunchString` is used for both `cppvsdbg` and `cppdbg`. It will need to switch based on which configuration it is generating for. 

The snippet above should fix that issue.

sean-mcmanus:
> The issue is that `createLaunchString` is used for both `cppvsdbg` and `cppdbg`. It will need to switch based on which configuration it is generating for.
> 
> The snippet above should fix that issue.

There's another spot in the code -- I found it. I'll submit a 2nd PR soon.

PR #7051, opened 2021-02-25
Fix "console" squiggle with "cppdbg".

### LINKED ISSUE #7048: "console" property in launch.json for cppdbg squiggles with "Property console is not allowed."

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/7048
Follow up to https://github.com/microsoft/vscode-cpptools/pull/6794

### DIFF
--- a/Extension/tools/OptionsSchema.json
+++ b/Extension/tools/OptionsSchema.json
@@ -561,7 +561,7 @@
             "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
           ],
           "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
-          "default": "externalTerminal"
+          "default": "internalConsole"
         },
         "sourceFileMap": {
           "type": "object",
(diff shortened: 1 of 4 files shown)


Extension/tools/OptionsSchema.json
```diff
@@ -561,7 +561,7 @@
             "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
           ],
           "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
-          "default": "internalConsole"
+          "default": "externalTerminal"
```
sean-mcmanus: Is this correct?

WardenGnaw: No. I believe `default` should indicate the value it will be if it is not set. 

So the default value of console is `internalConsole`.

sean-mcmanus: but package.json has it as "internalConsole" -- you mean the value in package.json is incorrect?

WardenGnaw: `internalConsole` is correct and `package.json` properties for the debugger are generated from `OptionsSchema.json`. So the value before this change is correct.

WardenGnaw (approved):

PR #7109, opened 2021-03-05
Add autocomplete.addParenthesesAfterFunctionCalls

### LINKED ISSUE #882: Auto adding parentheses for function calls
During auto-completion for methods, brackets do not get inserted automatically. Most of the time they are required and it's easier to delete them than insert them manually. Calling methods is common than passing them as arguments.

Eg: Tab or Return after "size_t n = name.len" should autocomplete to "size_t n = name.length()" rather than "size_t n = name.length" with the cursor between/after (according to whether there is function argument)  the brackets.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/882

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1158,10 +1158,10 @@
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
         },
-        "C_Cpp.autocompleteAddParenthesesAfterFunctionCalls": {
+        "C_Cpp.autocompleteAddParentheses": {
           "type": "boolean",
           "default": true,
-          "description": "%c_cpp.configuration.autocompleteAddParenthesesAfterFunctionCalls.description%",
+          "description": "%c_cpp.configuration.autocompleteAddParentheses.description%",
           "scope": "resource"
         }
       }
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1158,10 +1158,10 @@
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
         },
-        "C_Cpp.autocomplete.addParenthesesAfterFunctionCalls": {
+        "C_Cpp.autocompleteAddParenthesesAfterFunctionCalls": {
           "type": "boolean",
           "default": true,
-          "description": "%c_cpp.configuration.autocomplete.addParenthesesAfterFunctionCalls.description%",
+          "description": "%c_cpp.configuration.autocompleteAddParenthesesAfterFunctionCalls.description%",
           "scope": "resource"
         }
       }
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -1157,6 +1157,12 @@
           "default": "Enabled",
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
+        },
+        "C_Cpp.autocomplete.addParenthesesAfterFunctionCalls": {
+          "type": "boolean",
+          "default": true,
+          "description": "%c_cpp.configuration.autocomplete.addParenthesesAfterFunctionCalls.description%",
+          "scope": "resource"
         }
       }
     },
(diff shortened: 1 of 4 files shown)


Extension/package.nls.json
```diff
@@ -162,6 +162,7 @@
     "c_cpp.configuration.vcpkg.enabled.markdownDescription": "Enable integration services for the [vcpkg dependency manager](https://aka.ms/vcpkg/).",
     "c_cpp.configuration.addNodeAddonIncludePaths.description": "Add include paths from nan and node-addon-api when they're dependencies.",
     "c_cpp.configuration.renameRequiresIdentifier.description": "If true, 'Rename Symbol' will require a valid C/C++ identifier.",
+    "c_cpp.configuration.autocomplete.addParenthesesAfterFunctionCalls.description": "If true, autocomplete will automatically add ( after function calls, in which case ) may also be added, depending on editor.autoClosingBrackets.",
```
bobbrow: suggestion: "...depending on the value of the editor.autoClosingBrackets setting"

Extension/package.json
```diff
@@ -1157,6 +1157,12 @@
           "default": "Enabled",
           "description": "%c_cpp.configuration.codeFolding.description%",
           "scope": "window"
+        },
+        "C_Cpp.autocompleteAddParenthesesAfterFunctionCalls": {
```
bobbrow: This is a REALLY long setting name.  Can we remove "autocomplete" from the name?  I think it's implied and will at least be described in the description.

Extension/package.nls.json
sean-mcmanus: Yeah, I was thinking the same thing. I wasn't sure which phrasing was better.

Extension/package.json
sean-mcmanus: The longer name matches what other languages use and the UI separates the long text into words, so I'd prefer to keep "autocomplete", but we can see what Julia/others think:
image

bobbrow: Python just calls it "addBrackets" (not saying it's better, but it is shorter)
image

I think I would be happy with keeping `autocomplete` in the name if we made it shorter like `autocompleteAddParentheses`.  The description is the better place to add additional detail.  We had similar problems with naming the vcFormatting settings and we just have to rely on the description.

sean-mcmanus: Yeah, I like "autoCompleteAddParentheses".

Extension/src/LanguageServer/client.ts
```diff
@@ -1292,8 +1300,11 @@ export class DefaultClient implements Client {
                     },
                     space:  vscode.workspace.getConfiguration("C_Cpp.vcFormat.space", this.RootUri),
                     wrap:  vscode.workspace.getConfiguration("C_Cpp.vcFormat.wrap", this.RootUri)
-                },
-                tabSize: vscode.workspace.getConfiguration("editor.tabSize", this.RootUri)
+                }
+            },
+            editor: {
+                tabSize: otherSettingsFolder.editorTabSize,
```
Colengms: tabSize has been removed on the native side, as it wasn't needed.  (The tab size is also included in format requests).  So, I think this handling of tabSize can be removed from the TypeScript side.

Extension/src/LanguageServer/client.ts
```diff
@@ -1194,6 +1198,9 @@ export class DefaultClient implements Client {
                 files: {
                     encoding: settings_filesEncoding
                 },
+                editor: {
```
Colengms, line 1201: If you choose to keep tabSize (which is no longer read on the native side), I think it would need to be added to this editor node, to match the change you made in sendAllSettings().  Though, I think it would be better to remove tabSize entirely.  (Also, the set_tab_size() declaration can be removed from settings.h on the native side, as it's not defined).

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I noticed that. I can remove it.

Colengms (approved):

bobbrow (approved):

ultraGentle:
Hi, do you happen to know when this feature will be included in insiders?  And does it apply to languages outside C++?  For some reason, I'm unable to get parenthesis in javascript, even with "completeFunctionCalls" enabled.  (Sorry if this is the wrong place to ask -- wasn't sure!)

sean-mcmanus:
@ultraGentle We don't know yet...maybe within the next couple weeks? We're still working on other stuff for 1.3.0-insiders (https://github.com/microsoft/vscode-cpptools/milestone/59). The feature will just be applied to C/C++ documents. You'd have to file an issue with VS Code for the javascript issue: https://github.com/microsoft/vscode/issues .

PR #7140, opened 2021-03-10
Avoid running the node-addon code if the setting isn't enabled.

sean-mcmanus:


Extension/src/LanguageServer/configurations.ts
```diff
@@ -691,14 +694,15 @@ export class CppProperties {
             return;
         }
         const settings: CppSettings = new CppSettings(this.rootUri);
+        const userSettings: CppSettings = new CppSettings();
```
bobbrow, line 705: Shouldn't the correct value should already be in `settings`?  I thought that the workspace/folder doesn't have the ability to override an 'application' setting but it should reflect the current value of that setting.

sean-mcmanus: Yeah, I think that works too, but I think it's better to explicitly use a non-resource scoped settings object when accessing a non-resource scoped setting.

bobbrow: If it's intentional, then I won't push back.

Colengms (approved):

Issue #5733, opened 2020-07-07
Spurious asterisk is inserted on new line if the previous one starts from the space(s) followed by asterisk


UR3IQO:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 8.0
- VS Code Version: 1.46.1
- C/C++ Extension Version: 0.28.3
When editing C code expression spurious asterisk is inserted on the new line if the previous line starts from the space(s) followed by asterisk

**Steps to reproduce**
Create new .c file. Enter the following code:
```
void a(void)
{
   int x = 1 
         * 1 + 1
}
```
Put cursor before the `+` and hit enter. If there is one or more space(s) before the aterisk, spurious asterisk will be inserted on the new line. See the record:
![vscode7](https://user-images.githubusercontent.com/25171239/86745208-dd378800-c042-11ea-8aeb-ec491fd1432b.gif)

The very bad thing is that the code with the asterisk added may compile without any warnings or errors, but you will get wrong result (for example multiplication instead of addition or subtraction).

The problem roots are probably from the automatic asterisk insertion function when editing comments in the following form:
```
/*************
 *
 *
 **************/
```
If so, that function should be activated only if the comment is edited.

sean-mcmanus:
The code that implements this is at https://github.com/microsoft/vscode-cpptools/blob/release/Extension/src/LanguageServer/languageConfig.ts#L229 if anyone wants to debug/fix it.

I debugged it, but I didn't see what was wrong with the generated regex's (and I don't have currently have time to continue investigating).

bobbrow:
When I implemented this, I emulated what VS Code was doing for TypeScript.  I seem to recall this being a problem for TypeScript too, but since I tried it again, it looks like they have a way to disable it for the case you've shared.

I believe this is the set of rules for TypeScript: https://github.com/microsoft/vscode/blob/e5ceb315da2462b0fea70716e79cc82ce708b438/extensions/typescript-language-features/src/features/languageConfiguration.ts#L21

EDIT: I think this is the commit that changed the behavior: https://github.com/microsoft/vscode/commit/81f58337a9d6fe5a1c199417b17ba1ae150085e8#diff-5fa357ec086244b9f81089781bd6e25b.  There appears to be a new property that can be used to ensure the ` * ` is not added.

bobbrow:
Also mentioned in #2072

UR3IQO:
[...]

PR #7238, opened 2021-03-26
Fix comment continuation.

sean-mcmanus:
Follow up to https://github.com/microsoft/vscode-cpptools/issues/5733

### DIFF
--- a/Extension/src/LanguageServer/languageConfig.ts
+++ b/Extension/src/LanguageServer/languageConfig.ts
@@ -55,7 +55,7 @@ function getMLSplitAfterPattern(): string {
 
 function getMLPreviousLinePattern(insert: string): string | undefined {
     if (insert.startsWith("/*")) {
-        return `\\A(?=^(\\s*(\\/\\*\\*|\\*)).*)(?=(?!(\\s*\\*\\/)))`;
+        return `(?=^(\\s*(\\/\\*\\*|\\*)).*)(?=(?!(\\s*\\*\\/)))`;
     }
     return undefined;
 }
--- a/Extension/test/integrationTests/languageServer/languageServer.integration.test.ts
+++ b/Extension/test/integrationTests/languageServer/languageServer.integration.test.ts
@@ -28,7 +28,7 @@ suite("multiline comment setting tests", function(): void {
         },
         {   // e.g.  * ...|
             beforeText: /^(\t|[ ])*\ \*(\ ([^\*]|\*(?!\/))*)?$/,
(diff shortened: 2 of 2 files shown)


sean-mcmanus:
@elahehrashedi Do you know if there was a reason for the `\\A`? That matches the letter A. Was it just a typo or was something else intended?

Extension/src/LanguageServer/languageConfig.ts
```diff
@@ -55,7 +55,7 @@ function getMLSplitAfterPattern(): string {
 
 function getMLPreviousLinePattern(insert: string): string | undefined {
     if (insert.startsWith("/*")) {
-        return `\\A(?=^(\\s*(\\/\\*\\*|\\*)).*)(?=(?!(\\s*\\*\\/)))`;
```
bobbrow, line 58: The `\A` was added to fix an issue with false positives on multi-line arithmetic.  It seems it breaks for comments too?
```
int n = 3
    * 4
    * 5  // Enter here results in an asterisk being added without \A
```

sean-mcmanus: Ah, yeah, I repro that bug. What was the reasoning behind adding the "\A"? That appears to be just break "everything", right?

Anyway, I'll cancel this PR since it sounds like the actual fix is more complicated.

sean-mcmanus: @bobbrow TypeScript appears to have the same bug. So, I'm thinking we keep that bug and file a VS Code bug? It at least fixes the particular case mentioned in the original bug repro, which might be more common (i.e. the bug repro no requires 2 lines starting with the "*" instead of one).

And the more common case would be the 2 lines of * being part of a comment continuation (so it's better to not break the more common case).

bobbrow: I tend to agree.  I would rather have the comments working than not.  @elahehrashedi do you have any objections?

bobbrow: To answer your earlier question about \A:
image

sean-mcmanus: Where are you getting that from? The regex parsers I have tried do not have that `'\A'` listed as valid, i.e. how do you know the `\A` is valid for the TypeScript regex?

bobbrow: My mistake. I forgot to set the Language mode on regex101.com.

elahehrashedi:
This change will break this scenario, but we agreed that the current scenario is more common. 
image

elahehrashedi (approved):

PR #7239, opened 2021-03-26
Update webpack dependencies.

sean-mcmanus:


sean-mcmanus:
@Colengms Do you know if this breaks anything in regards to vscode-nls? I wasn't able to detect any issue.

bobbrow (approved):

Extension/webpack.config.js
```diff
@@ -54,14 +54,4 @@ const config = {
     }
 }
 
-if (process.argv.includes('--vscode-nls')) {
```
Colengms, line 57: Isn't this removing the nls rewriting, which is required for localization?

Colengms (changes requested):

Extension/webpack.config.js
Colengms: Was there a reason this was removed?  Does it cause a failure now?  Localization of strings (the ones implemented with localize() in TypeScript code) cannot work when webpacked without this.

sean-mcmanus: It doesn't work. I get some error like "webpack doesn't support --vscode-nls". How do I test the localization case? There might be some new way to get it to work.

sean-mcmanus: I'm seeing translated strings in most cases, but not in the Output window.

sean-mcmanus: Yeah, Output window and Find All References "type" are not translated.

sean-mcmanus: The error is
```
[webpack-cli] Error: Unknown option '--vscode-nls'
[webpack-cli] Run 'webpack --help' to see available commands and options
```
I didn't see what change was causing this in the migration docs at https://webpack.js.org/blog/2020-10-10-webpack-5-release/ , but it sounds like it's handling command line arguments different.

Do you have any idea what to change to make it work?

Colengms (approved):

Extension/.vscode/tasks.json
```diff
@@ -95,9 +93,7 @@
             "command": "yarn",
             "args": [
                 "run",
-                "compile-watch",
-                "--loglevel",
-                "silent"
+                "compile-watch"
```
bobbrow, line 96: You may want to take a look at this PR by Xinyu.  We had to fix an issue with the "watch" mode after upgrading webpack in CMake Tools.

https://github.com/microsoft/vscode-cmake-tools/pull/1685/files

Extension/webpack.config.js
bobbrow: We had to do something similar in CMake Tools.  It looks like you landed on about the same approach.
[PR](https://github.com/microsoft/vscode-cmake-tools/pull/1475/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json)

sean-mcmanus: I've verified the loc stuff works.

Extension/.vscode/tasks.json
sean-mcmanus: Ah, thanks. I see those errors now.

sean-mcmanus: I removed the flags that weren't working, but the watcher scripts don't work for me. Webpack gets silently stuck. Might be a webpack bug? I never use the watcher modes. Does anyone else care?

sean-mcmanus: I get `error: EPERM: operation not permitted, watch	at FSEvent.FSWatcher._handle.onchange ` ...seems like it's not our issue.

Colengms: I never use the watcher modes either.  Maybe just remove them?  If they're needed in the future, we can figure it out then.

bobbrow: You have to change the background parser to make it work because webpack changed the messages.  That was the other part of Xinyu's PR.

sean-mcmanus: Ah, okay, that makes sense now.

Issue #4509, opened 2019-10-29
Header autocomplete and header/source switch for "*.cu" "*.cuh" files


vakokako:
Hi, I'm developing in Cuda right now and just adding `*.cu` and `*.cuh` to `files.associations` in settings works for syntax parsing. 
However, when I try to autocomplete `#include <Header.cuh>`, it doesn't show files with `.cuh`/`.cu` extensions. 
Also switching between `.cuh`/`.cu` as header/source doesn't work.

Is there any setting that can control this?

sean-mcmanus:
Unfortunately, the types are currently hardcoded. We might need to add some setting to indicate which files are headers versus source files in addition to just files.associations. The switch header source is being tracking with https://github.com/microsoft/vscode-cpptools/issues/3199 , but we can use this issue to track the completion.

It might just be easier to add those types to our hardcoded list.

PR #7255, opened 2021-03-29
Add CUDA support

Colengms:
A global `supportCuda` is set only on activation.  If the language ID is not registered, all CUDA support should effectively be disabled.  Since the cuda-cpp language ID is being added to VS Code itself, we don't expect the availability of the language ID to change at runtime.

Eventually, we could remove the supportCuda setting, when/if we update the version requirement for VS Code to a version that include the cuda-cpp language ID.


Extension/package.json
```diff
@@ -2209,13 +2224,23 @@
           "command": "workbench.action.showAllSymbols",
           "group": "other1_navigation@4"
         },
+        {
+          "when": "editorLangId == cuda-cpp",
```
bobbrow: interesting that we add an object for each languageId.  I think we can do `||` in these `when` blocks, but that can be something we look into for a subsequent PR.

sean-mcmanus: Can we just try that now and see if it works? I recall I tried a long time ago and it didn't work but may be it works now?

Colengms: Could we defer that to a future PR, just in case?

sean-mcmanus: Isn't it easy to check to see if it works? I tried and it works. The keyboard command case needs to be formatted like `editorLangId == 'cpp' && editorTextFocus || editorLangId == 'c' && editorTextFocus` without `()` grouping.

sean-mcmanus: Oh, did you not change the keyboard switch header case? Is that supposed to work with CUDA? I see an issue at https://github.com/microsoft/vscode-cpptools/issues/4509

sean-mcmanus: Seems like the keybinding should be updated since the context menus exist.

sean-mcmanus (changes requested):
keybindings change?

sean-mcmanus (approved):

Issue #5040, opened 2020-03-02
async functions should not be using Promise APIs


bobbrow:
There are a number of places in the extension where async/await is being mixed with Promise/.then (example: `githubAPI.ts`).  The preferred paradigm is async/await.  Using Promises directly with .then is discouraged.

PR #2895, opened 2018-12-05
Add refresh button to process picker

mbise1993:
Issue [#2885](https://github.com/Microsoft/vscode-cpptools/issues/2885)

PR #7277, opened 2021-04-01
refactoring async/await

elahehrashedi:
issue: https://github.com/microsoft/vscode-cpptools/issues/5040
Unit tests are passing.
Integration tests are passing.


Extension/src/githubAPI.ts
```diff
@@ -154,42 +154,38 @@ export interface BuildInfo {
...
+    } else if (userVersion.isExtensionVersionGreaterThan(new PackageVersion(targetBuild.name))) {
+        // downgrade
+        telemetry.logLanguageServerEvent("UpgradeCheck", { "action": "downgrade", "newVersion": targetBuild.name });
+    } else {
+        // upgrade
+        telemetry.logLanguageServerEvent("UpgradeCheck", { "action": "upgrade", "newVersion": targetBuild.name });
+    }
+
+    if (!targetBuild) {
+        return Promise.resolve(undefined);
```
bobbrow: Because this function is async, you should be able to `return undefined;` directly instead of putting it in a `Promise.resolve(undefined)` #Resolved

Extension/src/githubAPI.ts
```diff
@@ -154,42 +154,38 @@ export interface BuildInfo {
...
+    try {
+        const platformInfo: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+        const vsixName: string = vsixNameForPlatform(platformInfo);
+        const downloadUrl: string = getVsixDownloadUrl(targetBuild, vsixName);
+        if (!downloadUrl) {
+            return undefined;
+        }
+        return { downloadUrl: downloadUrl, name: targetBuild.name };
+    } catch (error) {
+        return Promise.reject(error);
```
bobbrow: Promise.reject translates to `throw` in an `async` function, so I think you should be able to just throw the error (or in this case the right fix might be to remove the try/catch since you'd only be catching it just to throw it again). #Resolved

Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -30,44 +30,45 @@ export interface AttachItem extends vscode.QuickPickItem {
...
-        new Promise<string>((resolve, reject) => {
-            const quickPick: vscode.QuickPick<AttachItem> = vscode.window.createQuickPick<AttachItem>();
-            quickPick.title = localize("attach.to.process", "Attach to process");
-            quickPick.canSelectMany = false;
-            quickPick.matchOnDescription = true;
-            quickPick.matchOnDetail = true;
-            quickPick.placeholder = localize("select.process.attach", "Select the process to attach to");
+export async function showQuickPick(getAttachItems: () => Promise<AttachItem[]>): Promise<string | undefined> {
+    const processEntries: AttachItem[] = await getAttachItems();
+    return new Promise<string>((resolve, reject) => {
```
Colengms: I don't think it's necessary to construct a Promise here anymore.  Since the result of an async function is always a promise, you can move the contained code into the main body of the function.  Instead of calling resolve(), return the value (which would be implicitly wrapped in a Promise for you).  Instead of calling reject, throw an exception.

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -15,7 +15,7 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
 
     public provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
         return new Promise<vscode.TextEdit[]>((resolve, reject) => {
-            this.client.notifyWhenReady(() => {
+            this.client.notifyWhenReady(async () => {
```
Colengms: If converting all calls to notifyWhenReady to use async lambdas, I think you can remove the override of notifyWhenReady that accepts non-async lambdas.

Extension/src/LanguageServer/Providers/foldingRangeProvider.ts
```diff
@@ -21,36 +21,34 @@ export class FoldingRangeProvider implements vscode.FoldingRangeProvider {
             uri: document.uri.toString()
         };
         return new Promise<vscode.FoldingRange[]>((resolve, reject) => {
```
Colengms: I think you should be able to convert provideFoldingRanges() (and any other VS Code API that returns a ProviderResult) into an async function.  If so, you could remove this creation of a Promise.

Extension/src/LanguageServer/Providers/onTypeFormattingEditProvider.ts
```diff
@@ -17,7 +17,7 @@ export class OnTypeFormattingEditProvider implements vscode.OnTypeFormattingEdit
         return new Promise<vscode.TextEdit[]>((resolve, reject) => {
```
Colengms: Same as previous comment.  I think this provider API override can be made async.

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
```diff
@@ -18,7 +18,7 @@ export class SemanticTokensProvider implements vscode.DocumentSemanticTokensProv
 
     public async provideDocumentSemanticTokens(document: vscode.TextDocument, token: vscode.CancellationToken): Promise<vscode.SemanticTokens> {
         return new Promise<vscode.SemanticTokens>((resolve, reject) => {
```
Colengms: Another.  I think this Promise is unnecessary.  This function is already async.

Extension/src/LanguageServer/client.ts
```diff
@@ -1850,9 +1843,9 @@ export class DefaultClient implements Client {
      * before attempting to send messages or operate on the client.
      */
 
-    public queueTask<T>(task: () => Thenable<T>): Thenable<T> {
+    public async queueTask<T>(task: () => Thenable<T>): Promise<T> {
         if (this.isSupported) {
-            const nextTask: () => Thenable<T> = async () => {
+            const nextTask: () => Promise<T> = async () => {
```
Colengms, line 1854: Was this change needed?  I thought Promise is derived from Thenable.

Extension/src/LanguageServer/extension.ts
```diff
@@ -866,63 +848,82 @@ function onSelectConfiguration(): void {
...
-function onSelectConfigurationProvider(): void {
+async function onSelectConfigurationProvider(): Promise<void> {
     onActivationEvent();
     if (!isFolderOpen()) {
         vscode.window.showInformationMessage(localize("configuration.provider.select.first", 'Open a folder first to select a configuration provider'));
     } else {
-        selectClient().then(client => client.handleConfigurationProviderSelectCommand(), rejected => {});
+        try {
+            const client: Client = await selectClient();
+            client.handleConfigurationProviderSelectCommand();
```
Colengms: I think the previous form would have throw an exception from here if encountered within handleConfigurationProviderSelectComment().  This change will hide any exception that occurs there.

Colengms: The same seems to apply to all uses of selectClient().

Colengms: And getActiveClient()

Extension/src/LanguageServer/extension.ts
```diff
@@ -866,63 +848,82 @@ function onSelectConfiguration(): void {
...
-function onGenerateEditorConfig(): void {
+async function onGenerateEditorConfig(): Promise<void> {
     onActivationEvent();
     if (!isFolderOpen()) {
         generateEditorConfig();
     } else {
-        selectClient().then(client => generateEditorConfig(client.RootUri));
+        try {
+            const client: Client = await selectClient();
+            generateEditorConfig(client.RootUri);
```
Colengms: This call to selectClient() did not have a rejected => {}, so it would have thrown an exception.  Also, any exception from generateEditorConfig() should be allowed to propagate.

Extension/src/LanguageServer/extension.ts
```diff
@@ -866,63 +848,82 @@ function onSelectConfiguration(): void {
...
+        } catch (err) {}
     }
 }
 
-function onGoToNextDirectiveInGroup(): void {
+async function onGoToNextDirectiveInGroup(): Promise<void> {
```
Colengms: Similar to last comment, here and again below.

Extension/src/LanguageServer/ui.ts
```diff
@@ -290,14 +289,15 @@ export class UI {
         };
 
         if (this.configurationUIPromise) {
-            this.configurationUIPromise = this.configurationUIPromise.then(result => {
+            const result: ConfigurationResult = await this.configurationUIPromise;
+            this.configurationUIPromise = new Promise<ConfigurationResult>((resolve, reject) => {
```
Colengms: Why is a new Promise being added here?  There is no async within its contents, so making a promise out of it doesn't seem to accomplish anything.  It was previously a promise in order to get chained(?) after the previous configurationUIPromise, but that was awaited on the previous line.  Since it was used for chaining, I'm not sure if an await on the previous one is the right thing to do.  You may want to talk to Michelle, who I think wrote the code in this area.  She was using promise chaining to make some functionality serial.  It's unclear to me if configurationUIPromise will be needed after your changes.

Extension/src/common.ts
```diff
@@ -728,29 +730,27 @@ export function isExecutable(file: string): Promise<boolean> {
 }
 
 export function allowExecution(file: string): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
+    return new Promise<void>(async (resolve, reject) => {
```
Colengms: Is it valid to pass an async lambda to a Promise?  Shouldn't this Promise be removed and allowExecution be converted to async?

Extension/src/main.ts
```diff
@@ -428,7 +424,7 @@ async function postInstall(info: PlatformInformation): Promise<void> {
 
     // If there is a download failure, we shouldn't continue activating the extension in some broken state.
     if (!installSuccess) {
-        return Promise.reject<void>("");
+        throw new Error(``);
```
Colengms: Hmm.. Maybe we should use a meaningful error message here.

Extension/src/packageManager.ts
```diff
@@ -376,7 +370,7 @@ export class PackageManager {
 
                 zipfile.readEntry();
 
-                zipfile.on('entry', (entry: yauzl.Entry) => {
+                zipfile.on('entry', async (entry: yauzl.Entry) => {
```
Colengms: Does zipfile.on calls this lambda as async?  If not, the lambda may not be awaited, so would return to the caller in an incomplete state.  (Not sure if that is an issue in this particular case)

Extension/src/packageManager.ts
```diff
@@ -356,7 +350,7 @@ export class PackageManager {
         });
     }
 
-    private InstallPackage(pkg: IPackage): Promise<void> {
+    private async InstallPackage(pkg: IPackage): Promise<void> {
         this.AppendLineChannel(localize("installing.package", "Installing package '{0}'", pkg.description));
 
         return new Promise<void>((resolve, reject) => {
```
Colengms: This async function is creating a Promise.  It's contents could  be moved into the enclosing function instead, right?

Colengms (changes requested):

Colengms:
Another interesting difference between use of async and `.then()` is that a function called with async will not effectively return to the (async) calling function until it is complete.  But, a function called with `.then` actually returns immediately, and queues up the remainder (passed to then()) to be called later, when the promise completes.  Code after a `.then` (not passed to it) previously would likely have executed before to the contents passed to `.then`.  We may need to watch out for cases in which that change of order could cause different behavior.

bobbrow:
> Another interesting difference between use of async and `.then()` is that a function called with async will not effectively return to the (async) calling function until it is complete. But, a function called with `.then` actually returns immediately, and queues up the remainder (passed to then()) to be called later, when the promise completes. Code after a `.then` (not passed to it) previously would likely have executed before to the contents passed to `.then`. We may need to watch out for cases in which that change of order could cause different behavior.

I just synced with @elahehrashedi on this and one guiding principle we established is that this PR should not change the return type of a function from non-promise/thenable to a promise/thenable.  Functions like those (e.g. accepting a callback as an arg) would likely need a redesign and I think that's out of scope for this change.  We just want to do the mechanical ones here where there is no additional code after a `.then` statement unless it is chaining another `.then` statement.  And also that `.then` should not accept an `async` function as an argument.

Extension/src/Debugger/attachQuickPick.ts
elahehrashedi: I applied this change in some other places, but here as the resolve and reject were coming from the event listener I couldn't apply this change.

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
elahehrashedi: in some places in client.ts, `notifyWhenReady` uses accepts normal non-async lambdas

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
elahehrashedi: now `return new Promise` are removed in other places wherever the function is async.

Extension/src/LanguageServer/ui.ts
elahehrashedi: @michelleangela can you take a look at the latest change?

Extension/src/main.ts
elahehrashedi: I added a message here. does handleError take care of sending the good error message in telemetry?

elahehrashedi: or should I throw a `new PackageManagerError`

Extension/src/packageManager.ts
elahehrashedi: let's discuss about this, will call you. couldn't remove the return new promise in this case.

sean-mcmanus (changes requested):
It's hitting linter errors.

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -43,66 +42,64 @@ export class RemoteAttachPicker {
...
+                matchOnDescription: true,
+                placeHolder: localize("select.process.attach", "Select the process to attach to")
+            };
 
-                        return vscode.window.showQuickPick(processes, attachPickOptions)
-                            .then(item => item ? item.id : Promise.reject<string>(new Error(localize("process.not.selected", "Process not selected."))));
-                    });
-            }
-        });
+            return vscode.window.showQuickPick(processes, attachPickOptions)
```
Colengms: Should this be await'ed also?

Extension/src/Debugger/extension.ts
```diff
@@ -76,55 +76,55 @@ export function initialize(context: vscode.ExtensionContext): void {
...
-                } else {
                     return Promise.resolve();
-                    // TODO uncomment this when single file mode works correctly.
-                    // const buildTasks: vscode.Task[] = await getBuildTasks(true);
-                    // const task: vscode.Task = buildTasks.find(task => task.name === selection.configuration.preLaunchTask);
-                    // await vscode.tasks.executeTask(task);
-                    // delete selection.configuration.preLaunchTask;
                 }
+            } else {
+                return Promise.resolve();
```
Colengms: Is this `Promise.resolve()` needed (and others in this function) ?  Since this code is within an async lamba (being passed to `registerTextEditorCommand`), I think these can just be `return` without having to use `Promise.resolve()`.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -436,9 +436,12 @@ export class CppProperties {
     private async readNodeAddonIncludeLocations(rootPath: string): Promise<void> {
         let error: Error | undefined;
         let pdjFound: boolean = false;
-        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
-            .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
-            .catch(e => (error = e));
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8");
+        try {
+            pdjFound = true; return JSON.parse(package_json);
```
Colengms: Could you separate these into 2 lines?

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,110 +496,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
-                } else {
-                    cmdFile = 'code.cmd';
-                }
-                const vsCodeExeDir: string = path.dirname(process.execPath);
-                return path.join(vsCodeExeDir, 'bin', cmdFile);
-            } else if (platformInfo.platform === 'darwin') {
-                return path.join(process.execPath, '..', '..', '..', '..', '..',
-                    'Resources', 'app', 'bin', 'code');
+    const platformInfo: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+    const vsCodeScriptPath = (platformInfo: any): string => {
```
Colengms: vsCodeScriptPath was previously a string (which was initialized by a function that was defined and called in the same statement).  Since it's now a lambda, could you choose a new name for it?  Maybe GetVsCodeScriptPath?

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,110 +496,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
+    } catch (err) {
+        throw new Error('Failed to find VS Code script');
+    }
 
+    // 1.28.0 changes the CLI for making installations.  1.27.2 was immediately prior.
+    const oldVersion: PackageVersion = new PackageVersion('1.27.2');
+    if (userVersion.isVsCodeVersionGreaterThan(oldVersion)) {
+        let process: ChildProcess;
+        try {
+            process = spawn(vsCodeScriptPath(platformInfo), ['--install-extension', vsixLocation, '--force']);
```
Colengms: This is invoking the lambda a second time.  Could you instead use the value returned by the lambda the previous time called?

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,110 +496,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
-                return;
-            }
-            stdout.on('data', () => {
-                if (sentOverride) {
-                    return;
-                }
-                const stdin: Writable | null = process.stdin;
-                if (!stdin) {
-                    reject(new Error("Failed to communicate with VS Code script process for installation"));
+            process.on('exit', (code: number) => {
```
Colengms: `process.on` is registering a callback to be called when the process exits.  (i.e. Like an event handler).  Since that is not going to occur immediately in the context of this call, the call stack will be different when it's called.  So, this should not throw an error, as that would wind up being processed somewhere else.  I think you'll need to restore the use of a Promise and calls to resolve/reject here, so it properly completes the original function when the process exits.  This is like one of the 'event handler' scenario.

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,110 +496,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
+            if (process.pid === undefined) {
+                throw new Error();
+            }
+        } catch (error) {
+            throw new Error('Failed to launch VS Code script process for installation');
+        }
+    } else {
+        let process: ChildProcess;
+        try {
+            process = spawn(vsCodeScriptPath(platformInfo), ['--install-extension', vsixLocation]);
```
Colengms: Another call to the lambda that could use the original returned value.

Extension/src/LanguageServer/extension.ts
```diff
@@ -498,110 +496,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
+        }, 30000);
+
+        // If downgrading, the VS Code CLI will prompt whether the user is sure they would like to downgrade.
+        // Respond to this by writing 0 to stdin (the option to override and install the VSIX package)
+        let sentOverride: boolean = false;
+        const stdout: Readable | null = process.stdout;
+        if (!stdout) {
+            throw new Error("Failed to communicate with VS Code script process for installation");
+        }
+        stdout.on('data', () => {
```
Colengms: Similar to the above.  `stdout.on` is an event that may occur later.  In order for it's completion/failure to trigger completion/failure of the original function, it will need use of resolve/reject restored.

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -23,7 +23,7 @@ export function createProtocolFilter(clients: ClientCollection): Middleware {
     /* tslint:enable */
 
     return {
-        didOpen: (document, sendMessage) => {
+        didOpen: async (document, sendMessage) => {
```
Colengms: Is it valid to make this lambda async?  The prototype for didOpen does not seem to expect a Promise, so it will not await for this async function to complete.  That actually might not cause an issue, but it changes the behavior.  Instead of deferring only what is in the then() block until the setTextDocumentLanguage() is complete, it will defer everything after any call to await until the awaited operation is complete.  In this case, I think it will work out fine, but leaving this comment to make you aware of this potential pitfall.

Extension/src/LanguageServer/settings.ts
```diff
@@ -688,21 +688,28 @@ function populateEditorConfig(rootUri: vscode.Uri | undefined, document: vscode.
         const lastPosition: vscode.Position = document.lineAt(document.lineCount - 1).range.end;
         edits.insert(document.uri, lastPosition, remainingSettingsText);
     }
-    vscode.workspace.applyEdit(edits).then(() => vscode.window.showTextDocument(document));
+    (async () => {
```
Colengms: This looks more complicated than the original line of code.  Can the enclosing function `populateEditorConfig` be made async?  If not, I would leave use of `then` instead of wrapping things in an async lambda that is itself not awaited.

Extension/src/LanguageServer/settings.ts
```diff
@@ -688,21 +688,28 @@ function populateEditorConfig(rootUri: vscode.Uri | undefined, document: vscode.
...
-            vscode.workspace.openTextDocument(uri).then(
-                (document) => populateEditorConfig(rootUri, document),
-                () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
-        }, () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
+
+        try {
+            await vscode.workspace.applyEdit(edits);
+        } finally {
+            document = await vscode.workspace.openTextDocument(uri);
+            populateEditorConfig(rootUri, document);
```
Colengms: All of the calls to `populateEditorConfig` are in this function.  So, I think `populateEditorConfig` could be made async, to address my previous comment.

Extension/src/common.ts
```diff
@@ -727,34 +729,29 @@ export function isExecutable(file: string): Promise<boolean> {
...
-                            });
+            const exists: boolean = await checkFileExists(file);
+            if (exists) {
+                const isExec: boolean = await isExecutable(file);
+                if (isExec) {
+                    return;
+                } else {
+                    fs.chmod(file, '755', (err: NodeJS.ErrnoException | null) => {
+                        if (err) {
+                            throw new Error(err.message);
```
Colengms: `fs.chmod` is a file operation, so it may not complete within the context of this call.  So, the lambda being passed in may execute later with a different call stack.  I don't think it's safe to throw errors here.  It's similar to an event handler scenario, where you would need to continue to use resolve/reject in order to control completion of the calling function.

Or, you could switch to using `chmodSync()` to simplify the code, as a very slight impact to performance.

Extension/src/common.ts
```diff
@@ -806,11 +803,12 @@ export function promptForReloadWindowDueToSettingsChange(): void {
 
 export function promptReloadWindow(message: string): void {
     const reload: string = localize("reload.string", "Reload");
-    vscode.window.showInformationMessage(message, reload).then((value?: string) => {
+    (async () => {
```
Colengms: I'd avoid using a async lambda to wrapper awaited calls.  If you don't also await the lamda, it's not really accomplishing anything.  If promptReloadWindow cannot be made async, this will need to continue to use `then`.

Extension/src/main.ts
```diff
@@ -104,43 +104,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
         } else if (!(await util.checkInstallJsonsExist())) {
             // Check the Json files to declare if the extension has been installed successfully.
             errMsg = localize("jason.files.missing", "The C/C++ extension failed to install successfully. You will need to reinstall the extension for C/C++ language features to function properly.");
             const downloadLink: string = localize("download.button", "Go to Download Page");
-            vscode.window.showErrorMessage(errMsg, downloadLink).then(async (selection) => {
-                if (selection === downloadLink) {
-                    vscode.env.openExternal(vscode.Uri.parse(releaseDownloadUrl));
-                }
-            });
+            const selection: string | undefined = await vscode.window.showErrorMessage(errMsg, downloadLink);
```
Colengms: This looks kinda dangerous.  The previous behavior was to show the dialog, and IF someone clicked on the URL, to open it.  The new behavior is to pause async execution of the extension UNTIL someone clicks on the URL (or dismisses the dialog?).  I think this should be restored to the previous approach.  You might want to check to make sure you're not waiting for the results of dialogs elsewhere, in similar scenarios.

Extension/src/packageManager.ts
```diff
@@ -104,30 +104,27 @@ export class PackageManager {
...
-                    progress.report({ message: localize("downloading.progress.description", "Downloading {0}", pkg.description), increment: this.GetIncrement(count, packages.length) });
-                    count += 1;
-                    return p;
-                });
-            });
+    public async DownloadPackages(progress: vscode.Progress<{ message?: string; increment?: number }>): Promise<void | null> {
+        const packages: IPackage[] = await this.GetPackages();
+        let count: number = 1;
+        return this.BuildPromiseChain(packages, (pkg): Promise<void> => {
+            const p: Promise<void> = this.DownloadPackage(pkg);
```
Colengms: I think this could be an async lambda that first displays the progress info then awaits the result of `DownloadPackage`.

Extension/src/packageManager.ts
```diff
@@ -104,30 +104,27 @@ export class PackageManager {
...
-                    progress.report({ message: localize("installing.progress.description", "Installing {0}", pkg.description), increment: this.GetIncrement(count, packages.length) });
-                    count += 1;
-                    return p;
-                });
-            });
+    public async InstallPackages(progress: vscode.Progress<{ message?: string; increment?: number }>): Promise<void | null> {
+        const packages: IPackage[] = await this.GetPackages();
+        let count: number = 1;
+        return this.BuildPromiseChain(packages, (pkg): Promise<void> => {
+            const p: Promise<void> = this.InstallPackage(pkg);
```
Colengms: Similar to previous comment.

Colengms (changes requested):

Extension/src/packageManager.ts
elahehrashedi: Colen opened a new bug 
I will revert it, for it to be fixed in another issue.

sean-mcmanus (changes requested):
There are linter errors and some conflicting changes need to be resolved. We should probably try to get this checked in soon so further merge conflicts don't occur.

Extension/src/main.ts
elahehrashedi: this one is resolved. I will check all other places before closing this comment.

Extension/src/LanguageServer/ui.ts
elahehrashedi: Update: working with Michelle on this

Extension/src/common.ts
```diff
@@ -736,13 +738,7 @@ export function allowExecution(file: string): Promise<void> {
...
                             resolve();
                         } else {
-                            fs.chmod(file, '755', (err: NodeJS.ErrnoException | null) => {
-                                if (err) {
-                                    reject(err);
-                                    return;
-                                }
-                                resolve();
-                            });
+                            chmodAsync(file, '755');
```
bobbrow: I think this whole function should be made async now that you have `chmodAsync` implemented.  You can await this here as well as `checkFileExists` and `isExecutable`. #Resolved

Extension/src/githubAPI.ts
```diff
@@ -309,30 +303,32 @@ async function getReleaseJson(): Promise<Build[] | undefined> {
...
-                throw new Error('Failed to download release JSON');
-            }
-        });
-    if (!data) {
-        return Promise.resolve(undefined);
-    }
-
-    // Parse the file
-    let releaseJson: any;
+    let data: string;
```
bobbrow: I think you can declare this inside the `try` block #Resolved

Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -30,44 +30,45 @@ export interface AttachItem extends vscode.QuickPickItem {
...
+        quickPick.matchOnDescription = true;
+        quickPick.matchOnDetail = true;
+        quickPick.placeholder = localize("select.process.attach", "Select the process to attach to");
+        quickPick.items = processEntries;
+        quickPick.buttons = [new RefreshButton()];
+
+        const disposables: vscode.Disposable[] = [];
+
+        quickPick.onDidTriggerButton(async button => {
+            const processEntries: AttachItem[] = await getAttachItems();
```
sean-mcmanus: I think this should be simplified to `quickPick.items = await getAttachItems();` or `(async button => quickPick.items = await getAttachItems());` if that works.

sean-mcmanus (changes requested):

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -25,14 +25,13 @@ export interface AttachItemsProvider {
...
-    public ShowAttachEntries(): Promise<string | undefined> {
-        return util.isExtensionReady().then(ready => {
-            if (!ready) {
-                util.displayExtensionNotReadyPrompt();
-            } else {
-                return showQuickPick(() => this.attachItemsProvider.getAttachItems());
-            }
-        });
+    public async ShowAttachEntries(): Promise<string | undefined> {
+        const ready: boolean = await util.isExtensionReady();
```
sean-mcmanus: I think this should be simplified to `if (!await util.isExtensionReady())`...same for the other cases below (I see several cases).

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -43,66 +42,67 @@ export class RemoteAttachPicker {
...
+                matchOnDetail: true,
+                matchOnDescription: true,
+                placeHolder: localize("select.process.attach", "Select the process to attach to")
+            };
 
-                        return vscode.window.showQuickPick(processes, attachPickOptions)
-                            .then(item => item ? item.id : Promise.reject<string>(new Error(localize("process.not.selected", "Process not selected."))));
-                    });
+            const item: AttachItem | undefined = await vscode.window.showQuickPick(processes, attachPickOptions);
+            if (item) {
```
sean-mcmanus, line 98: I'd prefer `if (await vscode.window.showQuickPick(processes, attachPickOptions))`

sean-mcmanus (changes requested):

bobbrow:
>             promiseChain = promiseChain.then(() => promiseBuilder(item));

Should this just be `await promiseBuilder(item);`?

---
Refers to: Extension/src/packageManager.ts:150 in 32c152d. [](commit_id = 32c152d6666da697cd03580846c7c2acec3e3b75, deletion_comment = False)

bobbrow:
>             promiseChain = promiseChain.then(() => promiseBuilder(item));

(and you don't need `promiseChain` at all)

---
In reply to: [825985485](https://github.com/Microsoft/vscode-cpptools/pull/7277#issuecomment-825985485) [](http://example.com/codeflow?ancestors=825985485)

---
Refers to: Extension/src/packageManager.ts:150 in 32c152d. [](commit_id = 32c152d6666da697cd03580846c7c2acec3e3b75, deletion_comment = False)

bobbrow:
>             promiseChain = promiseChain.then(() => promiseBuilder(item));

I would almost just have you delete this function and do the
```
for (...) {
   // progress report
   /// await download or install
}
```
inline in `DownloadPackages` and `InstallPackages` since I think it's way more clear.

---
In reply to: [825985674](https://github.com/Microsoft/vscode-cpptools/pull/7277#issuecomment-825985674) [](http://example.com/codeflow?ancestors=825985674,825985485)

---
Refers to: Extension/src/packageManager.ts:150 in 32c152d. [](commit_id = 32c152d6666da697cd03580846c7c2acec3e3b75, deletion_comment = False)

Extension/src/platform.ts
```diff
@@ -47,10 +47,8 @@ export class PlatformInformation {
                 throw new Error(localize("unknown.os.platform", "Unknown OS platform"));
         }
 
-        return Promise.all([distributionPromise, versionPromise])
-            .then(([distro, version]) =>
-                new PlatformInformation(platform, architecture, distro, version)
-            );
+        const [distro, version] = await Promise.all([distributionPromise, versionPromise]);
```
bobbrow: I think this `Promise.all` just makes things more complicated.  We should just `await` the calls that matter in the switch..case statement and assign the variables directly there.

Then you just initialize `distro` and `version` as undefined instead of having `distributionPromise` and `versionPromise` #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -1486,34 +1483,31 @@
...
-                                    ask.Value = false;
-                                    break;
-                                }
-                                default: {
-                                    break;
-                                }
+                        const result: string | undefined = await vscode.window.showInformationMessage(message, allow, dontAllow, askLater);
+                        switch (result) {
+                            case allow: {
+                                await  this.configuration.updateCustomConfigurationProvider(provider.extensionId);
```
bobbrow: nit: delete one of the spaces after `await` #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -1550,8 +1544,8 @@
         });
     }
 
-    public updateCustomBrowseConfiguration(requestingProvider?: CustomConfigurationProvider1): Thenable<void> {
-        return this.notifyWhenReady(() => {
+    public async updateCustomBrowseConfiguration(requestingProvider?: CustomConfigurationProvider1): Promise<void> {
+        return this.notifyWhenReady(async () => {
```
bobbrow: it doesn't look like there was any need to make this or the line above `async`.  Can you revert it? #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -1883,11 +1879,11 @@
             pendingTask = new util.BlockingTask<T>(task, pendingTask);
             return pendingTask.getPromise();
         } else {
-            return Promise.reject(localize("unsupported.client", "Unsupported client"));
+            throw new Error (localize("unsupported.client", "Unsupported client"));
```
bobbrow, line 1886: Do you need to mark this function as `async` to ensure that this throw is converted to a promise rejection? #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -2116,39 +2112,37 @@
...
+                                        vscode.commands.executeCommand("workbench.action.problems.focus");
+                                    }
+                                    client.handleConfigurationEditUICommand();
+                                    telemetry.logLanguageServerEvent("SettingsCommand", { "toast": "ui" }, undefined);
+                                    break;
+                                case dontShowAgain:
+                                    showIntelliSenseFallbackMessage.Value = false;
+                                    break;
+                            }
+                            return Promise.resolve(true);
```
bobbrow: I think this should be `return true;` #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -2575,54 +2568,46 @@
...
-                            }
-                        });
-                });
+        this.notifyWhenReady(async () => {
+            const extensionId: string | undefined = await ui.showConfigurationProviders(this.configuration.CurrentConfigurationProvider);
+            if (extensionId === undefined) {
+                // operation was canceled.
+                return;
+            }
+            await this.configuration.updateCustomConfigurationProvider(extensionId);
```
bobbrow: this promise was not previously chained, but I think it's ok. #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -2651,20 +2636,18 @@
                 next: next
             };
 
-            this.languageClient.sendRequest(GoToDirectiveInGroupRequest, params)
```
bobbrow, line 2654: >sendRequest [](http://example.com/codeflow?start=32&length=11)

Hmm... I thought all `sendRequest` calls were supposed to happen in the context of one of the `whenReady` functions.  Something for us to revisit @colengms.

Extension/src/packageManager.ts
```diff
@@ -104,30 +104,27 @@ export class PackageManager {
...
-                });
-            });
+    public async DownloadPackages(progress: vscode.Progress<{ message?: string; increment?: number }>): Promise<void | null> {
+        const packages: IPackage[] = await this.GetPackages();
+        let count: number = 1;
+        return this.BuildPromiseChain(packages, async (pkg): Promise<void> => {
+            progress.report({ message: localize("downloading.progress.description", "Downloading {0}", pkg.description), increment: this.GetIncrement(count, packages.length) });
+            count += 1;
+            await this.DownloadPackage(pkg);
+            return;
```
bobbrow: return is not needed here #Resolved

Extension/src/packageManager.ts
```diff
@@ -104,30 +104,27 @@ export class PackageManager {
...
-                });
-            });
+    public async InstallPackages(progress: vscode.Progress<{ message?: string; increment?: number }>): Promise<void | null> {
+        const packages: IPackage[] = await this.GetPackages();
+        let count: number = 1;
+        return this.BuildPromiseChain(packages, async (pkg): Promise<void> => {
+            progress.report({ message: localize("installing.progress.description", "Installing {0}", pkg.description), increment: this.GetIncrement(count, packages.length) });
+            count += 1;
+            await this.InstallPackage(pkg);
+            return;
```
bobbrow: return is not needed here #Resolved

Extension/src/LanguageServer/client.ts
```diff
@@ -1863,13 +1855,18 @@ export class DefaultClient implements Client {
 
             if (pendingTask && !pendingTask.Done) {
                 // We don't want the queue to stall because of a rejected promise.
-                return pendingTask.getPromise().then(nextTask, nextTask);
+                try {
+                    await pendingTask.getPromise();
+                    return nextTask();
+                } catch (e) {
```
bobbrow: You had it correct with `finally` previously.  This should change back. #Closed

bobbrow: Wait, I take this back.  You have it right now.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -436,9 +436,13 @@ export class CppProperties {
     private async readNodeAddonIncludeLocations(rootPath: string): Promise<void> {
         let error: Error | undefined;
         let pdjFound: boolean = false;
-        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8")
-            .then(pdj => {pdjFound = true; return JSON.parse(pdj); })
-            .catch(e => (error = e));
+        const package_json: any = await fs.promises.readFile(path.join(rootPath, "package.json"), "utf8");
+        try {
```
bobbrow, line 440: the `await` should be inside the `try` block #Resolved

Extension/src/packageManager.ts
sean-mcmanus: Can you remove this "return;"? Same for the other cases below? #Resolved

Extension/src/LanguageServer/configurations.ts
```diff
@@ -436,9 +436,13 @@ export class CppProperties {
...
+            pdjFound = true;
+            return JSON.parse(package_json);
```
bobbrow: This `return` should actually be the assignment to the `package_json` variable #Resolved

bobbrow: And can you rename `package_json` -> `packageJson` while you're fixing this code?

Extension/src/LanguageServer/client.ts
```diff
@@ -1486,34 +1483,31 @@ export class DefaultClient implements Client {
...
-                                }
-                                case dontAllow: {
-                                    ask.Value = false;
-                                    break;
-                                }
-                                default: {
-                                    break;
-                                }
+                        const result: string | undefined = await vscode.window.showInformationMessage(message, allow, dontAllow, askLater);
+                        switch (result) {
```
sean-mcmanus: I'd prefer `switch (await vscode.window.showInformationMessage(message, allow, dontAllow, askLater))`.

There's are a 3 more similar cases below where the await result is only used once in a switch.

Extension/src/LanguageServer/extension.ts
```diff
@@ -499,110 +497,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
             }
-        }(platformInfo);
-        if (!vsCodeScriptPath) {
-            return Promise.reject(new Error('Failed to find VS Code script'));
-        }
-
-        // 1.28.0 changes the CLI for making installations.  1.27.2 was immediately prior.
-        const oldVersion: PackageVersion = new PackageVersion('1.27.2');
-        if (userVersion.isVsCodeVersionGreaterThan(oldVersion)) {
-            return new Promise<void>((resolve, reject) => {
```
bobbrow: I believe this needs to stay as a Promise.  Those `reject` statements that were changed to `throw` are not running in this context and as a result cannot be caught. #Resolved

Extension/src/LanguageServer/extension.ts
```diff
@@ -499,110 +497,99 @@ function installVsix(vsixLocation: string): Thenable<void> {
...
         }
+    };
+    let vsCodeScriptPath: string;
+    try {
+        vsCodeScriptPath = getVsCodeScriptPath(platformInfo);
+    } catch (err) {
+        throw new Error('Failed to find VS Code script');
+    }
 
-        return new Promise<void>((resolve, reject) => {
```
bobbrow: I believe this promise needs to stay as it was too. #Resolved

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -801,53 +788,49 @@ function onSwitchHeaderSource(): void {
...
+            const client: Client | undefined = clients.get(key);
+            if (client) {
+                return client;
+            } else {
+                console.assert("client not found");
             }
-            return Promise.reject<Client>(localize("client.not.found", "client not found"));
-        });
+        }
+        throw new Error (localize("client.not.found", "client not found"));
```
bobbrow: >  [](http://example.com/codeflow?start=23&length=1)

nit: remove space after `Error` #Resolved

sean-mcmanus:
I finished reviewing -- I just had some minor suggests on removing unneeded variables. Other than that I approve.

Extension/src/LanguageServer/protocolFilter.ts
bobbrow: I think it is safer not to convert this one to async/await

Extension/src/LanguageServer/settings.ts
```diff
@@ -691,18 +691,22 @@ function populateEditorConfig(rootUri: vscode.Uri | undefined, document: vscode.
...
         const uri: vscode.Uri = vscode.Uri.joinPath(rootUri, ".editorconfig");
         const edits: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
         edits.createFile(uri, { ignoreIfExists: true, overwrite: false });
-        vscode.workspace.applyEdit(edits).then(() => {
-            vscode.workspace.openTextDocument(uri).then(
-                (document) => populateEditorConfig(rootUri, document),
-                () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
-        }, () => vscode.workspace.openTextDocument().then((document) => populateEditorConfig(rootUri, document)));
+
+        try {
```
bobbrow, line 701: To keep the intent of the previous code, this should be rewritten as:
```
try {
  await vscode.workspace.applyEdit(edits);
  document = await vscode.workspace.openTextDocument(uri);
} catch {
  document = await vscode.workspace.openTextDocument();  // no uri...
}
populateEditorConfig(rootUri, document);
``` #Resolved

**Review by bobbrow — COMMENTED** [review]
Note for myself.  I have reviewed Extension/src/* and Extension/src/LanguageServer/* so far.

**Comment by sean-mcmanus** [comment]
> Note for myself. I have reviewed Extension/src/* and Extension/src/LanguageServer/* so far.

Isn't that what the "Viewed" UI is for?

In the future it might be better to split a large refactoring like this into multiple smaller chunks, e.g. Extension/src/*, LanguageServer, Debugger, maybe client.ts, etc.

**Comment by bobbrow** [comment]
> Isn't that what the "Viewed" UI is for?

I'm using CodeFlow and didn't expect it to carry over across sessions. It was also implicitly a note for @elahehrashedi but I didn't communicate it as such.

File: Extension/src/Debugger/attachToProcess.ts (hunk 37 above)
Comment (michelleangela) [reply Extension/src/Debugger/attachToProcess.ts:98]: `if (item)` is easier to read in my opinion because it breaks down the process.

Comment (sean-mcmanus) [reply Extension/src/Debugger/attachToProcess.ts:98]: Breaks down what process? The result is just more text to read. Using 1 line removes ~26 characters. It also introduces an extra item variable in the scope that is unnecessary.

Comment (michelleangela) [reply Extension/src/Debugger/attachToProcess.ts:98]: it breaks down the process of
1. get result of `await vscode.window.showQuickPick`
2. consume the result

But the coding style is decided by the person currently writing the code :)

File: Extension/src/LanguageServer/Providers/foldingRangeProvider.ts [hunk 55]
```diff
@@ -13,46 +13,42 @@ export class FoldingRangeProvider implements vscode.FoldingRangeProvider {
...
-                            });
-                            resolve(result);
-                        }
-                    });
+        return this.client.notifyWhenReady(async () => {
+            const ranges: GetFoldingRangesResult = await this.client.languageClient.sendRequest(GetFoldingRangesRequest, params);
+            if (ranges.canceled) {
+                throw new Error('');
+            } else {
                 token.onCancellationRequested(e => this.client.abortRequest(id));
```
bobbrow: This gets set too late.  The request has already completed at this point.  The old code would do this immediately after `sendRequest` (but without waiting for `sendRequest` to complete).  Should this be done before calling `sendRequest` or do we need to revert this back to using `.then`? #Resolved

Extension/src/LanguageServer/Providers/foldingRangeProvider.ts
```diff
@@ -13,46 +13,42 @@ export class FoldingRangeProvider implements vscode.FoldingRangeProvider {
...
-                                }
-                                result.push(foldingRange);
-                            });
-                            resolve(result);
-                        }
-                    });
+        return this.client.notifyWhenReady(async () => {
+            const ranges: GetFoldingRangesResult = await this.client.languageClient.sendRequest(GetFoldingRangesRequest, params);
+            if (ranges.canceled) {
+                throw new Error('');
```
bobbrow: It would be good to put some text in here, I think.  'providing folding ranges was cancelled' or something like that. #Resolved

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -13,53 +13,50 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
...
-                        range: {
-                            start: {
-                                character: 0,
-                                line: 0
-                            },
-                            end: {
-                                character: 0,
-                                line: 0
-                            }
+    public async provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
```
bobbrow, line 16: `async` is not needed here because the function it decorates does not `await` anything.  It just returns the Promise created by another function. #Resolved

bobbrow: This applies to all the "provide" functions in the Providers folder except `provideWorkspaceSymbols`

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
```diff
@@ -17,40 +17,36 @@ export class SemanticTokensProvider implements vscode.DocumentSemanticTokensProv
...
+            } else {
+                token.onCancellationRequested(_e => this.client.abortRequest(id));
+                const id: number = ++DefaultClient.abortRequestId;
+                const params: GetSemanticTokensParams = {
+                    id: id,
+                    uri: uriString
+                };
+                const tokensResult: GetSemanticTokensResult = await this.client.languageClient.sendRequest(GetSemanticTokensRequest, params);
+                if (tokensResult.canceled) {
+                    throw new Error('');
```
bobbrow: I think we should put some text in these Errors. #Resolved

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -33,21 +33,20 @@ export class CppdbgDebugAdapterDescriptorFactory extends AbstractDebugAdapterDes
...
-    createDebugAdapterDescriptor(session: vscode.DebugSession, executable?: vscode.DebugAdapterExecutable): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
-        return util.isExtensionReady().then(ready => {
-            if (ready) {
-                let command: string = path.join(this.context.extensionPath, './debugAdapters/OpenDebugAD7');
-
-                // Windows has the exe in debugAdapters/bin.
-                if (os.platform() === 'win32') {
-                    command = path.join(this.context.extensionPath, "./debugAdapters/bin/OpenDebugAD7.exe");
-                }
+    async createDebugAdapterDescriptor(session: vscode.DebugSession, executable?: vscode.DebugAdapterExecutable): Promise<vscode.ProviderResult<vscode.DebugAdapterDescriptor>> {
```
bobbrow: I seem to remember talking to you about this one. You can't change the return type for this because it's implementing an abstract method.  I'm surprised that this is compiling in CI.  We should just revert this one and let it keep using `.then`. #Resolved

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -58,21 +57,20 @@ export class CppvsdbgDebugAdapterDescriptorFactory extends AbstractDebugAdapterD
         super(context);
     }
 
-    createDebugAdapterDescriptor(session: vscode.DebugSession, executable?: vscode.DebugAdapterExecutable): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
+    async createDebugAdapterDescriptor(session: vscode.DebugSession, executable?: vscode.DebugAdapterExecutable): Promise<vscode.ProviderResult<vscode.DebugAdapterDescriptor>> {
```
bobbrow: I think we should revert this one too. #Resolved

Extension/src/Debugger/attachToProcess.ts
elahehrashedi: Will look for similar cases and will resolve this comment after.

Extension/src/Debugger/attachToProcess.ts
elahehrashedi: in this case, I am using the `item.id` later, so I kept this one, but I'm changing similar cases.

sean-mcmanus: Oops, didn't see that.

Extension/src/packageManager.ts
elahehrashedi: I changed the way `BuildPromiseChain `was working (I couldn't leave this comment under the relevant comment that you left). Please take a look at `SequentialResolve` function. it sequentially resolves a sequence of promises.

Extension/src/LanguageServer/client.ts
elahehrashedi: it turned out that we can't await on `showInformationMessage`, so I reverted this part of code and similar cases.

Extension/src/platform.ts
elahehrashedi: please take a look at the modified version.

Extension/src/LanguageServer/configurations.ts
elahehrashedi: please check the modified code.

sean-mcmanus (approved):
I'm approving because I'm OOF Wed.

Extension/src/LanguageServer/settings.ts
elahehrashedi: please take a look at the modified code.

Extension/src/LanguageServer/Providers/foldingRangeProvider.ts
elahehrashedi: It looks like I meant to shift `token.onCancellationRequested(e => this.client.abortRequest(id));` to happen earlier. but I should have placed it before `await this.client.languageClient.sendRequest`, not in the else clause. How does it look now?

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
elahehrashedi: Do new messages make sense?

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
bobbrow: I take it back.  It looks like notifyWhenReady returns `Promise<Promise<>>` so the async is needed to strip the outer Promise.

That said, we talked offline and since `Promise<Promise<>>` is an extremely odd concept and we're not sure how it will behave in practice, let's go an alternate route here and not allow `async` callbacks to be passed in here.  Instead, let's do
```
  await this.client.notifyWhenReady(() => {});
```
and then put the body of the callback after that await.

I think in the future, a better pattern would be to eliminate callbacks as input to `notifyWhenReady` and require all callers to await its return and perhaps have it provide the `languageClient` as a return value since that is the whole purpose of this function to begin with (to protect access to `this.client.languageClient` before it is ready)

Extension/src/common.ts
```diff
@@ -1292,3 +1288,11 @@ export async function checkCuda(): Promise<void> {
     const langs: string[] = await vscode.languages.getLanguages();
     supportCuda = langs.findIndex((s) => s === "cuda-cpp") !== -1;
 }
+
+// Sequentially Resolve Promises.
+export async function sequentialResolve(items: any[], promiseBuilder: (item: any) => Promise<any>): Promise<void> {
```
bobbrow: To improve type safety with this method, can you change the signature to:
```
export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Promise<void>): Promise<void> {
```

(basically remove the `any` types)

It also doesn't need to be decorated with `async` #Resolved

Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -30,44 +30,42 @@ export interface AttachItem extends vscode.QuickPickItem {
...
+        quickPick.canSelectMany = false;
+        quickPick.matchOnDescription = true;
+        quickPick.matchOnDetail = true;
+        quickPick.placeholder = localize("select.process.attach", "Select the process to attach to");
+        quickPick.items = processEntries;
+        quickPick.buttons = [new RefreshButton()];
+
+        const disposables: vscode.Disposable[] = [];
+
+        quickPick.onDidTriggerButton(async () => { quickPick.items = await getAttachItems(); });
```
bobbrow: You need to add back in the `undefined` and `disposables` arguments to `onDidTriggerButton` so that the event listener can be cleaned up later. #Resolved

Extension/src/LanguageServer/client.ts
bobbrow: Why can't we await this?  It seems to compile for me.  (and I actually preferred the original approach with the local variable defined before the switch, but either way is fine :))

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
elahehrashedi: please check the modified code.
There were other usages of `notifyWhenReady` in client.ts that  Ididn't change.

bobbrow: I will. Yeah, we only need to update the `notifyWhenReady` calls that pass in an `async` function as the argument.

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
Colengms: Bob, I think you may be mistaken.  You can provide an overload with a compatible type with the abstract function's return type.  ProviderResult<T> is:
```
export type ProviderResult<T> = T | undefined | null | Thenable<T | undefined | null>;
```

So, this could return a `Promise<vscode.DebugAdapterDescriptor>`, as it's compatible with `Thenable<vscode.DebugAdapterDescriptor>`, and can also be async.  The same applies for various other provider API overloads.  For example, look at `provideCodeActions` in `CodeActionProvider`.  It's async and returns a `Promise<(vscode.Command | vscode.CodeAction)[]>`, yet the API function it's overloading returns a `ProviderResult<(Command | CodeAction)[]>`.

Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -30,44 +30,42 @@ export interface AttachItem extends vscode.QuickPickItem {
...
-
-            quickPick.onDidTriggerButton(button => {
-                getAttachItems().then(processEntries => quickPick.items = processEntries);
-            }, undefined, disposables);
-
-            quickPick.onDidAccept(() => {
-                if (quickPick.selectedItems.length !== 1) {
-                    reject(new Error(localize("process.not.selected", "Process not selected.")));
-                }
+export async function showQuickPick(getAttachItems: () => Promise<AttachItem[]>): Promise<string | undefined> {
```
Colengms, line 34: It might be a good idea to add a comment on async/Promise functions when the function is not intended to have a top-level await.  For example, any function that returns a promise acquired from `showInformationMessage`/`showWarningMessage`/`showErrorMessage` should not have a top level await.  By 'top level await', I mean that it's OK for a function to await on this function, but that chain must be broken at some point. 
 Otherwise, the topmost await will cause the extension to be stalled, and there are some async actions we should never stall waiting for.  This particular function does not resolve until quickPick UI is used, so should never have a top level await.  Having a comment to make that clear might help someone in the future who sees an async function and thinks they are fixing a bug by adding an await to it.

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -43,66 +41,66 @@ export class RemoteAttachPicker {
...
+            let pipeProgram: string | undefined;
 
-                if (os.platform() === 'win32' &&
-                    pipeTransport.pipeProgram &&
-                    !fs.existsSync(pipeTransport.pipeProgram)) {
-                    const pipeProgramStr: string = pipeTransport.pipeProgram.toLowerCase().trim();
-                    const expectedArch: debugUtils.ArchType = debugUtils.ArchType[process.arch as keyof typeof debugUtils.ArchType];
+            if (os.platform() === 'win32' &&
+                pipeTransport.pipeProgram &&
+                !fs.existsSync(pipeTransport.pipeProgram)) {
```
Colengms: Since this is an async function now, could you replace the calls to `fs.existsSync` in this function with `await util.checkFileExists`?

Extension/src/Debugger/extension.ts
```diff
@@ -76,55 +76,55 @@ export function initialize(context: vscode.ExtensionContext): void {
...
-                    Telemetry.logDebuggerEvent("buildAndDebug", { "success": "true" });
-                } catch (e) {
-                    Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
-                }
+                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
             }
-        });
+        } catch (e) {
+            try {
+                vscode.debug.startDebugging(folder, selection.configuration);
```
Colengms: Not your change, but I think there is an error here.  There should be an await before this call to startDebugging.  Otherwise, the try block that contains it may fail to catch errors that occur within it.

Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -30,44 +30,42 @@ export interface AttachItem extends vscode.QuickPickItem {
...
+        quickPick.canSelectMany = false;
+        quickPick.matchOnDescription = true;
+        quickPick.matchOnDetail = true;
+        quickPick.placeholder = localize("select.process.attach", "Select the process to attach to");
+        quickPick.items = processEntries;
+        quickPick.buttons = [new RefreshButton()];
+
+        const disposables: vscode.Disposable[] = [];
+
+        quickPick.onDidTriggerButton(async () => { quickPick.items = await getAttachItems(); }, undefined, disposables);
```
Colengms, line 47: Not your change, but do you know if this code is actually necessary?  `quickPick.items` was already populated by a call to `getAttachItems()`, and it don't see any code here that would cause it to return something different the second time.

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
```diff
@@ -17,40 +17,36 @@ export class SemanticTokensProvider implements vscode.DocumentSemanticTokensProv
...
+            } else {
+                token.onCancellationRequested(_e => this.client.abortRequest(id));
+                const id: number = ++DefaultClient.abortRequestId;
+                const params: GetSemanticTokensParams = {
+                    id: id,
+                    uri: uriString
+                };
+                const tokensResult: GetSemanticTokensResult = await this.client.languageClient.sendRequest(GetSemanticTokensRequest, params);
+                if (tokensResult.canceled) {
+                    throw new Error('Requst for providing semantic tokens is cancelled.');
```
bobbrow: [sp] Requst -> Request #Resolved

Extension/src/LanguageServer/Providers/semanticTokensProvider.ts
```diff
@@ -17,40 +17,36 @@ export class SemanticTokensProvider implements vscode.DocumentSemanticTokensProv
...
-                                    tokensResult.tokens.forEach((token) => {
-                                        builder.push(token.line, token.character, token.length, token.type, token.modifiers);
-                                    });
-                                    const tokens: vscode.SemanticTokens = builder.build();
-                                    this.tokenCaches.set(uriString, [tokensResult.fileVersion, tokens]);
-                                    resolve(tokens);
-                                }
-                            }
+                    if (tokensResult.fileVersion !== openFileVersions.get(uriString)) {
+                        throw new Error('The semantic token are not related to the current version of the document.');
```
bobbrow: [sp] token -> tokens #Resolved

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -13,53 +13,49 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
...
-                    configCallBack(undefined);
-                } else {
-                    const editorConfigSettings: any = cachedEditorConfigSettings.get(filePath);
-                    if (!editorConfigSettings) {
-                        editorConfig.parse(filePath).then(configCallBack);
-                    } else {
-                        cachedEditorConfigSettings.set(filePath, editorConfigSettings);
-                        configCallBack(editorConfigSettings);
+    public async provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
+        await this.client.notifyWhenReady(() => { });
```
Colengms: This looks odd.  Perhaps we need a new function for awaiting ready, that we don't have to pass an empty lambda to?

```
    public awaitReady(): Thenable<void> {
        const task: () => Thenable<void> = () => new Promise<void>(resolve => {
            resolve();
        });
        return this.queueTask(task);
    }
``` #Resolved

File: Extension/src/LanguageServer/protocolFilter.ts (hunk 25 above)
Comment (bobbrow) [inline Extension/src/LanguageServer/protocolFilter.ts]: We should revert this to not use `async` for the reason @colengms stated. #Resolved

File: Extension/src/main.ts [hunk 70]
```diff
@@ -414,7 +414,7 @@ async function postInstall(info: PlatformInformation): Promise<void> {
 
     // If there is a download failure, we shouldn't continue activating the extension in some broken state.
     if (!installSuccess) {
-        return Promise.reject<void>("");
+        throw new Error("Failed installing dependencies");
```
Colengms: Could you use a call to localize() here, so the new string can be localized?

Colengms (changes requested):

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
bobbrow: Sure. Another idea would be to make the argument optional and keep the current function.

Long term... my thought was that we'd want to revisit this pattern and have the "when ready" functions return access to the client's `languageClient` variable which is what we're trying to enforce here.  We have a lot of functions that accept callbacks and I think several of those functions shouldn't have been written that way (I'm pointing the finger at myself as the author of many of those).

We also have a couple of functions in cpptools that assume calling `new Promise` is deferring execution of code when in fact it does not.  We wrote a utility function in Makefile Tools that does true deferrals/scheduling for us and I think we should copy it back here.

(sorry for the tangential comments @elahehrashedi, you don't need to do anything about my future thoughts in this PR)

elahehrashedi: I made a small change to the names, `notifyWhenLanguageClientReady ` and `awaitUntilLanguageClientReady`

Extension/src/Debugger/extension.ts
elahehrashedi: seems right,  I added the await.

Extension/src/Debugger/attachQuickPick.ts
Colengms: This is not critical.  I'm OK with this getting checked in without such comments added.

Extension/src/Debugger/attachQuickPick.ts
elahehrashedi: I think the first call to `getAttachItems` was unnecessary, i.e. this line: `const processEntries: AttachItem[] = await getAttachItems();`
I removed the first call.

WardenGnaw: > Not your change, but do you know if this code is actually necessary? quickPick.items was already populated by a call to getAttachItems(), and it don't see any code here that would cause it to return something different the second time.

Yes. This was to handle refreshing using the refresh button and not having to close out of the quickPickMenu and restart a attach launch. It will be different based on the output of `wmic` or `ps`.

See https://github.com/microsoft/vscode-cpptools/pull/2895/files

The first call is important for the window to be initially populated.

Extension/src/Debugger/attachToProcess.ts
elahehrashedi: sure. What I changed was replacing `fs.chmod` with `chmodAsync`. There was only one place that I needed to use `chmodAsync`.

elahehrashedi: so right now in any async function, replace `fs.existsSync` is replaced with `await util.checkFileExists`

Extension/src/Debugger/attachQuickPick.ts
elahehrashedi: added some comments to **client.ts**:
```
/** Note: We should not await on the following functions,
 * or any funstion that returns a promise acquired from them,
 * vscode.window.showInformationMessage, vscode.window.showWarningMessage, vscode.window.showErrorMessage
*/
```

added some comments to **logger.ts** and **attachQuickPick.ts**:
`// We should not await on this function.`

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
bobbrow: You're right.  I think `Promise<vscode.DebugAdapterDescriptor>` is the way to go.

Colengms (approved):

Issue #7194, opened 2021-03-18
Retention Telemetry


andreeis:
Implement extra telemetry that would be useful to analyze retention for CMake build architects (as opposed to exclusively C/C++ developers).

Add a new event to cpptools ms-vscode.cpptools/cmakefilewrite that is identical to the existing cmake-tools event (see https://github.com/microsoft/vscode-cmake-tools/issues/1723). This is because our definition of a CMake user includes anyone using cpptools with a CMakeLists.txt, and this cohort is currently excluded from our existing cmakefilewrite telemetry.

PR #7296, opened 2021-04-06
CMake retention telemetry

andreeis:
This PR addresses work item https://github.com/microsoft/vscode-cpptools/issues/7194.
Corresponding work-item and PRs in CMake Tools are:
https://github.com/microsoft/vscode-cmake-tools/issues/1723
https://github.com/microsoft/vscode-cmake-tools/pull/1740
https://github.com/microsoft/vscode-cmake-tools/pull/1744

Implementing such telemetry in CppTools (while having identical support in CMake Tools) is necessary for when CMake Tools is not installed.

I wasn't able to easily determine if a CMakeLists.txt, CMakeCache.txt or .cmake are directly associated or not with any project in the (multi-root) workspace. I only considered if they are inside or outside the subfolders, based on paths. For CMake Tools analysis of this, I could benefit of variables like sourcesDirectory (pointing to CMakeLists.txt anywhere) or binaryDirectory (pointing to CMakeCache.txt anywhere). This "outside" property is not critically important but I thought to mention in case I missed something regarding how to implement it more accurately in CppTools.


bobbrow (dismissed):

sean-mcmanus (changes requested):
I don't understand. OnDidSaveTextDocument is only sent once and it's called on the Client associated with the workspace folder the file exists in.

sean-mcmanus:
> I don't understand. OnDidSaveTextDocument is only sent once and it's called on the Client associated with the workspace folder the file exists in.

Oh, I figured it out.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -243,6 +243,33 @@ export class CppProperties {
         vscode.workspace.onDidChangeTextDocument((e: vscode.TextDocumentChangeEvent) => {
```
sean-mcmanus: Hmm, this existing code seems incorrect. It's being called on every document edit instead of on save. I think there's an existing bug where c_cpp_properties.json edits are being handled before the file is saved.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I think this should be changed to "vscode.workspace.onDidSaveTextDocument((doc: vscode.TextDocument) => {``

michelleangela:
Which release is this change for?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -243,6 +243,33 @@ export class CppProperties {
...
             if (e.document.uri.fsPath === settingsPath) {
```
sean-mcmanus: This settingsPath should have the location of the c_cpp_properties.json, so you should be able to call vscode.workspace.getWorkspaceFolder to check if it's outside a workspace folder.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Actually, this code may have been intended to handle file creation/deletion as well...but that's already handled by the file watchers...it appears to be for when file watchers fail.

It might be better to handle this in the a didSave handler called on the client (i.e. add one to protocolFilter.txt). You could also add a onDidSaveTextDocument handler like I originally suggested. I'm just worried about calling the code on every edit.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
andreeis: Yes my mistake. I didn't realize I was in onDidChange somehow I thought I was in onDidSave all the time like the counterpart fix in CMake Tools. I am correcting this in an update.

Extension/src/LanguageServer/configurations.ts
andreeis: Great tip, thanks.

andreeis:
> Which release is this change for?

Erika needs this as soon as possible, so Bob said it would be ok for the upcoming 1.3. Of course, if we find anything complicated about this we can postpone for the next release. But I think it's ok and low risk.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -246,6 +246,40 @@ export class CppProperties {
             }
         });
 
+        vscode.workspace.onDidSaveTextDocument((doc: vscode.TextDocument) => {
+            // For multi-root, the "onDidSaveTextDocument" will be received once for each project folder.
+            // To avoid misleading telemetry (for CMake retention) skip if the notifying folder
+            // is not the same workspace folder of the modified document.
+            // Exception: if the document does not belong to any of the folders in this workspace,
+            // getWorkspaceFolder will return null and we report this as "outside".
```
sean-mcmanus: null->undefined

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
andreeis: Yes, of course. Thanks.

sean-mcmanus (approved):

PR #7303, opened 2021-04-06
Add check and message for x64 VSIX on M1 Mac

Colengms:
Add a prompt to inform the user they are using an Intel Mac version of the extension on an M1 Mac.

This change (and changes to the release pipeline) repurpose the `install.lock` file to contain a JSON object containing the installed (or bundled) platform and architecture of native binaries.

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -99,7 +99,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
         if (installedPlatformAndArchitecture.platform === 'darwin' && installedPlatformAndArchitecture.architecture === "x64" && arch === "arm64") {
             if (promptForMacArchictureMismatch.Value) {
                 // Display a message specifically referring the user to the ARM64 Mac build on ARM64 Mac.
-                errMsg = localize("native.binaries.mismatch.osx", "This {0} version of the extension has been installed.  Since you are on an Apple Silicon Mac, we recommend installing the \"{1}\" version of the extension.");
+                errMsg = localize("native.binaries.mismatch.osx", "This {0} version of the extension has been installed.  Since you are on an Apple Silicon Mac, we recommend installing the \"{1}\" version of the extension.", GetOSName(installedPlatformAndArchitecture.platform), vsixName);
                 promptForMacArchictureMismatch.Value = false;
                 vscode.window.showErrorMessage(errMsg, downloadLink).then(async (selection) => {
                     if (selection === downloadLink) {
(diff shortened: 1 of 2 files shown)


sean-mcmanus (approved):

michelleangela (approved):

Extension/src/main.ts
```diff
@@ -75,38 +75,70 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
+                    if (selection === downloadLink) {
+                        vscode.env.openExternal(vscode.Uri.parse(releaseDownloadUrl));
+                    }
+                });
             }
-        });
+        } else {
+            // Reset the persistent boolean tracking whether to warn the user of architecture mismatch on OSX.
+            promptForMacArchictureMismatch.Value = true;
+            errMsg = localize("native.binaries.not.supported", "This {0} version of the extension is incompatible with your OS. Please download and install the \"{1}\" version of the extension.", GetOSName(installedPlatformAndArchitecture.platform), vsixName);
```
michelleangela, line 114: `GetOSName(installedPlatformAndArchitecture.platform)` only returns "darwin", also add the arch `GetOSName(installedPlatformAndArchitecture.arch)`. 

and "darwin" might not be a common name. for the actual release, should we use "macOS"?

Colengms: I changed this to refer directly to "Intel" and "Apple Silicon", since this message is only for that scenario.

Extension/src/main.ts
```diff
@@ -75,38 +75,71 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
+    const fileContents: string = await util.readFileText(util.getInstallLockPath());
+    let installedPlatformAndArchitecture: util.InstallLockContents;
+    // Just in case we're debugging with an existing install.lock that is empty, assume current platform if empty.
+    if (fileContents.length === 0) {
+        installedPlatformAndArchitecture = {
+            platform: process.platform,
+            architecture: arch
+        };
+    } else {
+        installedPlatformAndArchitecture = <util.InstallLockContents>JSON.parse(fileContents);
```
bobbrow, line 90: You should check that the properties exist instead of asserting that they are there.  People can mess with files and invalidate your assumptions.

Extension/src/main.ts
sean-mcmanus: What? It returns MacOS, right?

Extension/src/main.ts
Colengms: The values would equal undefined (or whatever they put in there), and fail to match.  The message might show incorrectly if we don't understand the platform type, but I don't think that is terrible if they are mucking with our data.  We don't really have a failure path here, other than to throw an exception.

Issue #6804, opened 2021-01-17
"Configure Default Build Task" - Task Generation Only Lists Top-Most Project As Active In Multi-Root Workspace


vindicatorr:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Linux x64 5.8.7-dirty
- VS Code Version: 1.53.0-insider (9b6aaf1e86c8fec4c86c664edf5f713583c70fe5)
- C/C++ Extension Version: v1.1.3
- Other extensions you installed (and if the issue persists after disabling them): Only C/C++ extension is enabled.
- Does this issue involve using SSH remote to run the extension on a remote machine?: N/A
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).: "Tasks: Configure Default Build Task" always lists the top-most project as the "active file" for generating a new task.

**Steps to reproduce**

1. Create a workspace
2. Add multiple root workspace projects
3. Make one of the projects NOT contain a "tasks.json" file, but do NOT make that project the top/first project. (not really necessary, but highlights the issue)
4. Tasks: Configure Default Build Task
5. Close the workspace and only open the project WITHOUT a "tasks.json" file.
6. Tasks: Configure Default Build Task

**Expected behavior**

I'd expect step 4 to show the project that doesn't have a "tasks.json" file like it does in step 6, as opposed to showing the top root workspace project for task generation (screenshots should make it clear).

<details>
  <summary><strong>Logs</strong></summary>

```
N/A (nothing showing relating to the "Configure Default Build Task" when activated)
```
</details>

**Screenshots**

Multi-root incorrect:
![Screenshot_20210117_000720_multiRoot_highlight](https://user-images.githubusercontent.com/4075850/104832782-5bc67b00-5859-11eb-9eca-4ba3e243ab46.jpg)
It looks to always be the top-most project that gets the "generate" option for tasks. It seems to always consider it the "active file".
Even when I open the "app.cpp" for the "firmware" project which already has a "tasks.json" file, it still considers the top Arduino project as the "active file".

Solo project correct:
![Screenshot_20210117_000823_solo_highlight](https://user-images.githubusercontent.com/4075850/104832778-5701c700-5859-11eb-9ff3-58cf95b4c02f.jpg)

**Additional context**

sean-mcmanus:
@elahehrashedi Do you remember if this was a known TODO for build tasks? I seem to recall it was.

[...]

PR #7315, opened 2021-04-07
Modify "Configure Task" behavior for multiroot

elahehrashedi:
Consider this multi-root workspace:
image

if I want to configure a new task for a .cpp file in Root3, the API shows  the Root1 task to the user:
image

with this change, root3 options will be available to the user (it still shows other workspace roots' tasks as well, but at least it is better than only showing the root1).
image

For a complete solution, we are waiting for vscode response. follow from here: https://github.com/microsoft/vscode-cpptools/issues/6804 and https://github.com/microsoft/vscode/issues/120133


Colengms (approved):

WardenGnaw:
nit: Please use `Squash and merge` instead of the merge commit unless each commit is an independent feature or bug fix.
image

E,g. linter error fix would not be for the whole main branch, just to the PR
image

PR #7363, opened 2021-04-15
Changing default launch.json from "${workspaceFolder}" to "${fileDirname}"

### LINKED ISSUE #7362: Change default launch configuration from "${workspaceFolder}" to "${fileDirname}", for a better debugging experience with input files out of the box
Bug type: Debugger

**System**
- OS and Version: Windows 20H2 (OS: Windows_NT x64 10.0.19042) with WSL Ubuntu G++ 9.3.0
- VS Code Version: 1.55.2 (user setup)
                               Commit: 3c4e3df9e89829dce27b7b5c24508306b151f30d
                              Date: 2021-04-13T09:35:57.887Z
                              Electron: 11.3.0
                              Chrome: 87.0.4280.141
                              Node.js: 12.18.3
                              V8: 8.7.220.31-electron.0

- C/C++ Extension Version: 1.3.0
- Other extensions you installed (and if the issue persists after disabling them): WSL extension, which is required for the above set up

**A clear and concise description of what the bug is.**

Currently the default launch.json file has a cwd value of: `${workspaceFolder}`. While this works fine a lot of the time, in some cases the debugger is unable to read variable values when taking input from files and if the file being debugged is not at the root of the workspace. This is due to it being launched at the root of the workspace and not at the location of the file being debugged. This can be rectified by changing the default launch location from `${workspaceFolder}` to `${fileDirname}`. Changing this should not have a large impact on most workflows as in most cases both the values will have the same result and if not, the file selected for debugging is likely the one that the developer is interested in.

**To Reproduce**
Sample code
_**main.cpp**_
```
#include <bits/stdc++.h>

using namespace std;

int main() {
    freopen("testin.txt", "r", stdin);
    freopen("testout.txt", "w", stdout);
    
    int a, b, ans;
    cin >> a >> b;
    ans = a * b;

    cout << ans << "\n";
}
```
The code takes 2 numbers as input from a text file, then stores their product in a variable and outputs it

_**testin.txt**_
`3 4`

_**testout.txt**_
_file should be empty_

_**Original launch.json**_
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "g++ - Build and debug active file",
            "type": "cppdbg",
            "request": "launch",
            "program": "${fileDirname}/${fileBasenameNoExtension}",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${workspaceFolder}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ],
            "preLaunchTask": "C/C++: g++ build active file",
            "miDebuggerPath": "/usr/bin/gdb"
        }
    ]
}
```

_**Modified launch.json**_
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "g++ - Build and debug active file",
            "type": "cppdbg",
            "request": "launch",
            "program": "${fileDirname}/${fileBasenameNoExtension}",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${fileDirname}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ],
            "preLaunchTask": "C/C++: g++ build active file",
            "miDebuggerPath": "/usr/bin/gdb"
        }
    ]
}
```

**Steps**
1. Set up your workspace in such a way that _main.cpp_, _testin.txt_ and _testout.txt_ are not at the root folder of the workspace
2. Place a breakpoint in main.cpp at the start of line 13 (before the cout statement)
3. Set the launch.json to the original one provided above (or however close you can without breaking anything specific to your machine and install)
4. Build and debug main.cpp, when the breakpoint hits you will see something like this:
incorrect values
As you can see, this is different from our expected values of 3, 4 and 12 respectively
5. Change the launch.json to the modified one from above (again, adjust lines to your platform if needbe, without changing the `cwd` value)
6. Delete the compiled file (I don't think this step is essential, but I needed to do it for changes to be registered when building)
7. Again, build and debug main.cpp, this time when the breakpoint hits, you will see the correct values:
image

HackintoshwithUbuntu:
Bug fix: #7362
Old behavior:
incorrect values
New behavior:
image


HackintoshwithUbuntu:
Seems like the Mac check had a network error?

sean-mcmanus:
> Seems like the Mac check had a network error?

Yeah, that's normal. Retry usually fixes it. We haven't investigated how to stop that from happening yet.

sean-mcmanus (changes requested):
Can you change the reference to `${workspaceFolder}` in vscode-cpptools\Extension\src\LanguageServer\cppBuildTaskProvider.ts ? That code is hit with Configure Default Build Task.

elahehrashedi:
I have added some comments in the original bug report: #7362

changing the cwd from `${workspaceFolder}` will break other scenarios with the cpp extension task provider.

elahehrashedi (approved):

sean-mcmanus (approved):

HackintoshwithUbuntu:
Just noticed that the vs code docs, https://github.com/microsoft/vscode-docs/tree/main/docs/cpp will also need to be updated accordingly

PR #7371, opened 2021-04-15
Attempt to fix random yarn timeout on Mac.

sean-mcmanus:


sean-mcmanus:
It gives: The workflow is not valid. .github/workflows/ci_mac.yml (Line: 24, Col: 9): Unexpected value 'continueOnError' .github/workflows/ci_mac.yml (Line: 29, Col: 9): Unexpected value 'timeoutInMinutes'

sean-mcmanus:
Looks like the GitHub actions yml schema is different from the Azure one.

sean-mcmanus:
I can't tell if this actually fixes the issue or not because I can't get it to repro currently.

michelleangela (approved):

Issue #7405, opened 2021-04-22
Address race conditions in InstallPackage()


Colengms:
In `InstallPackage()`, in `packageManager.ts`, `zipfile.on()` is used to install callbacks to manage file extraction.  In response to an "end" event, `InstallPackage()` is being completed, despite async operations that have started in response to the "entry" events not yet having completed.

Opening this issue to track ensuring `InstallPackage()` is not completed until both the "end" event has been received and all async operations that have started in response to "entry" events have also completed.

A similar issue may exist in `initVcpkgDatabase()`, which also uses `zipfile.on()`.

It's possible this may be contributing to installation issues that are occasionally reported, as we will indicate completion (by dismissing the install progress indicator) before the installation is actually complete.

Colengms:
This issue isn't as bad as I thought, now that I better understand how yauzl is used asynchronously w/lazyEntries.  There are some issues due to potentially rejecting the same promise multiple times, completing before the input file is closed, leaving the input file open on error, etc., which I can clean up.

PR #7410, opened 2021-04-23
Some cleanup of use of zipfile

Colengms:
Addresses https://github.com/microsoft/vscode-cpptools/issues/7405 , which does not actually appear to be a likely cause of install failures.

Fixes issue in which the promise could be rejected multiple times.
Ensures promise is not resolved until the file is closed (safe to be deleted).
Prevent renaming of a file to destination file name if unzip read/write of the file operation had actually failed.
Fixed an issue in which the FD of the zip file could be left open indefinitely.  (fromFd is not autoClose by default).
Fixes to initVcpkgDatabase(), such as ensuring to advance to the next entry instead of orphaning the operation if there happened to be multiple files in the VCPkgHeadersDatabase.zip.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -71,37 +71,37 @@ function initVcpkgDatabase(): Promise<vcpkgDatabase> {
                 resolve(database);
             });
             zipfile.on('entry', entry => {
-                if (entry.fileName === 'VCPkgHeadersDatabase.txt') {
-                    zipfile.openReadStream(entry, (err?: Error, stream?: Readable) => {
-                        if (err || !stream) {
-                            zipfile.close();
+                if (entry.fileName !== 'VCPkgHeadersDatabase.txt') {
+                    zipfile.readEntry();
+                    return;
+                }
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -60,49 +60,50 @@ let vcpkgDbPromise: Promise<vcpkgDatabase>;
...
-                    if (err || !stream) {
-                        resolve({});
-                        return;
-                    }
-                    const database: vcpkgDatabase = {};
-                    const reader: rd.ReadLine = rd.createInterface(stream);
-                    reader.on('line', (lineText: string) => {
-                        const portFilePair: string[] = lineText.split(':');
-                        if (portFilePair.length !== 2) {
+                if (entry.fileName === 'VCPkgHeadersDatabase.txt') {
```
sean-mcmanus: Can you use

```
                if (entry.fileName !== 'VCPkgHeadersDatabase.txt') {
                    zipFile.readEntry();
                    return;
                }
```

?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

Issue #7433, opened 2021-04-27
Onboard new localization pipeline


Colengms:
Opening this issue to track work needed internally to adopt new localization pipeline/process.

PR #7441, opened 2021-04-30
Switch to new localization pipeline

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7433

The import script now pulls translations from the Loc repo instead of a file share.  Multi-checkout was not working for me, possible due to spanning GitHub and DevOps.  So, it clones the loc repo using a PAT.

The import script was also updated to work properly when deployed via YML.

The export script now uses the OneLocBuild DevOps.  (I think I will also need to point them to the build pipeline that uses it, after this is checked in, as it just generates artifacts that I think they will need to pick up).

Removed a "\n" from a description string, as somewhere in the pipeline a "\r" is also being added to that.  No other descriptions add new lines, so removed that instance.


Extension/translations_auto_pr.js
```diff
@@ -66,22 +70,16 @@ function sleep(ms) {
...
 directories.forEach(folderName => {
-    let sourcePath = `${rootSourcePath}\\${folderName}\\Projects\\Src\\vscode-extensions\\vscode-${locProjectName}.${folderName}.xlf`;
+    let sourcePath = `${rootSourcePath}\\${folderName}\\vscode-cpptools.xlf`;
     let destinationPath = `../vscode-translations-import/${folderName}/vscode-extensions/vscode-${locProjectName}.xlf`;
-    console.log(`Copying "${sourcePath}$" to "${destinationPath}"`);
+    console.log(`Copying "${sourcePath}" to "${destinationPath}"`);
     fs.copySync(sourcePath, destinationPath);
 });
 
+
```
sean-mcmanus: Is this extra newline intentional?

Extension/translations_auto_pr.js
```diff
@@ -115,14 +113,21 @@ if (!hasAnyChanges()) {
     return;
 }
 
-// Commit changes files.
-console.log(`Commiting changes (git commit -m "${commitComment}")`);
-cp.execSync(`git commit -m "${commitComment}"`);
+// Set up user and permissions(Never run this locally)
```
sean-mcmanus: What happens if it's run locally? Should there be some guard against that in case?

sean-mcmanus (approved):

michelleangela:
Is there public documentation for how the "new localization pipeline" works that we can save/reference in our internal docs for future understanding or maintenance?

Extension/translations_auto_pr.js
```diff
@@ -15,9 +15,12 @@ let repoName = process.argv[3];
 let locProjectName = process.argv[4];
 let authUser = process.argv[5];
 let authToken = process.argv[6];
+let userFullName = process.argv[7];
+let userEmail = process.argv[8];
+let locRepoPath = process.argv[9];
 
 if (!repoOwner || !repoName || !locProjectName || !authUser || !authToken) {
```
michelleangela: Also add check for the new vars `user_full_name` and `user_email` and `loc_repo_path`? Or add comment that states which args are required at minimum to be correct?

Extension/translations_auto_pr.js
Colengms: I set the values using script args, so if run locally the right local values could be passed in.  I had to set them as there is an error if unset on commit, on the build vm.  I'm not sure how to guard against it.  I could do so more digging to find a way.

Colengms:
> Is there public documentation for how the "new localization pipeline" works that we can save/reference in our internal docs for future understanding or maintenance?

Unfortunately, no, not really.  I'm not getting a lot of support from the loc team, so I've actually been struggling to ensure we can continue to keep this process automated.  There is this: https://dev.azure.com/ceapex/CEINTL/_git/SoftwareLocalization?path=%2Fsrc%2FOneLocBuild  (Which Jonathan referred me to, and the Loc team did not).  Which we're able to leverage to generate files for submitting into Juno, but we can't use its functionality to generate PRs, as that only works for DevOps repos and would not do the additional step we need of actually ingesting the strings into the i18n directory.  I did get some 'instructions' from the loc team, which arrived after I had composed scripts to do the same things.

Build/loc/TranslationsImport.yml
```diff
@@ -0,0 +1,33 @@
...
+    workingDirectory: '$(Build.SourcesDirectory)'
+
+- task: CmdLine@2
+  inputs:
+    script: 'cd Extension && yarn install'
+    workingDirectory: '$(Build.SourcesDirectory)'
+
+- task: CmdLine@2
+  inputs:
+    script: 'cd Extension && node ./translations_auto_pr.js microsoft vscode-cpptools cpptools csigs $(csigsPat) "cpptools service account" cpptools@microsoft.com $(Build.SourcesDirectory)/../VSCodeExt'
```
bobbrow: Can you remove the direct reference to the internal account as well?  We are retiring this account anyway.

Build/loc/TranslationsExport.yml
```diff
@@ -0,0 +1,39 @@
+# ==================================================================================
+# Pipeline for VsCodeExtension-Localization build definition
+# Runs OneLocBuild task to localize xlf file
+#
+# Triggered only by daily schedule
```
sean-mcmanus: This comment seems no longer valid.

sean-mcmanus (changes requested):

Build/loc/TranslationsImport.yml
Colengms: I need to set the email and user name of the git commit to something.  That something will be displayed in the (public) git log, so there is no use in hiding it here.  Do you have a suggestion on what that user name and email should be?

sean-mcmanus (approved):

Build/loc/TranslationsImport.yml
Colengms: Previously, we were picking up the default configured git user name and email from the build machine.  i.e.:
DevDiv Build Lab - Dev14 <dlab14@microsoft.com>

But, when using YML, the build machine is not configured with a user name or email, and cannot commit unless I set them.

Provided that the PR is squashed when committed, that info will be replaced with the name and email of associated with the PR (csigs) :

csigs <csigs@users.noreply.github.com>

I suppose I could use that same account information for the commit.

michelleangela (approved):

bobbrow (approved):

PR #7463, opened 2021-05-04
Consolidate localization export and import into a single yml

Colengms:
Instead of using a repo provided by the Loc team to retrieve localized strings, this approach picks up those same strings as generated by the OneLocBuild task, so in the same pipeline as strings are being exported.  

With this change, I'm checking in our LCL files.  These are internal files used by the loc system, which are processed to generate XLF files, which are then ingested (into i18n directory, etc.).  LCL files will be *directly submitted* into our repo by csigs, but we will still receive PRs to manage ingesting strings in the format used by cpptools.


Colengms:
Cancelling this for now, as we are investigating the option of having the loc team host these LCL files for us.

PR #7466, opened 2021-05-05
Update ESLint.

sean-mcmanus:


michelleangela (approved):

sean-mcmanus:
We could potentially add more cases to @typescript-eslint/naming-convention later, but I didn't want to change a lot of existing code that is using inconsistent casing (e.g. camel versus pascal versus snake case, leading underscore etc.).

PR #7479, opened 2021-05-06
Consolidate localization import and export into same YML.  Use LCL hosted by the Loc team.

Colengms:
Instead of using a repo provided by the Loc team to retrieve localized strings, this approach picks up those same strings as generated by the OneLocBuild task, so in the same pipeline as strings are being exported.


Extension/translations_auto_pr.js
```diff
@@ -13,24 +13,34 @@ const pullRequestTitle = '[Auto] Localization - Translated Strings';
...
+    console.error(`   loc_root_path - The path to the folder with language-specific directories (containing localized xlf files).`);
+    console.error(`   loc_sub_path - A sub-path after the language-specific directory, where the xlf to import is located.  This should not include the name of the xlf file to import.)`);
     return;
 }
 
 console.log(`repoOwner=${repoOwner}`);
 console.log(`repoName=${repoName}`);
-console.log(`locProjectName=${locProjectName}`);
 console.log(`authUser=${authUser}`);
 console.log(`authToken=${authToken}`);
```
bobbrow: Credentials/secrets should not appear in logging.

Colengms: They are actually automatically stripped out by the build pipeline. i.e.:

```
authUser=csigs
auth***=***
```

Colengms: Similarly, when passed as an arg to a command line, the echoing of that command line is masked:
```
cd Extension && node ./translations_auto_pr.js microsoft vscode-cpptools csigs *** csigs csigs@users.noreply.github.com "D:\a\1\a/loc" vscode-extensions-localization-export/vscode-extensions && cd ..
```

bobbrow (approved):

PR #7486, opened 2021-05-07
Add capabilities.

sean-mcmanus:
Handle https://github.com/microsoft/vscode-cpptools/issues/7483 (virtualWorkspaces) and https://github.com/microsoft/vscode/issues/120251 (untrustedWorkspaces)


sean-mcmanus:
Switching "untrustedWorkspaces" from false to "limited" is a TODO task for 1.5.0.

Extension/package.nls.json
```diff
@@ -1,4 +1,5 @@
 {
+    "c_cpp.capabilities.untrustedWorkspaces.description": "Multiple settings can cause processes local to the workspace to be executed, e.g. C_Cpp.clang_format_path, C_Cpp.addNodeAddonIncludePaths, C_Cpp.default.compilerPath, C_Cpp.default.configurationProvider, and C_Cpp.default.compileCommands, and the equivalent properties in c_cpp_properties.json.",
```
bobbrow, line 2: ~~I don't think configuration provider counts because the trust support is delegated to that extension, and we don't explicitly activate it for the user.~~

Do we also need to mention the debugger paths supported by `launch.json`?

bobbrow: Well, I take it back about config provider since it does send a compiler path back to us

bobbrow: The question is whether we can trust the config provider?

sean-mcmanus: No -- debugger stuff is already disabled with untrusted workspaces.

sean-mcmanus: If they have a configuration provider set, then the workspace itself (cmake or makefile) could reference a compiler path that is local to the workspace folder, so technically the cmake or makefile extensions would not be the ones that are executing the untrusted code, it would be our extension.

Colengms (approved):

PR #7487, opened 2021-05-07
check if the directory exists

elahehrashedi:
Some of the `checkFileExists` are replaced with `checkDirectoryExists`.
This will solve the issue for syncing between the configuration content in JSON and UI

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -173,7 +173,7 @@ export async function activate(activationEventOccurred: boolean): Promise<void> 
     if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
         for (let i: number = 0; i < vscode.workspace.workspaceFolders.length; ++i) {
             const config: string = path.join(vscode.workspace.workspaceFolders[i].uri.fsPath, ".vscode/c_cpp_properties.json");
-            if (await util.checkFileExists(config)) {
+            if (await util.checkDirectoryExists(config)) {
                 cppPropertiesExists = true;
                 const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(config);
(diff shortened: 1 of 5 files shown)


michelleangela (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -173,7 +173,7 @@ export async function activate(activationEventOccurred: boolean): Promise<void>
     if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
         for (let i: number = 0; i < vscode.workspace.workspaceFolders.length; ++i) {
             const config: string = path.join(vscode.workspace.workspaceFolders[i].uri.fsPath, ".vscode/c_cpp_properties.json");
-            if (await util.checkFileExists(config)) {
+            if (await util.checkDirectoryExists(config)) {
```
sean-mcmanus: Isn't this a file and not a directory?

Extension/src/packageManager.ts
```diff
@@ -384,7 +384,7 @@ export class PackageManager {
                             zipfile.readEntry();
                         });
                     } else {
-                        util.checkFileExists(absoluteEntryPath).then((exists: boolean) => {
+                        util.checkDirectoryExists(absoluteEntryPath).then((exists: boolean) => {
```
sean-mcmanus: Isn't this a file and not a directory? The "if" part above appears to handle directories.

sean-mcmanus (changes requested):

Extension/src/packageManager.ts
elahehrashedi: I am not sure about this one.

Extension/src/LanguageServer/extension.ts
elahehrashedi: correct.

sean-mcmanus (approved):

PR #7512, opened 2021-05-11
Add Getting Started Walkthrough

### LINKED ISSUE #7273: Getting Started experience
VS Code is working on a "getting started" workflow (https://github.com/microsoft/vscode/issues/118402) that would enable us to plug in and provide information for new and returning users.  This would make it easier for folks to get onboarded to using the extension with links to tutorials, etc.

Colengms:
Adds 'walkthrough' content to package.json.  That content is currently in English.  Because we don't expect to get localized strings before shipping 1.4.0, we plan to stick with English only content for now.  (Also, I will need to update localization scripts to enable localization of the MD files).  This walkthrough is mostly static, and the plan is to update it considerably as additional support we've requested is added to VS Code.

Currently VS Code Insiders is needed to see the Getting Started UI, as well as setting "workbench.welcomePage.experimental.extensionContributions": true .

Added args to Configuration UI/JSON commands to allow display within a specified ViewColumn.  This allows opening of the Configuration UI 'beside' the walkthrough.

Added a CheckForCompiler command that is triggered by the walkthrough and displays a dialog listing found compilers, or a warning that no compilers were found.

Resolves https://github.com/microsoft/vscode-cpptools/issues/7273


Extension/package.json
```diff
@@ -42,6 +42,168 @@
   ],
   "main": "./dist/main",
   "contributes": {
+    "walkthroughs": [
+      {
+        "id": "cppWelcome",
+        "title": "Get Started with C++ Development",
```
sean-mcmanus, line 55: Is this supposed to be "Getting"?

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+        "description": "Your first steps to set up powerful C++ tools in a lightweight, performant editor!",
+        "steps": [
```
sean-mcmanus, line 57: Using the "powerful" adjective seems odd to me...maybe "helpful" would be better if it's referring to IntelliSense? Or if "tools" are supposed to include the compiler maybe there should be no adjective at all. Compilers are required tools, but I wouldn't normally describe them as "powerful".

Extension/package.json
```diff
@@ -42,6 +42,168 @@
   ],
   "main": "./dist/main",
   "contributes": {
+    "walkthroughs": [
+      {
+        "id": "cppWelcome",
+        "title": "Get Started with C++ Development",
+        "description": "Your first steps to set up powerful C++ tools in a lightweight, performant editor!",
```
sean-mcmanus: Use of the adjective "performant" seems unnecessarily verbose...maybe "fast, lightweight editor" would be better?

But in some cases C++ IntelliSense may not be fast (due to implementation details requiring tons of parsing of headers or having to resolve recursive includes across thousands of folders, etc.), particularly compared to TypeScript. so I'm hesitant to setup the fast or performant expectation.

sean-mcmanus: I believe "!" marks are supposed to be avoided in a technical context.

Extension/package.json
sean-mcmanus, line 55: What about users targeting C development?

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+          {
+            "id": "verify.compiler.mac",
+            "title": "Install a C++ compiler",
+            "description": "The C++ extension uses the C++ compiler on your computer to configure IntelliSense for your project. [Check for C++ compilers](command:C_Cpp.CheckForCompiler)",
+            "when": "isMac",
+            "doneOn": {
+              "command": "C_Cpp.CheckForCompiler"
+            },
+            "media": {
+              "altText": "markdown file with instructions for installing Clang on macOS",
```
sean-mcmanus: Is altText supposed to be formatted as a sentence starting with a capital letter and ending with a period?

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+          },
+          {
+            "id": "setup.intellisense",
+            "title": "Configure C++ IntelliSense",
+            "description": "Verify important IntelliSense settings like Compiler Path, Compiler Args, IntelliSense Mode, and Include Path. \n[Open IntelliSense Configuration](command:C_Cpp.ConfigurationEditUI?-2)",
+            "doneOn": {
+              "command": "C_Cpp.ConfigurationEditUI"
+            },
+            "media": {
+              "altText": "screenshot that shows IntelliSense configurations UI with important settings highlighted",
```
sean-mcmanus: "shows the IntelliSense"

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+              }
+            }
+          },
+          {
+            "id": "build.project.mac",
+            "title": "Build your C++ file",
+            "description": "Create a default build task for your project. \n[Learn more](https://code.visualstudio.com/docs/cpp/config-clang-mac#_build-helloworldcpp)",
+            "when": "isMac",
+            "media": {
+              "altText": "markdown file with instructions for configuring default build task on Mac with Clang",
```
sean-mcmanus: "configuring the default"...same with the additional cases below...

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            "description": "Create a debug configuration for your project. \n[Learn more](https://code.visualstudio.com/docs/cpp/config-msvc#_debug-helloworldcpp)",
+            "when": "isWindows",
+            "media": {
+              "altText": "markdown file with instructions for starting a debugging session on Windows",
+              "path": "walkthrough/run-and-debug-project-windows.md"
+            }
+          },
+          {
+            "id": "install.cmake.tools",
+            "title": "Install CMake Tools",
```
sean-mcmanus, line 177: "build and debug your CMake project" (or something else that is grammatically correct)

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+              "path": {
+                "light": "walkthrough/cmake-tools-light.png",
+                "dark": "walkthrough/cmake-tools-dark.png",
+                "hc": "walkthrough/cmake-tools-hc.png"
+              }
+            }
+          },
+          {
+            "id": "watch.tutorials",
+            "title": "Lean back and get started",
```
sean-mcmanus, line 193: Some people with different backgrounds may not understand what this title means in reference to "lean back". Maybe it could be changed to something more literal that refers to watching videos?

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            "description": "Watch this series of short and practical videos about setting up C++ IntelliSense, building, and debugging C++ projects. \n[Watch Tutorials](https://code.visualstudio.com/docs/cpp/introvideos-cpp)",
+            "media": {
+              "altText": "screenshot with preview of video tutorial and play button",
```
sean-mcmanus: The grammar seems a little too informal, maybe something like "with a preview of the video tutorial and play button".

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,23 @@ export class DefaultClient implements Client {
         }
     }
 
+    public async handleCheckForCompiler(): Promise<void> {
+        await this.awaitUntilLanguageClientReady();
```
sean-mcmanus, line 2665: Unrelated to this PR, but this naming seems odd since the method itself doesn't "await"...not sure what a better name would be though.

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,23 @@ export class DefaultClient implements Client {
...
+        const compilers: configs.KnownCompiler[] | undefined = await this.getKnownCompilers();
+        if (!compilers || compilers.length === 0) {
+            const compilerName: string = process.platform === "win32" ? "MSVC" : (process.platform === "darwin" ? "Clang" : "GCC");
+            vscode.window.showInformationMessage(localize("no.compilers.found", "No C++ compilers were found on your system. Install {0} using the instructions in the editor.", compilerName), { modal: true });
```
sean-mcmanus: What does "instructions in the editor" mean?

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,23 @@ export class DefaultClient implements Client {
...
+        } else {
+            const header: string = localize("compilers.found", "We found the following C++ compiler(s) on your system:");
+            let message: string = header + "\n";
+            compilers.forEach(compiler => {
+                message += "\n" + compiler.path;
+            });
+            message += "\n\n" + localize("compilers.found.message", "You can specify which compiler to use in your project's IntelliSense Configuration.");
```
sean-mcmanus: Seems like Configuration should be lowercase. Are users going to know what "your project's IntelliSense configuration" means?

sean-mcmanus: Is this message valid if only 1 compiler is found? Maybe there could be a separate "else if" for the 1 compiler case?

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,23 @@ export class DefaultClient implements Client {
         }
     }
 
+    public async handleCheckForCompiler(): Promise<void> {
+        await this.awaitUntilLanguageClientReady();
+        const compilers: configs.KnownCompiler[] | undefined = await this.getKnownCompilers();
+        if (!compilers || compilers.length === 0) {
+            const compilerName: string = process.platform === "win32" ? "MSVC" : (process.platform === "darwin" ? "Clang" : "GCC");
```
sean-mcmanus, line 2668: I'm concerned about potentially confusing users by referring to only MSVC on Windows, Clang on Mac, and GCC on Linux, when Clang/GCC are available on all 3 OS's...maybe add some comment about those being "recommended" as a default for your OS if you don't know which one to choose?

Extension/src/LanguageServer/settingsPanel.ts
```diff
@@ -82,10 +82,12 @@ export class SettingsPanel {
...
-    public createOrShow(configSelection: string[], activeConfiguration: config.Configuration, errors: config.ConfigurationErrors): void {
-        const column: vscode.ViewColumn | undefined = vscode.window.activeTextEditor
-            ? vscode.window.activeTextEditor.viewColumn
-            : undefined;
+    public createOrShow(configSelection: string[], activeConfiguration: config.Configuration, errors: config.ConfigurationErrors, viewColumn?: vscode.ViewColumn): void {
+        const column: vscode.ViewColumn | undefined = viewColumn !== undefined
+            ? viewColumn
+            : vscode.window.activeTextEditor
+                ? vscode.window.activeTextEditor.viewColumn
+                : undefined;
```
sean-mcmanus: `viewColumn ?? vscode.window.activeTextEditor?.viewColumn;`

Extension/walkthrough/build-project-linux.md
```diff
@@ -0,0 +1,13 @@
+# Build your C++ file with GCC on Linux
+
+1. Open the C++ source file that you want to build. Make sure this file is active (currently displayed) in the editor.
```
sean-mcmanus: "Currently displayed" doesn't mean active, i.e. there can be multiple files displayed but only one is active.

sean-mcmanus (changes requested):

Extension/package.json
Colengms: This text was copied (and modified to refer to C++) from the text used by the Java extension.

Extension/walkthrough/run-and-debug-project-linux.md
```diff
@@ -0,0 +1,13 @@
...
+
+1. Open the C++ source file that you want to run and debug. Make sure this file is active (currently displayed) in the editor.
+
+2. From the main menu, select **Run > Add Configuration...** and choose **C++ (GDB/LLDB)**.
+
+3. Choose **g++ build and debug active file.**
+
+    ![Dropdown showing C++ debug configurations for GCC on Linux](linux-build-and-debug-active-file.png)
+
+4. To start a debugging session, Press `F5` or from the main menu choose **Run > Start Debugging**.
```
sean-mcmanus: Lowercase "press"

Extension/package.json
sean-mcmanus: See https://docs.microsoft.com/en-us/style-guide/punctuation/exclamation-points .

Extension/package.json
sean-mcmanus: Can you change the highlighting color to green? It's currently red (usually red is used to highlight an error).

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+    "walkthroughs": [
+      {
+        "id": "cppWelcome",
+        "title": "Get Started with C++ Development",
+        "description": "Your first steps to set up powerful C++ tools in a lightweight, performant editor!",
+        "steps": [
+          {
+            "id": "verify.compiler.mac",
+            "title": "Install a C++ compiler",
+            "description": "The C++ extension uses the C++ compiler on your computer to configure IntelliSense for your project. [Check for C++ compilers](command:C_Cpp.CheckForCompiler)",
```
sean-mcmanus: There is no \n here but the other 2 OS's uses a \n, before the `[Check`

sean-mcmanus: "C++ extension" should be "C/C++ extension".

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            },
+            "media": {
+              "altText": "markdown file with instructions for installing MSVC on Windows",
+              "path": "walkthrough/install-compiler-windows.md"
+            }
+          },
+          {
+            "id": "open.project",
+            "title": "Open your project",
+            "description": "Open an existing folder that contains your C++ project, or create a new folder. \n[Open Folder](command:workbench.action.files.openFolder)",
```
sean-mcmanus: Is referring to "C++ project" going to confuse anyone who may associate that with VS C++ project files or users who may wonder "how do I get a C++ project" in the first place? Should it be "C++ files" and should the title be "Open your C++ folder"? Since VS Code lacks a concept of "projects" it seems clearer to remove all references to the word?

sean-mcmanus: Maybe if we show users how to create a workspace we could change all references of project to workspace? But introducing the workspace concept seems unnecessary unless multiple folders are needed.

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            "when": "isWindows",
+            "media": {
+              "altText": "markdown file with instructions for starting a debugging session on Windows",
+              "path": "walkthrough/run-and-debug-project-windows.md"
+            }
+          },
+          {
+            "id": "install.cmake.tools",
+            "title": "Install CMake Tools",
+            "description": "Do you build your project with CMake? Install the CMake Tools extension to seamlessly build and debug CMake project. \n[Install CMake Tools](command:workbench.extensions.installExtension?%22ms-vscode.cmake-tools%22)",
```
sean-mcmanus: Should we add a reference to Make and the Makefile Tools extension too?

Extension/package.json
jureid: Following the pattern VS Code has with their other walkthroughs, it should be "Get" (for example, they have "Learn the fundamentals" instead of "Learning the fundamentals" and "Customize your setup" instead of "Customizing")

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+                "light": "walkthrough/cmake-tools-light.png",
+                "dark": "walkthrough/cmake-tools-dark.png",
+                "hc": "walkthrough/cmake-tools-hc.png"
+              }
+            }
+          },
+          {
+            "id": "watch.tutorials",
+            "title": "Lean back and get started",
+            "description": "Watch this series of short and practical videos about setting up C++ IntelliSense, building, and debugging C++ projects. \n[Watch Tutorials](https://code.visualstudio.com/docs/cpp/introvideos-cpp)",
```
sean-mcmanus: Should be something like "setting up C++ IntelliSense and building and debugging C++ projects", "setting up C++ IntelliSense, building and debugging C++ projects, and `<some other topic>`", or "configuring, building, and debugging C++ projects". The reason is because in the original text the comma is used as a delimiter of a sequence, but the 2nd item, "building", is detached from the required object, "C++ projects".

Extension/package.json
jureid: Yes, we were following what Java does. But to Sean's point, I don't see exclamation marks in any of the other VS Code walkthroughs. And on second thought, I think we can make this description simpler. Let's change it to this:

"Dive into VS Code's rich C++ development experience."

Thoughts? We can iterate on this description in A/B experiments down the road.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Can we add an option for users to choose which type (MSVC/Clang/gcc) of compiler they want to use? Or is that better to leave for a V2?

Extension/package.json
sean-mcmanus: Should the root level "question" be something like "Do you want to do C or C++ development?" or is that better for a V2 or not possible yet?

Extension/package.json
jureid: I think we should keep "Install CMake Tools" as the title for now to diversify the titles of the tasks a bit, since the previous two start with "Build" and "Debug." Especially since this task doesn't actually tell users how to build or debug their CMake projects, it just tells them to install the extension. 

Down the road, we will only show the step about CMake Tools if the customer has a CMakeLists.txt in their project. And in that case, we'll hide the steps for building a C++ file and debugging a C++ file, because they wouldn't need to know how to create tasks.json and launch.json if they're using the CMake Tools extension. When we do that, I agree we should change the title to "Build and Debug your CMake project with CMake Tools" or something like that.

Extension/package.json
jureid: Honestly, not sure, but couldn't hurt to do so :) . Happy to change it to a full sentence

Extension/package.json
jureid: I agree that "configure" is technically more correct, but I think for beginners "create" is less overwhelming. "Configure" implies a lot of manual adjusting and makes it seem more complicated, in my opinion.

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            "media": {
+              "altText": "screenshot with preview of video tutorial and play button",
+              "path": "walkthrough/getting-started-video.PNG"
```
sean-mcmanus: Not sure if it matters, but the file extension is PNG here, but png in other cases. Seems simpler to stick with all lowercase?

Extension/package.json
jureid: That's a good point. This wording was copied from one of VS Code's walkthroughs, and they've seen high success with this task / wording. And since we're only shipping an English version to start, I think we should keep it for now. But I'll ask the VS Code team if they've considered this point. I wonder if the localization team would take care of this when they localize our content...

Extension/package.json
jureid: Thanks-- will fix

Extension/package.json
jureid: will update

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            }
+          },
+          {
+            "id": "install.cmake.tools",
+            "title": "Install CMake Tools",
+            "description": "Do you build your project with CMake? Install the CMake Tools extension to seamlessly build and debug CMake project. \n[Install CMake Tools](command:workbench.extensions.installExtension?%22ms-vscode.cmake-tools%22)",
+            "doneOn": {
+              "command": "workbench.extensions.installExtension"
+            },
+            "media": {
```
sean-mcmanus, line 182: ...of the CMake...

Extension/src/LanguageServer/client.ts
jureid: a markdown file with instructions is displayed in the editor (VS Code) next to the walkthrough

Extension/src/LanguageServer/client.ts
jureid: We capitalize Configuration everywhere else in VS Code when referencing IntelliSense Configurations. I think customers who aren't familiar with IntelliSense Configurations will be once they see the "Configure C++ IntelliSense" step next in the walkthrough.

jureid: Good call!

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+              "altText": "markdown file with instructions for installing GCC on Linux",
+              "path": "walkthrough/install-gcc-linux.md"
+            }
+          },
+          {
+            "id": "verify.compiler.windows",
+            "title": "Install a C++ compiler",
+            "description": "The C++ extension uses the C++ compiler on your computer to configure IntelliSense for your project. \n[Check for C++ compilers](command:C_Cpp.CheckForCompiler)",
```
sean-mcmanus: In the remote scenario, the remote computer's compilers are used...not sure if that's worth mentioning or if it would just complicate things.

Extension/package.json
jureid: In blog posts and conference talks we just say C++, and the audience seems to understand that it encompasses both C++ and C. I want to be consistent with our recent branding of calling it "the C++ extension," but I think this would be a good thing to experiment with!

Extension/src/main.ts
```diff
@@ -488,6 +488,7 @@ function rewriteManifest(): Promise<void> {
         "onCommand:C_Cpp.GenerateEditorConfig",
         "onCommand:C_Cpp.GoToNextDirectiveInGroup",
         "onCommand:C_Cpp.GoToPrevDirectiveInGroup",
+        "onCommand:C_Cpp.CheckForCompiler",
```
sean-mcmanus, line 491: This should be added to commandsToRegister in commands.ts.

sean-mcmanus (changes requested):

Extension/package.json
sean-mcmanus: Ok

Extension/src/LanguageServer/client.ts
jureid: That's a good point... instead of "Install {compilerName}..." we can say "For your platform, we recommend installing {compilerName}...".

Extension/walkthrough/build-project-linux.md
jureid: that's true.. "displayed and selected" is more accurate. Will update.

Extension/walkthrough/run-and-debug-project-linux.md
jureid: good catch

Extension/src/main.ts
sean-mcmanus: Actually, it looks like it doesn't belong in commandsToRegister since it isn't callable via the command palette.

Extension/package.json
jureid: Great catch! There should be one

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,23 @@ export class DefaultClient implements Client {
...
+            vscode.window.showInformationMessage(localize("no.compilers.found", "No C++ compilers were found on your system. Install {0} using the instructions in the editor.", compilerName), { modal: true });
+        } else {
+            const header: string = localize("compilers.found", "We found the following C++ compiler(s) on your system:");
```
sean-mcmanus, line 2671: This list will contain C and C++ compilers, i.e. gcc and g++ -- is that intentional or did we want to filter out the C versions here?

Extension/package.json
sean-mcmanus: According to https://www.towson.edu/about/accessibility/technology/alt-text.html it's recommended to use normal punctuation and sentence structuring.

Extension/package.json
jureid: We can! I'll try to do that if I have time today

Extension/package.json
jureid: Copying reply to your previous comment about this:

In blog posts and conference talks we just say C++, and the audience seems to understand that it encompasses both C++ and C. I want to be consistent with our recent branding of calling it "the C++ extension," but I think this would be a good thing to experiment with!

Extension/package.json
jureid: In the future, yes! We need a way to detect presence of a Makefile in a workspace.

Extension/package.json
jureid: Yeah, I struggled with that comma placement because I was worried that "and building and debugging" looked weird. But I agree that "setting up C++ IntelliSense and building and debugging C++ projects" is better. I'll change this

Extension/package.json
jureid: Yeah, let's use "system" instead of "computer" because that makes it more versatile.

Extension/package.json
jureid: Will update those file extensions if I have time

Extension/src/LanguageServer/client.ts
jureid: Intentional

Extension/package.json
jureid: True. Let's follow what VS Code has in their walkthrough which does a better job addressing this:
Title: "Open your project folder"
Button: "Pick a folder"

Extension/package.json
sean-mcmanus: Not sure about "rich", but that's okay if you want.

Extension/package.json
sean-mcmanus: Yeah, I noticed other text below used "system" too, so that'd be more consistent.

Extension/package.json
sean-mcmanus: ...my comment got placed on the wrong like, it's referring to `"screenshot of CMake Tools extension page",`

Extension/package.json
sean-mcmanus: Ok

Extension/src/LanguageServer/client.ts
sean-mcmanus: Ah, I haven't reviewed that file yet...

Extension/walkthrough/install-compiler-windows.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. To install MSVC, download **C++ Build Tools** from the Visual Studio [Downloads](https://visualstudio.microsoft.com/downloads#other) page. 
+
+2. In the Visual Studio Installer, check the **C++ build tools** workload and select **Install**.
+
+   >**Note**: You can use the C++ toolset from Visual Studio Build Tools along with Visual Studio Code to compile, build, and verify any C++ codebase as long as you also have a valid Visual Studio license (either Community, Pro, or Enterprise) that you are actively using to develop that C++ codebase.
+
+3. Check your MSVC installation by typing `cl` into the **Developer Command Prompt for VS**. You should see a copyright message with the version and basic usage description.
+
+   ![Developer Command Prompt](developer-cmd-prompt-menu.png)
```
sean-mcmanus: Your screenshot shows the VS 2017 command prompt. The VS 2019 one would be preferred.

Extension/walkthrough/install-compiler-windows.md
```diff
@@ -0,0 +1,18 @@
...
+
+
+   >**Note**: To use MSVC from the command line or VS Code, you must run from a **Developer Command Prompt for VS**. An ordinary shell such as PowerShell, Bash, or the Windows command prompt does not have the necessary path environment variables set
```
sean-mcmanus: Seems like this should end in a period? "...variables set."

Extension/walkthrough/install-gcc-linux.md
```diff
@@ -0,0 +1,20 @@
...
+    ```bash
+    sudo apt-get update
+    ```
+
+2. Install the GNU compiler tools and the GDB debugger with this command:
+
+    ```bash
+    sudo apt-get install build-essential gdb
+
+3. Verify GCC is installed by running the following command. You should see a copyright message with the version and basic usage description.
```
sean-mcmanus: I don't see a "basic usage description":
```
~$ gcc --version
gcc (Ubuntu 10.3.0-1ubuntu1~20.04~2) 10.3.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

sean-mcmanus:
select-build-task-*.png files don't appear to be used -- should those be deleted or is a reference missing?

Extension/walkthrough/install-clang-macos.md
```diff
@@ -0,0 +1,13 @@
+# Install a C++ compiler on macOS
+
+If you're doing C++ development for macOS, we recommend installing the Clang compiler. All you need to do is run the following command in a Terminal window to install the command line developer tools:
+
+```bash
+xcode-select --install
+```
+
+Then, to verify that clang is installed, run the following command in a Terminal window. You should see a copyright message with the version and basic usage description.
```
sean-mcmanus: I don't see a copyright message or basic usage:
```
~$ clang --version
clang version 10.0.0-4ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
just the version info. Well, that's on Linux, but Mac is the same other than the version, target, etc. being different.

sean-mcmanus (changes requested):

Extension/walkthrough/run-and-debug-project-windows.md
```diff
@@ -0,0 +1,13 @@
+# Run and debug your C++ file on Windows
+
+1. Open the C++ source file that you want to run and debug. Make sure this file is active (currently displayed) in the editor.
+
+2. From the main menu, select **Run > Add Configuration...** and choose **C++ (Windows)**.
```
sean-mcmanus: This (Add Configuration) only works if launch.json doesn't exist already (that might be a bug/oversight)...not sure if that matters in the context of this walkthrough though. Deleting any existing launch.json is a workaround.

Extension/package.json
sean-mcmanus: The image has a play button, so I thought it was a video I could click but clicking it didn't work...not sure if users would hit that in the context of the walkthrough though.

Extension/package.json
sean-mcmanus: Was this Configurations supposed to be capital too to be consistent with the other usage?

Extension/package.json
sean-mcmanus: I don't understand, my comment was in reference to the string "markdown file with instructions for configuring default build task on Mac with Clang" and I was not referring to "create" at all.

Extension/package.json
sean-mcmanus: You misunderstood my comment (GitHub might be placing them at the wrong spot). I was referring to the descripton text and not the title.

sean-mcmanus:
Can we use a different icon that is bigger and easier to see the C/C++ text? The C/C++ text is too small for me to see (GitHub is magnifiying the image): 
image

sean-mcmanus:
I clicked the "X" hide button on the left and now I can't get it back, i.e. I can't try out the walkthrough anymore :(

sean-mcmanus:
> I clicked the "X" hide button on the right and now I can't get it back, i.e. I can't try out the walkthrough anymore :(

Configure Welcome Page Content at the bottom fixed it :)

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,25 @@ export class DefaultClient implements Client {
         }
     }
 
+    public async handleCheckForCompiler(): Promise<void> {
+        await this.awaitUntilLanguageClientReady();
+        const compilers: configs.KnownCompiler[] | undefined = await this.getKnownCompilers();
```
sean-mcmanus, line 2666: @Colengms These compilers don't have the correct slash direction -- can you fix that via something like what is done in the setKnownCompilers method? Maybe the code should be refactored to do the slash direction fix earlier? Not sure, since doing it later allows the slash direction to change?

sean-mcmanus (changes requested):

sean-mcmanus:
How do I see the .md files in the UI? I don't see any.

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+            "when": "isWindows",
+            "doneOn": {
```
sean-mcmanus: @Colengms I'm getting squiggles: "doneOn is deprecated. By default steps will be checked off when their buttons are clicked, to configure further use completionEvents""

Extension/package.json
jureid: oh, my bad. I thought you were suggesting to change "create a default" to "configuring the default" in the title. Updating the altText now :)

Extension/package.json
jureid: Got it, sorry about that misunderstanding! GitHub is showing the id and title for this comment. I'll update the description to include "build and debug" not just "build" -- good idea

Extension/walkthrough/install-compiler-windows.md
jureid: Good catch, fixing

Extension/package.json
jureid: Got it, updating

jureid:
> select-build-task-*.png files don't appear to be used -- should those be deleted or is a reference missing?
> 
> Same as with the 4 run-and-debug images.

Those should be deleted, they're old. I'll remove them.

Extension/walkthrough/install-gcc-linux.md
jureid: Updated to "You should see a copyright message and information about the version of GCC you're using."

Extension/walkthrough/install-clang-macos.md
jureid: Thanks for checking. Changing it to "You should see a message with information about the version of Clang you're using."

Extension/package.json
jureid: I did the same thing when going through VS Code's walkthrough with this same screenshot. I'm going to ask them if there's a way to make the image clickable.

Extension/package.json
jureid: Yes, updating.

Extension/package.json
jureid: Oh, I just noticed the missing "your," that's probably what you were referring to. Fixing that, too.

Extension/walkthrough/run-and-debug-project-windows.md
jureid: This is a good thing to think about when we measure the success of the walkthrough on new users vs. returning users. New users likely won't run into the issue of already having a launch.json, but returning users might! Ideally, we'd automatically mark this step as done if we detect a launch.json in their workspace. VS Code is adding support for that, it's not currently available. Let's keep the walkthrough as is, since we're targeting new users, but keep an eye on what returning users do at this step.

Extension/package.json
jureid: From Jackson (software developer for VS Code): The new VS Code Insiders "allows checking off based on an extension being installed, link clicked, command run, or the step simply being selected. _The existing doneOn syntax will still work, but it's deprecated in favor of new "completionEvents"._ Also, by default, if neither of those are defined simply pressing any of the buttons on a step will check the step off (this seems to be the common case). (Exception: if the step has no buttons at all, selecting it will check the item off)."

More info and example code here: https://github.com/microsoft/vscode/issues/122570#issuecomment-840209548

Extension/package.json
jureid: Updated

michelleangela:
Would it be possible to organize the files in the "walkthrough" folder so it's not a giant list of files? Perhaps create a separate folder for images or create a folder per instruction subject.

Example:

**Walkthrough** folder
- images (folder for images not used in any markdown files)
- installcompiler (folder for files and images for installing compiler)
- configure (folder for files/images for configuring intellisense)
- buildanddebug (folder for files and images for build and debug)

jureid:
> Can we use a different icon that is bigger and easier to see the C/C++ text? The C/C++ text is too small for me to read (GitHub is magnfiying the image):
> image

Will reach out to design team and VS Code team for icons.

jureid:
> Would it be possible to organize the files in the "walkthrough" folder so it's not a giant list of files? Perhaps create a separate folder for images or create a folder per instruction subject.
> 
> Example:
> 
> **Walkthrough** folder
> 
> * images (folder for images not used in any markdown files)
> * installcompiler (folder for files and images for installing compiler)
> * configure (folder for files/images for configuring intellisense)
> * buildanddebug (folder for files and images for build and debug)

Yes! I've been wanting to do this, too. I have a few screenshots to update based on Sean's feedback, so when I add those new screenshots I'll organize this folder.

michelleangela:
> > Can we use a different icon that is bigger and easier to see the C/C++ text? The C/C++ text is too small for me to read (GitHub is magnfiying the image):
> > image
> 
> Will reach out to design team and VS Code team for icons.

If they're creating a new icon, can they make it so it's also visible for the tab icon?
image

Extension/walkthrough/run-and-debug-project-mac.md
```diff
@@ -0,0 +1,13 @@
...
+
+1. Open the C++ source file that you want to run and debug. Make sure this file is active (currently displayed and selected) in the editor.
+
+2. From the main menu, select **Run > Add Configuration...** and choose **C++ (GDB/LLDB)**.
+
+3. Choose **clang++ build and debug active file.**
+
+    ![Dropdown showing C++ debug configurations for Clang on macOS](clang-build-and-debug-active-file.png)
+
+4. To start a debugging session, Press `F5` or from the main menu choose **Run > Start Debugging**.
```
michelleangela: `Press` for "# Run and debug your C++ file on macOS" and "# Run and debug your C++ file on Windows" also needs to be lowercased.

Extension/package.json
```diff
@@ -42,6 +42,168 @@
...
+              "path": "walkthrough/build-project-linux.md"
+            }
+          },
+          {
+            "id": "build.project.windows",
+            "title": "Build your C++ file",
+            "description": "Create a default build task for your project. \n[Learn more](https://code.visualstudio.com/docs/cpp/config-msvc#_build-helloworldcpp)",
+            "when": "isWindows",
+            "media": {
+              "altText": "A markdown file with instructions for configuring the default build task on MSVC with Windows.",
```
michelleangela: For "configuring the default build task on MSVC with Windows" does it mean to say "build task on Windows with MSVC"?

sean-mcmanus:
> How do I see the .md files in the UI? I don't see any.

It started appearing today -- not sure what changed.

Extension/package.json
jureid: Good catch :)

Extension/walkthrough/run-and-debug-project-mac.md
jureid: Oh true, thanks!

Extension/package.json
sean-mcmanus: @Colengms Can we switch to completionEvents?

Extension/package.json
jureid: Fixed

michelleangela (approved):

jureid:
> > > Can we use a different icon that is bigger and easier to see the C/C++ text? The C/C++ text is too small for me to read (GitHub is magnfiying the image):
> > > image
> > 
> > 
> > Will reach out to design team and VS Code team for icons.
> 
> If they're creating a new icon, can they make it so it's also visible for the tab icon?
> image

Yup; I included the tab in the email I sent to them. For our extension page, we just have a generic icon in the tab. Maybe we can use that.

image

Extension/package.json
```diff
@@ -42,6 +42,153 @@
...
+                "hc": "walkthrough/intellisense-config-hc.png"
+              }
+            }
+          },
+          {
+            "id": "build.project.mac",
+            "title": "Build your C++ file",
+            "description": "Create a default build task for your project. \n[Learn more](https://code.visualstudio.com/docs/cpp/config-clang-mac#_build-helloworldcpp)",
+            "when": "isMac",
+            "media": {
```
sean-mcmanus, line 120: This reference to "Mac" should be "macOS" to be consistent with the other uses.

Extension/src/LanguageServer/client.ts
```diff
@@ -2660,6 +2661,31 @@ export class DefaultClient implements Client {
...
```
sean-mcmanus, line 2666: What happens if this list is too big to display in the UI? I'm worried about the "Ok" button being hidden off/below the screen. Should we limit the number of items? I get about 30 results on Linux...that could get higher if more versions of clang/gcc were to be installed.

Extension/walkthrough/run-and-debug-project-windows.md
sean-mcmanus: @jureid New users can easily hit this due to bug https://github.com/microsoft/vscode-cpptools/issues/7517 . They may not immediately know which compiler to select when the list appears, so when they cancel the empty launch.json will get created...maybe that's a reason to fix the issue to not create any launch.json, but I don't know if that's possible yet.

sean-mcmanus (changes requested):
Extension/walkthrough/msvc-build-active-file.png has a visually distracting orange character in the bottom left. Can you remove that?

image

Extension/src/LanguageServer/client.ts
Colengms: It looks like VS Code handles this by limiting the dialog to the height of the screen and putting the text in a scrollable field.

Extension/package.json
jureid: fixed

Extension/package.json
jureid: Done

Extension/walkthrough/install-compiler-windows.md
jureid: Decided to just delete this screenshot and to, instead, explain how to open the developer command prompt in another step.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Cool, thanks for checking.

sean-mcmanus (approved):

sean-mcmanus (changes requested):
I'm hitting some issues when the walkthrough is used "too early". The Check Compilers button may not work or the configuration can be blank: 
image

sean-mcmanus:
> I'm hitting some issues when the walkthrough is used "too early". The Check Compilers button may not work or the configuration can be blank:
> image

Well, the Check Compilers behavior seems okay, but I repeatedly get a blank configuration -- that could be an unrelated bug with the configuration UI? It happens randomly or something even long after startup.

sean-mcmanus:
> > I'm hitting some issues when the walkthrough is used "too early". The Check Compilers button may not work or the configuration can be blank:
> 
> Well, the Check Compilers behavior seems okay, but I repeatedly get a blank configuration -- that could be an unrelated bug with the configuration UI? It happens randomly or something even long after startup.

It appears to be an existing (?) bug that repros whenever the Configuration UI is opened -- i.e. it's randomly blank, but the repro probability seems a lot higher when invoked from the walkthrough.

sean-mcmanus:
> It appears to be an existing (?) bug that repros whenever the Configuration UI is opened -- i.e. it's randomly blank, but the repro probability seems a lot higher when invoked from the walkthrough.

Actually, the repro probability increases in the remote scenario, not with the walkthrough. I've opened bug https://github.com/microsoft/vscode-cpptools/issues/7523 .

sean-mcmanus (approved):

jureid:
> > It appears to be an existing (?) bug that repros whenever the Configuration UI is opened -- i.e. it's randomly blank, but the repro probability seems a lot higher when invoked from the walkthrough.
> 
> Actually, the repro probability increases in the remote scenario, not with the walkthrough. I've opened bug #7523 .

I'm seeing the same bug repro in the walkthrough locally

PR #7555, opened 2021-05-19
Cleanup IsExplicit flags.

Colengms:
This addresses an issue in which a fully default config (no `c_cpp_properties.json`, and no `C_Cpp.default.*` settings) would appear as explicitly set.

These flags are intended to indicate whether a value has been explicitly specified by a user (or config provider, etc.) rather than an automatic default value, so that a Configuration Warning can be suppressed if an automatic default value needs to be corrected.  This change ensures that these flags are always set to false when a new configuration is set up, always set to align with a configuration after it's been read, updated to reflected whether a user-specified `C_Cpp.default` setting was used, and never written to `c_cpp_properties.json`.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -766,17 +766,18 @@ export class CppProperties {
             configuration.cStandard = this.updateConfigurationString(configuration.cStandard, settings.defaultCStandard, env);
             configuration.cppStandard = this.updateConfigurationString(configuration.cppStandard, settings.defaultCppStandard, env);
             configuration.intelliSenseMode = this.updateConfigurationString(configuration.intelliSenseMode, settings.defaultIntelliSenseMode, env);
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1206,23 +1209,26 @@ export class CppProperties {
                     delete (<any>e).knownCompilers;
                     dirty = true;
                 }
-                if ((<any>e).compilerPathIsExplicit !== undefined) {
-                    delete (<any>e).compilerPathIsExplicit;
+            });
+
+            for (let i: number = 0; i < this.configurationJson.configurations.length; i++) {
+                if (this.configurationJson.configurations[i].compilerPathIsExplicit !== undefined) {
```
michelleangela: Since all set `dirty = true`, consolidate 4 conditionals into one?

michelleangela (approved):

sean-mcmanus (approved):

PR #7556, opened 2021-05-20
Add symbol scopes.

### LINKED ISSUE #7120: Feature Request: Allow outline panel to show access specifier (i.e. private, public, protected)
I found #3117 and microsoft/vscode#113845 discussing similar problem. Actually I only want some extra small, grey text showing `private` or `public` beside the names of class members, just like what "declaration" does now.

**Screenshots**
Screenshot from 2021-03-05 21-02-02

Displaying other keywords, such as `override`, `virtual` would also be helpful.

_Originally posted by @YalandHong in https://github.com/microsoft/vscode-cpptools/discussions/7112_

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/7120

image

Is the "private: ", "protected:" labeling okay? I figure we actually want https://github.com/microsoft/vscode/issues/113845 .

### DIFF
--- a/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
@@ -22,9 +22,17 @@ export class DocumentSymbolProvider implements vscode.DocumentSymbolProvider {
             symbols.forEach((symbol) => {
                 let detail: string = util.getLocalizedString(symbol.detail);
                 if (symbol.scope === SymbolScope.Private) {
-                    detail = "private" + (detail.length === 0 ? "" : localize("c.cpp.documentsymbolscope.separator", ", ")) + detail;
+                    if (detail.length === 0) {
+                        detail = "private";
+                    } else {
+                        detail = localize("c.cpp.symbolscope.separator", "{0}, {1}", "private", detail);
+                    }
                 } else if (symbol.scope === SymbolScope.Protected) {
-                    detail = "protected" + (detail.length === 0 ? "" : localize("c.cpp.documentsymbolscope.separator", ", ")) + detail;
+                    if (detail.length === 0) {
+                        detail = "protected";
+                    } else {
+                        detail = localize("c.cpp.symbolscope.separator", "{0}, {1}", "protected", detail);
+                    }
                 }
                 const r: vscode.Range = new vscode.Range(symbol.range.start.line, symbol.range.start.character, symbol.range.end.line, symbol.range.end.character);
                 const sr: vscode.Range = new vscode.Range(symbol.selectionRange.start.line, symbol.selectionRange.start.character, symbol.selectionRange.end.line, symbol.selectionRange.end.character);
--- a/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
@@ -3,7 +3,7 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import * as vscode from 'vscode';
-import { DefaultClient, LocalizeDocumentSymbol, GetDocumentSymbolRequestParams, GetDocumentSymbolRequest } from '../client';
+import { DefaultClient, LocalizeDocumentSymbol, GetDocumentSymbolRequestParams, GetDocumentSymbolRequest, SymbolScope } from '../client';
 import * as util from '../../common';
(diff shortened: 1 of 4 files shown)


Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
```diff
@@ -16,7 +16,12 @@ export class DocumentSymbolProvider implements vscode.DocumentSymbolProvider {
         const documentSymbols: vscode.DocumentSymbol[] = [];
         if (symbols) {
             symbols.forEach((symbol) => {
-                const detail: string = util.getLocalizedString(symbol.detail);
+                let detail: string = util.getLocalizedString(symbol.detail);
+                if (symbol.scope === SymbolScope.Private) {
+                    detail = "private" + (detail.length === 0 ? "" : ": ") + detail;
```
michelleangela: Do "private" and "protected" need to be localized or are they C++ keywords in this context?

sean-mcmanus: They directly refer to C++ keywords so shouldn't be localized, but I'll look into adding a localization for the punctuation.

sean-mcmanus: I added a localizable string for the punctuation.

michelleangela (approved):

Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
```diff
@@ -30,9 +30,9 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
             let name: string = symbol.name;
             if (suffix.length) {
                 if (symbol.scope === SymbolScope.Private) {
-                    suffix = "private" + localize("c.cpp.workspacesymbolscope.separator", ", ") + suffix;
+                    suffix = localize("c.cpp.symbolscope.separator", "{0}, {1}", "private", suffix);
```
Colengms: You might want to add a 'hint' here for translators, who might not understand what they should do with `{0}, {1}`.

Adding a hint would look something like:

```
suffix = localize({ key: "c.cpp.symbolscope.separator", comment: ["Used in a list as a separator between keywords.  Example: ..."] }, "{0}, {1}", "private", suffix);
```

Colengms (changes requested):

Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
sean-mcmanus: Sure, I wasn't sure if "hint" was available or not yet.

Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
```diff
@@ -22,10 +26,21 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
...
         // Convert to vscode.Command array
         symbols.forEach((symbol) => {
-            const suffix: string = util.getLocalizedString(symbol.suffix);
+            let suffix: string = util.getLocalizedString(symbol.suffix);
             let name: string = symbol.name;
             if (suffix.length) {
+                if (symbol.scope === SymbolScope.Private) {
+                    suffix = localize("c.cpp.symbolscope.separator", "{0}, {1}", "private", suffix);
+                } else if (symbol.scope === SymbolScope.Protected) {
+                    suffix = localize("c.cpp.symbolscope.separator", "{0}, {1}", "protected", suffix);
```
Colengms: Also, could you assign "private" or "protected" to a variable, so only one call to localize() is necessary?  Multiple calls with the same ID should be fine, but I'm not entire certain what happens in that case.

Colengms (changes requested):

Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
sean-mcmanus: I created a util function reused by all 4 cases.

Extension/src/common.ts
```diff
@@ -1049,7 +1050,7 @@ export function escapeForSquiggles(s: string): string {
                 newResults += "\\";
             }
             lastWasBackslash = false;
-            lastBackslashWasEscaped  = false;
```
sean-mcmanus, line 1052: I ran file formatting.

Colengms (approved):

Issue #7542, opened 2021-05-18
Browse configuration is not preserved when config provider is auto-detected


Colengms:
With https://github.com/microsoft/vscode-cpptools/issues/6150 , we started automatically using a configuration provider if only one is available, and there is no `c_cpp_properties.json`.  When this occurs, we appear to be failing to persist the 'browse configuration' from the config provider, such that a subsequent reload of the window will fall back to default settings instead of a preserved browse configuration.

Colengms:
We may also be neglecting to request the browse configuration.  Code in various places which checks the currently registered config provider would seem to be failing for automatic providers.

Colengms:
The real issue here is that cpptools is not aware of a config provider until it is registered (if not specified explicitly in `c_cpp_properties.json`).  Automatic config provider detection cannot use a config provider by default, until/unless it's registered.  If using `"cmake.configureOnOpen": true`, CMake Tools will register itself early.  Otherwise, it does not register itself until a command such as `CMake: Configure` is run, at which point we receive a new browse configuration anyway.

So, I don't think this is something that can be addressed in cpptools.  Perhaps if CMake Tools could always activate if there is a CMakeTools.txt file in the workspace (if it doesn't already), and always register with cpptools at startup, that could address the issue.

PR #7562, opened 2021-05-20
Use provider ID associated with cached browse config to auto-detect provider

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7542

Uses the provider ID associated with the last received browse config, to auto-detect a config provider, if none is registered.

Fixes an issue with the cached browse config not being removed after switching to a different config (that does not use the same config provider), or a different config provider is auto-detected.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -881,7 +881,7 @@ export class CppProperties {
                         }
                     } else if (this.lastCustomBrowseConfigurationProviderId !== undefined
                         && !!this.lastCustomBrowseConfigurationProviderId.Value) {
-                        // Use the last configure provider we received a browse config from as the provider ID.
+                        // Use the last configuration provider we received a browse config from as the provider ID.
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -843,25 +851,45 @@ export class CppProperties {
...
+                let keepCachedBrowseConfig: boolean = true;
+                if (hasEmptyConfiguration) {
+                    if (providers.size === 1) {
+                        providers.forEach(provider => { configuration.configurationProvider = provider.extensionId; });
+                        if (this.lastCustomBrowseConfigurationProviderId !== undefined) {
+                            keepCachedBrowseConfig = configuration.configurationProvider === this.lastCustomBrowseConfigurationProviderId.Value;
+                        }
+                    } else if (this.lastCustomBrowseConfigurationProviderId !== undefined
+                        && !!this.lastCustomBrowseConfigurationProviderId.Value) {
+                        // Use the last configure provider we received a browse config from as the provider ID.
```
sean-mcmanus: configure -> configuration?

sean-mcmanus (approved):

PR #7581, opened 2021-05-24
Fix package.nls.json issues.

sean-mcmanus:


Colengms (approved):

Extension/package.nls.json
```diff
@@ -188,8 +188,8 @@
     "c_cpp.debuggers.logging.engineLogging.description": "Optional flag to determine whether diagnostic debug engine messages should be logged to the Debug Console. Defaults to false.",
     "c_cpp.debuggers.logging.trace.description": "Optional flag to determine whether diagnostic adapter command tracing should be logged to the Debug Console. Defaults to false.",
     "c_cpp.debuggers.logging.traceResponse.description": "Optional flag to determine whether diagnostic adapter command and response tracing should be logged to the Debug Console. Defaults to false.",
-    "c_cpp.debuggers.cppvsdbg.logging.threadExit.description": "Optional flag to determine whether thread exits messages should be logged to the Debug Console. Default: `false`.",
-    "c_cpp.debuggers.cppvsdbg.logging.processExit.description": "Optional flag to determine whether target process exits messages should be logged to the Debug Console., or debugging is stopped. Default: `true`.",
```
sean-mcmanus, line 192: @WardenGnaw What was this " Debug Console., or debugging is stopped. " text supposed to be? I removed the last part which didn't make sense.

Also, I changed "exits" to "exit" -- is that correct or is it supposed to be "exits"?

sean-mcmanus: If it's not correct, we can fix it later for 1.4.0.

WardenGnaw: This is fine. 

Thanks!

PR #7582, opened 2021-05-24
Fix the walkthrough for remote targets.

sean-mcmanus:


Extension/package.json
```diff
@@ -59,7 +59,7 @@
             "id": "verify.compiler.mac",
             "title": "Install a C++ compiler",
             "description": "The C++ extension uses the C++ compiler on your system to configure IntelliSense for your project. \n[Check for C++ compilers](command:C_Cpp.CheckForCompiler)",
-            "when": "isMac",
+            "when": "workspacePlatform == mac",
```
Colengms, line 62: Is this in the latest insiders yet?  I just tried it out, and it didn't seem to work.  Also, I'm curious what happens if no workspace is open.  Does it default to the host platform, or do we need some boolean logic to use the original flags if no workspacePlatform?

Colengms (changes requested):

Extension/package.json
sean-mcmanus: It should be in tonight's Insiders (not released yet). The commit is at https://github.com/microsoft/vscode/commit/196e79256dc9fcda1bdc93f0ac321223d8ff9d56 .

So from what I see, it doesn't appear to be dependent on a workspace being open.

sean-mcmanus: It works with the latest Insiders. If no workspace is open, it still works.

Colengms (approved):

michelleangela (approved):

PR #7612, opened 2021-05-27
Fix Walkthrough property (#7611)

sean-mcmanus:


sean-mcmanus:
Don't click "Update branch"

Issue #7538, opened 2021-05-17
Implement localization for MD files in Getting Started walkthrough


Colengms:
Opening this to track some additional scripting needed in gulp.js to handle localization export and import for MD files used by the Getting Started Walkthrough.  This would be similar to how HTML files and json schema files are currently handled.

PR #7620, opened 2021-05-28
Upgrade vscode-nls-dev module.  Add translation hints for strings containing markdown

Colengms:
Support for translation hints (comments) have been added to the latest vscode-nls-dev.

This does not yet localize the Getting Started walkthough.  We anticipate significant changes to the walkthrough in 1.5.0, so I don't want to make those strings localizable just yet.  (Also, I need to finish https://github.com/microsoft/vscode-cpptools/issues/7538 )

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2608,7 +2608,7 @@
     "typescript": "^3.5.3",
     "vscode-debugadapter": "^1.35.0",
     "vscode-debugprotocol": "^1.35.0",
-    "vscode-nls-dev": "^3.2.6",
+    "vscode-nls-dev": "^4.0.0-next.1",
     "vscode-test": "^1.3.0",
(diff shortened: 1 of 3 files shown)


sean-mcmanus (approved):

michelleangela (approved):

sean-mcmanus:
@Colengms This PR appears to break the settings UI so I had to undo it: https://github.com/microsoft/vscode-cpptools/pull/7694

Colengms:
> @Colengms This PR appears to break the settings UI so I had to undo it: #7694

Could you elaborate?  What was the exception?  Since the strings were getting properly extracted in the export XLF, the problem would seem to be in vscode-nls, so we need to open a bug against that, if so.

sean-mcmanus:
@Colengms Details are in the email I sent.

PR #7636, opened 2021-05-31
Change instances of OSX to macOS

TylerADavis:
Yesterday I noticed a popup that said `Downloading C/C++ language components (OS X)`. Apple [renamed OS X to macOS a few years back](https://www.theverge.com/2016/6/13/11906214/apple-mac-os-10-announced-features-siri-wwdc-2016), so I created this PR to change a few lingering instances of `OS X` to `macOS`.

## Changes
* Rename lingering instances of `OS X` to `macOS`
* Add `x86_64` to description for x86_64 version of macOS cpptools, to make architecture clear in messages
* VS Code automatically stripped some unnecessary whitespace from the debug instructions

## Questions
I noticed that `package.json` uses both `arm64` and `ARM64`. Should one of these be preferred over the other?

Additionally, a few formats are used to communicate architecture in descriptions, e.g. `Linux / ARM64` vs `macOS ARM64`. Is one format preferred?


Extension/package.json
```diff
@@ -2685,7 +2685,7 @@
       "integrity": "186614524A13EC75BF68798861818D25CE82FDE293C57B725C11F62B0694EC55"
     },
     {
-      "description": "C/C++ language components (OS X)",
+      "description": "C/C++ language components (macOS x86_64)",
```
sean-mcmanus: Yeah, I'd say use macOS / x86_64 to be consistent with the other cases.

Extension/package.json
```diff
@@ -2802,7 +2802,7 @@
       "integrity": "AEB24D394118EAD87087DCC651F2EE329FC335ECE88ED6C8C2A9E93ED60DFECD"
     },
     {
-      "description": "ClangFormat (OS X arm64)",
+      "description": "ClangFormat (macOS arm64)",
```
sean-mcmanus: I'd go with ARM64 instead of arm64.

Extension/package.json
```diff
@@ -2788,7 +2788,7 @@
       "integrity": "CD5578A21C8D515D15C0611621E44C03DE4E667EFB4EE1A0DE18B08FD9B96794"
     },
     {
-      "description": "ClangFormat (OS X)",
+      "description": "ClangFormat (macOS)",
```
sean-mcmanus: This should be macOS / x86_64 as well.

Extension/package.json
sean-mcmanus: Should the x86_64 be x64 instead for macOS?

TylerADavis: I'm not certain, but I see on [this developer documentation page](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) that apple refers to `x86_64`.

Extension/package.json
TylerADavis: Done

Extension/package.json
TylerADavis: Done

Extension/package.json
TylerADavis: Done

TylerADavis:
Thanks for your review @sean-mcmanus . I've changed `arm64` to `ARM64` and added the separating slashes inbetween OS name and architecture. I believe the name `x86_64` is used on Mac over `x64`, though I'm not entirely certain.

Please let me know if you'd like to see any other changes.

sean-mcmanus (approved):

michelleangela (approved):

Issue #7275, opened 2021-04-01
Support the experimentation platform


bobbrow:
We want this for our "getting started" work.

sean-mcmanus:
What is the experimentation platform? Is it fixed by https://github.com/microsoft/vscode-cpptools/pull/7512 ?

Colengms:
The module involved is `vscode-tas-client`.  It would be a replacement for our A/B testing approach and use of cpptools.json.  

It allows retrieval of variables from 'control tower', which can be varied based on the 'treatment' of the client (whether the client is selected for an experiment or not).  

There is a TAS-Client repo in DevOps (in DivDev) that contains some documentation.  (I don't want to link it here, since it's internal).

It was suspect that this might be required for the Getting Started walkthrough work, but experiments related to who that is enabled for are handed by VS Code and not the extension itself.  Since it's relatively simple to use, I'd like to wait until we have an experiment to run (so we have a reason to actually call into it) before doing this work.  Otherwise, referencing the module and initializing it would seem have no particular use.

PR #7659, opened 2021-06-08
Integrate experimentation platform.  Remove ABTesting.

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7275

Enables use of the experimentation platform, which involves wrapping our telemetry calls such that experimentation properties get included in all telemetry events.  (I copied the IExperimentationTelemetry implementation from [here](https://github.com/assujith2255/a/blob/c89861f76f8525e106cd2421ed104ab88d78ca42/lib/vscode/extensions/github-authentication/src/experimentationService.ts).)

Since the experimentation platform provides similar capabilities as ABTesting, I removed ABTesting.  Only one of the ABTesting values was still used - the minimum VS Code version to allow Insiders to be downloaded.  This has been changed to use an experiment that returns true for a 'allowInsiders' variable only for users with a minimum version of VS Code (specified in Control Tower).


Extension/src/telemetry.ts
```diff
@@ -5,52 +5,121 @@
...
-export function logDebuggerEvent(eventName: string, properties?: { [key: string]: string }): void {
-    if (telemetryReporter) {
-        const eventNamePrefix: string = "cppdbg/VS/Diagnostics/Debugger/";
-        telemetryReporter.sendTelemetryEvent(eventNamePrefix + eventName, properties);
+export async function deactivate(): Promise<void> {
+    if (initializationPromise) {
+        try {
+            await initializationPromise;
+        } catch (e) {
+            // Continue even if we were not able to initialize the experimention platform.
```
jogo-: Typo: experimention -> experimen**ta**tion

Extension/src/telemetry.ts
```diff
@@ -5,52 +5,121 @@
...
 
-export function logLanguageServerEvent(eventName: string, properties?: { [key: string]: string }, metrics?: { [key: string]: number }): void {
-    if (telemetryReporter) {
-        const eventNamePrefix: string = "C_Cpp/LanguageServer/";
-        telemetryReporter.sendTelemetryEvent(eventNamePrefix + eventName, properties, metrics);
+export async function logDebuggerEvent(eventName: string, properties?: { [key: string]: string }): Promise<void> {
+    try {
+        await initializationPromise;
+    } catch (e) {
+        // Continue even if we were not able to initialize the experimention platform.
```
jogo-: Typo: experimention -> experimen**ta**tion

Extension/src/telemetry.ts
```diff
@@ -5,52 +5,121 @@
...
-    if (util.extensionContext) {
-        const packageInfo: IPackageInfo = getPackageInfo();
-        if (packageInfo) {
-            return new TelemetryReporter(packageInfo.name, packageInfo.version, appInsightsKey);
-        }
+export async function logLanguageServerEvent(eventName: string, properties?: { [key: string]: string }, metrics?: { [key: string]: number }): Promise<void> {
+    try {
+        await initializationPromise;
+    } catch (e) {
+        // Continue even if we were not able to initialize the experimention platform.
```
jogo-: Typo: experimention -> experimen**ta**tion

Extension/src/LanguageServer/extension.ts
```diff
@@ -331,25 +331,32 @@ function realActivation(): void {
 
     vcpkgDbPromise = initVcpkgDatabase();
 
-    PlatformInformation.GetPlatformInformation().then(info => {
+    PlatformInformation.GetPlatformInformation().then(async info => {
```
michelleangela, line 334: Can `PlatformInformation.GetPlatformInformation().then(async info =>` be refactored so it doesn't need to use `.then`? For example, use `await` on `PlatformInformation.GetPlatformInformation()`.

Or is this one of those cases where `.then` is required to achieve the correct behavior?

Extension/src/telemetry.ts
```diff
@@ -5,52 +5,121 @@
...
+                } else {
+                    targetPopulation = TargetPopulation.Internal;
+                }
+                experimentationTelemetry = new ExperimentationTelemetry(new TelemetryReporter(packageInfo.name, packageInfo.version, appInsightsKey));
+                initializationPromise = getExperimentationServiceAsync(packageInfo.name, packageInfo.version, targetPopulation, experimentationTelemetry, util.extensionContext.globalState);
+            }
+        }
     } catch (e) {
-        // can't really do much about this
+        // Can't really do much about this.
```
michelleangela: Suggest to update comment "Can't really do much about this." to something that is not colloquial.

Example: "Handle error with a try/catch, but do nothing for errors."

Extension/src/LanguageServer/extension.ts
Colengms: This is intentional, to chain functionality without waiting for it to complete.  It could take up to 6 seconds for the `getExperimentationServiceAsync` call to complete, and we don't want to stall extension activation until that is done.  (Or, we don't need to yet.  If we ever add an experiment variable that we need to read prior to activation to set things up properly, this would need to be awaited, and we'd just have to accept the delay.)

sean-mcmanus:
Hmm...should we keep the abTesting code to the minimumVSCodeVersion or is the experimentation platform going to be sufficient for that, i.e. is it easy to update for all users before we release a new insiders that has a higher requirement or could there be issues?

Colengms:
> Hmm...should we keep the abTesting code for the minimumVSCodeVersion or is the experimentation platform going to be sufficient for that, i.e. is it easy to update for all users before we release a new insiders that has a higher requirement or could there be issues?

I think it should be easier to update the value of the variable in the control tower, than to publish a new cpptools.json.  Julia is looking into how to do this.

michelleangela (approved):

sean-mcmanus (approved):

PR #7716, opened 2021-06-17
Fix overlapping folding ranges.

### LINKED ISSUE #5521: Code folding will collapse entire line containing closing bracket
**Type: LanguageService**

**Describe the bug**
- OS and Version: Linux, 5.6.13-arch1-1
- VS Code Version: 1.44.2
- C/C++ Extension Version: 0.28.0
- Other extensions you installed (and if the issue persists after disabling them): github.vscode-pull-request-github james-yu.latex-workshop liviuschera.noctis Yes, issue persists
- Does this issue involve using SSH remote to run the extension on a remote machine?: Dunno how to check
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc): in all types of workspaces, when you use formatting style which puts "else if"'s in the same line as closing bracket to previous conditional, folding hides too many lines (look at screenshots).

**Steps to reproduce**

1. Type this code 
```cpp
if(cond) {
//
} else {
//
}
```
2. Try to fold the "if" conditional
3. See error: "else" gets folded with the previous line

**Expected behavior**

fold is only applied to what's inside the if, not touching the "else" line

<details>
  <summary><strong>Logs</strong></summary>

Nothing shows.
</details>

**Screenshots**

Before folding: 
image
Folding with extensions disabled: 
image
Folding with C++ extension:
image

**Additional context**

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/5521

image

@Colengms Is there some problem/bug with this fix and/or a reason this wasn't done initially?


Colengms:
@sean-mcmanus This seems to be related to: https://github.com/microsoft/vscode-cpptools/issues/5521    Currently, it's by design that we collapse the entire line containing the closing bracket.  Having the fold take as few lines as possible when collapsed (similar to VS) was desired.  Also, if we did not include the last line, code immediately preceding a closing brace would not get included in the fold.

```cpp
	if (i == 0) {
		i = 1 
		j = 2 }
	}
```

But, not including the last line only when overlapping the start of another folding range, seems OK to me.

Colengms (approved):

sean-mcmanus:
Ah, I see, it behaves unideally when 
```
if (i == 0) {
    i = 1 
    j = 2 } else {
    return 0;
}
```
is used; however, that seems like more desirable behavior than the pre-existing one which hides the else/else if, which is likely very important for the user to see and not be collapsed.

PR #7726, opened 2021-06-19
CUDA/CPP extensions

jogo-:
* Missing dot: "cu" -> ".cu"
* Useless uppercase ".C" before needed lowercase ".c"
* Add same CPP extensions as here:
https://github.com/microsoft/vscode-cpptools/blob/f82b1beaa0899b67685441325592690a9b664be9/Extension/src/LanguageServer/cppBuildTaskProvider.ts#L83


sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -89,7 +89,7 @@ export async function getRawJson(path: string | undefined): Promise<any> {
 
 export function fileIsCOrCppSource(file: string): boolean {
     const fileExtLower: string = path.extname(file).toLowerCase();
-    return ["cu", ".C", ".c", ".cpp", ".cc", ".cxx", ".mm", ".ino", ".inl"].some(ext => fileExtLower === ext);
+    return [".cu", ".c", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".tcc", ".mm", ".ino", ".ipp", ".inl"].some(ext => fileExtLower === ext);
```
Colengms, line 92: There may be some more code here that needs updating.  The ".C" being upper case was intentional, as only uppercase ".C" is associated with C++ by default.

sean-mcmanus, line 92: I don't think so -- the input file is made lowercased and the function name is COrCpp. I don't see any C++ specific code that uses this function.

Colengms (approved):

PR #7745, opened 2021-06-24
Use setTextDocumentLanguage to temporarily set language ID for CUDA sources, and headers

Colengms:
There is a PR on the native side for the bulk of this change.

Adds a new protocol message from the native side which invokes `vscode.languages.setTextDocumentLanguage`.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1993,7 +1993,6 @@ export class DefaultClient implements Client {
     }
 
     private async setTemporaryTextDocumentLanguage(params: SetTemporaryTextDocumentLanguageParams): Promise<void> {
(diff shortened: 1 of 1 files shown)


michelleangela (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1984,6 +1992,15 @@ export class DefaultClient implements Client {
         }
     }
 
+    private async setTemporaryTextDocumentLanguage(params: SetTemporaryTextDocumentLanguageParams): Promise<void> {
+        // const uri: vscode.Uri = vscode.Uri.file(params.path);
```
sean-mcmanus: Why is this here?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

Issue #7723, opened 2021-06-18
Don't squiggle a `compilerPath` of `cl` or `cl.exe` in `c_cpp_properties.json`


Colengms:
It's OK to specify a `compilerPath` of `cl` or `cl.exe`.  We don't invoke `cl.exe`, but it's location would tell us where to find system headers.  If only `cl` or `cl.exe` is specified, we should  search for an install of VS on the system and use those system headers.  So, this is valid, doesn't result in any configuration errors/warnings, and doesn't need to be squiggled in `c_cpp_properties.json`

sean-mcmanus:
Oh, this must be a regression -- it didn't squiggle cl/cl.exe previously.

sean-mcmanus:
I had fixed this bug previous at https://github.com/microsoft/vscode-cpptools/pull/3263

PR #7752, opened 2021-06-28
don't squiggle for "cl.exe" or "cl" path

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cpptools/issues/7723

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1656,8 +1656,10 @@ export class CppProperties {
             let compilerPathNeedsQuotes: boolean = false;
             let compilerMessage: string | undefined;
             const compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(compilerPath);
-            // Don't squiggle invalid cl.exe paths because it could be for an older preview build.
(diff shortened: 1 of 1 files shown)


sean-mcmanus (changes requested):
Actually, we want any compilerPath than ends with cl.exe or cl to be acceptable, because otherwise, it'll frequently squiggle when the user installs a preview build with a cl.exe version that no longer exists, i.e. we should accept the old path.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1656,8 +1656,12 @@ export class CppProperties {
             let compilerPathNeedsQuotes: boolean = false;
             let compilerMessage: string | undefined;
             const compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(compilerPath);
-            // Don't squiggle invalid cl.exe paths because it could be for an older preview build.
-            if (compilerPathAndArgs.compilerName.toLowerCase() !== "cl.exe" && compilerPathAndArgs.compilerPath !== undefined) {
+            const clCompilerLowerCase: any = compilerPathAndArgs.compilerName.toLowerCase();
```
sean-mcmanus: Can this "any" be string?

sean-mcmanus (approved):

sean-mcmanus (approved):

Issue #7796, opened 2021-07-07
Log the logging level


Colengms:
This is a feature request to add the logging level itself to the log, so we can tell what level of logging we are looking at when looking at logs submitted by users.

Colengms:
This should be addressed in 1.6.0-insiders.

PR #7807, opened 2021-07-09
Add logging of loggingLevel

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7796


Extension/src/logger.ts
```diff
@@ -70,6 +71,11 @@ let outputChannel: vscode.OutputChannel | undefined;
 export function getOutputChannel(): vscode.OutputChannel {
     if (!outputChannel) {
         outputChannel = vscode.window.createOutputChannel("C/C++");
+        const settings: CppSettings = new CppSettings();
+        const loggingLevel: string | undefined = settings.loggingLevel;
+        if (!!loggingLevel && loggingLevel !== "None" && loggingLevel !== "Error") {
```
sean-mcmanus, line 76: Can you save the original logging level and re-output it if it's changed to a value that would affect a difference (i.e. switching from None to Error wouldn't matter)?

sean-mcmanus (changes requested):

sean-mcmanus:
Also, would it be good to add the loggingLevel to the Log Diagnostics logging in cases where the users clear their log or only provide a subset of the C/C++ log?

Extension/src/logger.ts
Colengms: Ok.  I'm not sure adding it also to Log Diagnostics would be useful.  A user could potentially generate the Log Diagnostics from a different session than the output log, making the value unreliable.

sean-mcmanus: What do you mean by "different session"? I was thinking if the user is in log gathering mode then there's a high probability the Log Diagnostics and C/C++ logging would be close in time, but there is a possibility they change the loggingLevel before or after they do the Log Diagnostics, so it wouldn't be 100% reliable, but could hint at what's probably happening if C/C++ logs appear to be missing info.

sean-mcmanus (approved):

michelleangela (approved):

PR #7813, opened 2021-07-09
Same C/CPP extensions as VSCode

jogo-:
Support same C and CPP extensions as VSCode: https://github.com/microsoft/vscode/blob/60ddedc0503dfec6cb1aac0814f7fc09dc52d2b3/extensions/cpp/package.json#L15-L66


sean-mcmanus (changes requested):
I believe the method fileIsCOrCppSource is intended to return only source files and not header files. Do you see a usage where header files are also intended?

jogo-:
You're right.
Sorry

Issue #6877, opened 2021-01-30
files.exclude doesn't apply to references to workspace folders outside of the current workspace folder in a multi-root workspace


MmAaXx500:
**Type: LanguageService**

**Describe the bug**
- OS and Version: Windows 10 20H2 and Debian testing
- VS Code Version: 1.52.1
- C/C++ Extension Version: 1.1.3
- Other extensions you installed (and if the issue persists after disabling them): Only this extension is enabled

**Steps to reproduce**

1. Download the [example-workspace](https://github.com/microsoft/vscode-cpptools/files/5898784/example-workspace.zip)
2. Open it but DO NOT open any files in them
3. Examine the directories. You should see `dir1/A`, `dir1/C` and `dir2`
4. Enable the C/C++ extension
5. Open `abc.cpp` inside `dir2`
6. Ctrl click on the header and it will point to the header in folder `B` (this folder is excluded)
7. Examine the directories again. You should now see `dir1/B` among other folders

If you close all files, reopen vscode and open `abc.cpp` in `dir1/A` the IntelliSense using the correct paths (both in `dir1` and `dir2`).

**Expected behavior**

The header should be pointing in any case to folder `C`

<details>
  <summary><strong>Logs</strong></summary>

```
cpptools/didChangeCppProperties
MSVC intelliSenseMode specified.  Configuring for compiler cl.exe.
MSVC intelliSenseMode specified.  Configuring for compiler cl.exe.
  Folder: C:/PROGRAM FILES (X86)/MICROSOFT VISUAL STUDIO/2019/COMMUNITY/VC/TOOLS/MSVC/14.28.29333/ATLMFC/INCLUDE/* will be indexed
  Folder: C:/PROGRAM FILES (X86)/MICROSOFT VISUAL STUDIO/2019/COMMUNITY/VC/TOOLS/MSVC/14.28.29333/INCLUDE/* will be indexed
  Folder: C:/PROGRAM FILES (X86)/WINDOWS KITS/10/INCLUDE/10.0.18362.0/CPPWINRT/ will be indexed
  Folder: C:/PROGRAM FILES (X86)/WINDOWS KITS/10/INCLUDE/10.0.18362.0/SHARED/ will be indexed
  Folder: C:/PROGRAM FILES (X86)/WINDOWS KITS/10/INCLUDE/10.0.18362.0/UCRT/ will be indexed
[...]

Issue #6870, opened 2021-01-29
Make punctuation marks in description of property and configuration settings consistent


michelleangela:
Some property descriptions have periods at the end and some don't. Example:

With "." period
![image](https://user-images.githubusercontent.com/38734287/106331785-5fcfa100-623a-11eb-932e-2972a2986a9b.png)

Without "." period
![image](https://user-images.githubusercontent.com/38734287/106331838-77a72500-623a-11eb-838b-a5762e029a63.png)

Some property descriptions use double quotes, and some use single quotes. Example:

This uses single quotes:
![image](https://user-images.githubusercontent.com/38734287/106331904-986f7a80-623a-11eb-89e6-2257b4bc8248.png)

This uses double quotes:
![image](https://user-images.githubusercontent.com/38734287/106331917-9c030180-623a-11eb-8521-412bb6c40d9d.png)

sean-mcmanus:
I think it should be normalized to end with periods and use double quotes.

sean-mcmanus:
I was going to do this for 1.5.0, but I think it'll break too many translations before we want to ship.

Issue #5029, opened 2020-02-28
Cache results of Find All References


jandyman:
I'm not sure if others are having this problem, but I'm finding Find All References to be very slow on several projects which aren't even that big. In addition, I haven't figured a way limit the scope of the searches. In all other senses, the navigation and Intellisense are better than Eclipse (my previous goto for this), but Find All References is both too slow and something I use ALL the time. It's a continual thorn in my side. Is there a solution to this? It seems like once an index is built for a project, this function should be instantaneous, as it is in Eclipse

sean-mcmanus:
Find All References scope limitation is being tracked at https://github.com/microsoft/vscode-cpptools/issues/4102 .

We need more info on what stage is too slow for you. What does the status indicator in the bottom right show for the majority of your wait time? Does it mention "files confirmed"?

Our telemetry indicates that most of our users are experiencing acceptable performance for Find All References, but some user's with particular projects or hardware are not. During the operation, what does your memory and CPU usage look like? We scale "confirmation" via the core count, so if your machine has too few cores, that could be one factor, or if your machine has too little memory with a high core count, that could be the cause of slowness.

The index we build is able to quickly get a subset of "potential" references, but our implementation has to compile the code in order to "confirm" references (i.e. to type check it) in the target files where references are suspected (same as what VS does). You can click the search icon to get preview of the results.

We're tracking Find All References performance with https://github.com/microsoft/vscode-cpptools/issues/4051 .

jandyman:
I moved my comment about scope to #4102

I'm running a brand new Macbook Pro, so there is no problem with cores or such. I suspect that the reason I see a difference with Eclipse is that it builds an index once and reuses it, whereas VS code is compiling all the files. I don't see anything but a progress bar at the top of the UI until the Find All References search is almost completely done, so I don't really know what you are talking about WRT the "search icon". Maybe there is one in the dialog box that pops up in the lower right corner when the scan is almost completely done, but it is there so briefly (a small portion of the total time to Find All References) that I can't tell. 

I hope that helps, but I feel that I'm misunderstanding something.

sean-mcmanus:
You're supposed to see a UX like the following after 3 seconds:
![image](https://user-images.githubusercontent.com/19859882/75598831-b1ccdd00-5a53-11ea-805e-0f5b2f4477f9.png)

How long is Find All References taking for you?

jandyman:
[...]

Issue #4811, opened 2020-01-03
Add setting (and/or auto-heuristic) to limit the maximum number of IntelliSense processes to spawn


sean-mcmanus:
Based on a user comment in https://github.com/microsoft/vscode-cpptools/issues/3220 for 12 IntelliSense processes using >600 MB each (or more?) on a machine with 8 GB RAM.

Right now, our heuristic uses core count and not total available memory.

https://github.com/microsoft/vscode-cpptools/issues/4036 is tracking the Find All References case.

https://github.com/microsoft/vscode-cpptools/issues/4038 (Find All References) caching could be added more easily if this were fixed.

joaquin-herrero:
Not sure if it is 100% related with the limit of spawn processes but here I have a machine with 12 cores and 12GB of RAM running Debian 9 and VSCode 1.43. 

I have 23 processes of cpptools eating all my RAM so I need to reboot the machine again and again. 

Te two things I can do are:
Set Intelli Sense Engine to tag parser seems to stop cpptools processes to eat all the memory.
Set the memory system limits to avoid the system hang up, but this makes me unable to use cpp features on VSCode.

sean-mcmanus:
@joaquin-herrero What version of our C/C++ extension are you using? I assume 0.27.0-insiders3? Are you using multi-root? Are you using multiple VS Code instances? Do you mean 23 cpptools processes or 23 cpptools-srv processes (or the old-named Microsoft.VSCode.Cpp proceses). Only 1 cpptools process is expected per VS Code instance. Having 23 cpptools-srv processes open is not expected unless you have 12 C++ files open and then run a Find All References operation. If you close VS Code do the processes go away (i.e. are there any dangling processes)? If you close files you aren't using, the cpptools-srv processes should go away (there's a few second delay).

Setting the IntelliSenseEngine to "Tag Parser" disables a lot of our functionality so it's not recommended.

Setting the CPU affinity that controls which cores a process can run on might also be a workaround, depending on your OS's support for that.

terriblefire:
Happens for me too. Makes my macOS system unusable when opening a large project (e.g. Mame) in VSCode.

sean-mcmanus:
@terriblefire This issue is only about the number of IntelliSense processes to spawn, and currently only 1 process spawns per file open up to around the core count -- do you have lots of files open? If you're seeing an issue with only 1 file open, then you're seeing a different issue (you may be seeing multiple threads from the same process).

[...]

PR #7816, opened 2021-07-10
Add clang-tidy support

### LINKED ISSUE #2908: Implement enhanced integration with clang-tidy
Implement support for clang extra tools.

- clang-include-fixer
- clang-format
- clang-refactor
- clang-analyzer
- clang-tidy
- clang-rename

EDIT:  This issue will just cover clang-tidy.  The remaining tools are covered by issue #7276

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/2908
- Update Clang tools fwlinks to 12.0.1.
- Add new commands and settings.
- Change settings to use markdownDescription and have descriptions end with periods: https://github.com/microsoft/vscode-cpptools/issues/6870

Also, https://github.com/microsoft/vscode-cpptools/issues/4036, https://github.com/microsoft/vscode-cpptools/issues/4811, https://github.com/microsoft/vscode-cpptools/issues/5029, https://github.com/microsoft/vscode-cpptools/issues/6877.


Extension/src/LanguageServer/settings.ts
```diff
@@ -68,55 +68,105 @@ export class CppSettings extends Settings {
...
+            const clangName: string = isFormat ? this.clangFormatName : this.clangTidyName;
+            const setCachedClangPath: (path: string | null) => void = isFormat ? setCachedClangFormatPath : setCachedClangTidyPath;
+            path = which.sync(clangName, { nothrow: true });
+            setCachedClangPath(path);
             if (!path) {
                 return undefined;
             } else {
-                // Attempt to invoke both our own version of clang-format to see if we can successfully execute it, and to get it's version.
-                let clangFormatVersion: string;
+                // Attempt to invoke both our own version of clang-* to see if we can successfully execute it, and to get it's version.
```
jogo-: Typo: it's -> its

Extension/package.nls.json
```diff
@@ -21,6 +21,36 @@
...
+    "c_cpp.configuration.codeAnalysis.clangTidy.fallbackConfig.markdownDescription": { "message": "Specifies a `clang-tidy` configuration in YAML/JSON format to be be used as a fallback when `#C_Cpp.codeAnalysis.clangTidy.config#` is not set and no `.clang-tidy` file is found: `{Checks: '-*,clang-analyzer-*', CheckOptions: [{key: x, value: y}]}`.", "comment": [ "Words 'key' and 'value' in '{key: value, ...}' should be translated, but all other markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
jogo-: Typo: to be **be** used -> to be used

Extension/src/LanguageServer/settings.ts
sean-mcmanus: Thanks. Fixed with https://github.com/microsoft/vscode-cpptools/pull/7816/commits/374060eacf860d77a89a3e32f94b65f6175dc82b

Extension/package.nls.json
sean-mcmanus: Thanks. Fixed with https://github.com/microsoft/vscode-cpptools/pull/7816/commits/374060eacf860d77a89a3e32f94b65f6175dc82b

Extension/package.json
```diff
@@ -241,6 +241,1312 @@
...
+          "type": "string",
+          "markdownDescription": "%c_cpp.configuration.codeAnalysis.clangTidy.config.markdownDescription%",
+          "scope": "resource"
+        },
+        "C_Cpp.codeAnalysis.clangTidy.fallbackConfig": {
+          "type": "string",
+          "markdownDescription": "%c_cpp.configuration.codeAnalysis.clangTidy.fallbackConfig.markdownDescription%",
+          "scope": "resource"
+        },
+        "C_Cpp.codeAnalysis.clangTidy.fix.warnings": {
```
bobbrow: I don't think we're doing "fixes" in v1, so let's hide these settings for now.

Extension/src/LanguageServer/settingsTracker.ts
```diff
@@ -177,7 +177,7 @@ export class SettingsTracker {
                     break;
                 }
                 default: {
-                    if (key === "clang_format_path" || key === "intelliSenseCachePath" || key.startsWith("default.")) {
+                    if (key === "clang_format_path" || key === "codeAnalysis.clangTidy.path" || key === "codeAnalysis.clangTidy.buildPath" || key === "intelliSenseCachePath" || key.startsWith("default.")) {
```
bobbrow: I think we should also scrub the `codeAnalysis.clangTidy.args`/`config`/`fallbackConfig`

Extension/package.nls.json
```diff
@@ -22,6 +22,37 @@
...
+    "c_cpp.configuration.codeAnalysis.clangTidy.fallbackConfig.markdownDescription": { "message": "Specifies a `clang-tidy` configuration in YAML/JSON format to be used as a fallback when `#C_Cpp.codeAnalysis.clangTidy.config#` is not set and no `.clang-tidy` file is found: `{Checks: '-*,clang-analyzer-*', CheckOptions: [{key: x, value: y}]}`.", "comment": [ "Words 'key' and 'value' in '{key: value, ...}' should be translated, but all other markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
bobbrow, line 47: How do you provide YAML in a json property for `config`/`fallbackConfig`?  Should we just say JSON and only support that?

Extension/package.nls.json
```diff
@@ -22,6 +22,37 @@
...
+    "c_cpp.configuration.codeAnalysis.clangTidy.fix.notes.markdownDescription": { "message": "If `true`, `clang-tidy` fixes for warnings are automatically applied, along with fixes found through an associated diagnostic note.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.clangTidy.headerFilter.markdownDescription": { "message": "Regular expression matching the names of the headers to output diagnostics from. Diagnostics from the main file of each translation unit are always displayed. This option overrides the `HeaderFilterRegex` option in a `.clang-tidy` file, if any.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
bobbrow: Maybe rename to `includeHeaders`?  I'm not sure why but "filter" usually means "filter out" to me so I get confused when I look at this setting.

bobbrow: Though maybe 'include' is too overloaded here and could be equally bad. 🤷‍♂️

sean-mcmanus: I agree that the name is confusing, but the headerFilter matches the corresponding clangTidy setting name and there's a direct correspondence.

Extension/package.nls.json
sean-mcmanus: The same as with the corresponding clang-format settings.

Extension/src/LanguageServer/settingsTracker.ts
sean-mcmanus: Sure

Extension/package.json
sean-mcmanus: Yeah

Extension/package.nls.json
sean-mcmanus: Maybe something like headerFilterInclude?

Extension/src/LanguageServer/settingsTracker.ts
sean-mcmanus: I also added headerFilter.

Extension/package.nls.json
bobbrow: If it's called "header filter" in clang-tidy, then no change should be made here.

Extension/package.nls.json
bobbrow: We support YAML in clang-format?  How?  YAML requires newlines.

sean-mcmanus: It's this format: `{Checks: '-*,clang-analyzer-*', CheckOptions: [{key: x, value: y}]}` -- maybe that's called YAML/JSON? It's like YAML but without newlines. This setting/format is directly from clang-tidy.

bobbrow: Ok. Weird.

Extension/package.json
```diff
@@ -241,6 +241,1312 @@
       "type": "object",
       "title": "C/C++",
       "properties": {
+        "C_Cpp.maxConcurrentThreads": {
+          "type": [
+            "integer",
```
Colengms, line 246: Does using both integer and null work as expected in VS Code's settings UI?  It looks like null is represented as an empty field in the settings UI.  You might want to add verbiage to the settings description to refer to a blank/empty value in the UI addition to referring to 'null'.  Or, consider using an out of range value as a magic number, to avoid null types.

Extension/src/Debugger/configurations.ts
```diff
@@ -29,9 +29,9 @@ export function indentJsonString(json: string, numTabs: number = 1): string {
 }
 
 function formatString(format: string, args: string[]): string {
-    for (const arg in args) {
-        format = format.replace("{" + arg + "}", args[arg]);
-    }
+    args.forEach((arg: string, index: number) => {
```
Colengms, line 32: One issue with using forEach instead of for-in loops is that you can't automatically step into them without a breakpoint.  What was the issue causing the switch to forEach?

Extension/src/LanguageServer/client.ts
```diff
@@ -148,23 +149,51 @@ function showWarning(params: ShowWarningParams): void {
...
 function publishDiagnostics(params: PublishDiagnosticsParams): void {
-    if (!diagnosticsCollection) {
-        diagnosticsCollection = vscode.languages.createDiagnosticCollection("C/C++");
+    const is_intelliSense: boolean = params.diagnosticsType === DiagnosticsType.IntelliSense;
+    if (is_intelliSense) {
+        if (!diagnosticsCollectionIntelliSense) {
+            diagnosticsCollectionIntelliSense = vscode.languages.createDiagnosticCollection("C/C++");
+        }
+    } else if (!diagnosticsCollectionCodeAnalysis) {
+        diagnosticsCollectionCodeAnalysis = vscode.languages.createDiagnosticCollection("clang-tidy");
```
Colengms, line 184: Is the name passed to createDiagnosticCollection used for anything?  The docs say, "Every diagnostic from this collection will be associated with this name", but I don't see it displayed in the problems list with 1.7.0-dogfood2.

Extension/src/LanguageServer/client.ts
```diff
@@ -148,23 +149,51 @@ function showWarning(params: ShowWarningParams): void {
 }
 
 function publishDiagnostics(params: PublishDiagnosticsParams): void {
```
Colengms, line 151: Could you break this into 2 separate functions?  I use a separate function for CDD diagnostics.  That should also address the call to timeTelemetryCollector.setUpdateRangeTime(), which should only be called for IntelliSense diagnostics.

Extension/src/LanguageServer/settings.ts
```diff
@@ -68,55 +68,112 @@ export class CppSettings extends Settings {
...
+    public get clangTidyPath(): string | undefined {
+        return this.getClangPath(false);
     }
 
     public get clangFormatPath(): string | undefined {
-        let path: string | undefined | null = super.Section.get<string>("clang_format_path");
+        return this.getClangPath(true);
+    }
+
+    private getClangPath(isFormat: boolean): string | undefined {
```
Colengms, line 111: There is an open issue about the logic in this function: https://github.com/microsoft/vscode-cpptools/issues/6294#issuecomment-858205479  Not sure if you want to address that or not.  But, FYI, that issue would extend to clang-tidy as well.

Extension/src/Debugger/configurations.ts
sean-mcmanus: Because of the new ts lint rule: "@typescript-eslint/no-for-in-array": "error", .

Extension/src/LanguageServer/client.ts
sean-mcmanus: It's not displayed to the user...i.e. not sure why the API needs it. It's so we can clear/set the errors independently. I could look into it more...

Extension/src/Debugger/configurations.ts
Colengms: I guess I'm asking why it was necessary to add that rule.  :)

Extension/src/LanguageServer/client.ts
Colengms: I used the same string ("C/C++") for CDD, and was still able to clear/set the errors independently, due to using a separate DiagnosticsCollection object.  ... My guess is that the name is shown in the UI if another field is unset.

Extension/src/Debugger/configurations.ts
bobbrow: Because it's [bad practice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in) to use `in` with arrays.  You can use `for..of` instead.

Colengms: Ok.  I suggest we always prefer `for..of` over `forEach`, due to the behavior when stepping.

Extension/package.json
sean-mcmanus: I can follow VS Code's search.maxResults example:
image

Extension/src/Debugger/configurations.ts
sean-mcmanus: But in this case the "index" provided by the .foreach is desired, so it seems better (less code) than for...of.

Extension/src/LanguageServer/client.ts
sean-mcmanus: It looks like something is broken currently such that only 1 set is visible at a time (IntelliSense or clang-tidy) -- might be caused by the newer version of VS Code, since I'm pretty sure this was working previously?

sean-mcmanus: Oh, I think my recent change just regressed it...

sean-mcmanus: I fixed the regression bug. The name doesn't seem to have any effect I can tell. Did you think the name should be changed?

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure...

Extension/package.nls.json
```diff
@@ -22,6 +22,38 @@
...
+    "c_cpp.configuration.codeAnalysis.runInBackground.markdownDescription": { "message":  "If `true`, code analysis will run automatically in the background.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.runOnBuild.markdownDescription": { "message": "If `true`, code analysis will run automatically after a build is completed.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.clangTidy.enabled.markdownDescription": { "message": "If `true`, code analysis using `clang-tidy` will be enabled.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
bobbrow: I think the description for `enabled` should reference that it enables clangTidy for `runInBackground`.

Extension/package.nls.json
```diff
@@ -22,6 +22,38 @@
...
+    "c_cpp.configuration.codeAnalysis.excludeBoolean.markdownDescription": { "message": "The glob pattern to match file paths against. Set to `true` or `false` to enable or disable the pattern.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.excludeWhen.markdownDescription": { "message": "Additional check on the siblings of a matching file. Use `$(basename)` as variable for the matching file name.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.runInBackground.markdownDescription": { "message":  "If `true`, code analysis will run automatically in the background.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
bobbrow: I wonder if we should call this `runAutomatically` instead. It doesn't really matter that it runs in the "background."  Can we also update the description to add that code analysis runs automatically when files are edited?  We don't have to explain all the rules for cancelation, etc.  It's enough just to say that code analysis will start after files change.

Extension/package.json
```diff
@@ -241,6 +241,1312 @@
...
+          },
+          "scope": "resource"
+        },
+        "C_Cpp.codeAnalysis.runInBackground": {
+          "type": "boolean",
+          "markdownDescription": "%c_cpp.configuration.codeAnalysis.runInBackground.markdownDescription%",
+          "default": true,
+          "scope": "resource"
+        },
+        "C_Cpp.codeAnalysis.runOnBuild": {
```
bobbrow: In case this was missed, I don't think we are doing `runOnBuild` for v1 either.

sean-mcmanus: Yes

Extension/package.nls.json
sean-mcmanus: Sure, I could add that.

Extension/package.nls.json
sean-mcmanus: It's not after edit, it's after it's saved or opened -- runOnOpenAndSave would be the most explicit. There is "refresh" and cancel/retry logic that can re-run it on settings changes or cancelation, but that applies regardless of the runInBackground setting. @jureid was using the "in the background" to be intentionally vague. I like being more specific...she might have more arguments for why the "in the background" might be preferred...I think there was thought that the "in background" behavior might change in the future maybe?

bobbrow: I don't think we need to be explicit in the setting name.  `runAutomatically` is as vague as `runInBackground` but slightly more descriptive ("run in background" could mean that you have to invoke it manually, but it runs in the background).  The description can go into details if we want to explain *when* it is run "automatically".

sean-mcmanus: That seems fine to me. Does @jureid have any concern with runAutomatically? We could also potentially split it into runOnOpen and runOnSave settings in case the user only wants it to happen in one of the cases but not the other, such as if they open a lot of files for browsing and only want clang-tidy for files they're editing, or they just want it to run on open but not repeatedly while editing/saving.

michelleangela (approved):

PR #7819, opened 2021-07-13

csigs:


sean-mcmanus (changes requested):
Don't check in the FRA regressions to ":" spacing...and the json.files.missing loc key changes.

sean-mcmanus (approved):

Issue #7773, opened 2021-07-03
Add .cpp at the end of the Textmate Colorization scopes.


REYNEP:
Type: Feature Request

Currently none of the textmate Scopes [a.k.a Semantic token] have .cpp in them. Without that, no Theme can target Specific Language. And After using a lot of themes, it feels like, Most of the Theme Authors does that. Vanilla VSCode textMate scopes does have `.cpp` like language specifications at the end. And that helps a lot.... 🙂

sean-mcmanus:
Are you referring to semantic token scopes (https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide) or lexical/syntax scopes (https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide)? Our extension doesn't provide any lexical scopes (they're obtained by VS Code from https://github.com/jeff-hykin/cpp-textmate-grammar/issues), and semantic tokens we provide are mapped to fallback textmate scopes by VS Code (https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#predefined-textmate-scope-mappings).

Colengms:
@REYNEP When VS Code looks up a TextMate scope, it uses a fully qualified scope (including ".cpp" if the file is C++).  I don't believe it's necessary for the C++ extension to specify this anywhere.  There is a fallback approach that is used to match a scope in a Theme.  It will try the fully qualified scope first, then reduce it until a match is found.  For example: `a.b.c.cpp` is tried first, then `a.b.c`, `a.b`, then `a`.

That said, is there a specific scenario in which this lookup is not working as expected for you?  Where exactly are you suggesting ".cpp" needs to be added?

Are you seeing an issue related to the TextMate fallback scopes used by our semantic token provider?  Our semantic token provider supports both C and C++ sources, so cannot specify ".c" or ".cpp" in the fallback Textmate scopes for semantic tokens in `semanticTokenScopes` in our `package.json`.  If VS Code is failing to add ".cpp" to those scopes when looking them up, that may be a VS Code issue.

REYNEP:
@Colengms & Sean-mcmanus, I apologize for late reply....

I think even if `a.b.c.cpp` is tried, it still doesn't work, In details, even if `.cpp` in the end is available, still yet the SEMANTIC TOKEN SCOPE added by the Enhanced Colorization won't contain .cpp

[...]

PR #7829, opened 2021-07-14
Provide separate semantic token fallback scopes for each supported language

Colengms:
This addresses: https://github.com/microsoft/vscode-cpptools/issues/7773

The root cause of the issue is: https://github.com/microsoft/vscode/issues/128565

Since fixing it in VS Code is not a high priority, this works around the issue by providing separate TextMate fallback scopes for all of our custom semantic tokens as well as standard semantic tokens.  When/If the issue is addressed in VS Code, this change can be reverted.  It should not cause a problem even if not reverted.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2641,60 +2641,356 @@
     ],
     "semanticTokenScopes": [
       {
+        "language": "c",
         "scopes": {
+          "namespace": [
+            "entity.name.namespace.c"
+          ],
+          "type": [
+            "entity.name.type.c"
+          ],
+          "type.defaultLibrary": [
+            "support.type.c"
+          ],
+          "struct": [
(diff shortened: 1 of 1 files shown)


sean-mcmanus (changes requested):
Should the original non-language scopes still exist, such as entity.name.namespace ?

Colengms:
> Should the original non-language scopes still exist, such as entity.name.namespace ?

I don't think so.  I should work properly with the most-qualified name.  A theme style is looked up first with the fully qualified scope, and the fallback process removes suffixes.  For example, `entity.name.namespace.cpp` is searched for first.  If no style is found, `entity.name.namespace` is tried, then `entity.name`, then `entity`.

sean-mcmanus (approved):

Issue #7840, opened 2021-07-17
Add property for a `.config` file on `c_cpp_properties.json` to parse defines


microhobby:
Type: Feature Request

Some bare metal systems, RTOS and Linux use a `.config` file with defines needed to build the system. It would be really nice to have a configuration property where I can pass this file path to be the base of defines for the extension. Something like:

```json
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**",
                "${workspaceFolder}/../nuttx/**"
            ],
            "dotconfig": "${workspaceFolder}/.config",
            "defines": [],
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "gnu17",
            "cppStandard": "gnu++14",
            "intelliSenseMode": "linux-gcc-x64"
        }
    ],
    "version": 4
}
```

The pattern of the `.config` file is quite simple, one define/config per line ignoring the `#` comment lines:

```Kconfig
#
# Console Configuration
#
CONFIG_NSH_CONSOLE=y
# CONFIG_NSH_ALTCONDEV is not set
CONFIG_NSH_ARCHINIT=y
CONFIG_NSH_CMDLINE=""
# CONFIG_NSH_LOGIN is not set
# CONFIG_NSH_CONSOLE_LOGIN is not set
```

microhobby:
with a little more research I conclude that this might be somewhat unnecessary. Along with the `.config` the tools usually generate a `.h` with the defines, so it's just a matter of putting that `.h` in the include path.

microhobby:
Ok, changing my mind, not all `CONFIG_` defines are in `.h`. Another case use of this property would be to browse the code and have the structures and macros without having to compile it a first time, just by sharing the `.config` it is possible to have the same flow.

microhobby:
Closed by https://github.com/microsoft/vscode-cpptools/pull/7845

PR #7845, opened 2021-07-19
Add dotconfig property for c_cpp_properties.json

microhobby:
Some bare metal systems, RTOS and Linux use a generated .config file
with defines needed to build the system and switch features. This patch
add the `dotconfig` property that will store the path for read a
.config file and will concat the defines from file to the extension
defines.

Signed-off-by: Matheus Castello <matheus@castello.eng.br>


microhobby:
I thinking out loud: https://github.com/microsoft/vscode-cpptools/issues/7840

Extension/src/LanguageServer/configurations.ts
```diff
@@ -764,6 +781,10 @@ export class CppProperties {
                 configuration.includePath = includePath.concat(this.nodeAddonIncludes.filter(i => includePath.indexOf(i) < 0));
             }
             configuration.defines = this.updateConfigurationStringArray(configuration.defines, settings.defaultDefines, env);
+
+            configuration.dotconfig = this.updateConfigurationString(configuration.dotconfig, settings.defaultDotconfig, env);
+            configuration.defines = configuration.defines?.concat(this.getDotconfigDefines(configuration.dotconfig as string));
```
sean-mcmanus: Can configuration.defines be null or undefined? If so, I don't think this uses the .config defines.

sean-mcmanus:
@microhobby Is this ready for code review?

Does anyone know what "This branch is out-of-date with the base branch" means?

microhobby:
@sean-mcmanus I will rebase it with the base branch and remove the WIP from the title, so i think it's good for the review. Open to suggestions from maintainers.

Extension/src/LanguageServer/configurations.ts
microhobby: hmm, yeah the linter complained if I don't use the `?` here, you are right, let me check this

Extension/c_cpp_properties.schema.json
```diff
@@ -82,6 +82,10 @@
                         "type": "string",
                         "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8\\.1$"
                     },
+                    "dotconfig": {
```
bobbrow: Please update the name to `dotConfig` to match our property naming style.

Extension/c_cpp_properties.schema.json
```diff
@@ -82,6 +82,10 @@
...
+                        "description": "A path to .config file",
```
Colengms: Hi @microhobby .  Could you provide more information here about `.config` files in this description field?  Is the convention of using `.config` files specific to Linux kernel development?  Perhaps refer to that in the description, for users who are unfamiliar with the file and may not otherwise understand what this setting would do.  Also, since the title of the PR contains "WIP", I assume it's not ready to be signed off on.  When you are ready for us to sign off, could you remove that from the title?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -746,6 +747,22 @@ export class CppProperties {
...
+        const isWindows: boolean = os.platform() === 'win32';
+
+        try {
+            if (dotconfigPath !== undefined) {
+                const path: string = this.resolvePath(dotconfigPath, isWindows);
+                const configContent: string[] = fs.readFileSync(path, "utf-8").split("\n");
+                return configContent.filter(i => !i.startsWith("#") && i !== "");
+            }
+        } catch (err) {
+            const failedToResolve: string = localize("cannot.resolve.dotconfig.path", 'Invalid input, cannot resolve .config path');
```
Colengms: Since only the readFileSync call would throw, could you wrap only that with the try/catch?  Also, could you remove ".config" from the localizable portion of the string (to avoid having it translated)?  i.e.: `localize("cannot.resolve.dotconfig.path", "Cannot resolve {0} path: {1}", ".config");`  Also, consider logging an error to our output channel using  `getOutputChannelLogger().appendLine()` instead of display an error dialog.  (Users seem to dislike error dialogs).

Colengms (dismissed):

microhobby:
@bobbrow @sean-mcmanus and @Colengms thanks for the review, I will address the comments and as soon as I send the changes commits I will remove the WIP

Extension/src/LanguageServer/configurations.ts
microhobby: @Colengms I updated it and also added another commit to validate the path to check if the path is valid and if the path is a valid file, adding warning diagnostics to point it to the user during the file edition. Let me know what you think.

Extension/c_cpp_properties.schema.json
```diff
@@ -82,6 +82,10 @@
                         "type": "string",
                         "pattern": "^\\d{2}\\.\\d{1}\\.\\d{5}\\.\\d{1}$|^8\\.1$"
                     },
+                    "dotConfig": {
+                        "description": "A path to a .config file created by Kconfig system. Kconfig system generate a file with all the defines, CONFIG options selected, to build a project. Examples of projects that use Kconfig system is Linux Kernel and NuttX RTOS.",
```
sean-mcmanus: generate->generates

sean-mcmanus: I don't understand how this ", CONFIG options selected," fits into the sentence -- should the commas be removed and the 1st comma replaced by an "and" or is there another intended meaning?

sean-mcmanus: is->are

sean-mcmanus: "Linux kernel"->"the Linux kernel"

sean-mcmanus (changes requested):

Extension/c_cpp_properties.schema.json
microhobby: @sean-mcmanus Proposed changes applied, thanks for the review

Extension/ui/settings.ts
```diff
@@ -260,6 +261,7 @@ class SettingsApp {
 
             (<HTMLInputElement>document.getElementById(elementId.intelliSenseMode)).value = config.intelliSenseMode ? config.intelliSenseMode : "${default}";
             (<HTMLInputElement>document.getElementById(elementId.includePath)).value = joinEntries(config.includePath);
+            (<HTMLInputElement>document.getElementById(elementId.dotConfig)).value = joinEntries(config.dotConfig);
```
sean-mcmanus: Can you move this to the end of the settings (to the Advanced section)? Same for the case above with the elementId.

sean-mcmanus (changes requested):

Extension/ui/settings.ts
microhobby: @sean-mcmanus proposed changes applied, thanks for the review.

sean-mcmanus (approved):

sean-mcmanus:
@microhobby FYI, 1.6.0 is locked down for shipping next week, but this can go into 1.7.0.

sean-mcmanus:
@bobbrow @Colengms Can we check this in now?

sean-mcmanus (changes requested):
It seems like the package.json is missing the "default.dotConfig" setting and the package.nls.json is missing a description for that. I have a bunch of settings added in my branch at https://github.com/microsoft/vscode-cpptools/pull/7816/files if you want an example, but you don't need to use markdownDescription since that is only used in that branch and I can update it later.

microhobby:
> It seems like the package.json is missing the "default.dotConfig" setting and the package.nls.json is missing a description for that. I have a bunch of settings added in my branch at https://github.com/microsoft/vscode-cpptools/pull/7816/files if you want an example, but you don't need to use markdownDescription since that is only used in that branch and I can update it later.

@sean-mcmanus thanks for the examples, I add a commit adding it to the package and package.nls.json. Let me know if this is ok.

sean-mcmanus (approved):

sean-mcmanus:
@Colengms @michelleangela @bobbrow Does someone else want to review this too?

michelleangela (approved):

michelleangela (changes requested):
This property will also need to be added in the .html file of the IntelliSense Configuration UI.

See https://github.com/microsoft/vscode-cpptools/blob/main/Extension/ui/settings.html and the UI element for "Compile commands" property for example of creating a field for one file path entry.

Place UI element for dotConfig after "Forced include" in the UI. Use title "Dot Config" for the new property.

Also add this new property to https://github.com/microsoft/vscode-cpptools/blob/main/Extension/src/LanguageServer/settingsPanel.ts so that the UI will be updated with values from c_cpp_properties.json

See examples for `elementId.compileCommands` in settingsPanel.ts.

https://github.com/microsoft/vscode-cpptools/blob/5dcb3081e502f84185b71f773f3d106b5d27f63c/Extension/src/LanguageServer/settingsPanel.ts#L41
https://github.com/microsoft/vscode-cpptools/blob/5dcb3081e502f84185b71f773f3d106b5d27f63c/Extension/src/LanguageServer/settingsPanel.ts#L325

michelleangela:
@microhobby 
Here is how strings are localized in HTML.

To localize HTML content, the following 2 attributes are used on HTML nodes:

`data-loc-id`
Specifying a unique string key value using this attribute causes the content of the node to be localized associated with that key
	
`data-loc-id-<attribute-name>`
Specifying a unique string key value using this attribute causes the specified attribute of this node to be localized.  For example, to localize the title attribute of a node, use: data-loc-id-title="some.title.id"

The following is an example of a span with localized content, containing a sub-node with a localized title attribute.

```
<span data-loc-id="switch.to.json">Switch to the <a href="command:C_Cpp.ConfigurationEditJSON" title="Edit configurations in JSON file" data-loc-id-title="edit.configurations.in.json">c_cpp_properties.json</a> file by clicking on the file link or using the command:</span>
```

Child nodes are not included when content of a parent node is localized.  In the outer nodes of the above example, the following string will be extracted for localization:

"Switch to the {0} file by clicking on the file link or using the command:"

Extension/ui/settings.ts
```diff
@@ -270,6 +271,7 @@ class SettingsApp {
             (<HTMLInputElement>document.getElementById(elementId.compileCommands)).value = config.compileCommands ? config.compileCommands : "";
             (<HTMLInputElement>document.getElementById(elementId.configurationProvider)).value = config.configurationProvider ? config.configurationProvider : "";
             (<HTMLInputElement>document.getElementById(elementId.forcedInclude)).value = joinEntries(config.forcedInclude);
+            (<HTMLInputElement>document.getElementById(elementId.dotConfig)).value = joinEntries(config.dotConfig);
```
michelleangela: Since `config.dotConfig` will only have one file path, this does not need `joinEntries`.

For setting elements with one value:
```
(<HTMLInputElement>document.getElementById(elementId.dotConfig)).value = config.dotConfig ? config.dotConfig : "";
```

microhobby:
@michelleangela thanks for the review, I will work on it.

sean-mcmanus:
FYI, we've changed our settings strings to prefer using markdownDescription, with a hint string -- see https://github.com/microsoft/vscode-cpptools/pull/8129 for examples, so for this change you'd use backticks around .config and replace the \"\" with ``.

microhobby:
@sean-mcmanus  @michelleangela first of all sorry for the delay, I rebased it and addressed the proposed changes.
@michelleangela let me know if the html UI changes looks good to you, or if I missed something.

sean-mcmanus (approved):

sean-mcmanus:
It says "This branch is out-of-date with the base branch" -- can that be updated?

michelleangela:
@microhobby 
I tested the Configuration UI changes. The input field works and updates the JSON file and vice versa. It is only missing the error message on the UI when the path file is invalid. The UI element for the error message is already there in the `settings.html' file but the string for the error message is missing.

Note that the error message for invalid paths on JSON file works.

**Example of expected error on UI:**
image

**Where to get error message for the UI and updating errors on UI:**

1. UI error messages are created in function `private getErrorsForConfigUI(configIndex: number)` (https://github.com/microsoft/vscode-cpptools/blob/main/Extension/src/LanguageServer/configurations.ts#L1365)
2. In `getErrorsForConfigUI` add something like `errors.dotConfig = this.validatePath(config.dotConfig, false);` similar to `errors.compileCommands = this.validatePath(config.compileCommands, false);`
3. Another string property will need to be added to `interface ConfigurationErrors` for dotConfig errors.
4. Errors on the UI are updated via the `private updateErrors(errors: any)` function in https://github.com/microsoft/vscode-cpptools/blob/main/Extension/ui/settings.ts. See examples `this.showErrorWithInfo(elementId.compileCommandsInvalid, errors.compileCommands);` and `compileCommandsInvalid: "compileCommandsInvalid",` for adding another error message to update.

microhobby:
@michelleangela thank you very much for the valuable tips, I push a new commit, tested here and now the error message appears as expected.

michelleangela (approved):

michelleangela:
@microhobby Could you update your branch again so it is up-to-date with the base branch? Then I'll re-run the checks and check-in the PR.

michelleangela:
@microhobby 
Your changes to add `dotConfig` property has been checked-in. Thank you for your contribution to the extension!

PR #7850, opened 2021-07-20
Fix search not finding results in package.nls.json.

sean-mcmanus:


sean-mcmanus:
@Colengms There wasn't a reason for `**/*.nls.json` right? It seems like only `**/*.nls.*.json` is what we want. It also seems like `**/nls.*.json` is not used, but it isn't causing any issue so I just kept it.

Colengms:
> @Colengms There wasn't a reason for `**/*.nls.json` right? It seems like only `**/*.nls.*.json` is what we want. It also seems like `**/nls.*.json` is not used, but it isn't causing any issue so I just kept it.

It might be left over from before I was using 'bundles' for localized strings.  But, yes, this should be OK to remove.

Colengms (approved):

Issue #4827, opened 2020-01-09
Alpine operating system support


jaudiger:
Type: General
I'm not sure if this issue is already tracked somewhere else. I checked before to open this one, and I only found pull requests talking about disabling the support of the extension under Alpine and ARM environment:
- https://github.com/microsoft/vscode-cpptools/pull/4027
- https://github.com/microsoft/vscode-cpptools/pull/4066

**Describe the bug**

- OS and Version: Alpine 3.11
- VS Code Version: 1.41
- C/C++ Extension Version: 2.26
- C/C++ Extension does not support Alpine environment due to the use of musl library.

I don't know what are the dependencies of the C/C++ Extension since this is not an open source extension. But what is the cost to add the support of musl library in addition of the usual libc library ?

I'm currently looking to mount a development Docker environment based on an Alpine operating system.

lodo1995:
+1 on this.
I think this feature might get more and more requested given that 1) Remote-Containers is getting more popular to develop on Windows machines without cluttering them 2) Alpine is getting very popular because of its speed, and is starting to support more libraries and tools.

whitetigle:
Same here, would definitely love to see that coming so that I can use it on gitpod using Alpine. Thanks!

Trass3r:
So what exactly isn't working and how does it depend on musl?

sean-mcmanus:
@Trass3r Are you asking us? I think the cpptools/cpptools-srv doesn't run. @andreeis Might know more.

Trass3r:
Yes, so we get that message alpine is not supported but it's not clear which features are affected.
vscode-cpptools consists of different components. I personally only use the debugger as there are other solutions for LSP.

sean-mcmanus:
@WardenGnaw Do you know anything about Alpine support for the debugger?

WardenGnaw:
If there is a Mono Runtime (using musl) that works on Alpine, we can ship the debugger to work on Alpine.

From when I last checked, there was not an officially supported Mono Runtime for Alpine.

jaudiger:
There is a mono package (https://pkgs.alpinelinux.org/package/edge/testing/x86_64/mono), but it's not ready for production right now. The package is only available on edge branch, not on the 3.12 release. And it's tagged as *testing*.

Trass3r:
Indeed. It can be installed via
```sh
sed -i -e 's/v[[:digit:]]\..*\//edge\//g' /etc/apk/repositories
[...]

PR #7858, opened 2021-07-22
Update Debugger Packages

WardenGnaw:
Debugger now ships using .NET 5 Runtime to run OpenDebugAD7.

Supports:
- win7-x86
- win-x64
- win10-arm64
- linux-x64
- linux-musl-x64
- linux-arm
- linux-arm64
- osx-x64

Summary:
- Cleaned up unused setup files.
- Removed 'removeUnnecessaryFile' step since that removed Windows .NET Framework configurations for non-Windows (mono) OpenDebugAD7.


Extension/package.json
```diff
@@ -3309,67 +3309,116 @@
       "integrity": "6756365C82B5BD90C6A40C13CF8E3599E4A1DB50628111FFDB3A76FF85256BA4"
     },
     {
-      "description": "Mono Framework Assemblies",
-      "url": "https://go.microsoft.com/fwlink/?LinkId=2027135",
+      "description": "OpenDebugAD7 (Windows x86)",
+      "url": "https://download.visualstudio.microsoft.com/download/pr/65d2e4a8-b9e4-40e8-a131-51b3b6af7b43/de9fa59805117816e8a5f94c8266f08f/win7-x86.zip",
```
sean-mcmanus, line 3313: Aren't we supposed to use fwlinks?

Extension/package.json
```diff
@@ -3309,67 +3309,116 @@
...
-        "./debugAdapters/mono.linux-arm64"
+        "./debugAdapters/bin/OpenDebugAD7"
       ],
-      "integrity": "946C54C8C6BF5BF79AC05D4E152F8D8647700FA786C21505832B3C79988339B4"
+      "integrity": "2EB2F631C70AB4C6717D5AC61009051F6EFD4BFAF68E7364D95533CA2A7F2F03"
     },
     {
-      "description": "Mono Runtime (macOS)",
-      "url": "https://go.microsoft.com/fwlink/?LinkId=2027403",
+      "description": "OpenDebugAD7 (Alpine Linux x64)",
```
sean-mcmanus, line 3396: Should we disable the "Alpine containers are not supported." message and instead just disable the language service stuff (as if intelliSenseEngine was "Disabled") so this can be used?

sean-mcmanus (changes requested):

Extension/package.json
WardenGnaw: I was not sure if we wanted to ship without language service. 

Theres a couple issues with the package manager needing to detect Alpine Linux and the ConfigurationProvider seems to be integrated with the language service. I will need to either remove the configuration provider during alpine launch or refactor so it does not use the language service in that scenario.

Extension/package.json
WardenGnaw: Is there a reason to do so? If we need to ship a new version of the debugger, there will be a new link anyways.

We ship with the VS CDN in the C# extension: https://github.com/OmniSharp/omnisharp-vscode/blob/724c565ac5ba7276b2d0a54ca7204b396a44501e/package.json#L220

Extension/package.json
sean-mcmanus: Now or later? The issue tracking alpine is at https://github.com/microsoft/vscode-cpptools/issues/4827

sean-mcmanus (approved):

Extension/package.json
WardenGnaw: I'll do it in another check-in. For now, lets get our currently supported platforms updated.

Trass3r:
>     * linux-musl-x64

The problem is there is no actual musl build: https://github.com/microsoft/vscode-cpptools/releases/tag/1.6.0-insiders

sean-mcmanus:
@Trass3r That's being tracked by https://github.com/microsoft/vscode-cpptools/issues/4827...not sure if we want to enable debugging on Alpine if the language service isn't available.

Trass3r:
I personally only need the debugging functionality.
Clangd does the language server part.

PR #7880, opened 2021-07-27
Retry more.

sean-mcmanus:
Add more retries. The failure issue today would fix itself if it retried more, but I'm not sure what the ideal retry amount is because the issue stopped reproing. Maybe we could check telemetry afterwards to see if the number of failures decreased.

### DIFF
--- a/Extension/src/packageManager.ts
+++ b/Extension/src/packageManager.ts
@@ -189,7 +189,7 @@ export class PackageManager {
         let success: boolean = false;
         let lastError: any = null;
         let retryCount: number = 0;
-        const MAX_RETRIES: number = 5;
+        const MAX_RETRIES: number = 10;
 
         // Retry the download at most MAX_RETRIES times with 2-32 seconds delay.
         do {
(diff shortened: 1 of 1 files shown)


Colengms:
Depending on what caused the error, retries might not help.  The outage seemed to repro consistently for certain systems, and for quite a while (an hour?).  If a DNS issue, a client machine might keep retrying with the cached IP it got from an earlier DNS request.

Colengms (approved):

sean-mcmanus:
> Depending on what caused the error, retries might not help. The outage seemed to repro consistently for certain systems, and for quite a while (an hour?). If a DNS issue, a client machine might keep retrying with the cached IP it got from an earlier DNS request.

Yeah, but in this particular incident, I was seeing the failures be random, so retrying fixed it if enough retries were done.

bobbrow:
I think it's too many retries.  This is supposed to be about the developer's connection being bad, not about potential CDN incidents.

sean-mcmanus:
@bobbrow I reduced the max delay between retries to 15 seconds: https://github.com/microsoft/vscode-cpptools/pull/7883 .

PR #7887, opened 2021-07-28
Use new modal w/ detail dialog for compiler list in Walkthrough

Colengms:
VS Code added a new modal dialog which displays details less prominently than the main message.  This change switches to using that UI for the detected compiler list that can be displayed via the Getting Started walkthrough.

It was (on Windows):
image

It's now (on Windows):
image

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -11,7 +11,7 @@
   },
   "license": "SEE LICENSE IN LICENSE.txt",
   "engines": {
-    "vscode": "^1.57.0"
+    "vscode": "^1.58.0"
   },
   "bugs": {
     "url": "https://github.com/Microsoft/vscode-cpptools/issues",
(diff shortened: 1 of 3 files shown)


Extension/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "license": "SEE LICENSE IN LICENSE.txt",
   "engines": {
-    "vscode": "^1.57.0"
+    "vscode": "^1.58.0"
```
sean-mcmanus, line 14: Is this going to be too recent for 1.6?

Colengms: The current released version of VS Code is 1.58.2.  If we need to lag behind publicly available versions, how far do we need to lag behind?

sean-mcmanus: I'm not sure. It somewhat depends on how much we need the new feature too. Maybe ask Bob? Bob said the change to the 1.57.0 dependency was okay.

jureid:
Let's shorten the message now that it's a heading. Can we try this:

"We found the following C++ compilers on your system. Choose a compiler in your project's IntelliSense Configuration."

I don't think we need the parentheses in compiler(s). If the list only contains 1, I think it still reads well.

sean-mcmanus (approved):

PR #7895, opened 2021-07-30
parse arguments in tasks.json

### LINKED ISSUE #7891: Linux: Cannot use `pkg-config` arguments in `task.json`
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 
- OS Version: Pop! OS 20.04 LTS

Steps to Reproduce:

1. Add a `pkg-config` instruction to the task.json args list.
2. Run Build Task

This is a known issue, but it was never fixed, but closed because a workaround was in place (inlining the args into the compiler command). https://github.com/Microsoft/vscode/issues/47538

hello.c
```c
#include "stdio.h"
#include "glib.h"

int main (void) {
    printf("Hello, World!\n");

    return 0;
}
```
task.json
```json
{
	"version": "2.0.0",
	"tasks": [
        {
            "type": "cppbuild",
            "label": "C/C++: gcc build active file",
            "command": "/usr/bin/gcc",
            "args": [
                "-g",
                "`pkg-config --cflags glib-2.0`",
                "${file}",
                "`pkg-config --libs glib-2.0`",
                "-o",
                "${fileDirname}/${fileBasenameNoExtension}"
            ],
            "options": {
                "cwd": "${fileDirname}"
            },
            "problemMatcher": [
                "$gcc"
            ],
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "detail": "compiler: /usr/bin/gcc"
        }
    ]
}
```

**OUTPUT:**
```none
> Executing task: C/C++: gcc build active file <

Starting build...
/usr/bin/gcc -g "`pkg-config --cflags glib-2.0`" /home/zak/Development/test/c/glib/ray.c "`pkg-config --libs glib-2.0`" -o /home/zak/Development/test/c/glib/ray
/home/zak/Development/test/c/glib/ray.c:2:10: fatal error: glib.h: No such file or directory
    2 | #include "glib.h"
      |          ^~~~~~~~
compilation terminated.

Build finished with error(s).
The terminal process failed to launch (exit code: -1).

Terminal will be reused by tasks, press any key to close it.
```

working `args`:
```json
"args": [
    "-g",
    "-I/usr/include/glib-2.0",
    "-I/usr/lib/x86_64-linux-gnu/glib-2.0/include",
    "${file}",
    "-lglib-2.0",
    "-o",
    "${fileDirname}/${fileBasenameNoExtension}"
],
```

> _Originally reported on wrong repository and forwarded here:_
> https://github.com/microsoft/vscode/issues/126682

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/7891
also partially related to: https://github.com/microsoft/vscode-cpptools/issues/6773

@sean-mcmanus  the part related to showing errors and warnings might need some modification.


Extension/src/common.ts
```diff
@@ -1304,3 +1304,16 @@ export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Pr
         return promiseBuilder(nextItem);
     }, Promise.resolve());
 }
+
+// Quote a file path if it includes space.
+export function quoteFilePaths(path: string): string {
+    // Check if the path is not already quoted.
+    if (!/".*?"/.test(path) && !/'.*?'/.test(path)) {
```
Colengms: * #7891 relates to use of backtick quoting (which is only supported on mac and linux), but I don't see any handling of backtick quoting here.
* It looks like you are using a pattern to check if the entire path is surrounded by quotes.  Quoted portions may be contained anywhere within an argument. If there is already quoting within an arg, placing quotes around the whole arg would requiring escaping those internal quotes.  Beware if detecting internal quotes within an arg, as they should not be considered as quoting if escaped (`\"`)
* It's unnecessary to detect if the arg contains a path, as args that contain paths are no different from any other arg contents with regards to whether or not they require quoting.  i.e. macros/defines may have spaces.  (So, this function should be something like `quoteArgument`).  Since it's perfectly acceptable to quote an argument that does not contain spaces, consider quoting ALL arguments (if not already fully quoted), and escaping any quoting within it.  That might be easier than detecting if quoting is necessary.  Backticks might be different - I suspect that must always fully quote an arg and cannot be escaped.
* Single-quotes are supported only on Linux.  They behave as literals on Windows, so shouldn't be checked for on Windows.

Colengms (changes requested):

Extension/src/common.ts
elahehrashedi: based on our previous discussion:
Assumption: Backtick quoting must be around the entire argument. (Whereas single-quote and double-quote quoting can start in the middle of an argument. For example: -DTEST="TEST TEST", or -DTEST='TEST TEST' on Linux/Mac



The algorithm:



1. Trim the arg to remove leading and trailing spaces.
2. (On Linux/Mac) If completely enclosed within backtick quoting, leave it alone.
3. Check if the string contains any escaped double-quote characters. If so, leave the arg alone, as the user is doing the quoting themselves. Adding additional quoting would change the behavior. (i.e. -DTEST="TEST TEST" is different from "-DTEST=\"TEST TEST\"")
4. On Linux or Mac, check if the string contains any unescaped single-quote characters. If so, leave the arg alone. (same reason as above).
5. Scan for any spaces not preceded by a slash. Only if there is an un-escaped space is quoting needed.
6. If any un-escaped space is found, remove the "\" from any escaped spaces ("\ ", in case there were some as they will no longer be valid if quoted), and put double-quotes around the whole thing.

sean-mcmanus:
What do you mean by "@sean-mcmanus the part related to showing errors and warnings might need some modification."? Can you test the scenarios to make sure they all work okay still?

sean-mcmanus (approved):

elahehrashedi:
@sean-mcmanus one scenario is not working fine. In a test project, GCC is throwing an error, but `error` is empty and `stderr && !stdout` is true (i.e. stderr has the error message in it and stdout doesn't have any content). But the current code identifies that as a warning. 
I am using the previous code. other scenarios were fine.

sean-mcmanus:
> @sean-mcmanus one scenario is not working fine. In a test project, GCC is throwing an error, but `error` is empty and `stderr && !stdout` is true (i.e. stderr has the error message in it and stdout doesn't have any content). But the current code identifies that as a warning. other scenarios were fine.

Can you change the code to correctly detect the error?

sean-mcmanus (approved):

Extension/src/common.ts
```diff
@@ -1304,3 +1304,21 @@ export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Pr
         return promiseBuilder(nextItem);
     }, Promise.resolve());
 }
+
+export function normalizeArg(arg: string): string {
+    arg = arg.trimLeft().trimRight();
+    // Check if the arg is enclosed in backtick,
+    // or includes escaped double-quotes,
```
Colengms: This is a really hard problem, actually...  I think we really need a test case for this, so we can plug in various different scenarios to ensure they are supported. 

Also, contrary to what I may have said (wrong) in a previous conversation, I think you want to check for any "unescaped" double-quotes here.  We should ignore any arg with un-escaped quotes in it, as adding quotes around that (even if escaping them internally) changes the arg...  For example:

```
$ g++ -E -dD -xc++ -DTESTDEFINE="asd asd" /dev/null | grep TESTDEFINE
#define TESTDEFINE asd asd
```

would become:

```
$ g++ -E -dD -xc++ "-DTESTDEFINE=\"asd asd\"" /dev/null | grep TESTDEFINE
#define TESTDEFINE "asd asd"
```

Colengms (changes requested):

sean-mcmanus:
@elahehrashedi Is this ready for code review now?

Colengms (approved):

Issue #7920, opened 2021-08-05
Support standard `.editorconfig` fields


Colengms:
Since `cpp_*` fields in `.editorconfig` files are supported (when `"C_Cpp.formatting": "vcFormat"`), users may expect the standard `.editorconfig` fields to also be applied.  Currently, this requires installing another extension specifically for that support.  (i.e. the "EditorConfig for Visual Studio Code" extension)

Supporting those fields would seem to involve setting the equivalent VS Code settings, as they are applied in scenarios other than processing a "Format" operation.  Also, unlike `.editorconfig`, the VS Code setting equivalent of `indent_style`, `indent_size`, `tab_width`, `end_of_line`, etc., could not be more granular than workspace-level.

ChrisGuzak:
This is very important for my intended use in the OS.2020 repo. thanks for filing this.

bobbrow:
Does it present a problem for us to support those if the EditorConfig extension is installed?  IOW, will the changes be applied twice?

Colengms:
@bobbrow Based on it's code [here](https://github.com/editorconfig/editorconfig-vscode), it looks like it will apply the settings directly to the document in response to notifications from `onDidChangeActiveTextEditor` or `onDidChangeWindowState`.  With that approach instead of modifying settings, we can support finer (per folder) granularity.  I assume the extension that changes the options last would take precedence.  As long as each extension is setting the same values for these settings, it should work as expected.

Colengms:
This should be addressed in 1.6.0-insiders.

bl-ue:
>This is very important for my intended use in the OS.2020 repo. thanks for filing this.

@ChrisGuzak what's that?

PR #7927, opened 2021-08-06
Add support for most standard `.editorconfig` settings

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7920

Adds support for:
* `indent_style`
* `indent_size`
* `tab_size`
* `end_of_line` (applied any time the document is activated, as an unsaved edit.  This is equivalent to selecting the newline mode from the menu at the bottom of the window).
* `insert_final_newline` (applied only when "Format Document" is used, so also on save if using `editor.formatOnSave`.)

vcFormat appears to always do `trim_trailing_whitespace`.

That leaves `charset`.  Support for which appears to be blocked on: https://github.com/microsoft/vscode/issues/824


Extension/src/LanguageServer/client.ts
```diff
@@ -2387,10 +2390,47 @@ export class DefaultClient implements Client {
         return this.requestWhenReady(() => this.languageClient.sendRequest(SwitchHeaderSourceRequest, params));
     }
 
+    private async updateActiveDocumentTextOptions(): Promise<void> {
+        const editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+        if (editor
+            && editor.document
+            && editor.document.uri.scheme === "file"
```
sean-mcmanus: You could remove `editor && editor.document` if you use `editor?.document?.uri.scheme`

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -42,6 +42,31 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
                     newText: textEdit.newText
                 });
             });
+            // Apply insert_final_newline from .editorconfig
+            if (document.lineCount > 0 && editorConfigSettings !== undefined && editorConfigSettings.insert_final_newline) {
+                // Check if there is already a newline at the end.  If so, formatting edits should not replace it.
+                if (!document.lineAt(document.lineCount - 1).isEmptyOrWhitespace) {
+                    const endPosition: vscode.Position = document.lineAt(document.lineCount - 1).range.end;
```
sean-mcmanus: You could avoid this 2nd call if you saved the 1st call in a variable.

Extension/src/LanguageServer/client.ts
```diff
@@ -2387,10 +2390,47 @@ export class DefaultClient implements Client {
...
+            && (editor.document.languageId === "c"
+                || editor.document.languageId === "cpp"
+                || editor.document.languageId === "cuda-cpp")) {
+            // If using vcFormat, check for a ".editorconfig" file, and apply those text options to the active document.
+            const settings: CppSettings = new CppSettings();
```
sean-mcmanus: I believe you should be passing the rootUri here or else you might not get the settings for the current workspace folder.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: Hmm.  I'm wondering if this is correct in a few other places.  Should the existing occurrence in `onDidChangeSettings()` also use `this.RootUri`?

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: It's incorrect whenever a "resource" or "machine-overridable" scope setting is accessed, which is the case for "formatting" in onDidChangeSettings -- we could have other cases too.

PR #7933, opened 2021-08-06
Telemetry tweaks

bobbrow:
While reviewing the exp platform changes for CMake Tools, a few issues were noticed that we wanted to bring fixes back to cpptools for:
* Incorrect placement of undefined in type annotation. For some reason TypeScript does not complain about `getExperimentationService` returning undefined
* `experimentationTelemetry` should always be disposed
* Prefer not to use await for logging telemetry. It's not an issue in cpptools right now, but if the no-floating-promises linter rule is enabled it wants you to annotate every call to `log*Event`.


Extension/src/telemetry.ts
```diff
@@ -92,34 +92,50 @@ export async function deactivate(): Promise<void> {
...
+    const sendTelemetry = () => {
+        if (experimentationTelemetry) {
+            const eventNamePrefix: string = "cppdbg/VS/Diagnostics/Debugger/";
+            experimentationTelemetry.sendTelemetryEvent(eventNamePrefix + eventName, properties);
+        }
+    };
+
+    if (initializationPromise) {
+        try {
+            // Use 'then' instead of 'await' because telemetry should be "fire and forget".
```
Colengms, line 111: `logDebuggerEvent` and `logLanguageServerEvent` were already "fire and forget" due to not awaiting calls to it.  We don't use `no-floating-promises`.  There are many cases in which async functions are not being awaited, and cannot be awaited (such as when registered as a callback to a `registerCommand` or when registering for various VS Code events.  There are other intentional cases a well).  Are you sure you want to take this change an incur more calls to `then`?

bobbrow: My intention was to keep the two implementations in sync.  If you don't want this part of the change I can back it out.  There's nothing wrong with using `.then` for intentional "fire and forget" cases.  We still have several `.then` calls in the extension.

Colengms (approved):

sean-mcmanus (approved):

sean-mcmanus:
@bobbrow Are you going to check this in now?

sean-mcmanus:
@bobbrow Did you want to check this in?

bobbrow:
Sorry, I thought you were going to merge this for me.

bobbrow:
Thanks @Colengms.  I had updated the branch, and then forgot again.

Issue #7929, opened 2021-08-06
Default to vcFormat if an `.editorconfig` containing vcFormat entries was found nearer than any `.clang-format` file


Colengms:
Currently, the `Default` setting of `Cpp.formatting` is the same as `clangFormat`.

This would redefine `Default` to be dynamic selection of either clang-format or vcFormat depending on whether a `.clang-format` file or an `.editorconfig` file (containing vcFormat entries) are found, or which is nearer the file to be formatted.

Defaulting to clang-format if both are found, and only using an `.editorconfig` if it contains vcFormat entries, would avoid changing the behavior for users who either intended to use clangFormat, are configured for both (so it shouldn't matter which is used), or might otherwise not appreciate having their formatting arbitrarily changed.  The existance of the vcFormat entries and no `.clang-format` file (at better or equivalent distance), would seem a reasonable indication that vcFormat is intended to be used.

Instead of sending the value of `Cpp.formatting` to the native side with other settings, it could be sent only per formatting request, allowing it to be varied on the TypeScript side when set to `Default`.

sean-mcmanus:
I don't think the "distance" should be a factor.

Colengms:
@sean-mcmanus .  Considering these two scenarios:

```
workspace_root\.clang-format
workspace_root\subdirectory1\.editorconfig
workspace_root\subdirectory1\subdirectory2\source.cpp
```

```
workspace_root\.editorconfig
workspace_root\subdirectory1\.clang-format
workspace_root\subdirectory1\subdirectory2\source.cpp
```

By "distance", I'm suggesting the first example would use `.editorconfig` and the second example use `.clang-format`, based on being nearer the source file.  

Are you suggesting that both examples should use `.clang-format` or both should use `.editorconfig`?

sean-mcmanus:
Okay, changed my mind, distance seems okay to consider.

Colengms:
This is addressed in 1.6.0-insiders.

PR #7941, opened 2021-08-09
Add conditional use of vcFormat when `formatting` is set to `Default`

Colengms:
Implement conditional enabling of vcFormat support if an `.editorConfig` file is found (with vcFormat entries) and a `.clang-format` file is not found nearer the file.  This algorithm  is described in: https://github.com/microsoft/vscode-cpptools/issues/7929

The results of searching is cached on a per-file basis.  (As the relevance of contents in `.editorconfig` can vary based on a filename pattern).

Does not search further up the path for `.editorconfig` if `root` is found with a value of `true`.

Does not attempt to switch to vcFormat if `clang_format_style` is not set to `file`.  (`Default` formatting with a non-default `clang_format_style` will continue to use clang-format).

Moved some EditorConfig related code into the `CppSettings` class.

Changed the default value of the `formatting` setting to `Default` instead of `clangFormat`.

Removed the `formattingEngine` setting from the block of settings sent to the native side, and instead always indicates whether to use clang-format or vcFormat in the formatting request.

Also fixed a bug in which standard `.editorconfig` settings were not applied as a file is opened on startup (where we don't get notification of a change of active document).

### DIFF
--- a/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
@@ -19,7 +19,7 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
         const useVcFormat: boolean = settings.useVcFormat(document);
         const configCallBack = async (editorConfigSettings: any | undefined) => {
             const params: FormatParams = {
-                settings: { ...editorConfigSettings },
+                editorConfigSettings: { ...editorConfigSettings },
                 useVcFormat: useVcFormat,
                 uri: document.uri.toString(),
                 insertSpaces: options.insertSpaces,
--- a/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
@@ -15,7 +15,7 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
     public async provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
         await this.client.awaitUntilLanguageClientReady();
         const filePath: string = document.uri.fsPath;
-        const settings: CppSettings = new CppSettings();
+        const settings: CppSettings = new CppSettings(this.client.RootUri);
         const useVcFormat: boolean = settings.useVcFormat(document);
         const configCallBack = async (editorConfigSettings: any | undefined) => {
             const params: FormatParams = {
--- a/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
@@ -3,9 +3,8 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import * as vscode from 'vscode';
-import { DefaultClient, FormatParams, FormatDocumentRequest, cachedEditorConfigSettings } from '../client';
-import { CppSettings } from '../settings';
-import * as editorConfig from 'editorconfig';
+import { DefaultClient, FormatParams, FormatDocumentRequest } from '../client';
+import { CppSettings, getEditorConfigSettings } from '../settings';
 
 export class DocumentFormattingEditProvider implements vscode.DocumentFormattingEditProvider {
     private client: DefaultClient;
@@ -16,9 +15,12 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
     public async provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
         await this.client.awaitUntilLanguageClientReady();
         const filePath: string = document.uri.fsPath;
+        const settings: CppSettings = new CppSettings();
+        const useVcFormat: boolean = settings.useVcFormat(document);
         const configCallBack = async (editorConfigSettings: any | undefined) => {
             const params: FormatParams = {
                 settings: { ...editorConfigSettings },
+                useVcFormat: useVcFormat,
                 uri: document.uri.toString(),
                 insertSpaces: options.insertSpaces,
                 tabSize: options.tabSize,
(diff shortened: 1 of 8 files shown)


Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -16,9 +15,12 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
     public async provideDocumentFormattingEdits(document: vscode.TextDocument, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
         await this.client.awaitUntilLanguageClientReady();
         const filePath: string = document.uri.fsPath;
+        const settings: CppSettings = new CppSettings();
```
sean-mcmanus: The resource this.RootUri should be passed, same for the other cases below.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/settings.ts
```diff
@@ -452,6 +461,224 @@ export class CppSettings extends Settings {
...
+            const editorConfigPath: string = path.join(parentPath, ".editorconfig");
+            if (fs.existsSync(editorConfigPath)) {
+                const editorConfigSettings: any = getEditorConfigSettings(document.uri.fsPath);
+                const keys: string[] = Object.keys(editorConfigSettings);
+                for (let i: number = 0; i < keys.length; ++i) {
+                    if (keys[i].startsWith("cpp_")) {
+                        foundEditorConfigWithVcFormatSettings = true;
+                        const didEditorConfigNotice: PersistentState<boolean> = new PersistentState<boolean>("Cpp.didEditorConfigNotice", false);
+                        if (!didEditorConfigNotice.Value) {
+                            vscode.window.showInformationMessage(localize("editorconfig.default.behavior",
```
sean-mcmanus, line 644: Is this notice really necessary/important to show? Seems to me like it's unnecessary.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/settings.ts
```diff
@@ -452,6 +461,224 @@ export class CppSettings extends Settings {
...
+                        return true;
+                    }
+                }
+                if (editorConfigSettings.root?.toLowerCase() === "true") {
+                    return true;
+                }
+            } else {
+                const clangFormatPath1: string = path.join(parentPath, ".clang-format");
+                if (fs.existsSync(clangFormatPath1)) {
+                    return true;
```
bobbrow, line 657: Should this be false?  Based on how it's used in the do..while, I thought you expected this to only return true for .editorconfig found.

Extension/src/LanguageServer/settings.ts
```diff
@@ -452,6 +461,224 @@ export class CppSettings extends Settings {
...
+            if (fs.existsSync(editorConfigPath)) {
+                const editorConfigSettings: any = getEditorConfigSettings(document.uri.fsPath);
+                const keys: string[] = Object.keys(editorConfigSettings);
+                for (let i: number = 0; i < keys.length; ++i) {
+                    if (keys[i].startsWith("cpp_")) {
+                        foundEditorConfigWithVcFormatSettings = true;
+                        const didEditorConfigNotice: PersistentState<boolean> = new PersistentState<boolean>("Cpp.didEditorConfigNotice", false);
+                        if (!didEditorConfigNotice.Value) {
+                            vscode.window.showInformationMessage(localize("editorconfig.default.behavior",
+                                "vcFormat was selected because an '.editorconfig' file was found containing vcFormat entries and 'Cpp.formatting' is set to 'Default'."));
```
bobbrow: I think this message should only appear when both .editorconfig and .clang-format are detected in the same workspace, so we shouldn't pop it up here.

Message suggestion: "Code formatting will use settings from .editorconfig instead of .clang-format because the configuration is closer to this file. For more information, see the documentation for the 'C_Cpp.formatting' setting."

bobbrow: (and we need to update `package.nls.json` with the more detailed information promised) 😄

Extension/src/LanguageServer/settings.ts
Colengms: I'm returning true from the lambda to indicate that a file was found (either `.clang-format` or `.editorconfig`).  Which one found is specified by `foundEditorConfigWithVcFormatSettings`.  If false is returned, it moves up to check the next parent directory.

Extension/src/LanguageServer/settings.ts
Colengms: Note that the scan continues up the directory structure beyond the workspace folder (which is by design).  Once we find one, we don't currently continue scanning for the other.

Lack of both files would still default to `clangFormat`, so the behavior for the user would change whenever a `.editorconfig` is selected (regardless of whether we found a `.clang-format` as well).  So, I think it would make more sense to display a message any time `.editorconfig` is selected - any time (er, the first time) the behavior may have been changed out from under them.  If you agree, would you like to suggest new text?  :)

Extension/src/LanguageServer/settings.ts
Colengms: Bob requested this message.  As mentioned in my response to Bob's comment, my intention was to display a message only the first time the new behavior is triggered.

Extension/src/LanguageServer/settings.ts
bobbrow: "Code formatting is using settings from .editorconfig instead of .clang-format. For more information, see the documentation for the 'C_Cpp.formatting.Default' setting."

Documentation for the "Default" enum value should be something like: "By default, clang-format will be used to format the code. However, the Visual C++ formatting engine will be used if an .editorconfig file with relevant settings is found nearer to the code being formatted and clang_format_style is the default value: 'file'."

Extension/src/LanguageServer/settings.ts
```diff
@@ -452,6 +461,224 @@ export class CppSettings extends Settings {
...
+            if (fs.existsSync(editorConfigPath)) {
+                const editorConfigSettings: any = getEditorConfigSettings(document.uri.fsPath);
+                const keys: string[] = Object.keys(editorConfigSettings);
+                for (let i: number = 0; i < keys.length; ++i) {
+                    if (keys[i].startsWith("cpp_")) {
+                        foundEditorConfigWithVcFormatSettings = true;
+                        const didEditorConfigNotice: PersistentState<boolean> = new PersistentState<boolean>("Cpp.didEditorConfigNotice", false);
+                        if (!didEditorConfigNotice.Value) {
+                            vscode.window.showInformationMessage(localize("editorconfig.default.behavior",
+                                "Code formatting is using settings from .editorconfig instead of .clang-format. For more information, see the documentation for the 'C_Cpp.formatting.Default' setting."));
```
sean-mcmanus: The "Default" is technically a value and not a setting...not sure is worth clarifying or not.

sean-mcmanus (approved):

PR #8003, opened 2021-08-18
Fix rewrite of updateChannel for remote VS Code insiders

Colengms:
The default value of `updateChannel` is rewritten in our package.json, when VS Code Insiders is detected.  That check was failing to detect VS Code Insiders when used in a remote scenario.

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -425,7 +425,7 @@ async function finalizeExtensionActivation(): Promise<void> {
     const packageJson: any = util.getRawPackageJson();
     let writePackageJson: boolean = false;
     const packageJsonPath: string = util.getExtensionFilePath("package.json");
-    if (packageJsonPath.includes(".vscode-insiders") || packageJsonPath.includes(".vscode-exploration")) {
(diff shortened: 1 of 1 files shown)


Extension/src/main.ts
```diff
@@ -425,7 +425,7 @@ async function finalizeExtensionActivation(): Promise<void> {
     const packageJson: any = util.getRawPackageJson();
     let writePackageJson: boolean = false;
     const packageJsonPath: string = util.getExtensionFilePath("package.json");
-    if (packageJsonPath.includes(".vscode-insiders") || packageJsonPath.includes(".vscode-exploration")) {
+    if (packageJsonPath.includes(".vscode-insiders") || packageJsonPath.includes(".vscode-server-insiders") || packageJsonPath.includes(".vscode-exploration")) {
```
sean-mcmanus: Add .vscode-server-exploration too.

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #8012, opened 2021-08-19
Merge rewrites of package.json so it's only done once

Colengms:
Also fixes a string that I neglected to save changes to before commit it in the loc PR.


sean-mcmanus (approved):
Can you file a loc bug for the CLANG-FORMAT issue?

PR #8016, opened 2021-08-20
Add support for localization hints in schema files

Colengms:
Also adds a hint.

The additional field does not appear to cause an issue for VS Code.


Extension/c_cpp_properties.schema.json
```diff
@@ -188,6 +188,7 @@
         "env": {
             "type": "object",
             "description": "Custom variables that can be reused anywhere in this file using the ${variable} or ${env:variable} syntax.",
+            "descriptionHint": "The contents within curly braces ('{variable}', and '{env:variable}') should not be translated.",
```
sean-mcmanus: This is incorrect -- only "env" should not be translated...well, and the punctuation marks ${:} should be preserved. See https://github.com/microsoft/vscode-cpptools/pull/8018 ...you may also want to emphasize that "variable" should be translated (e.g. German may not have been translated, i.e. the casing might be wrong, not 100% yet sure though).

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #8077, opened 2021-08-31
add ${execpath} and ${pathSeparator}

### LINKED ISSUE #7753: Can not use ${execPath} in c_cpp_properties.json file!
Bug type: Language Service

**Describe the bug**
Version: 1.55.0
Commit: c185983a683d14c396952dd432459097bc7f757f
Date: 2021-03-30T16:03:00.337Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT ia32 6.1.7601

**Steps to reproduce**

In c_cpp_properties.json file, I can use ${workspaceFolder}, But I can not use ${execPath}! Why?Thanks!

**Expected behavior**

**Code sample and logs**

- Code sample

- Configurations in `c_cpp_properties.json`

- Logs from running `C/C++: Log Diagnostics` from the VS Code command palette

- Logs from [the language server logging](https://code.visualstudio.com/docs/cpp/enable-logging-cpp#_enable-logging-for-the-language-server)

**Screenshots**

**Additional context**

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/7753

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1207,6 +1207,8 @@ export class CppProperties {
                 delete this.configurationJson.env['workspaceRoot'];
                 delete this.configurationJson.env['workspaceFolder'];
                 delete this.configurationJson.env['workspaceFolderBasename'];
+                delete this.configurationJson.env['execPath'];
+                delete this.configurationJson.env['pathSeparator'];
                 delete this.configurationJson.env['default'];
             }
 
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -403,6 +403,8 @@ export class CppProperties {
         }
 
         result["workspaceFolderBasename"] = this.rootUri ? path.basename(this.rootUri.fsPath) : "";
+        result["execPath"] = process.execPath;
+        result["pathSeparator"] = (os.platform() === 'win32') ? "\\" : "/";
         return result;
     }
 
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -335,11 +335,7 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
                             if (name === "execPath") {
                                 newValue = process.execPath;
                             } else if (name === "pathSeparator") {
-                                if (os.platform() === 'win32') {
-                                    newValue = "\\";
-                                } else {
-                                    newValue = "/";
-                                }
+                                newValue = (os.platform() === 'win32') ? "\\" : "/";
                             }
                         }
                     }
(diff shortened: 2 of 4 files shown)


Extension/src/common.ts
```diff
@@ -331,6 +331,17 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
                     }
                     if (newValue === undefined) {
                         newValue = process.env[name];
+                        if (!newValue) {
+                            if (name === "execPath") {
+                                newValue = process.execPath;
+                            } else if (name === "pathSeparator") {
+                                if (os.platform() === 'win32') {
```
michelleangela:
```
 if (os.platform() === 'win32') {
    newValue = "\\";
 } else {
    newValue = "/";
 }
```

Can be shorten to

```
newValue = (os.platform() === 'win32') ? "\\" : "/";
```

Extension/src/common.ts
```diff
@@ -331,6 +331,17 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
                     }
                     if (newValue === undefined) {
                         newValue = process.env[name];
+                        if (!newValue) {
```
bobbrow: These special named things are normally added to the "additionalEnvironment."  Instead of adding special code for them here, I would recommend searching the codebase for callers of this function and add them to the "additionalEnvironment" instead.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -403,6 +403,8 @@ export class CppProperties {
         }
 
         result["workspaceFolderBasename"] = this.rootUri ? path.basename(this.rootUri.fsPath) : "";
+        result["execPath"] = process.execPath;
```
michelleangela, line 406: Will `process.execPath` always have a value, or can it be undefined? Does it need to be `""` if it's undefined?

bobbrow, line 406: Sorry for iterating one comment at a time, but there are a couple more places that need to be updated:
https://github.com/microsoft/vscode-cpptools/blob/5dcb3081e502f84185b71f773f3d106b5d27f63c/Extension/src/LanguageServer/client.ts#L759-L761
https://github.com/microsoft/vscode-cpptools/blob/5dcb3081e502f84185b71f773f3d106b5d27f63c/Extension/src/LanguageServer/configurations.ts#L1205-L1208
https://github.com/microsoft/vscode-cpptools/blob/5dcb3081e502f84185b71f773f3d106b5d27f63c/Extension/c_cpp_properties.schema.json#L193

elahehrashedi: `execPath` is the path to the running VS Code executable, do you think there is a situation that it might be undefined? (if it is undefined it won't be causing an error later on). In my tests it always had a value in it.

elahehrashedi: thank you for finding those.

bobbrow (approved):

PR #8081, opened 2021-08-31
skip relative paths in forceInclude

### LINKED ISSUE #7822: "Force Include" config UI should not report validation errors.  Whether a forced include can be resolved is unknown until compiler is queried.
Bug type: Language Service

**Describe the bug**
- OS and Version: Windows
- C/C++ Extension Version: 1.5.1

**Steps to reproduce**

1. Enter an invalid path in "Force include" property in IntelliSense config UI. There should be error saying cannot find path.
2. Open the c_cpp_properties.json file
3. Observe there is not error in json file for invalid path
image

**Expected behavior**
UI and json file should show the same errors if the current selected active configuration is the one also show on config UI

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/7822


sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1440,7 +1440,9 @@ export class CppProperties {
 
             // Check if resolved path exists
             if (!fs.existsSync(resolvedPath)) {
-                if (!this.rootUri) {
+                if (skipRelativePaths) {
```
bobbrow: This doesn't check that the resolved path is relative though.

bobbrow: It could be an absolute path that failed the 'exists' test

bobbrow: Try this: https://nodejs.org/api/path.html#path_path_isabsolute_path

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1622,13 +1624,22 @@ export class CppProperties {
...
                     for (const curPath of pathArray) {
                         paths.push(`${curPath}`);
                     }
                 }
             }
+            // Skip the relative forcedInclude files.
+            if (currentConfiguration.forcedInclude) {
+                for (const file of currentConfiguration.forcedInclude) {
+                    const resolvedFilePath: string = this.resolvePath(file, isWindows);
+                    if (util.checkFileExistsSync(resolvedFilePath) || util.checkDirectoryExistsSync(resolvedFilePath)) {
```
bobbrow: You don't need to check it exists here.  Use this API instead to decide: https://nodejs.org/api/path.html#path_path_isabsolute_path

bobbrow (changes requested):

bobbrow (approved):

michelleangela (approved):

PR #8095, opened 2021-09-02
detect c/c++ source files when running a build a task

### LINKED ISSUE #7892: cppbuild tasks should be able to detect the current file is not a source file
A common cause of compilation failure in the "build and debug active file" command/task is that the current file is not a C++ source file.  Is it possible to check the active file when the task is run and see if it is a source file or not and warn the developer/abort the compilation?

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/7892

Which error message would you prefer to see?
image


bobbrow:
> Which error message would you prefer to see?
If it works to just put it in the Terminal window, I think that's better than having a

michelleangela (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -345,6 +345,12 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
     }
 
     async open(_initialDimensions: TerminalDimensions | undefined): Promise<void> {
+        const editor: TextEditor | undefined = window.activeTextEditor;
+        if (editor && !util.fileIsCOrCppSource(editor.document.fileName)) {
+            this.writeEmitter.fire(localize("cannot.build.non.cpp", 'Cannot build and debug because the active file is not a C or C++ source file.') + this.endOfLine);
+            this.closeEmitter.fire(-1);
+            return Promise.resolve();
```
bobbrow, line 352: This function is `async` so it should just `return;`  There is no need to call `Promise.resolve()`

PR #8101, opened 2021-09-03
Add support for Debugging on Alpine

WardenGnaw:
+ @gregg-miskelly 

This enables Debug Launch/Attach for Linux-x64 Alpine.

Addresses debugging scenario for #4827 

Added notes via `TODO(LS-alpine)` when Language Service has alpine support so we can have a full activation instead of a debugger only version.

Tested using `remote-container` with the following configurations:
<details>
<summary>Dockerfile</summary>

```
FROM mcr.microsoft.com/vscode/devcontainers/base:dev-alpine-3.14

RUN apk add --no-cache gdb file g++ clang coreutils
```
</details>

<details>
<summary>devcontainer.json</summary>

```
{
	"name": "Alpine",
	"build": {
		"dockerfile": "Dockerfile",
		// Update 'VARIANT' to pick an Alpine version: 3.11, 3.12, 3.13, 3.14
		"args": { "VARIANT": "3.14" }
	},
	
	// Set *default* container specific settings.json values on container create. 
	"settings": {},

	// Add the IDs of extensions you want installed when the container is created.
	// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
	"extensions": [
		"ms-vscode.cpptools"
	],

	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	// "forwardPorts": [],

	// Use 'postCreateCommand' to run commands after the container is created.
	// "postCreateCommand": "uname -a",

	// Uncomment when using a ptrace-based debugger like C++, Go, and Rust
	// "runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],

	// Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
	"remoteUser": "vscode"
}
```

</details>

### DIFF
--- a/Extension/src/Debugger/extension.ts
+++ b/Extension/src/Debugger/extension.ts
@@ -24,6 +24,26 @@ export function buildAndDebugActiveFileStr(): string {
     return ` - ${localize("build.and.debug.active.file", 'Build and debug active file')}`;
 }
 
+/*
+   Separate initialize method for Alpine until we can use the regular 'initialize' method.
+
+   This enables attach and cppdbg for alpine without snippets or 'Build and Debug Active file'
+   since we do not have a language service.
+
+   TODO(LS-alpine): Remove initializeForAlpine
+*/
+export function initializeForAlpine(context: vscode.ExtensionContext): void {
+    const attachItemsProvider: AttachItemsProvider = NativeAttachItemsProviderFactory.Get();
+    const attacher: AttachPicker = new AttachPicker(attachItemsProvider);
+    disposables.push(vscode.commands.registerCommand('extension.pickNativeProcess', () => attacher.ShowAttachEntries()));
+    const remoteAttacher: RemoteAttachPicker = new RemoteAttachPicker();
+    disposables.push(vscode.commands.registerCommand('extension.pickRemoteNativeProcess', (any) => remoteAttacher.ShowAttachEntries(any)));
+
+    disposables.push(vscode.debug.registerDebugAdapterDescriptorFactory(CppdbgDebugAdapterDescriptorFactory.DEBUG_TYPE, new CppdbgDebugAdapterDescriptorFactory(context)));
+
+    vscode.Disposable.from(...disposables);
+}
+
 export function initialize(context: vscode.ExtensionContext): void {
     // Activate Process Picker Commands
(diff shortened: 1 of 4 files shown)


gregg-miskelly (approved):

WardenGnaw:
Also a huge thanks to @Trass3r for the help and validation for this scenario.

Extension/src/platform.ts
```diff
@@ -22,6 +23,16 @@ export function GetOSName(processPlatform: string | undefined): string | undefin
     }
 }
 
+let _isAlpine: boolean | undefined;
+
+export function IsAlpine(): boolean {
```
bobbrow, line 28: functions should be named like `isAlpine`

Extension/src/Debugger/extension.ts
```diff
@@ -24,6 +24,26 @@ export function buildAndDebugActiveFileStr(): string {
     return ` - ${localize("build.and.debug.active.file", 'Build and debug active file')}`;
 }
 
+/*
+   Separate initialize method for Alpine until we can use the regular 'initialize' method.
+
+   This enables attach and cppdbg for alpine without snippets or 'Build and Debug Active file'
+   since we do not have a language service.
```
bobbrow, line 31: Do we want one?  What would it take for us to make it work?  It would be better if we could just support the platform across the board.

WardenGnaw: Lets discuss this in our next meeting.

Trass3r: Well the current one is glibc-based.

Trass3r:
Any updates?

WardenGnaw:
@Trass3r We are planning on fully adding Alpine support for both the Language Service and Debugger.

The current plan is targeted for [1.8.0](https://github.com/microsoft/vscode-cpptools/milestone/75).

Trass3r:
I see. But that will take a while.
In the meantime couldn't it "simply" be changed to reference the correct debugger package so [these steps](https://github.com/microsoft/vscode-cpptools/issues/4827#issuecomment-910033294) become unnecessary?

Colengms:
@WardenGnaw Can this PR be abandoned, now that we support Alpine x64?

WardenGnaw:
Closing in favor for 1.8.0-insiders3

Issue #8100, opened 2021-09-03
VSCode C++ outline view in sources (*.cpp) does not group methods by class names (in hierarchy)


raymond-wh-leung:
Bug type: Language Service

**Describe the bug**
I came across the same issue reported from this thread for VSCode C++ outline view.
(https://github.com/microsoft/vscode/issues/107161).

In short, the C++ outline view does not show class methods in hierarchy if the editor is in the source (*.cpp), it will show something like below.   

   o MyClassName::Method1()
   o MyClassName::Method2()
   o ...
   o MyClassName::MethodN()

(every method is appended with the class name MyClassName::) 

It happens very often that our sources have very long class name, this makes the outline view almost unusable because what you’ll see there is a list of “LooooongClassName::Method”, and those method name will get cut off (unless you make the outline view very wide). (there is no horizontal scrollbar there.)

And I wonder if we can make the outline view group by class name even in source *.cpp?
- Trial
   o  MyClass
       - Method1()
       - Method2()
       - …
 
This is one major thing that I missed a lot when switching over to VSCode from Source Insight 3.5 (Source Insight 3.5 grouped the methods by class name and that makes the outline very handy.)
 
VSCode (with C++ extension) is great and I like it a lot. If we can address the above, it will make the experience much better!!

- OS and Version:
Windows 10

- VS Code Version:
Version: 1.59.1

- C/C++ Extension Version:
Latest (1.6.0)

- Other extensions you installed (and if the issue persists after disabling them):
No other extensions installed that would affect that C++ outline view.

- If using SSH remote, specify OS of remote machine:

- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).
C++ outline view when editor is in source (*.cpp) shows all methods in the from of "MyClassName::Method". Very often there are long class name, and it will show "LooooongClassName::Method" in the outline view, and the method name will get cut off unless you make the outline view very wide. (that makes the outline view not usable in that case, there is no horizontal scrollbar there.) The experience will be much better if the methods shown in the outline view is grouped by class name (in hierarchy). (Some other editors like source insight 3.5 does that and it makes the outline view very handy.)

**Steps to reproduce**
1. Create a new C++ class (with a header *.h and source *.cpp) with a long class name, like "MyLoooooooooongClassName".
2. Add some methods there, like Method1(), Method2(), Method3().
3. Open the source with *.cpp
[...]

PR #8111, opened 2021-09-07
Move the scope of document symbols to the details.

### LINKED ISSUE #7785: Outline of class's methods in a .cpp fie is difficult to read
When working in a .cpp file with lots of methods of the same class, the class name is shown for each method making it difficult to find a method by name. For example:

image

Now compare that to viewing the same class's header file:

image

It would be great to have the same kind of the view in the .cpp file that we have in the header file.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/7785

image

### DIFF
--- a/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
+++ b/Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
@@ -30,6 +30,20 @@ export class DocumentSymbolProvider implements vscode.DocumentSymbolProvider {
                         detail = util.getLocalizedSymbolScope("protected", detail);
                     }
                 }
+
+                // Move the scope in the name to the detail.
+                if (detail.length === 0) {
+                    let offset_paren: number = symbol.name.indexOf("(");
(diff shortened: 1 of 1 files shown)


michelleangela (approved):

Colengms (approved):

bobbrow:
This doesn't seem right.  Won't this be confusing for people by hiding the scoping information?  It looks like `foo` is defined in `f1` instead of showing it under
```
f1
+--f2
   +--foo
       +--func
```

sean-mcmanus:
The outline view is just designed to have nodes for the {} ranges, i.e. for navigation -- the hiding of the full scope on the right is not a priority for the outline view. They would see that either via clicking the item or viewing the right-side if it's visible.

i.e. the reason it should be moved to the right is that it's lower priority info.

I might re-open https://github.com/microsoft/vscode-cpptools/issues/8100 for the more difficult solution of creating a node for the "ghost" scopes.

Issue #3727, opened 2019-06-04
Add way to restart language server for a specific file


ThadHouse:
https://github.com/microsoft/vscode-cpptools/issues/3603 and https://github.com/microsoft/vscode-cpptools/issues/2966 are still open, and I still randomly see at least 2966, which seems like a race condition somewhere. Closing the file, waiting for the language server for that file to close, and reopening always solves both issues, so it would be nice if there was a way to manually reset the language server on a file. That way the delay wouldn't be needed, and could just be immediately restarted.

sean-mcmanus:
By language server I assume you mean the Microsoft.VSCode.CPP.IntelliSense.Msvc.exe process -- you could trying do some a "kill Microsoft.VSCode.CPP.IntelliSense.Msvc.exe" command.

It sounds like fixing the root bugs might be a better solution. I think they're caused by our didOpen event getting delayed for the custom config provider, but then we receive other messages that aren't delayed.

ThadHouse:
I don't want to kill the whole process, I notice in the debug logs that after about 20 seconds it receives a message saying `shutting down language server for file *`. So I don't know if its just 1 process with multiple files, or 1 process per file, but a way to restart just the server parsing for that file.

Another option would be a button to reforce sending the API data to get intellisense.

I do agree fixing the root bugs would be the best fix, but I know that doing that might take a while.

sean-mcmanus:
We never output "shutting down language server for file" -- do you mean `Shutting down IntelliSense server: <file>`? That shutdown terminates the IntelliSense process for the translation unit and not the main extension process. It's one process per translation unit.

bobbrow:
This would be a relatively easy thing to add.

sean-mcmanus:
The fix is available with https://github.com/microsoft/vscode-cpptools/releases/tag/1.7.0-insiders2 (C/C++: Restart IntelliSense for Active File command).

PR #8128, opened 2021-09-09
Restart Language Server for active file

elahehrashedi:
feature-request: https://github.com/microsoft/vscode-cpptools/issues/3727


Extension/src/LanguageServer/client.ts
```diff
@@ -2437,6 +2439,15 @@ export class DefaultClient implements Client {
         });
     }
 
+    /**
+     * notifications to the language server
```
michelleangela: "notifications to the language server" is generic as there are other functions that send notifications to the language server. Suggest to change to something specific that describes what `restartLanguageServerActiveFile` is uniquely doing.

Extension/src/LanguageServer/extension.ts
```diff
@@ -796,9 +796,24 @@ export function registerCommands(): void {
...
     disposables.push(vscode.commands.registerCommand('cpptools.activeConfigCustomVariable', onGetActiveConfigCustomVariable));
     disposables.push(vscode.commands.registerCommand('cpptools.setActiveConfigName', onSetActiveConfigName));
+    disposables.push(vscode.commands.registerCommand('C_Cpp.RestartLanguageServerActiveFile', onRestartLanguageServerActiveFile));
+
     getTemporaryCommandRegistrarInstance().executeDelayedCommands();
 }
 
+async function onRestartLanguageServerActiveFile(): Promise<void> {
+    const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+    if (!activeEditor || !activeEditor.document) {
```
michelleangela: should `activeEditor.document.uri.scheme !== "file"` also be added to this conditional?

see function `onDidChangeActiveTextEditor` for example.

Extension/package.nls.json
```diff
@@ -12,6 +12,7 @@
     "c_cpp.command.resetDatabase.title": "Reset IntelliSense Database",
     "c_cpp.command.takeSurvey.title": "Take Survey",
     "c_cpp.command.buildAndDebugActiveFile.title": "Build and Debug Active File",
+    "c_cpp.command.restartLanguageServerActiveFile.title": "Restart Language Server on Active File",
```
sean-mcmanus: "Language Server" references should be renamed "IntelliSense" or possibly "IntelliSense Server" or "IntelliSense Process".

sean-mcmanus (changes requested):

Extension/package.nls.json
bobbrow: In Visual Studio, the menu item is simply "Rescan File".  I think the use of "Active" in this menu item (and the one above it) could be dropped.

sean-mcmanus: The potential problem with removing the "active" part is that in other context such as the Command Palette it may not as obvious that it's referring to the active file. Also there are a bunch of VS Code "File" commands that refer to the active file.

Also, the term "scan" or "rescan" is not used anywhere else with our extension and it's not really explicitly saying what exactly is being done. Also, "scanning" in the context of text processing usually refers to a quick O(N) operation and not a long parsing operation.

Colengms: IMHO, we should remove the "Active" part.  Perhaps word it as "ForFile"/"for file", to indicate it doesn't set it for all files.  It would be relatively simple to add an optional command argument indicating which file to reset.  That would be extensible, allowing other extension developers to reset IntelliSense for a specific file (though, I'm not sure the user case).

sean-mcmanus: Your comment seems to be a reason to not remove the Active part, makes it more confusing in the event that we decide to add a new command to take a file argument, i.e. "Rescan Active File" versus "Rescan File..." (user navigates to the file or it's passed via some arg).

sean-mcmanus: image

without the "active", it would be reasonable in this context for the user to mistakenly believe they're allowed to choose the file and not make one active first. In particular, we wanted to avoid users running those commands with the wrong file active, so having "active" in the command hopefully makes that more obvious, and it leaves open the possibility of adding a non-active "file..." command as mentioned before (without the confusing of "Rescan file" and "Rescan file..." doing different things).

Colengms: I'm OK with leaving "Active" in the text description that is displayed in the UI, but I think the command itself could so easily be extended, that it should not contain "Active" in the implementation of the code.

I have similar feedback in the native PR, where "Active" is referred to despite the URI being passed in from the TypeScript side.  In that case, since the file is provided in the message, the code shouldn't refer to the "Active" file.

sean-mcmanus: Yeah, that sounds fine.

bobbrow: So we can't have the command say "active file" without also having the item in the context menu say "active file"?  I'm fine with leaving "active" in the command where the context is less obvious.  It was the context menu item I was hoping to remove it from because it seems unnecessary.

sean-mcmanus: @bobbrow The command has a string "title" that is shared by all contexts, Command Palette and context menu.

bobbrow: That's too bad

elahehrashedi: image
This is what the user will see, but the functions itself will send the document info to the language server, so inside the code the name "active" is not used, to keep the function more general.

sean-mcmanus: Ok

Extension/src/LanguageServer/client.ts
```diff
@@ -2437,6 +2439,15 @@ export class DefaultClient implements Client {
         });
     }
 
+    /**
+     * send notifications to the language server to restart IntelliSense for the selected file.
+     */
+    public async restartIntelliSenseForFile(document: vscode.TextDocument): Promise<void> {
+        this.notifyWhenLanguageClientReady(() => {
```
sean-mcmanus: This can be simplified to 
await this.awaitUntilLanguageClientReady();

A few other spots could also be simplified.

Extension/src/LanguageServer/extension.ts
```diff
@@ -796,9 +796,20 @@ export function registerCommands(): void {
     disposables.push(vscode.commands.registerCommand('cpptools.activeConfigName', onGetActiveConfigName));
     disposables.push(vscode.commands.registerCommand('cpptools.activeConfigCustomVariable', onGetActiveConfigCustomVariable));
     disposables.push(vscode.commands.registerCommand('cpptools.setActiveConfigName', onSetActiveConfigName));
+    disposables.push(vscode.commands.registerCommand('C_Cpp.RestartIntelliSenseForFile', onRestartIntelliSenseForFile));
+
     getTemporaryCommandRegistrarInstance().executeDelayedCommands();
 }
 
+async function onRestartIntelliSenseForFile(): Promise<void> {
```
sean-mcmanus: Does this need to be async?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
elahehrashedi: Sean, I think this is fine, I know we talked about it. But I can see in the code that wherever there is a notification, we have called `notifyWhenLanguageClientReady`, and we called `awaitUntilLanguageClientReady` otherwise.

```
    public notifyWhenLanguageClientReady<T>(notify: () => T): Promise<T> {
        const task: () => Promise<T> = () => new Promise<T>(resolve => {
            resolve(notify());
        });
        return this.queueTask(task);
    }

    public awaitUntilLanguageClientReady(): Thenable<void> {
        const task: () => Thenable<void> = () => new Promise<void>(resolve => {
            resolve();
        });
        return this.queueTask(task);
    }
```

sean-mcmanus: Calling the notify* isn't necessary in async functions. It makes the code simpler to call `await this.await*`. If other async methods are also calling the notify* they could also be changed (I already changed some of those in my clangTidy branch).

Extension/package.json
```diff
@@ -2543,6 +2548,11 @@
           "when": "editorLangId == 'c' || editorLangId == 'cpp' || editorLangId == 'cuda-cpp'",
           "command": "C_Cpp.BuildAndDebugActiveFile",
           "group": "other2_debug@1"
+        },
+        {
+          "when": "editorLangId == 'c' || editorLangId == 'cpp' || editorLangId == 'cuda-cpp'",
+          "command": "C_Cpp.RestartIntelliSenseForFile",
+          "group": "other2_debug@2"
```
Colengms: Not sure this should go into a 'debug' related group, as it doesn't involve debugging.  But, it's also not navigation related.  I'm not sure how these groups work, but you might want to create a new one.

Colengms (changes requested):

Extension/package.json
bobbrow: These are extension-specific commands that don't align with the standard LSP tasks. I think it makes sense to keep them together.  If the group name is the problem, it can be changed.

sean-mcmanus: Run Code Analysis on Active File is also in this "debug" section -- it might be renamable to "other".

Extension/src/LanguageServer/client.ts
sean-mcmanus: Well, notify could be okay too if you wanted to make it all on one line and remove the 2 extra lines, i.e. `() => this.languageClient.sendNotification(RestartIntelliSenseForFileNotification, this.languageClient.code2ProtocolConverter.asTextDocumentIdentifier(document)); `.

Colengms (approved):

michelleangela (approved):

sean-mcmanus (approved):

Issue #7358, opened 2021-04-14
Add descriptions for enums to package.json


Colengms:
A lot of enum settings in `package.json` lack `enumDescriptions`.  Opening this issue to track adding a description for each enum value.

sean-mcmanus:
Some cases of this are fixed in my clang-tidy branch (along with other description fixes). The other cases didn't seem like an enum description would be much help. Were there particular cases you thought should have an enum description?

sean-mcmanus:
I'm considering this fixed...let me know if you think other enums should have descriptions. One thing to be aware of is that the enumDescriptions aren't currently copyable via the settings UI, which can make some descriptions that contain "sample text" better placed in the "description" instead.

PR #8129, opened 2021-09-09
String/settings fixes

### LINKED ISSUE #4544: Use the markdownDescription property for settings
settings descriptions in `package.json` that reference enum values should use the `markdownDescription` property instead of `description`.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/4544, https://github.com/microsoft/vscode-cpptools/issues/6870, https://github.com/microsoft/vscode-cpptools/issues/7358 .

Also, fixed a bunch of CppSettings that were using the wrong context in regards to workspace settings versus workspace folder settings.

I moved these out of my clangTidy branch.


Extension/gulpfile.js
```diff
@@ -192,7 +192,7 @@ const processHtmlFiles = () => {
 const traverseJson = (jsonTree, descriptionCallback, prefixPath) => {
     for (let fieldName in jsonTree) {
         if (jsonTree[fieldName] !== null) {
-            if (typeof (jsonTree[fieldName]) == "string" && fieldName === "description") {
+            if (typeof (jsonTree[fieldName]) == "string" && (fieldName === "description" || fieldName === "markdownDescription")) {
```
Colengms, line 195: It might be a good idea to automatically add a hint whenever a `markdownDescription` is detected, indicating that the string could contain markdown and the translator should avoid breaking it.  Since the hints are stored in an array, you can add to it instead of replacing one if existing.  Or, alternatively, you could add a `descriptionHint` field to each `markdownDescription` in the json schema file.

Extension/package.nls.json
```diff
@@ -22,49 +22,49 @@
...
     "c_cpp.configuration.formatting.description": "Configures the formatting engine",
-    "c_cpp.configuration.formatting.clangFormat.description": "clang-format will be used to format code.",
-    "c_cpp.configuration.formatting.vcFormat.description": "The Visual C++ formatting engine will be used to format code.",
-    "c_cpp.configuration.formatting.Default.description": "By default, clang-format will be used to format the code. However, the Visual C++ formatting engine will be used if an .editorconfig file with relevant settings is found nearer to the code being formatted and clang_format_style is the default value: 'file'.",
-    "c_cpp.configuration.formatting.Disabled.description": "Code formatting will be disabled.",
-    "c_cpp.configuration.vcFormat.indent.braces.description": "Braces are indented by the amount specified in the Editor: Tab Size setting.",
-    "c_cpp.configuration.vcFormat.indent.multiLineRelativeTo.description": "Determines what new line indentation is relative to",
+    "c_cpp.configuration.formatting.clangFormat.markdownDescription": "`clang-format` will be used to format code.",
```
Colengms: Do you want to add hints for these, indicated that they may contain markdown?  I think the issue blocking proper processing of the new syntax has been addressed, though it may requiring updating the minimum required version of VS Code.

Colengms (changes requested):

Extension/gulpfile.js
sean-mcmanus: These changes rely on the translators not touching text inside `` unless otherwise directed to by a hint. I think we just need to make sure that is communicated to the localization team if they don't already know.

Extension/package.nls.json
sean-mcmanus: I'd prefer to get the translators to understand the 1 rule "don't translate text inside `` unless otherwise hinted at" instead of doing things the other way around.

Extension/gulpfile.js
Colengms: I suspect there are quite a few people involved in localization - multiple people per language.  It would be a good idea to add hints whenever possible.  They are likely not highly technical people and keeping them trained to recognize requirements like these would seem to be an ongoing issue.

Extension/package.nls.json
Colengms: I just don't think we can count on that.  I had a thread a while back with Cristiano about potential issues with markdown contents in strings.  The conclusion seemed to be that translators are not sufficiently technical with regards to recognizing markdown and knowing how to avoid breaking it.  Even with hints, it may be an issue.  :/

michelleangela (approved):

Extension/package.nls.json
sean-mcmanus: But I think that requires VS 1.60.0 for package.nls.json strings, which was just recently released.

Colengms: Updating the minimum required version to 1.60.0 for our next insiders seems OK to me.  It looks like it was released last week.  I'm just trying to minimize the future work required when we inevitably get corrupted markdown from translators.  :/

sean-mcmanus: Okay, sounds the like the loc team wants a lot of hints.

Extension/gulpfile.js
sean-mcmanus: I just added the hints explicitly instead of via code, so that's it's more visible.

Colengms (approved):

Extension/c_cpp_properties.schema.json
```diff
@@ -78,19 +83,22 @@
...
+                        "markdownDescription": "The IntelliSense mode to use that maps to a platform and architecture variant of MSVC, gcc, or Clang. If not set or if set to `${default}`, the extension will choose the default for that platform. Windows defaults to `windows-msvc-x64`, Linux defaults to `linux-gcc-x64`, and macOS defaults to `macos-clang-x64`. IntelliSense modes that only specify `<compiler>-<architecture>` variants (e.g. `gcc-x64`) are legacy modes and are converted automatically to the `<platform>-<compiler>-<architecture>` variants based on the host platform.",
+                        "descriptionHint": "Uses of words 'platform', 'compiler', and 'architecture' are meant to be descriptive placeholders for literal values, so they should be translated, but without changing the surrounding `<>-<>` symbols. All other text inside `` should not translated.",
```
jogo-, line 101: Missing **be**: should not translated. -> should not **be** translated.

sean-mcmanus: Thanks: https://github.com/microsoft/vscode-cpptools/pull/8145

PR #8136, opened 2021-09-10
Fix quick info integration test

michelleangela:
Running test for Windows on Azure pipeline does not return full comment for quick info results. So for Windows only, validate result for symbol and no comment. This was already being done for the Doxygen comment test too.


sean-mcmanus (approved):

sean-mcmanus:
This is just a workaround for the test, right? We don't know why the result is different on the Azure machine? And I assume we don't have any real world repros so we assume it's just some test issue?

michelleangela:
> This is just a workaround for the test, right? We don't know why the result is different on the Azure machine? And I assume we don't have any real world repros so we assume it's just some test issue?

Yes, this is just a test issue. For some reason it worked before for normal comments, but it also didn't work for Doxygen comment either. So the workaround that was done for Doxygen comments is now being applied for normal comments.

PR #8155, opened 2021-09-15

csigs:


sean-mcmanus (changes requested):
Don't check this unless the errors are reverted.

bobbrow (approved):

Extension/ui/settings.html
```diff
@@ -671,7 +671,7 @@
             <div class="section-title" data-loc-id="merge.configurations">Merge configurations</div>
             <div>
                 <input type="checkbox" id="mergeConfigurations" style="vertical-align: middle; transform: scale(1.5)">
-                    <span data-loc-id="merge.configurations.description">When true (or checked), merge include paths, defines, and forced includes with those from a configuration provider.</span>
+                    <span data-loc-id="merge.configurations.description">When `true` (or checked), merge include paths, defines, and forced includes with those from a configuration provider.</span>
```
bobbrow, line 674: Does markdown like this render correctly in the html file?

bobbrow: Looks like no. I'll fix it.
image

PR #8159, opened 2021-09-15
Add error string

Colengms:
Adds an error string needed by a PR I have open on the native side.

### DIFF
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -232,5 +232,6 @@
     "inline_macro": {
         "text": "Inline macro",
         "hint": "'Inline' is a command and not an adjective, i.e. like 'Expand macro'."
-    }
+    },
+    "unable_to_access_browse_database": "Unable to access browse database. ({0})"
 }


Extension/src/nativeStrings.json
```diff
@@ -232,5 +232,6 @@
     "inline_macro": {
         "text": "Inline macro",
         "hint": "'Inline' is a command and not an adjective, i.e. like 'Expand macro'."
-    }
+    },
+    "unable_to_access_browse_database": "Unable to access browse database. ({0})"
```
sean-mcmanus, line 236: maybe use "...database in {0}." instead?

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
Colengms: I'd kinda prefer to keep it this way, so we don't have to worry about how translators would treat it within a sentence.

sean-mcmanus: I was thinking it would be better to make it a real sentence or the translators may be get confused by the ". ()".

sean-mcmanus (approved):

Extension/src/nativeStrings.json
michelleangela: The logging is mostly for us to track where it failed to access the database, so putting `{0}` in ( ) is sufficient.

michelleangela (approved):

PR #8174, opened 2021-09-19
Allow merge of C_Cpp_Properties & configuration provider.

### LINKED ISSUE #6909: Need a way to provide system defines and system includes when compiler querying is disabled (with compilerPath of "").
It's possible to disable compiler querying (and disable fallback to a different compiler on failure) by specifying a `compilerPath` of `""`.  This also works in combination with a custom configuration provider or `compile_commands.json`, to prevent use of the compiler they indicate (in case it's not cl.exe, gcc, or a gcc variant we can query).  However, the defines and includes from `c_cpp_properties.json` are not currently getting merged into the configuration in those cases.  Since no compiler will be queried, it seems necessary to support a way for the user to provide the missing system includes and systems defines.

This issue tracks merging in the includes and defines from `c_cpp_properties.json` when a `compilerPath` of `""` is used in combination with a custom configuration provider or `compile_commands.json`.

willson556:
Partially resolves #6909 (doesn't yet work with `compile_commands.json`,
just custom configuration providers).

Adds a configuration that enables the include paths, defines, and forced
includes from `c_cpp_properties.json` to be merged with those provided
by a configuration provider.

This is especially helpful for projects that use unsupported compilers
as defines/includePaths/forcedIncludes can be provided as shims for
intellisense.


gerhardol:
A step in the right direction I guess, but not helping in my use case as the cpp/cmake provider has no information (CMake could generate `compile_commands.json`, that would be useful if it could be merged),
Code looks reasonable.

michelleangela:
The new property `mergeConfigurations` in c_cpp_properties.json will need to be added to the IntelliSense configuration UI (use command `C/C++: Edit configurations (UI)` to view UI) as well. under the Advanced section, right after compile commands property.

See https://github.com/microsoft/vscode-cpptools/blob/main/Extension/ui/settings.html and the UI element for "see `Browse: limit symbols to included headers` for example of creating a boolean checkbox. Also add to corresponding TS file https://github.com/microsoft/vscode-cpptools/blob/main/Extension/ui/settings.ts.

Also add this new property to https://github.com/microsoft/vscode-cpptools/blob/main/Extension/src/LanguageServer/settingsPanel.ts so that the UI will be updated with values from c_cpp_properties.json.

**How to localize string sin HTML.**
To localize HTML content, the following 2 attributes are used on HTML nodes:

`data-loc-id`
Specifying a unique string key value using this attribute causes the content of the node to be localized associated with that key
	
`data-loc-id-<attribute-name>`
Specifying a unique string key value using this attribute causes the specified attribute of this node to be localized.  For example, to localize the title attribute of a node, use: data-loc-id-title="some.title.id"

The following is an example of a span with localized content, containing a sub-node with a localized title attribute.

```
<span data-loc-id="switch.to.json">Switch to the <a href="command:C_Cpp.ConfigurationEditJSON" title="Edit configurations in JSON file" data-loc-id-title="edit.configurations.in.json">c_cpp_properties.json</a> file by clicking on the file link or using the command:</span>
```

Child nodes are not included when content of a parent node is localized.  In the outer nodes of the above example, the following string will be extracted for localization:

"Switch to the {0} file by clicking on the file link or using the command:"

Extension/c_cpp_properties.schema.json
```diff
@@ -155,6 +155,10 @@
                         "description": "The id of a VS Code extension that can provide IntelliSense configuration information for source files.",
                         "type": "string"
                     },
+                    "mergeConfigurations": {
+                        "description": "Set to `true` to merge include paths, defines, and forced includes with those from a configuration provider.",
```
sean-mcmanus: This should be "markdownDescription" instead of "description". Also, copy the "descriptionHint".

sean-mcmanus (changes requested):

sean-mcmanus:
Can a "C_Cpp.default.mergeConfigurations" setting be added? See https://github.com/microsoft/vscode-cpptools/pull/7845/files for an example.

Extension/ui/settings.html
```diff
@@ -667,6 +667,15 @@
             </div>
         </div>
 
+        <div class="section">
+            <div class="section-title" data-loc-id="merge.configurations">Merge configurations</div>
+            <div>
+                <input type="checkbox" id="mergeConfigurations" style="vertical-align: middle; transform: scale(1.5)">
+                    <span data-loc-id="merge.configurations.description">Set to `true` to merge include paths, defines, and forced includes with those from a configuration provider.</span>
```
michelleangela: Since this is a checkbox and not an input value, we'll have to change the wording that describes checking the box:

"When true (or checked), merge include paths, defines, and forced includes with those from a configuration provider."

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1797,9 +1797,36 @@ export class DefaultClient implements Client {
...
                             const tuUri: vscode.Uri = vscode.Uri.parse(candidate);
                             if (await provider.canProvideConfiguration(tuUri, tokenSource.token)) {
-                                const configs: SourceFileConfigurationItem[] = await provider.provideConfigurations([tuUri], tokenSource.token);
+                                const configs: util.Mutable<SourceFileConfigurationItem>[] = await provider.provideConfigurations([tuUri], tokenSource.token);
                                 if (configs && configs.length > 0 && configs[0]) {
-                                    return configs;
+                                    const fileConfiguration: configs.Configuration | undefined = this.configuration.CurrentConfiguration;
+                                    if (fileConfiguration?.mergeConfigurations ?? false) {
+                                        configs.forEach(config => {
+                                            fileConfiguration?.includePath?.forEach(p => {
```
willson556: Does anybody have any idea why these lint errors exist? It _seems_ like a false positive to me but similar code in other places in this file isn't causing this error.

willson556: > Can a "C_Cpp.default.mergeConfigurations" setting be added? See https://github.com/microsoft/vscode-cpptools/pull/7845/files for an example; however, that PR has been updated to use "markdownDescription" instead of "description".

Added

willson556:
> Can a "C_Cpp.default.mergeConfigurations" setting be added? See https://github.com/microsoft/vscode-cpptools/pull/7845/files for an example; however, that PR has been updated to use "markdownDescription" instead of "description".

Added

Extension/ui/settings.html
willson556: Good call

Extension/c_cpp_properties.schema.json
willson556: Yep

Extension/package.json
```diff
@@ -45,7 +45,36 @@
     "virtualWorkspaces": false
   },
   "activationEvents": [
-    "*"
+    "onLanguage:c",
```
michelleangela: The value in `activationEvents` should be reverted back to `"*"`

Extension/src/LanguageServer/client.ts
michelleangela: @sean-mcmanus do you know about these lint errors?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -741,6 +742,18 @@ export class CppProperties {
         return util.resolveVariables(property, env);
     }
 
+    private updateConfigurationBoolean(property: boolean | undefined | null, defaultValue: boolean | undefined | null): boolean | undefined {
+        if (property === null || property === undefined) {
+            property = defaultValue;
+        }
+
+        if (property === null) {
+            return property = undefined;
```
michelleangela: Is this supposed to return `undefined` instead of `property = undefined`?

michelleangela (changes requested):

Extension/package.json
willson556: Ugh, missed that in the last iteration. Fixed now.

Extension/src/LanguageServer/configurations.ts
willson556: Yep, silly mistake, thanks.

michelleangela (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I'm not sure why the linter is giving these errors, but they appear to be fixable via changing the forEach calls with for loops, e.g. `for (const p in fileConfiguration.includePath) {`

Extension/src/LanguageServer/client.ts
```diff
@@ -1799,9 +1799,36 @@ export class DefaultClient implements Client {
                             const candidate: string = response.candidates[i];
                             const tuUri: vscode.Uri = vscode.Uri.parse(candidate);
                             if (await provider.canProvideConfiguration(tuUri, tokenSource.token)) {
-                                const configs: SourceFileConfigurationItem[] = await provider.provideConfigurations([tuUri], tokenSource.token);
+                                const configs: util.Mutable<SourceFileConfigurationItem>[] = await provider.provideConfigurations([tuUri], tokenSource.token);
                                 if (configs && configs.length > 0 && configs[0]) {
-                                    return configs;
+                                    const fileConfiguration: configs.Configuration | undefined = this.configuration.CurrentConfiguration;
+                                    if (fileConfiguration?.mergeConfigurations ?? false) {
```
sean-mcmanus: The `?? false` seems unnecessary. Is there a reason to use that?

Extension/src/LanguageServer/client.ts
```diff
@@ -1799,9 +1799,36 @@ export class DefaultClient implements Client {
...
+                                        configs.forEach(config => {
+                                            fileConfiguration?.includePath?.forEach(p => {
```
sean-mcmanus: `fileConfiguration?.` can just be `fileConfiguration.` since the surrounding `if` already handles the case where fileConfiguraiton is null/undefined.

Extension/src/LanguageServer/client.ts
willson556: just habit from C#, fixed :)

Extension/src/LanguageServer/client.ts
willson556: yep, you're right

Extension/src/LanguageServer/client.ts
willson556: Before microsoft/vscode-cpptools@621ca46, I was still getting an unused expression error from the linter on line 1825 on the old side of that commit.

Seems the ?. expressions might confuse the linter?

willson556: Yep, the issue seems to be https://github.com/eslint/eslint/issues/12642.

Extension/src/LanguageServer/client.ts
```diff
@@ -1799,9 +1799,38 @@ export class DefaultClient implements Client {
...
+                                    const fileConfiguration: configs.Configuration | undefined = this.configuration.CurrentConfiguration;
+                                    if (fileConfiguration?.mergeConfigurations) {
+                                        configs.forEach(config => {
+                                            for (const p in fileConfiguration.includePath) {
+                                                if (!config.configuration.includePath.includes(p)) {
+                                                    config.configuration.includePath.push(p);
+                                                }
+                                            }
+
+                                            for (const d in fileConfiguration.defines) {
```
michelleangela: @willson556 
Thanks for the addressing the change requests.

I tested the latest changes and found the `d` variable in `for (const d in fileConfiguration.defines)` was returning the index number of array `fileConfiguration.defines` instead of the values in the array. It was also the same case for `for (const p in fileConfiguration.includePath)` and probably for `for (const i in fileConfiguration.forcedInclude)`.  Could you check if this is also happening when you test and debug it? Unless I'm wrong, the Typescript syntax seems to be correct for accessing the values in an array.

bobbrow: I got bit by this in CMake Tools once.  We added the linter rule `"@typescript-eslint/no-for-in-array": "error",`
cpptools should add this rule too.

sean-mcmanus (changes requested):
See Michelle's comment.

Extension/src/LanguageServer/client.ts
willson556: Sorry, should have retested -- that looked like a straightforward change to satisfy the linter. I've put it back to a `.foreach()` call but with explicit `if (array) {}` checks to avoid confusing the linter and retested.

michelleangela: I retested the changes. It works now. Thank you!

sean-mcmanus (approved):

michelleangela:
@willson556 these changes are now merged and will be available in **1.7.0-insiders3**. Thank you for your work and contribution to the C++ extension.

willson556:
Thank you @michelleangela and @sean-mcmanus for taking the time to review and merge these changes!

This is going to significantly improve our experience with IAR and VSCode!

PR #8243, opened 2021-10-08
Update telemetry.ts

jureid:
it looks like in [package.json line 5](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/package.json#L5), we add a suffix, "main", to the version. When we assign customers to target populations for experiments, we were trying to assign customers without any suffix to Public, which resulted in no one being included. Instead, I think we should be checking if the suffix is "main" and assign those customers to Public.

### DIFF
--- a/Extension/src/telemetry.ts
+++ b/Extension/src/telemetry.ts
@@ -65,7 +65,7 @@ export function activate(): void {
             if (packageInfo) {
                 let targetPopulation: TargetPopulation;
                 const userVersion: PackageVersion = new PackageVersion(packageInfo.version);
-                if (userVersion.suffix === "") {
+                if (userVersion.suffix === "main") {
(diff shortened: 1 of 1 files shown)


sean-mcmanus (changes requested):
What? In general, the only users who should have the main suffix are internal developers. When we build the public vsix, the main is removed.

bobbrow:
I think the problem is that suffix is `undefined`, not `""`.  Looking at the constructor for PackageVersion, I think that's what's causing problems for you.  Try manually removing the "-main" from package.json for your local testing and you should be able to confirm if that's the case.

bobbrow (approved):

sean-mcmanus (approved):

Colengms (approved):

michelleangela (approved):

PR #8316, opened 2021-10-22
Add svdPath field to OptionsSchema

gcampbell-msft:
There is a need to allow for the svdPath field to be set through the launch.json for cppdbg scenarios. This PR will allow for this and allow for intellisense to be active for the svdPath.


Extension/package.nls.json
```diff
@@ -212,6 +212,7 @@
     "c_cpp.debuggers.customLaunchSetupCommands.description": "If provided, this replaces the default commands used to launch a target with some other commands. For example, this can be \"-target-attach\" in order to attach to a target process. An empty command list replaces the launch commands with nothing, which can be useful if the debugger is being provided launch options as command line options. Example: \"customLaunchSetupCommands\": [ { \"text\": \"target-run\", \"description\": \"run target\", \"ignoreFailures\": false }].",
     "c_cpp.debuggers.launchCompleteCommand.description": "The command to execute after the debugger is fully setup in order to cause the target process to run. Allowed values are \"exec-run\", \"exec-continue\", \"None\". The default value is \"exec-run\".",
     "c_cpp.debuggers.cppdbg.visualizerFile.description": ".natvis file to be used when debugging this process. This option is not compatible with GDB pretty printing. Please also see \"showDisplayString\" if using this setting.",
+    "c_cpp.debuggers.cppdbg.svdPath.description": "Specifies the path to the users's SVD file.",
```
bobbrow: typo `user's`

But perhaps we can describe this with a slightly different detail instead.  Would it make sense to say `The full path to an embedded device's SVD file'?  (you can remove the word "specifies" because that is implied already)  I would like for it to reference IoT or embedded devices in some way because many users may have never heard of this file and searching online doesn't bring up relevant results without additional keywords.

Extension/package.json
```diff
@@ -3098,6 +3103,7 @@
     "vscode-nls": "^4.1.1",
     "vscode-tas-client": "^0.1.22",
     "which": "^2.0.2",
+    "yarn": "^1.22.15",
```
bobbrow: Is it necessary to add `yarn` here?  I don't see any dependency changes in this PR.

Extension/tools/OptionsSchema.json
```diff
@@ -196,9 +196,7 @@
     "CppdbgLaunchOptions": {
       "type": "object",
       "default": {},
-      "required": [
-        "program"
-      ],
+      "required": ["program"],
```
bobbrow: Did you format this file and it put the array all on one line?  That seems unusual.  Can you double check?

bobbrow (changes requested):

Extension/package.nls.json
gcampbell-msft: That makes sense, changing now.

Extension/package.json
gcampbell-msft: You're right, the only change was the svdPath, so no updating of dependencies is necessary.

Extension/tools/OptionsSchema.json
gcampbell-msft: I fixed the unnecessary changes.

bobbrow (approved):

sean-mcmanus (approved):

PR #8385, opened 2021-11-11

csigs:


Colengms:
I see evidence that the issue causing unwanted linefeeds is still present.  Please do not submit before that is fixed.

PR #8397, opened 2021-11-13
Replace VSO with GitHub Codespaces, because VSO is retired

ChaseKnowlden:
VSO is retired.


bobbrow:
Thank you for this, but we actually just removed that extension and a few others from the pack based on feedback.

PR #8418, opened 2021-11-19
Update codeAnalysis.maxConcurrentThreads description.

sean-mcmanus:


sean-mcmanus:
The "half" should eventually be fixable, but I don't know when. The checks to avoid too much CPU and memory usage haven't been added yet.

bobbrow (approved):

PR #8437, opened 2021-11-23
Fix F5 debug configuration JSON parse error.

sean-mcmanus:
Fix F5 debugging regression from https://github.com/microsoft/vscode-cpptools/pull/8368 . json.Parse was throwing an exception (SyntaxError: Unexpected token in JSON at position 538).


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -563,7 +563,7 @@ class WindowsConfigurationProvider extends DefaultConfigurationProvider {
         "ignoreFailures": true
     },
     {
-        "description":  "${localize("enable.intel.disassembly.flavor", "Set Disassembly Flavor to {0}", "Intel").replace(/\"/g, "\\\"")},
+        "description":  "${localize("enable.intel.disassembly.flavor", "Set Disassembly Flavor to {0}", "Intel").replace(/\"/g, "\\\"")}",
```
bobbrow, line 566: There's a lot of double quotes in there.  Can we change the strings to use single quotes in this case?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -602,7 +602,7 @@ class LinuxConfigurationProvider extends DefaultConfigurationProvider {
...
```
bobbrow, line 605: Same here. Can we use single quotes?

bobbrow (approved):

Extension/src/Debugger/configurationProvider.ts
bobbrow: And, I'm just referring to the outermost quotes to change to single quotes

sean-mcmanus: Why? I don't recall seeing JSON with single quotes....that might only work with TypeScript/JavaScript strings?

sean-mcmanus: The whole string is part of a `` interpolated string.

WardenGnaw: Its similar to the `description` field above. 

One way to fix this string mess is to move it to be a javascript object and ToString it to the caller.

WardenGnaw (approved):

Extension/src/Debugger/configurationProvider.ts
bobbrow: My bad. I thought this was a JavaScript object.  No need to make changes then. I already approved.

sean-mcmanus:
Well, I'll check this in, but feel free to refactor the string stuff later if you want.

PR #8462, opened 2021-11-29
Revert XLoc tools to working version

Colengms:
Our loc contact was mistaken about this issue having been addressed.  We still need to pin to a specific version of XLoc tools to avoid issues with unwanted LFs in our localized strings.


bobbrow:
Could you also update CMake Tools?  I saw a bunch of `\r\n` in the loc PR this morning and was going to ask you about it.

bobbrow (approved):

Issue #4036, opened 2019-08-08
Add a setting for the max threads to use for Find All References


sean-mcmanus:
1. Do Find All References (Rename would have the same issue after it's released).

Bug: It uses all the available cores (minus 1), which is the same as VS's default, but VS also has a setting to set the max cores. E.g. C_Cpp.maxReferencesThreads (between 2-31).

Upvote this issue if you think this is important to add.

Lennon925:
Currently it's slow to find all references if there are many code files in your project.

sean-mcmanus:
FYI, adding this setting wouldn't make Find All References faster -- it would just enable users to use less cores, making Find All References slower (not sure why a user would want that...maybe if the 100% CPU usage causes their machine to freeze up?). And there are certain points in the operation that will use less cores.

sean-mcmanus:
Actually, if a user's machine doesn't have enough memory compared to cores, it could cause memory thrashing, so limiting it to fewer threads could increase performance in those cases.

heartacker:
@sean-mcmanus Should I change the default value to boost my cpptools speed?

heartacker:
![image](https://user-images.githubusercontent.com/1876302/143987970-1e50a0b7-5c30-4fe7-a232-542b02199b65.png)
ALSO,  `Max Cached Process` is set to 0 by default. not null. Is there something we should concern

bobbrow:
It looks like a bug in the translation.  It should be 0.  I saw something similar for the German translation when I checked a while back, but missed this one.

bobbrow:
Sorry, I got it backwards. `null` is the default value when the setting is not set.  Setting it to 0 will pick our lowest acceptable value for the setting (which I think is 2).  Use the gear icon to clear the value if it's showing 0.  Are you saying that it will always show 0 there?

heartacker:
no， it can be changed, and the default is 0, not null

bobbrow:
Ah hah, I was looking at the wrong one.  This is for the **References** Max Cached Processes.

bobbrow:
Yes, it's a bug.  We'll fix it.

sean-mcmanus:
Sorry for the confusion -- the 0 default is currently "by design". We could potentially change the default to null if we discover that it's a better default, but we don't know yet if that is the case or not, i.e. due to the increased memory usage and the fact that the lexer/search results are also not cached (which in some cases may be the bottleneck in Find All References performance).

[...]

PR #8468, opened 2021-11-30
Update description for references.maxCachedProcesses

bobbrow:
Addresses insiders feedback in #4036


sean-mcmanus:
No, the 0 default is "by design". I believe it should be opt-in and not to use the null default.

sean-mcmanus (changes requested):
...see my comment.

bobbrow:
Hmm... We need to update the description then.

sean-mcmanus:
Sure, I'm okay with the updated description -- but the default value was intentionally left out because it's not really necessary to describe the default since it shows in the UI...but we do for some settings, so I'm not sure what the best practice is in regards to that.

sean-mcmanus (approved):

bobbrow:
I think one of the problems here is that the other maxCachedProcesses settings do not disable the feature when you set it to 0.  They have a min value of 2 internally, but this one does not.  Setting it to 0 here turns it off.

sean-mcmanus:
> I think one of the problems here is that the other maxCachedProcesses settings do not disable the feature when you set it to 0. They have a min value of 2 internally, but this one does not. Setting it to 0 here turns it off.

What's the problem? Yeah, I wasn't sure about whether the internal min value of 2 should be propogated to the TypeScript's min value or not.

michelleangela (approved):

bobbrow:
The problem is that the settings all have the same name, but not the same defaults, so I wanted to document it better.

Issue #7380, opened 2021-04-16
C_Cpp.files.exclude should be removed from telemetry


bobbrow:
Bug type: Language Service

We can see who sets it, but we shouldn't log any of the values.  Because I blocked this property, we'll need to create a new key name like we did for clang_format_style in settingsTracker.ts.

sean-mcmanus:
I'm not able to detect a problem, i.e. I can't get it to send any telemetry with files.exclude.

Colengms:
For telemetry, it looks like we're not handling settings which are objects (unless we add explicit handling per object setting).  The default route through the code will fail to find `C_Cpp.files.exclude.**/.vscode` (the first/default property in the `C_Cpp.files.exclude` object) and skip it.  It doesn't looks like `C_Cpp.files.exclude` should be showing up in the telemetry.

PR #8492, opened 2021-12-06
Proactively handle exclude settings.

sean-mcmanus:


Extension/src/LanguageServer/settingsTracker.ts
```diff
@@ -181,6 +181,10 @@ export class SettingsTracker {
                         || key === "codeAnalysis.clangTidy.path"
                         || key === "codeAnalysis.clangTidy.headerFilter" || key === "codeAnalysis.clangTidy.args"
                         || key === "codeAnalysis.clangTidy.config" || key === "codeAnalysis.clangTidy.fallbackConfig"
+
+                        // Note: An existing bug prevents these settings of type "object" from getting processed here,
+                        // so these checks are here just in case that bug gets fixed later on.
```
bobbrow, line 186: Do we have a bug tracking that issue?  If so, then I don't think we need this comment.

sean-mcmanus: I don't believe there's a bug tracking it.

sean-mcmanus: It originally was discovered from https://github.com/microsoft/vscode-cpptools/issues/7380

michelleangela (approved):

Issue #8152, opened 2021-09-14
Implement platform-specific extensions


Colengms:
- [x] Update minimum required VS Code version to 1.61
- [x] Replace "*" in package.json with actual events
- [x] Remove "*" activation logic (remove rewrite logic, merge activation and realActivation, remove TemporaryCommandRegistrar)
- [x] Remove code that downloads platform-binaries and checks `install.lock`
- [x] Remove 'runtime dependencies' from package.json
- [x] Update Offline release pipeline to add `--target <platform>` when packaging each vsix
- [x] Add manual intervention step after building vsix'es, before invoke `vsce publish` to publish each vsix
- [ ] Remove Online release pipeline

[Platform-Specific Extension Documentation](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions)

isidorn:
@Colengms thanks a lot for creating this
Refs: https://github.com/microsoft/vscode/issues/23251

heartacker:
c# has add this implement, https://github.com/OmniSharp/omnisharp-vscode/issues/4775   hope that we can earn it ASAP

bobbrow:
Our next release (1.8.0) will have it.

Colengms:
HI @isidorn , I noticed that `vsce login` doesn't accept a PAT directly.  Is there a way to script login such that no user interaction is required (so we can automate that step - publishing 10 VSIX's) ?

isidorn:
@Colengms great that you are automating this.
I suggest to follow the sample that @joaomoreno created here https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml

So yes, you can use the `VSCE_PAT` env variable. [Example](https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml#L76) how Joao sets it.

isidorn:
👏  🎉 🚀

puremourning:
After this change, is there any way to get the vscode-cpptools packages from a known URL ? the vscode marketplace  doesn't (to my knowledge) allow direct downloads for offline use

sean-mcmanus:
@puremourning The marketplace allows direct downloads, e.g. https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-vscode/vsextensions/cpptools/<version>/vspackage?targetPlatform=linux-x64

puremourning:
[...]

PR #8505, opened 2021-12-08
Remove support for 'online' binary downloads

Colengms:
This is to support platform-specific VSIX's - https://github.com/microsoft/vscode-cpptools/issues/8152 

Removes the installation phase entirely.  Real activation events are present in `package.json`.  Removed re-writing of package.json.   Removed temporary command registrar.  Removed `runtime dependencies` section from `package.json`.  Merged realActivation with activation.  Removed use of `install.lock`.  Removed code to download platform-specific binaries.  Removed all code that verified the correct architecture binaries were available, as that should now be handled by the marketplace.

Also removed a check for CUDA support, which is no longer needed due to dependency on sufficiently recent version of VS Code.

Because this removed the Insider installation channel, this should not be checked in until right before the final insiders is released.


bobbrow:
> TODO: Do we need to remove `C_Cpp.updateChannel`, or is there a way to connect that to how the marketplace will support insiders builds?

Yes, we should remove it.  Marketplace will not refer to us for information about pre-release versions.

Extension/package.json
```diff
@@ -45,7 +45,41 @@
     "virtualWorkspaces": false
   },
   "activationEvents": [
-    "*"
+		"onLanguage:c",
```
bobbrow: can you use spaces instead of TABs for this file?

Extension/src/telemetry.ts
```diff
@@ -64,11 +63,14 @@ export function activate(): void {
...
                 let targetPopulation: TargetPopulation;
-                const userVersion: PackageVersion = new PackageVersion(packageInfo.version);
-                if (!userVersion.suffix) {
-                    targetPopulation = TargetPopulation.Public;
-                } else if (userVersion.suffix === "insiders") {
+
+                // If insiders.flag is present, consider this an insiders build.
+                // If release.flag is present, consider this a release build.
+                // Otherwise, consider this an internal build.
+                if (util.checkFileExistsSync(util.getExtensionFilePath("insiders.flag"))) {
```
bobbrow: Can you promote this into a utility function? Other places in the code may want to know whether we are pre-release or not in the future and it would be nice not to have to remember the filenames in multiple locations.

Extension/src/main.ts
```diff
@@ -73,63 +47,31 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
     // Initialize the DebuggerExtension and register the related commands and providers.
     DebuggerExtension.initialize(context);
 
-    await processRuntimeDependencies();
+    const info: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+    await makeBinariesExecutable();
+    sendTelemetry(info);
```
bobbrow, line 57: I was thinking we only sent this after downloading binaries (i.e. indicating an installation). We may want to rethink what we do with this telemetry, or rename the event to more accurately depict the event.

amtadev:
Does this PR make C++ extension VSIX packages on the marketplace platform-specific and offline?

Colengms:
> Does this PR make C++ extension VSIX packages on the marketplace platform-specific and offline?

That is the plan.  There are other pieces involved, in addition to this PR.

Extension/src/main.ts
```diff
@@ -130,3 +133,37 @@ function sendTelemetry(info: PlatformInformation): void {
...
+        if (settings.updateChannel === "Insiders") {
+            if (!displayedInsidersPrompt.Value) {
+                displayedInsidersPrompt.Value = true;
+                const message: string = localize('updateChannel.changed', "The `C_Cpp.updateChannel` setting is deprecated. Do you want to enable install of pre-releases of the C/C++ extension via the Marketplace?");
+                const yes: string = localize("yes.button", "Yes");
+                const no: string = localize("no.button", "No");
+                vscode.window.showInformationMessage(message, yes, no).then((selection) => {
+                    switch (selection) {
+                        case yes:
+                            vscode.commands.executeCommand("workbench.extensions.installExtension", "ms-vscode.cpptools", { installPreReleaseVersion: true });
```
bobbrow: What happens when there isn't a pre-release version available (e.g. when we release 1.8.0)?  Does this just sign them up for insiders?

Colengms: That is my understanding.  But, we should ask isidorn , just in case.

@isidorn , Can you comment on the behavior of installing with `installPreReleaseVersion` set to true?  If there is no pre-release version available, or if the user is already using a more recent (non-pre-)release version, will it remain on the current version of the extension (and update to a pre-release when it becomes available) ?

sandy081: Passing `installPreReleaseVersion` flag will install the latest version, it can be a pre-release version or a release version, which ever is greater. If it happens that the latest version is a pre-release version then the user gets signed up to pre-release versions and gets auto update to pre-releases. Otherwise not, user gets auto updated only to release versions. In summary, this is not a flag that will sign up users to pre-releases. Hope it is clear.

Edit: You can make sure that user sign up to pre-release version by passing this flag and have a latest pre-release version that is always bigger than latest release version.

Colengms: @sandy081 @isidorn .  It sounds like you understand what we are trying to do.  Our Insiders release channel is a very important part of the process of validating our extension.  However, it's a significant inconvenience to have to maintain an otherwise unnecessary prerelease version to ensure users are properly enrolled in receiving pre-release updates.  We may need to open a feature request to address this, and it may delay us from adopting platform-specific VSIX's.

sandy081: @Colengms I think your request makes sense and I will add that feature soon - https://github.com/microsoft/vscode/issues/139528

sean-mcmanus (approved):

isidorn:
🚀 👏 🎉

PR #8598, opened 2021-12-22
Add repository info to extension pack

Yash-Singh1:
This allows users to get linked to the repository through the Visual Studio Code Marketplace or when locally viewing the extension.

### DIFF
--- a/ExtensionPack/package.json
+++ b/ExtensionPack/package.json
@@ -16,6 +16,11 @@
     "categories": [
         "Extension Packs"
     ],
+    "repository": {
+        "type": "git",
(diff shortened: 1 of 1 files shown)


soreat26:
when we type printf or scanf, please add automatically this ("");

sean-mcmanus (approved):

sean-mcmanus:
> when we type printf or scanf, please add automatically this ("");

The problem with automatically adding the "" is that it may not be desirable in the case where you want to pass in a predefined "const char* variable instead of a new string literal. You could file a feature request though.

PR #8630, opened 2022-01-04
add cl.exe errors

### LINKED ISSUE #8467: Invalid success message when build task fails
Bug type: Debugger

**Describe the bug**
- OS and Version: Windows 11
- VS Code Version: 1.62.3
- C/C++ Extension Version: 1.7.1
- "Build finished successfully." message showing in Terminal when build fails

**To Reproduce**
1. Clone https://github.com/jureid/simple-fibonacci
2. Open project in VS Code using Developer Command Prompt for VS
3. Run vcpkg install in project directory
4. Open main.cpp and select Run > Start debugging
5. Choose Windows > cl.exe
6. See debugger error (main.exe doesn't exist because the build task failed)
7. Navigate to Terminal and see successful build task message even though it failed to find fmt/format.h
8. If you re-run the build task (Terminal > Run Build Task) you'll see it happen again

image

elahehrashedi:
Bug fix: https://github.com/microsoft/vscode-cpptools/issues/8467
Some cl.exe error messages are in stdout message instead of stderr.


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -447,13 +447,12 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
     }
 
     private printBuildSummary(error: string, stdout: string, stderr: string): void {
-        if (error || (!stdout && stderr && stderr.includes("error"))) {
+        if (error || (!stdout && stderr && stderr.includes("error")) ||
+            (stdout && stdout.includes("error C"))) { // cl.exe compiler errors
```
bobbrow, line 451: I wonder if we should change `printBuildSummary` to print something based on the `result` code instead. https://github.com/microsoft/vscode-cpptools/blob/1bd76ebd2d7de01f81f4306031a8afa8a9b4d63b/Extension/src/LanguageServer/cppBuildTaskProvider.ts#L415

There's no guarantee that the strings we're comparing will always be in English, so this might not work.

sean-mcmanus: All our problem matchers assume English results, at least for text like error/warning. Also, I believe cl.exe never has localized output. I seem to recall the result code wasn't sufficient, but someone could double check.

sean-mcmanus (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
bobbrow: The `result` seems to be correct for me:
#### Windows
```
D:\test>cl 1.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.31.30919 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

1.cpp
1.cpp(4): error C2065: 'asdfasdf': undeclared identifier
1.cpp(5): error C2143: syntax error: missing ';' before '}'

D:\test>echo %ErrorLevel%
2
```
#### Linux
```
~/test$ gcc 1.cpp
1.cpp: In function ‘int main(int, char**)’:
1.cpp:4:2: error: ‘asdfasdf’ was not declared in this scope
  asdfasdf
  ^~~~~~~~
~/test$ echo $?
1
```

bobbrow: If that works, then we can just print "Build Succeeded" or "Build Failed" based on the exit code of the compiler.

michelleangela: If actual compiler errors are displayed, it would help inform user of compiling problem. Or is there a reason to only show "Build Succeeded" or "Build Failed"?

bobbrow: The compiler errors will already have been printed to the terminal because we output stdout and stderr.  This is just a summary message at the end in case there is no output (e.g. gcc doesn't print anything sometimes when the build succeeds)

raghavthind2005 (approved):

PR #8636, opened 2022-01-05
Add check for fileBasenameNoExtension.

sean-mcmanus:


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -202,7 +202,10 @@ export class CppBuildTaskProvider implements TaskProvider {
             }
         }
 
-        const taskUsesActiveFile: boolean = definition.args.some(arg => arg.indexOf('${file}') >= 0); // Need to check this before ${file} is resolved
+        // Need to check this before ${file} is resolved
+        const taskUsesActiveFile: boolean = definition.args.some(arg =>
+            arg.indexOf('${file}') >= 0 || arg.indexOf('${fileBasenameNoExtension}') >= 0);
```
bobbrow, line 207: I left this one out because they could hard code the extension which would allow them to be in foo.h and still compile foo.cpp.  Did someone create an issue for this?  I don't think we released my changes yet.

PR #8637, opened 2022-01-05
Add cgmanifest for debugger components

WardenGnaw:
This PR adds in a cgmanifest.json file to comply with Component
Governance Compliance.

This adds two registrations to llvm-project and lldb-mi as components as
they are packaged and published with the macOS version of the VS Code
C/C++ Extension.

### DIFF
--- a/Extension/cgmanifest.json
+++ b/Extension/cgmanifest.json
@@ -1,0 +1,24 @@
+{
+    "Registrations": [
+        {
+            "Component": {
+                "Type": "git",
+                "Git": {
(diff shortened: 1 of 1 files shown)


michelleangela (approved):

bobbrow:
Can you add this to `.vscodeignore` so it doesn't get bundled into the VSIX?

WardenGnaw:
> Can you add this to `.vscodeignore` so it doesn't get bundled into the VSIX?

Addressed in https://github.com/microsoft/vscode-cpptools/pull/8641

Issue #7497, opened 2021-05-09
Add a button to build and run the program like Code Runner Extension


clicktodev:
Type: Feature Request

A button to easily build and run the file would be great, something like the java, python and code runner extensions provide.

sean-mcmanus:
Is the right-click "Build and debug active file" not sufficient? Is this a request to just add a UI button?

hichemfantar:
Yeah something like what the java extension provides would be great.
![image](https://user-images.githubusercontent.com/34947993/117833757-85733a00-b26e-11eb-9567-a0a7ff14ccf2.png)

hichemfantar:
Glad this was picked up again, would appreciate it if you added a "Build and Run active file" in addition to "Build and Debug active file".
Basically a way to run the file without the debugger.

Lunchb0ne:
Implementation depends on the OS, but generally, you can do something like `&& .\${fileBasenameNoExtension}` at the end of the build task to run it too, so you can then make the build task the default, letting you use Ctrl+B keybind to build and run it in a single step.

bobbrow:
I must have pushed a wrong button. I didn't mean to close this.

Zingam:
Will there be an unified experience, which unifies the UI regardless of the build system like with the CMake extension.
Like build, run, test, etc, which also take into account the selected launch config. Well maybe this should be more of a VSCode thing. Currently every language, debugger extension do their own thing. I'm comparing cpptools, cmake tools, codelldb, rust analyzer, the new Swift extension, etc. Maybe I'm wrong and I'm getting confused but well maybe it's confusing.

michelleangela:
This feature is now available to try in version 1.9.0 Insiders. 

To enable feature, set `C_Cpp.debugShortcut` to `true`.

PR #8651, opened 2022-01-11
Add Build and Debug button

elahehrashedi:
feature request: https://github.com/microsoft/vscode-cpptools/issues/7497  (
1. The run and debug shortcut button will be shown only for source files

2. after pressing the button, a list of launch tasks will be shown to the user to select from
    a. The recently chosen task will be on the top of the list
    b. The configured tasks from tasks.json will be below that
    c. The other detected tasks will be shown at the end of the list, if a detected task is already configured, it won't be shown.

3. If there is a default task chosen, the list won't show up, and the default build task will run.

4. In the case of single file mode, the list will always be shown to the user to choose from. The user can not set a default in a single mode file, because we decided not to write tasks in a temporary file, hence no default). But there is a recently used task.

image


Extension/src/Debugger/extension.ts
```diff
@@ -21,7 +21,11 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 const disposables: vscode.Disposable[] = [];
 
 export function buildAndDebugActiveFileStr(): string {
-    return ` - ${localize("build.and.debug.active.file", 'Build and debug active file')}`;
+    return `${localize("build.and.debug.active.file", 'Build and debug active file')}`;
+}
+
+export function CppExtensionSourceStr(): string {
+    return `${localize("Cpp.extension.source", 'C/C++')}`;
```
bobbrow: This does not need to be localized and since it's a constant string, you can just define a const string inside of `configurationProvider.ts` instead of making this a function. It doesn't look like this is used anywhere else.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -192,7 +192,7 @@ class CppConfigurationProvider implements vscode.DebugConfigurationProvider {
             newConfig.program = isWindows ? exeName + ".exe" : exeName;
             // Add the "detail" property to show the compiler path in QuickPickItem.
             // This property will be removed before writing the DebugConfiguration in launch.json.
-            newConfig.detail = task.detail ? task.detail : definition.command;
+            newConfig.detail = "preLaunch Task: " + task.name;
```
bobbrow: This would need to be localized unless you are using the JSON property name (e.g. remove the space between 'preLaunch' and 'Task')

Extension/package.json
```diff
@@ -2540,6 +2540,11 @@
             ]
           },
           "scope": "resource"
+        },
+        "C_Cpp.shortcutMenuBar.buildAndDebugFile": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "%c_cpp.configuration.shortcutMenuBar.buildAndDebugFile.markdownDescription%"
```
bobbrow: This setting name is kind of long and I don't anticipate us adding enough buttons to make it a "category". Plus, VS Code doesn't have any settings with the name "shortcut" in them, so I don't think this category would be commonly referred to this way.  Can it just be `C_Cpp.showDebugShortcut` or something shorter like that?

Extension/package.json
```diff
@@ -3797,6 +3803,13 @@
           "group": "navigation"
         }
       ],
+      "editor/title": [
+        {
+          "command": "C_Cpp.BuildAndDebugFile",
+          "when": "(editorLangId == 'c' || editorLangId == 'cpp' || editorLangId == 'cuda-cpp') && C_Cpp.shortcutMenuBar.buildAndDebugFile",
```
bobbrow: We didn't do this before, but is there an easy way to check if the active file is a header and disable the button by setting some context variable that could be checked here?

Extension/package.nls.json
```diff
@@ -26,6 +25,7 @@
     "c_cpp.command.RunCodeAnalysisOnOpenFiles.title": "Run Code Analysis on Open Files",
     "c_cpp.command.RunCodeAnalysisOnAllFiles.title": "Run Code Analysis on All Files",
     "c_cpp.command.ClearCodeAnalysisSquiggles.title": "Clear Code Analysis Squiggles",
+    "c_cpp.command.BuildAndDebugFile.title": "Build and Debug File",
```
bobbrow: Does this command still show up in the Command Palette?  It might still need to say "Active" for context.

Extension/package.nls.json
```diff
@@ -208,6 +208,7 @@
     "c_cpp.configuration.debugger.useBacktickCommandSubstitution.markdownDescription": { "message": "If `true`, debugger shell command substitution will use obsolete backtick ``(`)``.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.contributes.views.cppReferencesView.title": "C/C++: Other references results.",
     "c_cpp.contributes.viewsWelcome.contents": { "message": "To learn more about launch.json, see [Configuring C/C++ debugging](https://code.visualstudio.com/docs/cpp/launch-json-reference).", "comment": [ "Markdown text between () should not be altered: https://en.wikipedia.org/wiki/Markdown" ] },
+    "c_cpp.configuration.shortcutMenuBar.buildAndDebugFile.markdownDescription": { "message":"Show icon for 'Build and Debug' in shortcut menu bar", "comment": ["Markdown text between '' can be translated or localized"]},
```
bobbrow: `''` are not markdown characters, so you don't need the "comment"

bobbrow (changes requested):

Extension/package.json
Colengms: We used the following to set `when` context variables from code, for FAR:

```
vscode.commands.executeCommand('setContext', 'cppReferenceTypes:hasResults', hasResults);
...
vscode.commands.executeCommand('setContext', 'refView.isGroupedByFile', this.referencesModel.groupByFile);
```

The value would need to be updated as the active file changes.  Not sure if our `isHeader` function is sufficient, though.  It looks like we have a better check in `getTasks` (which should probably be moved into `isHeader`).

Extension/package.json
sean-mcmanus: Are those icons in the top right called "shortcuts" or somethign else?

bobbrow: I don't think they have an official name.  It's part of the `editor/title` section of `package.json` but using "title" seems cryptic.

Extension/package.json
```diff
@@ -2540,6 +2540,11 @@
             ]
           },
           "scope": "resource"
+        },
+        "C_Cpp.shortcutMenuBar.buildAndDebugFile": {
```
jureid: To ship this as an A/B experiment, we need to add:
```
"tags": [
     "experimental"
]
```

to this setting object (see [example in Python's package.json](https://github.com/microsoft/vscode-python/blob/e1773fc5c1bb3201bb4fe279c4850d3920ab1833/package.json#L554)) and change the default value to be false.

Then, I can override the setting's value in Control Tower to be true for the treatment group.

elahehrashedi: For this branch, I haven't applied this change, for testing purposes, but for shipping, I will apply it.

Extension/src/LanguageServer/client.ts
```diff
@@ -1718,6 +1718,7 @@ export class DefaultClient implements Client {
         if (document.uri.scheme === "file") {
             const uri: string = document.uri.toString();
             openFileVersions.set(uri, document.version);
+            vscode.commands.executeCommand('setContext', 'BuildAndDebug.isSourceFile', util.isCppOrCFile(document.uri.fsPath));
```
bobbrow: You might also want to set this to false if the document uri is not a "file".

Extension/src/LanguageServer/client.ts
```diff
@@ -2654,6 +2655,7 @@ export class DefaultClient implements Client {
             && (editor.document.languageId === "c"
                 || editor.document.languageId === "cpp"
                 || editor.document.languageId === "cuda-cpp")) {
+            vscode.commands.executeCommand('setContext', 'BuildAndDebug.isSourceFile', util.isCppOrCFile(editor.document.uri.fsPath));
```
bobbrow: Similar to above, you should set this to false if the uri is not "file."

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -662,3 +670,112 @@ export class ConfigurationSnippetProvider implements vscode.CompletionItemProvid
         return Promise.resolve(new vscode.CompletionList(items, true));
     }
 }
+
+export function buildAndDebugActiveFileStr(): string {
+    return `${localize("build.and.debug.active.file", 'Build and debug active file')}`;
+}
+
+export async function buildAndDebugShortCut(textEditor: vscode.TextEditor, cppVsDbgProvider: CppVsDbgConfigurationProvider | null, cppDbgProvider: CppDbgConfigurationProvider, debugModeOn: boolean = true): Promise<void> {
```
bobbrow: Function names should be a verb or verb phrase.  Can you remove "ShortCut" from the name?

elahehrashedi:
ready for review

Extension/src/common.ts
```diff
@@ -171,9 +155,38 @@ export function getVcpkgRoot(): string {
...
+    return (fileExt === ".C") || [".cu", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".ino", ".ipp", ".tcc"].some(ext => fileExtLower === ext);
+}
+
+export function isCFile (uri: vscode.Uri): boolean {
+    const fileExt: string = path.extname(uri.fsPath);
+    const fileExtLower: string = fileExt.toLowerCase();
+    return (fileExt === ".C") || fileExtLower === ".c";
+}
+
+export function isCppOrCFile(file?: string): boolean {
```
michelleangela: Can `isCppOrCFile` call the functions `isCppFile` and `isCFile`?

sean-mcmanus (approved):

michelleangela (approved):

Extension/package.nls.json
```diff
@@ -26,6 +25,8 @@
     "c_cpp.command.RunCodeAnalysisOnOpenFiles.title": "Run Code Analysis on Open Files",
     "c_cpp.command.RunCodeAnalysisOnAllFiles.title": "Run Code Analysis on All Files",
     "c_cpp.command.ClearCodeAnalysisSquiggles.title": "Clear Code Analysis Squiggles",
+    "c_cpp.command.BuildAndDebugFile.title": "Debug Active File",
```
jureid: We should add "C/C++" in the button titles in case customers have the Java, Python, or Code Runner extensions installed. I also don't think we need "Active" in the title if we have "C/C++" since the buttons will only be shown when a source file is active.

Let's use "Debug C/C++ File" and "Run C/C++ File"

jureid: This would be consistent with Python:
image

jureid (changes requested):

Extension/package.nls.json
elahehrashedi: good idea

PR #8661, opened 2022-01-12
Stop splitting ; or : in defines and compilerArgs.

### LINKED ISSUE #4028: IntelliSense process should not fail due to invalid characters in a MACRO name (i.e. ":")
- OS and Version: Windows 10
- VS Code Version: 1.36.1
- C/C++ Extension Version: 0.24.1

As the title says, I have a weird issue. A few days ago, Intellisense started behaving really badly, and ultimately stopped working on some of my projects. On the same setup, some other projects work perfectly, so it might be related to the complexity of the CMake structure ?

Anyway, in the C/C++ output, I just have the following error: `Failed to read response from server: 22`. And when I check the running processes related to VSCode, I noticed that `Microsoft.VSCode.CPP.Extension.exe` is running, but there is no `Microsoft.VSCode.Cpp.Intellisense.MSVC.exe` process on the projects where I have the problem. On the ones where intellisense works, this process is running as usual.

In my settings, I'm using

```
	"C_Cpp.default.intelliSenseMode": "msvc-x64",
	"C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools",
```

But as I said, those settings are common to all my projects, and only a few of them exhibit the problem so it might not be related.

Problem is that I have no idea what to do to diagnose where this error might come from ... I tried uninstalling / reinstalling the extension, no luck. Is there anything I can do to find where this problem comes from ?

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/4028


Colengms (approved):

bobbrow (changes requested):
Hang on a minute. I'd like to talk to you before you revert this behavior.

bobbrow (approved):
False alarm. This looks good to me.

Issue #8704, opened 2022-01-20
Implement mitigation to migrate Insiders users


Colengms:
The following issue will cause some users who were Insiders to not be migrated to the new Prerelease system correctly: https://github.com/microsoft/vscode/issues/140949  This issue repro'ed for users who got updated to 1.8.0 (insiders3) via the `updateChannel` mechanism.  (Due to the extension programmatically installing the new VSIX).

Once the VS Code issue is fixed, we should ship a release version with additional code to migrate users to the Prerelease system if their `updateChannel` was set to `Insiders` or if running a VS Code Insiders build and the `updateChannel` was unset (default would have been to use cpptools Insiders).

Colengms:
This should use a persistent var to avoid migrating the user more than once.  If the user manually switches from Prerelease to Release version, we don't want to keep moving them back to Prerelease.

PR #8705, opened 2022-01-20
Mitigate issue with migrating Insiders to Prerelease

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/8704
This works around https://github.com/microsoft/vscode/issues/140949

When existing VS Code Insiders users were updated to 1.8.0 (insiders3) via the `updateChannel` mechanism, the VSIX programmatically downloaded from GitHub was not correctly identified as a Prerelease.  Presumably, this means those users were not subscribed to receive further Prerelease versions.

This mitigation should take effect once those users receive a release version from the marketplace.  If running VS Code Insiders and `updateChannel` had not been modified, they will be subscribed to Prereleases.  This only occurs once, to avoid repeatedly moving the user to Prerelease in case they had manually switched to Release.

### DIFF
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -160,4 +160,20 @@ export function UpdateInsidersAccess(): void {
             migratedInsiders.Value = false;
         }
     }
+
+    // Mitigate an issue with VS Code not recognizing a programmatically installed VSIX as Prerelease.
+    // If using VS Code Insiders, and updateChannel is not explicitly set, default to Prerelease.
+    // Only do this once. If the user manually switches to Release, we don't want to switch them back to Prerelease again.
+    if (util.isVsCodeInsiders()) {
+        const insidersMitigationDone: PersistentState<boolean> = new PersistentState<boolean>("CPP.insidersMitigationDone", false);
+        if (!insidersMitigationDone.Value) {
+            if (vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
+                const settings: CppSettings = new CppSettings();
+                if (settings.getWithUndefinedDefault<string>("updateChannel") === undefined) {
+                    vscode.commands.executeCommand("workbench.extensions.installExtension", "ms-vscode.cpptools", { installPreReleaseVersion: true });
+                }
+            }
+            insidersMitigationDone.Value = true;
+        }
+    }
 }
--- a/Extension/src/main.ts
+++ b/Extension/src/main.ts
@@ -145,14 +145,16 @@ function sendTelemetry(info: PlatformInformation): void {
 }
 
 export function UpdateInsidersAccess(): void {
+    let installPrerelease: Boolean = false;
+
     // Only move them to the new prerelease mechanism if using updateChannel of Insiders.
     const settings: CppSettings = new CppSettings();
     const migratedInsiders: PersistentState<boolean> = new PersistentState<boolean>("CPP.migratedInsiders", false);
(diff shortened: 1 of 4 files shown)


Extension/src/main.ts
```diff
@@ -160,4 +160,20 @@ export function UpdateInsidersAccess(): void {
...
+
+    // Mitigate an issue with VS Code not recognizing a programmatically installed VSIX as Prerelease.
+    // If using VS Code Insiders, and updateChannel is not explicitly set, default to Prerelease.
+    // Only do this once. If the user manually switches to Release, we don't want to switch them back to Prerelease again.
+    if (util.isVsCodeInsiders()) {
+        const insidersMitigationDone: PersistentState<boolean> = new PersistentState<boolean>("CPP.insidersMitigationDone", false);
+        if (!insidersMitigationDone.Value) {
+            if (vscode.workspace.getConfiguration("extensions", null).get<boolean>("autoUpdate")) {
+                const settings: CppSettings = new CppSettings();
+                if (settings.getWithUndefinedDefault<string>("updateChannel") === undefined) {
```
sean-mcmanus, line 173: I'm not understanding why this needs to be checked.

Extension/src/main.ts
```diff
@@ -160,4 +160,20 @@ export function UpdateInsidersAccess(): void {
...
+                    vscode.commands.executeCommand("workbench.extensions.installExtension", "ms-vscode.cpptools", { installPreReleaseVersion: true });
```
sean-mcmanus: This still requires a VS Code Insiders fix first?

Extension/src/main.ts
Colengms: If `updateChannel` was explicitly set to `Insiders` it should already have been handled by the prior check in this function.  If `updateChannel` was explicitly set to something other than `Insiders`, we should not subscribe them to prereleases.  So, this check is only needed if `updateChannel` is not explicitly set, and they are running on a VS Code Insiders build, to apply the (now missing) `Insiders` default value for `updateChannel`.  The issue this is mitigating only occurs when the user had downloaded the build via the `updateChannel` mechanism using VS Code Insiders.

Extension/src/main.ts
Colengms: I don't think so.  It only occurs with VS Code Insiders, and this issue has been fixed: https://github.com/microsoft/vscode/issues/139528  We don't need to wait for https://github.com/microsoft/vscode/issues/140949 to be fixed, because this fix mitigates it.

Extension/src/main.ts
sean-mcmanus: But users who explicitly set the updateChannel to "Insiders" could still be using VS Code Insiders and they won't be fixed by this?

sean-mcmanus: It sounds like you're saying the fix for https://github.com/microsoft/vscode/issues/139528 would fix the explicitly set user case.

sean-mcmanus (approved):

sean-mcmanus (approved):

sean-mcmanus:
You might want to get another person to review this too.

Extension/src/main.ts
```diff
@@ -145,19 +145,40 @@ function sendTelemetry(info: PlatformInformation): void {
 }
 
 export function UpdateInsidersAccess(): void {
+    let installPrerelease: Boolean = false;
```
bobbrow: lowercase boolean?

bobbrow (approved):

PR #8739, opened 2022-01-29
Rework how cancelation is processed for semantic tokens and folding operations

Colengms:
When implementing support for semantic tokens and code folding, I had failed to leverage the existing mechanism for cancelling those requests (which is possible when a cancellation token is passed to calls to `sendRequest`).  We can also leverage this for other operations that involve `sendRequest`.  When invoked as part of a provider, the cancellation token passed to the provider can be passed through.

There is a corresponding PR for changes to the native process.


sean-mcmanus:
This is for 1.9.0, right?

Colengms:
> This is for 1.9.0, right?

Yes

sean-mcmanus (approved):

PR #8778, opened 2022-02-04
Update changelog for 1.8.4

Colengms:


sean-mcmanus (changes requested):
We normally end the descriptions with a period, i.e. `[` -> `. [`

sean-mcmanus (approved):

Issue #8762, opened 2022-02-02
Language service doesn't restart correctly "Invalid opened file instance. Ignoring IntelliSense message for file" with a multiroot workspace


sean-mcmanus:
Kill the cpptools process. Then try to use langauge service features.

Bug: Features stop working. Logging shows "Invalid opened file instance. Ignoring IntelliSense message for file".

UPDATE: Requires a multiroot workspace.

PR #8810, opened 2022-02-09
Fix issue with recovery from cpptools crash when using multiroot

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/8762

When recovering from the cpptools process crashing (which should hopefully be rare), the process of recreating client objects per workspace would skip clients that were not tracking any associated files.

In `ClientCollection.replace()`, the client objects had been getting opportunistically created in `getClientFor`, but no call was incurred if not the active client and no files were tracked.  This fix makes client creation somewhat more explicit, with a call to `getClientForFolder`.  (I moved client creation into this subroutine to keep that consolidated.)

Also fixed an issue with `defaultClient` not being reset causing the original (broken) client instance to linger and be used for files not associated with a workspace.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -689,6 +689,7 @@ export interface Client {
     RootPath: string;
     RootRealPath: string;
     RootUri?: vscode.Uri;
+    RootFolder?: vscode.WorkspaceFolder;
     Name: string;
     TrackedDocuments: Set<vscode.TextDocument>;
     onDidChangeSettings(event: vscode.ConfigurationChangeEvent, isFirstClient: boolean): { [key: string]: string };
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1483,15 +1483,15 @@ export class DefaultClient implements Client {
                     languageClientCrashedNeedsRestart = true;
                     telemetry.logLanguageServerEvent("languageClientCrash");
                     if (languageClientCrashTimes.length < 5) {
-                        allClients.forEach(client => { allClients.replace(client, true); });
+                        allClients.recreateClients(true);
                     } else {
                         const elapsed: number = languageClientCrashTimes[languageClientCrashTimes.length - 1] - languageClientCrashTimes[0];
                         if (elapsed <= 3 * 60 * 1000) {
                             vscode.window.showErrorMessage(localize('server.crashed2', "The language server crashed 5 times in the last 3 minutes. It will not be restarted."));
-                            allClients.forEach(client => { allClients.replace(client, false); });
+                            allClients.recreateClients(false);
                         } else {
                             languageClientCrashTimes.shift();
-                            allClients.forEach(client => { allClients.replace(client, true); });
+                            allClients.recreateClients(true);
                         }
                     }
                     return CloseAction.DoNotRestart;
--- a/Extension/src/LanguageServer/clientCollection.ts
+++ b/Extension/src/LanguageServer/clientCollection.ts
@@ -193,6 +193,10 @@ export class ClientCollection {
                 }
             });
 
+            // Note: VS Code currently reloads the extension whenever the primary (first) workspace folder is added or removed.
+            // So the following handling of changes to defaultClient are likely unnecessary, unless the behavior of
+            // VS Code changes. The handling of changes to activeClient are still needed.
+
             if (oldDefaultClient) {
                 const uri: vscode.Uri | undefined = oldDefaultClient.RootUri;
                 // uri will be set.
--- a/Extension/src/LanguageServer/clientCollection.ts
+++ b/Extension/src/LanguageServer/clientCollection.ts
@@ -104,33 +104,26 @@ export class ClientCollection {
     /**
      * creates a new client to replace one that crashed.
      */
-    public replace(client: cpptools.Client, transferFileOwnership: boolean): cpptools.Client | undefined {
-        let key: string | undefined;
-        for (const pair of this.languageClients) {
-            if (pair[1] === client) {
-                key = pair[0];
-                break;
-            }
-        }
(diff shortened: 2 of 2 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -116,22 +116,29 @@ export class ClientCollection {
...
 
+            const newClient: cpptools.Client = this.getClientForFolder(client.RootFolder);
+
             if (transferFileOwnership) {
                 // This will create a new Client for the workspace since we removed the old one from this.languageClients.
                 client.TrackedDocuments.forEach(document => this.transferOwnership(document, client));
                 client.TrackedDocuments.clear();
-            } else {
-                // Create an empty client that will continue to "own" files matching this workspace, but ignore all messages from VS Code.
-                this.languageClients.set(key, cpptools.createNullClient());
```
bobbrow, line 125: If you remove this, then isn't that going to restart the language server even though we said we wouldn't for this case?

Extension/src/LanguageServer/client.ts
```diff
@@ -689,6 +689,7 @@ export interface Client {
     RootPath: string;
     RootRealPath: string;
     RootUri?: vscode.Uri;
+    RootFolder?: vscode.WorkspaceFolder;
```
bobbrow, line 692: If you add this to the interface, I think you need to add it to the NullClient as well.

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -164,6 +171,10 @@ export class ClientCollection {
                         // this.activeClient.selectionChanged(Range.create(vscode.window.activeTextEditor.selection.start, vscode.window.activeTextEditor.selection.end);
                     }
 
+                    if (this.defaultClient === client) {
+                        this.defaultClient = this.getClientForFolder();
```
bobbrow: It looks like `this.getClientForFolder()` returns `this.defaultClient` when no folder is passed in so I'm not sure if this block is doing what you were expecting.

Extension/src/LanguageServer/client.ts
Colengms: I believe this is implicit.  The `?` indicates the value does not need to be defined, and it's correct for `NullClient` to not define it.  (It doesn't have a value to return).

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -154,14 +157,15 @@ export class ClientCollection {
...
                     if (this.activeClient === client && this.activeDocument) {
                         // Need to make a different client the active client.
                         this.activeClient = this.getClientFor(this.activeDocument.uri);
                         this.activeClient.activeDocumentChanged(this.activeDocument);
-                        // may not need this, the navigation UI should not have changed.
-                        // this.activeClient.selectionChanged(Range.create(vscode.window.activeTextEditor.selection.start, vscode.window.activeTextEditor.selection.end);
+                    }
+
+                    if (this.defaultClient === client) {
+                        this.defaultClient = this.getClientFor(folder.uri);
```
Colengms: Still working on this part.  Need to handle what happens if the user removes all folders and there isn't an existing client to make the default client.

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -198,13 +262,21 @@ export class ClientCollection {
...
-            defaultClient.sendAllSettings();
-            return newClient;
+            return this.createClient(folder);
+        }
+    }
+
+    public createClient(folder?: vscode.WorkspaceFolder, deactivated?: boolean): cpptools.Client {
+        const newClient: cpptools.Client = cpptools.createClient(this, folder);
+        if (deactivated) {
+            newClient.deactivate(); // e.g. prevent the current config from switching.
```
Colengms, line 277: I am still a little unclear on this bit here.  Should ALL clients be created in a deactivated state, and only activated in the context of processing an `activeDocumentChange` (and when we need to replace `this.activeClient` and there is a `this.activeDocument` present) ?

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -104,47 +104,55 @@ export class ClientCollection {
...
-        if (key) {
-            this.languageClients.delete(key);
+        for (const pair of oldLanguageClients) {
+            const client: cpptools.Client = pair[1];
 
+            let newClient: cpptools.Client;
             if (transferFileOwnership) {
-                // This will create a new Client for the workspace since we removed the old one from this.languageClients.
+                newClient = this.createClient(client.RootFolder);
                 client.TrackedDocuments.forEach(document => this.transferOwnership(document, client));
```
bobbrow, line 119: `this.transferOwnership` looks for a client in the map to transfer the file to, but line 111 just cleared the map.  I think you need to re-create the clients before you can transfer ownership.

Extension/src/LanguageServer/clientCollection.ts
sean-mcmanus: I believe the effect of having the client become active is that the config for that workspace folder will be shown in the status bar -- so you just have to decide if that's what you want to happen in your context. Usually it would make sense for all clients except for some "active" one to be deactivated...but if you don't detect an active one you can use the 1st workspace folder. In the crash scenario, it would be best to keep the active client the same as it was previously.

Extension/src/LanguageServer/clientCollection.ts
Colengms: transferOwnership will create clients.  Previously, it was how clients were being created here.  Though, now the immediately previous line creates the client (because I need it again later) that ownership of those files will be transferred to.  (Unlike the code that handles folder adding and removed, the transfers of ownership here should land in the client for the same WorkspaceFolder).

sean-mcmanus (approved):

PR #8819, opened 2022-02-09
add telemetry for debug and Run button

elahehrashedi:


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -679,6 +679,7 @@ export function buildAndDebugActiveFileStr(): string {
 
 export async function buildAndDebug(textEditor: vscode.TextEditor, cppVsDbgProvider: CppVsDbgConfigurationProvider | null, cppDbgProvider: CppDbgConfigurationProvider, debugModeOn: boolean = true): Promise<void> {
     const folder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(textEditor.document.uri);
+    const event: string = debugModeOn ? "buildAndDebug" : "buildAndRun";
```
bobbrow: Rather than have two different event names, I find it easier to query for Properties when I do the reporting.  I think it would be better if we have one event name, maybe something that matches what we're calling this (e.g. "launchShortcut"), and then add a "type" property where you can put "buildAndDebug" or "buildAndRun" as the values.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -724,6 +725,7 @@ export async function buildAndDebug(textEditor: vscode.TextEditor, cppVsDbgProvi
     }
 
     if (!selection) {
+        Telemetry.logDebuggerEvent(event + "UserCancelled");
```
bobbrow: Same idea here.  States that relate to the event (user click the shortcut button), should not be their own event.  Let's just add a property for this.  Maybe "cancelled": "true"/"false".

Extension/src/LanguageServer/client.ts
```diff
@@ -1667,6 +1667,14 @@ export class DefaultClient implements Client {
                     if (changedSettings["addNodeAddonIncludePaths"] && settings.addNodeAddonIncludePaths && this.configuration.nodeAddonIncludesFound() === 0) {
                         util.promptForReloadWindowDueToSettingsChange();
                     }
+                    if (changedSettings["debugShortcut"]) {
```
bobbrow: This is not needed.  We already log the changed settings on line 1680 below.

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: good idea. I will use LaunchPlayButton to match with what Julia says when referring to this button.

jureid (changes requested):
Can we also log an event when they run/debug in single file mode so we know how many customers are doing this? Other than that, looks great! Thanks a lot for doing this so quickly!

bobbrow:
> Can we also log an event when they run/debug in single file mode so we know how many customers are doing this? Other than that, looks great! Thanks a lot for doing this so quickly!

I recommend adding a property instead of a new event. :)

bobbrow (approved):

Issue #8755, opened 2022-02-01
Automatically set target architecture in tasks.json and launch.json to unblock build/debug on M1 chip


jureid:
Type: Feature Request

Unblock "Build and debug active file" on macOS ARM64 (M1 chip) by automatically adding "--target=x86_64-apple-darwin-gnu" to the args array in tasks.json and setting "targetArchitecture" to "x86_64" in launch.json.

If using CMake Tools, can we automatically set "-DCMAKE_OSX_ARCHITECTURES=x86_64"?

Next step would be to get the debugger binaries to run natively on Apple Silicon ARM64.: https://github.com/microsoft/vscode-cpptools/issues/7035

jureid:
@WardenGnaw can you confirm these are the correct properties / values?

WardenGnaw:
You will want `--target= x86_64-apple-darwin-gnu`. Possible cross compilation targets are listed in the clang reference. 

https://clang.llvm.org/docs/CrossCompilation.html

michelleangela:
>If using CMake Tools, can we automatically set "-DCMAKE_OSX_ARCHITECTURES=x86_64"?

Since this workaround is specific to the cpptool extension's default build settings that it defines in tasks.json, the extension should not also modify the default value of a build setting from another extension such as CMake Tools.

Issue #7035, opened 2021-02-23
Support Apple Silicon ARM64 architecture natively for debugger


michelleangela:
Type: Debugger

This feature request is to have debugger binaries to run natively on Apple Silicon ARM64.

The feature request to have the language server binaries to run natively on Apple Silicon ARM64 is tracked by 
https://github.com/microsoft/vscode-cpptools/issues/6595

**Note:** the cpptools, cpptools-srv, and debugger binaries built for macOS x64 may work on Apple Silicon's x64 emulator, however this may use more resources than having binaries run natively as ARM64 on Apple Silicon.

XIAZY:
Just some extra info here. VSCode (stable channel) has been released for ARM64-based Macs. Attempting to start debuggers on ARM64 Macs with `"targetArchitecture": "arm64"` will yield this error:
```
ERROR: Unable to start debugging. Unexpected LLDB output from command "-exec-run". process exited with status -1 (attach failed ((os/kern) invalid argument))
```

Lihp8270:
I'm also experiencing the same issue on the latest update on VSCode running natively on M1.

michelleangela:
@Lihp8270 
The feature to support native debugging on Apple Silicon ARM64 has not been done yet. You can run the current debugger binaries built for Apple Intel x64 using the Rosetta emulator. Note that the debugger for Apple Intel x64 can only debug x86_64 binaries.

jmarti44:
Any updates on this. I've been having issues running the debugger the debugger for C++

benjirer:
Does anyone know if this has been fixed? Quite new to vs code, would you recommend downloading the intel version and running through rosetta to use the debugger?

cloph:
> Does anyone know if this has been fixed? Quite new to vs code, would you recommend downloading the intel version and running through rosetta to use the debugger?

No, get the M1 version of Visual Studio Code - but the application you want to debug needs to be built for x86_64 and will be run via Rosetta.

benjirer:
Thanks for the hint! Building using Clang x86_64-apple-darwin and then debugging yields the same error. I think I'm way in over my head here, do I need to change anything else?

nicudo:
You need both the executable and the debugger to target x86_64.
I have: "targetArchitecture": "x86_64" in launch.json and "-DCMAKE_OSX_ARCHITECTURES=x86_64" in my build command (use --target=x86-64 if you just use clang).

kinkard:
[...]

PR #8820, opened 2022-02-10
Build and debug as x86_64 on Apple Silicon (macOS ARM64) by default

michelleangela:
This is a workaround (https://github.com/microsoft/vscode-cpptools/issues/8755) until https://github.com/microsoft/vscode-cpptools/issues/7035 is resolved.


Colengms (approved):

bobbrow:
We might actually get #7035 in 1.9 so maybe we just back this out?

michelleangela:
This is a temporary workaround and will get removed when #7035 is shipped.

Also,, this workaround is only applied when creating a new tasks.json or new task through the "C++ run and debug" feature.

PR #8826, opened 2022-02-10
resolve tasks in single mode

elahehrashedi:


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -55,6 +55,12 @@ export class CppBuildTaskProvider implements TaskProvider {
         return undefined;
     }
 
+    public resolveInsiderTask(_task: CppBuildTask): CppBuildTask | undefined {
```
sean-mcmanus, line 58: Should the "resolveTask" method be removed? It doesn't appear to be used anymore.

elahehrashedi: It is used by VsCode when running tasks from tasks.json

sean-mcmanus: Ok

sean-mcmanus (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
sean-mcmanus: TypeScript doesn't show any indication that the method is overriding a base class method (unlike C++'s override/virtual), so I didn't notice it was an override.

sean-mcmanus (approved):

PR #8850, opened 2022-02-12
Fix issue with cpptools restarting when it should not

Colengms:
This is a follow up my previous fix which was incomplete (didn't write the new clients to the languageClients map), leading to subsequent operations triggering launch of cpptools.  This change adds a state variable to ensure that any potential creation of clients is prevented once in this failsafe state.


sean-mcmanus:
Was this a user reported issue or you found from further testing?

sean-mcmanus (approved):

PR #8857, opened 2022-02-13
Make SwitchHeaderSource use workbench.editor.revealIfOpen setting

joelmsmith:
This change makes the SwitchHeaderSource command respect the workbench.editor.revealIfOpen setting.

~`C_Cpp.SwitchHeaderSource` has different behavior depending on whether the destination file is visible in a different editor pane or not.  This commit adds an option `headerSourcePaneJumping` which controls whether the command will jump to a different editor if the destination file is already visible there.  The default value of true preserves the command's original behavior.~

~Personally I do not want the default behavior; I want it to simply toggle back and forth between header / source in the same editor that is focused, regardless of what the other ones are displaying.~

I'm new to vscode and still finding my way around, so it is likely that I've missed or overlooked something.  Feedback/guidance is welcome.


Extension/src/LanguageServer/client.ts
```diff
@@ -1109,6 +1109,7 @@ export class DefaultClient implements Client {
         const settings_intelliSenseEngine: (string | undefined)[] = [];
         const settings_intelliSenseEngineFallback: (string | undefined)[] = [];
         const settings_errorSquiggles: (string | undefined)[] = [];
+        const settings_headerSourcePaneJumping: (boolean | undefined)[] = [];
```
sean-mcmanus: All the edits to client.ts are not needed and should be removed. Those client.ts settings are just for settings which need to be sent to our cpptools process.

sean-mcmanus (changes requested):
Seems fine to me, other than the one comment.

Extension/package.json
```diff
@@ -2111,6 +2111,12 @@
           "description": "%c_cpp.configuration.errorSquiggles.description%",
           "scope": "resource"
         },
+        "C_Cpp.headerSourcePaneJumping": {
+          "type": "boolean",
+          "default": true,
+          "description": "%c_cpp.configuration.headerSourcePaneJumping.description%",
+          "scope": "resource"
```
sean-mcmanus: This should be "application" scope instead of "resource" scope.

sean-mcmanus (changes requested):

sean-mcmanus (approved):
@bobbrow @jureid Any thoughts on the naming/text?

joelmsmith:
Hi @sean-mcmanus, thanks for looking at this.  I'm happy to change any of the naming/text if there's a a way to phrase things that better aligns with the project's terminology.

bobbrow:
Is there any desire for this to be a more general setting to cover any cases where we would open/focus a text editor?  (For example, goto definition)

joelmsmith:
As far as I can tell, Go to Definition, Go To Declaration, and Go To Type Definition all operate within the focused editor.

bobbrow:
If the definition is outside the current document, we should be opening a new file.  But I forget whether that is done by vscode or us.

sean-mcmanus (changes requested):
Instead of adding a new setting, maybe the switching should read the VS Code workbench.editor.revealIfOpen setting unless I'm misunderstanding that setting or there's a reason to separate out this particular case? I think that defaults to false. i.e. check the result of vscode.workspace.getConfiguration("workbench.editor.revealIfOpen ".

joelmsmith:
@sean-mcmanus - I'll take a look at this.  I did not know about this setting, but it sounds like making SwitchHeaderSource obey it will do what I want.
@bobbrow - a new file gets opened (or the tab focused, if it is already opened), but for those 'Go To' commands this happens in the same split (or what the Reval If Open setting refers to as "currently active editor group").

sean-mcmanus (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -484,14 +484,18 @@ async function onSwitchHeaderSource(): Promise<void> {
...
-    // If the document is already visible in another column, open it there.
-    vscode.window.visibleTextEditors.forEach((editor, index, array) => {
-        if (editor.document === document && !foundEditor) {
-            foundEditor = true;
-            vscode.window.showTextDocument(document, editor.viewColumn);
-        }
-    });
+    if (workbenchConfig.get("editor.revealIfOpen")) {
+        // If the document is already visible in another column, open it there.
+        vscode.window.visibleTextEditors.forEach((editor, index, array) => {
```
bobbrow: I know you didn't write this code, but if you wouldn't mind updating this line of the lambda to the more concise code below while you're in here, I'd appreciate it.  (totally optional though)
```
vscode.window.visibleTextEditors.forEach(editor => {
```
There's no need to declare `index` and `array`

bobbrow (approved):

sean-mcmanus (approved):

Issue #8773, opened 2022-02-03
Provide dynamic debug configuration for F5 instead of creating launch.json


jureid:
Type: Feature Request

For users that don't have tasks.json and launch.json, we should provide a dynamic `resolveDebugConfiguration` when they hit F5, which will provide a launch configuration on the fly, rather than generating tasks.json and launch.json
Here's an [example](https://github.com/microsoft/vscode-mock-debug/blob/main/src/activateMockDebug.ts#L161).

PR #8861, opened 2022-02-14
Dynamic debug config info

elahehrashedi:
Modification to "Run and Debug" button in debug panel to make it consistent with our play button:
1. keeping the Debug info in memory and skip writing it in launch.json
2. if the launch.json exists try to read the debug configs to check if the user has modified any debug info
3.Fixes are added for single-mode file debugging, however it is only working for cl.exe builds and msvc debugger for now

Feature  Request: https://github.com/microsoft/vscode-cpptools/issues/8773

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -116,7 +116,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
     resolveDebugConfigurationInternal(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration> | undefined {
         if (!config || !config.type) {
-            return undefined; // aborts debugging silently
+            return undefined;
         }
 
         if (config.type === 'cppvsdbg') {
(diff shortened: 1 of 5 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -31,15 +31,15 @@ function isDebugLaunchStr(str: string): boolean {
  * Ensures that the selected configuration's preLaunchTask (if existent) is populated in the user's task.json.
  * Automatically starts debugging for "Build and Debug" configurations.
  */
-export class QuickPickConfigurationProvider implements vscode.DebugConfigurationProvider {
-    private underlyingProvider: CppConfigurationProvider;
+export class CppDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
```
bobbrow: Let's not rename all this stuff.  Can you change it back?

Naming a class "Underlying" isn't a very helpful choice.  "Underlying" only has meaning within the context that it's used.  A standalone class is not "underlying" anything.

We shouldn't use the abbreviation `dbg` in variable names.  In general abbreviations should be avoided, but in this case it also causes confusion with the `cppdbg` debug type which is an actual thing.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -79,59 +79,94 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
-                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "true" });
+                Telemetry.logDebuggerEvent("buildAndDebug", { "type": "noLaunchConfig", "success": "true" });
             } catch (e) {
-                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
+                Telemetry.logDebuggerEvent("buildAndDebug", { "type": "noLaunchConfig", "success": "false" });
             }
         }
-        return [selection.configuration];
+
+        return Promise.resolve([selection.configuration]);
```
bobbrow: In an `async` function there is no need to wrap the result in `Promise.resolve`.  Just return the `[selection.configuration]`.

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: the name underlying is actually currently existing, I removed it in the next PR, in refactoring the code.

PR #8872, opened 2022-02-16
Ignore failing error code from VS Code

sean-mcmanus:
@michelleangela Does this seem okay or did you have any better ideas? I think it's more important that the test be passing by default so we can tell if there's a regression.


michelleangela:
If the error is only from VS Code's extension host and not from the test results, then it is okay to ignore errors from VS Code.

michelleangela (approved):

PR #8876, opened 2022-02-17
Dynamic debug config + Telemetry

elahehrashedi:
Modification to "Run and Debug" button in debug panel to make it consistent with our play button:

1. keeping the Debug info in memory and skip writing it in launch.json
2. if the launch.json exists try to read the debug configs to check if the user has modified any debug info

3. **Telemetry** is added for both `DebugPanel` and `playButton`.
4. the debug configurations are **resolved** for `playButton` as well.
5. the **recently used task** is shared between `DebugPanel` and `playButton`.

6. in refactoring, some code between two `DebugPanel` and `playButton` is reused so that they behave the same. 

7. Fixes are added for single-mode file debugging, however, it is only working for cl.exe builds and msvc debugger for now.

@jureid Telemetry will be as below:

there are two debuggerEvent: 

- "debugPanel" (the run and debug in debug panel)
- "launchPlayButton" (our play button)

both events will have these values:
```
{
"debugType": debug / run
"folderMode":  folder / singleFile
"config": noBuildConfig / noLaunchConfig / launchConfig
"success":  true/false
"cancelled": true
}
```
in config these are the 3 different modes (feel free to suggest more meaningful names):

- noBuildConfig: when the build config is not available.
- noLaunchConfig: when we don't have a launch config and we launch from memory (dynamic mode)
- launchConfig: when the user has modified the launch.json config and we are using that.

Feature Request: https://github.com/microsoft/vscode-cpptools/issues/8773


Extension/package.json
```diff
@@ -2763,7 +2762,6 @@
         "configurationAttributes": {
           "launch": {
             "type": "object",
-            "default": {},
```
bobbrow, line 2766: If updating the debug configuration, you should also update `tools/OptionsSchema.json`.  Remove the `default: {}` from "CppvsdbgLaunchOptions" and "CppdbgLaunchOptions" #Closed

Extension/src/Debugger/extension.ts
```diff
@@ -22,22 +23,24 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
...
+    const assetProvider: IConfigurationAssetProvider = ConfigurationAssetProviderFactory.getConfigurationProvider();
+
+    // Register DebugConfigurationProviders for "Run and Debug" in Debug Panel.
+    // On windows platforms, the cppvsdbg debugger will also be registered for initial configurations.
+    let cppVsDebugProvider: DebugConfigurationProvider | null = null;
     if (os.platform() === 'win32') {
-        cppVsDbgProvider = new CppVsDbgConfigurationProvider(configurationProvider);
-        disposables.push(vscode.debug.registerDebugConfigurationProvider('cppvsdbg', new QuickPickConfigurationProvider(cppVsDbgProvider)));
+        cppVsDebugProvider = new DebugConfigurationProvider(assetProvider, DebuggerType.cppvsdbg);
+        disposables.push(vscode.debug.registerDebugConfigurationProvider('cppvsdbg', cppVsDebugProvider, vscode.DebugConfigurationProviderTriggerKind.Dynamic));
```
bobbrow: Let's make the 'cppvsdbg' and 'cppdbg' strings into const variables. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
 
+    /**
+	 * Error checks the provided 'config' without any variables substituted.
+	 */
     resolveDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration> {
-        return this.underlyingProvider.resolveDebugConfiguration ? this.underlyingProvider.resolveDebugConfiguration(folder, config, token) : undefined;
+        if (!config || !config.type) {
+            // When DebugConfigurationProviderTriggerKind is Dynamic, this function will be called with an empty config.
+            // Providing debug configs, and debugging should be called manually.
+            this.provideDebugConfigurations(folder).then(async configs => {
```
bobbrow, line 118: This doesn't get returned, so the function returns `undefined` for this branch. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
 
         const selection: MenuItem | undefined = await vscode.window.showQuickPick(items, {placeHolder: localize("select.configuration", "Select a configuration")});
         if (!selection) {
-            throw Error(localize("debug.configuration.selection.canceled", "Debug configuration selection canceled")); // User canceled it.
+            // throw Error(localize("debug.configuration.selection.canceled", "Debug configuration selection canceled")); // User canceled it.
+            Telemetry.logDebuggerEvent(DebuggerEvent.debugPanel, { "debugType": "debug", "folderMode": folder ? "folder" : "singleMode", "cancelled": "true" });
```
bobbrow: Can we call it 'singleFile' instead of 'singleMode'? #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
 
         const selection: MenuItem | undefined = await vscode.window.showQuickPick(items, {placeHolder: localize("select.configuration", "Select a configuration")});
         if (!selection) {
-            throw Error(localize("debug.configuration.selection.canceled", "Debug configuration selection canceled")); // User canceled it.
+            // throw Error(localize("debug.configuration.selection.canceled", "Debug configuration selection canceled")); // User canceled it.
```
bobbrow: Please remove commented code #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
-                        vscode.window.showErrorMessage(localize("miDebuggerPath.not.available", "miDebuggerPath does not exist: {0}. Has a debugger been installed?", selection.configuration.miDebuggerPath));
-                        throw new Error();
-                    }
-                }
-                await vscode.debug.startDebugging(folder, selection.configuration);
-                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "true" });
-            } catch (e) {
-                Telemetry.logDebuggerEvent("buildAndDebug", { "success": "false" });
-            }
+        if (!this.isClAvailable(selection.label)) {
```
bobbrow: The return value is the same regardless of the result of this function. It sounds like you want the warning to print out in this case. Perhaps we should do it another way. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
-        return this.underlyingProvider.resolveDebugConfiguration ? this.underlyingProvider.resolveDebugConfiguration(folder, config, token) : undefined;
+        if (!config || !config.type) {
+            // When DebugConfigurationProviderTriggerKind is Dynamic, this function will be called with an empty config.
+            // Providing debug configs, and debugging should be called manually.
+            this.provideDebugConfigurations(folder).then(async configs => {
+                if (!configs || configs.length === 0) {
+                    Telemetry.logDebuggerEvent(DebuggerEvent.debugPanel, { "debugType": "debug", "folderMode": folder ? "folder" : "singleMode", "cancelled": "true" });
+                    return undefined; // aborts debugging silently
+                } else {
+                    await this.startDebugging(folder, configs[0], DebuggerEvent.debugPanel);
```
bobbrow, line 125: It might be good to log telemetry or `console.assert` that `configs.length === 1`.  Currently the provider function returns only 1 result, but if that ever changes, we could end up doing the wrong thing here.  At the very least, add a comment here that there's only 1 config expected in the array. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
+     * This is also ran after the tasks.json has completed.
+     *
+	 * Try to add all missing attributes to the debug configuration being launched.
+	 */
     resolveDebugConfigurationWithSubstitutedVariables(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration> {
-        return this.underlyingProvider.resolveDebugConfigurationWithSubstitutedVariables ? this.underlyingProvider.resolveDebugConfigurationWithSubstitutedVariables(folder, config, token) : undefined;
-    }
-}
+        if (!config || !config.type) {
+            return undefined;
```
bobbrow, line 161: Because we got bit by it, I think we should document that returning `null` here creates a launch.json and `undefined` is what we want. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +80,149 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
+
+        return config;
     }
 
-    /**
-	 * Returns a list of initial debug configurations based on contextual information, e.g. package.json or folder.
-	 */
-    async provideDebugConfigurations(folder?: vscode.WorkspaceFolder, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
-        const defaultConfig: vscode.DebugConfiguration = this.provider.getInitialConfigurations(this.type).find((config: any) =>
+    async provideDebugConfigurationsTypeSpecific(type: DebuggerType, folder?: vscode.WorkspaceFolder, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
```
bobbrow: rename suggestion: `provideDebugConfigurationsForType` #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -259,129 +350,36 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
-                delete config.externalConsole;
-            }
-
-            // Fail if cppvsdbg type is running on non-Windows
-            if (os.platform() !== 'win32') {
-                logger.getOutputChannelLogger().showWarningMessage(localize("debugger.not.available", "Debugger of type: '{0}' is only available on Windows. Use type: '{1}' on the current OS platform.", "cppvsdbg", "cppdbg"));
-                return undefined; // Stop debugging
-            }
+    private async loadDetectedTasks(): Promise<void> {
+        if (!DebugConfigurationProvider.detectedBuildTasks || DebugConfigurationProvider.detectedBuildTasks.length === 0) {
```
bobbrow, line 378: If the user installs a new compiler while VS Code is still open, we might need to refresh this list. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -259,129 +350,36 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
             });
         }));
         configs.push(defaultConfig);
+        // Sort tasks.
```
bobbrow: is this a TODO? #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -259,129 +350,36 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
-                            const helpURL: string = "https://aka.ms/vscode-cpptools/LLDBFrameworkNotFound";
-                            vscode.env.openExternal(vscode.Uri.parse(helpURL));
-                        }
-                    });
+    private buildAndDebugActiveFileStr(): string {
+        return `${localize("build.and.debug.active.file", 'Build and debug active file')}`;
+    }
 
-                return undefined;
+    private isClAvailable(configurationLabel: string): boolean {
```
bobbrow: This also returns true when cl is not in the selected configuration. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -505,17 +503,108 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
+
+        const folder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(textEditor.document.uri);
+        if (!util.isCppOrCFile(textEditor.document.uri)) {
+            vscode.window.showErrorMessage(localize("cannot.build.non.cpp", 'Cannot build and debug because the active file is not a C or C++ source file.'));
+            return;
+        }
+
+        // Get debug configurations for all debugger types.
+        let configs: vscode.DebugConfiguration[] = [];
+        if (os.platform() === 'win32') {
```
bobbrow, line 544: If the cppvsdbg provider can return `[]` for the non-win32 case, you may not need this check here.  And you can initialize `configs` with this value instead of `[]` #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -505,17 +503,108 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
-}
 
-export class CppDbgConfigurationProvider extends CppConfigurationProvider {
-    public constructor(provider: IConfigurationAssetProvider) {
-        super(provider, DebuggerType.cppdbg);
+    private async startDebugging(folder: vscode.WorkspaceFolder | undefined, configuration: vscode.DebugConfiguration, debuggerEvent: string, debugModeOn: boolean = true): Promise<void> {
+
+        const debugType: string = debugModeOn ? "debug" : "run";
+        const folderMode: string = folder ? "folder" : "singleMode";
+        if (configuration.preLaunchTask) {
```
bobbrow, line 601: I think it makes sense to separate this out from `startDebugging`.  It can either be its own method, or put it back in `buildAndDebug` #Closed

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: it doesn't work if we return from here. This is the error:
image

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -71,67 +94,154 @@ export class QuickPickConfigurationProvider implements vscode.DebugConfiguration
...
-            }
+
+        if (this.isClConfiguration(selection.label)) {
+            this.showErrorClNotAvailable(selection.label);
         }
+
         return [selection.configuration];
     }
 
+    /**
```
sean-mcmanus, line 108: The tabs/spacing are off?

bobbrow:
>           "${workspaceFolder}"

why is ${workspaceFolder} removed?

---
Refers to: Extension/package.json:261 in 513df2b. [](commit_id = 513df2b52ede5be769a042a0a5c113471a828ac1, deletion_comment = True)

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -262,126 +372,37 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
-                            const helpURL: string = "https://aka.ms/vscode-cpptools/LLDBFrameworkNotFound";
-                            vscode.env.openExternal(vscode.Uri.parse(helpURL));
-                        }
-                    });
+    private isClConfiguration(configurationLabel: string): boolean {
+        return configurationLabel.startsWith("C/C++: cl.exe") ? true : false;
+    }
 
-                return undefined;
+    private showErrorClNotAvailable(configurationLabel: string): boolean {
```
bobbrow: Can you add "if" to the function name?  `showErrorIfClNotAvailable`.  In the place where you call it, it is not clear that this is a conditional error, so the code can be confusing. #Closed

Extension/src/Debugger/configurationProvider.ts
bobbrow: Is that because we explicitly call `this.startDebugging` in the `else` case?  Should we be calling that here?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -262,126 +372,37 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
 
-                vscode.window.showErrorMessage(LLDBFrameworkMissingMessage, moreInfoButton)
-                    .then(value => {
-                        if (value === moreInfoButton) {
-                            const helpURL: string = "https://aka.ms/vscode-cpptools/LLDBFrameworkNotFound";
-                            vscode.env.openExternal(vscode.Uri.parse(helpURL));
-                        }
-                    });
+    private isClConfiguration(configurationLabel: string): boolean {
+        return configurationLabel.startsWith("C/C++: cl.exe") ? true : false;
```
bobbrow: I believe `startsWith` returns `true | false` so the extra `? true : false` should not be necessary. #Closed

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -262,126 +372,37 @@ export class CppConfigurationProvider implements vscode.DebugConfigurationProvid
...
+    }
 
-                return undefined;
+    private showErrorClNotAvailable(configurationLabel: string): boolean {
+        if (this.isClConfiguration(configurationLabel)) {
```
bobbrow: everywhere that I see the `showError`, there is already a call to `isClConfiguration`.

My recommendation to avoid the duplicate call is to just do it here, add "if" to the function name (mentioned in another comment), and return false instead of throwing an Error if the config is not for CL. #Closed

Extension/src/LanguageServer/extension.ts
```diff
@@ -30,6 +30,7 @@ nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFo
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 export const cppBuildTaskProvider: CppBuildTaskProvider = new CppBuildTaskProvider();
 export const CppSourceStr: string = "C/C++";
+export const CppSourceStrPrefix: string = "C/C++: ";
```
bobbrow: Is this the prefix of the task or launch config name?  I don't think "prefix" differentiates this much from the other string that doesn't have `: `.  I would call it `ConfigPrefix` or something like that.  (`Str` shouldn't really be in the name because we already know it's a string) #Closed

Extension/src/Debugger/configurationProvider.ts
bobbrow: This was not addressed yet.

Extension/package.json
```diff
@@ -258,8 +257,7 @@
         "source": "gcc",
         "owner": "cpptools",
         "fileLocation": [
-          "autoDetect",
-          "${workspaceFolder}"
+          "autoDetect"
```
sean-mcmanus: https://github.com/microsoft/vscode/blob/28ac2de7d0f2b5298957c0c42c3b06230e885726/src/vs/platform/markers/common/problemMatcher.ts -- I think maybe replace ${workspaceFolder} with ${cwd} since that's what VS Code seems to be doing. #Resolved

sean-mcmanus (dismissed):

bobbrow (approved):

Issue #8908, opened 2022-02-23
Add an error message if the wrong platform-specific VSIX is installed


Colengms:
Prior to switching to platform-specific VSIX's via the marketplace, we would display an error message if we detected that the wrong platform binaries were installed (if the user had manually downloaded and installed the wrong platform VSIX).  This error message should be added back in.  There is still the potential to download the wrong platform-specific VSIX from the marketplace web site.  There is also an issue in which VS Code may sometimes install the wrong VSIX: https://github.com/microsoft/vscode/issues/141696

There is a `TargetPlatform` field written to the `.vsixmanifest` file at the root of the extension directory (when installed from a VSIX), that could be used for this check.

PR #8910, opened 2022-02-23
Add compatibility check of VSIX TargetPlatform

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/8908

This should help catch issues with the wrong VSIX being deployed ( https://github.com/microsoft/vscode/issues/141696 ) before other wrong behaviors arise.

Adds an error dialog if the specified target is incompatible.  Adds a warning dialog if compatible but not matching (i.e. using x86 on x64 Windows).  Note that this check is done against the architecture of VS Code, so an x86 build of VS Code on x64 Windows will appear as x86.

Feedback on the message text would be welcome.

Uses regex to find the XML attribute, as Bob had suggested avoiding incurring use of an XML parser due to the frequency of security vulnerabilities in XML parser libraries.


Extension/src/LanguageServer/extension.ts
```diff
@@ -154,11 +155,86 @@ function sendActivationTelemetry(): void {
...
+                    isPlatformMatching = platform === "darwin" && arch === "arm64";
+                    // x64 binaries can also be run on arm64 macOS.
+                    isPlatformCompatible = platform === "darwin" && (arch === "x64" || arch === "arm64");
+                    break;
+                default:
+                    console.log("Unrecognized TargetPlatform in .vsixmanifest");
+                    break;
+            }
+            if (!isPlatformCompatible) {
+                vscode.window.showErrorMessage(localize("vsix.platform.incompatible", "The target platform {0} specifed in the C/C++ Extension VSIX is not compatible with your system.", vsixTargetPlatform));
```
bobbrow, line 222: If I were a VS Code user not working on the extension, I might not understand what this means for me.  Consider writing a message that says something along the lines of "It looks like you might have installed a version of the C/C++ extension that will not work with your OS. If this was a mistake, please uninstall and reinstall the C/C++ extension from the VS Code Marketplace to fix the problem."  You can also show them what we detected vs. the OS they have.

We might want to give them a way to suppress the message as well (in case we're wrong).

Extension/src/common.ts
```diff
@@ -1276,3 +1276,11 @@ export function isVsCodeInsiders(): boolean {
         extensionPath.includes(".vscode-exploration") ||
         extensionPath.includes(".vscode-server-exploration");
 }
+
+export function isAlpine(): boolean {
+    const output: string = child_process.spawnSync('ldd', ['--version']).stderr.toString();
```
bobbrow: We should already have the information we need in the LinuxDistribution class (`distroName` field).  According to [this](https://linuxhandbook.com/check-linux-version/#:~:text=The%20os-release%20file%20in%20the%20%2Fetc%20directory%20keeps,for%20Alpine%20Linux%20server%20running%20on%20Linode%20infrastructure.) the ID will be 'alpine'.

bobbrow: I think it would be safer to use that than this approach.

Extension/src/LanguageServer/extension.ts
Colengms: One of the issues we're seeing is that VS Code may install the wrong VSIX.  The wording should probably not suggest the issue is because they installed the VSIX themselves, or necessarily suggest that installing from the marketplace will fix the problem.

Colengms: Also, uninstall and reinstall is insufficient.  :(  VS Code caches installed extension directories, so reinstall winds up using the exact same one.  In this scenario, it would be necessary to instruct the user to delete the extension's installed directory before reinstalling.

There would seem to be a lot of information to try to pack into this message.  Perhaps it should link to a web page?

bobbrow: I was thinking that by the time they had run the extension and gotten the message, the Marketplace would have validated all the VSIXes and they'd get the right one next time.  But yeah, that's too much information for a popup.  We could add an explanation page to the docs folder and link to that.  Maybe put a button on the popup that navigates to that documentation.

Colengms: One problem with linking users to one of our docs via a URL is lack of localization. ...  I could include an md file with the extension, hook it up for localization, and the button/link could open it in VS Code.

Extension/src/LanguageServer/extension.ts
```diff
@@ -155,11 +156,85 @@ function sendActivationTelemetry(): void {
     telemetry.logLanguageServerEvent("Activate", activateEvent);
 }
 
+async function checkVsixCompatibility(): Promise<void> {
+    // Check to ensure the correct platform-specific VSIX was installed.
+    const vsixManifestPath: string = path.join(util.extensionPath, ".vsixmanifest");
+    // Skip the check if the file does not exist, such as when debugging cpptools.
+    if (await util.checkFileExists(vsixManifestPath)) {
```
sean-mcmanus, line 164: Can you early out instead?

Extension/src/LanguageServer/extension.ts
```diff
@@ -155,11 +156,85 @@ function sendActivationTelemetry(): void {
...
+        const content: string = await util.readFileText(vsixManifestPath);
+        const matches: RegExpMatchArray | null = content.match(/TargetPlatform="(?<platform>[^"]*)"/);
+        if (matches && matches.length > 0 && matches.groups) {
```
sean-mcmanus, line 167: Can you early out here too? It makes it more readable. Less blocks/indentation.

Extension/src/LanguageServer/extension.ts
```diff
@@ -155,11 +156,85 @@ function sendActivationTelemetry(): void {
...
+            const vsixTargetPlatform: string = matches.groups['platform'];
+            const platformInfo: PlatformInformation = await PlatformInformation.GetPlatformInformation();
+            let isPlatformCompatible: boolean = true;
+            let isPlatformMatching: boolean = true;
+            switch (vsixTargetPlatform) {
+                case "win32-x64":
+                    isPlatformMatching = platformInfo.platform === "win32" && platformInfo.architecture === "x64";
+                    // x64 binaries can also be run on arm64 Windows 11.
```
sean-mcmanus, line 175: Is this specific to Win11 and not Win10?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
Colengms: I would like to keep this as is.  I have changes to push after 1.9.1 to maintain a persistent value indicating whether or not to display the warning.  (The user can click an ignore button, if they intentionally want to use x86 on x64).  That persistent value is reset any time the scenario changes, so would need to be reset when these IFs are skipped.  That is cleaner to do when not doing an early exit.

Extension/src/LanguageServer/extension.ts
Colengms: See my reply to the other comment.

bobbrow (approved):

Extension/src/LanguageServer/extension.ts
bobbrow: Ok. Let's do that in the next PR.

sean-mcmanus (approved):

PR #8920, opened 2022-02-23
Add localized MD/HTML file for guidance when incompatible VSIX

Colengms:
Add localized MD/HTML file for guidance when incompatible/mismatching VSIX is detected.  This MD file is actually HTML, to enable localization.

Fixes an issue in which a redundant English version of each localized HTML file was included in the VSIX.

Adds a persistent var to suppress the warning when mismatched but compatible VSIX is detected.

I'm open to discussion on what to use as error/warning text and MD/HTML document contents.


Extension/Reinstalling the Extension.md
```diff
@@ -0,0 +1,32 @@
...
+
+<p data-loc-id="reinstall.extension.text5">On Windows:</p>
+<pre><code class="lang-bash">%USERPROFILE%\.vscode\extensions\ms-vscode.cpptools-1.9.0</code></pre>
+
+<p data-loc-id="reinstall.extension.text6">On Linux:</p>
+<pre><code class="lang-bash">$HOME/.vscode/extensions/ms-vscode.cpptools-1.9.0</code></pre>
+
+<p data-loc-id="reinstall.extension.text7">Then reinstall via the marketplace UI in VS Code.</p>
+
+<p data-loc-id="reinstall.extension.text8">If the correct version of the extension fails to be deployed by VS Code, the correct VSIX for your system can be <a href="https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools" data-loc-id="download.vsix.link.title">downloaded from the VS Code marketplace web site</a> and installed using the `Install from VSIX...` option in under the '...' menu in the marketplace UI in VS Code.</p>
```
sean-mcmanus: "in under"?

sean-mcmanus (changes requested):

sean-mcmanus:
@Colengms Was the vsix supposed to be built with this?

bobbrow:
> @Colengms Was the vsix supposed to be built with this?

No, this is for 1.9.2

Extension/src/LanguageServer/extension.ts
```diff
@@ -158,6 +158,9 @@ function sendActivationTelemetry(): void {
 }
 
 async function checkVsixCompatibility(): Promise<void> {
+    const ignoreMismatchedCompatibleVsix: PersistentState<boolean> = new PersistentState<boolean>("CPP.ignoreMismatchedCompatibleVsix", false);
```
bobbrow: Do we have a facility to make this value persist only for the current installation?  I would want it to reappear the next time they install the extension (if it's wrong). Maybe we could append the version number to the state key which would be close enough?

sean-mcmanus: We could reset the persistent state after a non-mismatch install, right?

Colengms: I made the persistent var specific to the current version number.

@sean-mcmanus Currently, we will reset the persistent var such that 'it does not ignore' mismatches (so reports a warning if there is a mismatch - somewhat of a double-negative) unless there is a mismatch detected while already set to ignore it...  So, as soon as we see a non-mismatching (er, matching) install, future mismatches will trigger the warning.

sean-mcmanus (changes requested):
occuring ->occurring

sean-mcmanus (approved):

Issue #8900, opened 2022-02-22
.editorconfig with root=true causes format-document to fail


barcharcraz:
I've got both a `.clang-format` and a `.editorconfig` file in one of my projects, and if `root = true` is set in the editorconfig file cpptools can't format any documents in my project. For "format document" it fails silently, for `Format document with" I get a popup with content:

`Command 'Format Document With...' resulted in an error (_a.toLowerCase is not a function)`

removing `root = true` from my .editorconfig or renaming the .editorconfig file resolves the issue.

This happens on linux (fedora 35) and windows 10, at least.

Colengms:
Hi @barcharcraz .  When `C_Cpp.formatting` is set to `Default`, it will attempt to use a `.clang-format` file or `.editorconfig` file (if the `.editorconfig` file contains any `c_cpp` formatting properties) that is relative/nearest to the file being formatted.

I'm not able to easily reproduce this issue.  Could you provide a simple repro example in the form of a zip of the folder, or a github repo?

barcharcraz:
[code_bug.zip](https://github.com/microsoft/vscode-cpptools/files/8128174/code_bug.zip)

Attached is a folder that demonstrates the issue. 

Open Folder
Open `main.c`
Ctrl-p->Format Document With...->C/C++
Error

Colengms:
Hi @barcharcraz .  Thanks, that helps a lot.  With your repro, I was able to pinpoint the problem and fix it pretty quickly.  Until we can get a release out with the fix included, you should be able to work around the issue by specifying "TRUE" or "FALSE" (in upper case) for the `root` value in your `.editorConfig` file.  That should be compatible with other tools, as the specification [here](https://editorconfig.org/) mentions:

> Currently all properties and values are case-insensitive.

Colengms:
I believe this should be fixed with 1.9.1, so closing this as fixed.  If you still repro an issue, let us know, and we can reopen the issue.

PR #8922, opened 2022-02-24
Fix a formatting bug

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/8900

The values in `.editorConfig` files are case insensitive.  However, if a lower case `true` or `false` is specified for a boolean property, it's treated as a boolean type (so no string functions are available on it) even as a property of an 'any' object.

This change handles the potential for `root` to be either of string or boolean type.

### DIFF
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -712,8 +712,16 @@ export class CppSettings extends Settings {
                         return true;
                     }
                 }
-                if (editorConfigSettings.root?.toLowerCase() === "true") {
-                    return true;
+                if (editorConfigSettings.root !== undefined) {
+                    if (typeof editorConfigSettings.root === "boolean") {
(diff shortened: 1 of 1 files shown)


michelleangela (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/settings.ts
```diff
@@ -712,8 +712,16 @@ export class CppSettings extends Settings {
                         return true;
                     }
                 }
-                if (editorConfigSettings.root?.toLowerCase() === "true") {
-                    return true;
+                if (editorConfigSettings.root !== undefined) {
+                    if (typeof editorConfigSettings.root === "boolean") {
+                        if (editorConfigSettings.root) {
```
bobbrow, line 717: If the type of this property is boolean, should we just return it instead of adding an `if` here?

PR #8947, opened 2022-02-28
Undo commits.

sean-mcmanus:


Extension/src/LanguageServer/settings.ts
```diff
@@ -712,29 +712,21 @@ export class CppSettings extends Settings {
...
-            } else {
-                const clangFormatPath1: string = path.join(parentPath, ".clang-format");
-                if (fs.existsSync(clangFormatPath1)) {
-                    return true;
-                } else {
-                    const clangFormatPath2: string = path.join(parentPath, "_clang-format");
-                    if (fs.existsSync(clangFormatPath2)) {
-                        return true;
-                    }
+                switch (typeof editorConfigSettings.root) {
```
bobbrow: ~~I don't think this should be backed out.  This wasn't part of @elahehrashedi's change.~~

bobbrow: ~~Just `configurationProvider.ts`~~

bobbrow: Nevermind. it looks like there was some other commit that changed this.

Issue #8930, opened 2022-02-24
Read the content of existing launch.json


elahehrashedi:
Currently, if the user changes the name of a debug configuration in launch.json, and plays the "Run and Debug" button, the config won't show in the list of available configs. (however, this renamed config will be shown in the list of available configs in "Run and Debug" panel)

sean-mcmanus:
@elahehrashedi Is this for 1.9?

elahehrashedi:
@sean-mcmanus yes

PR #8949, opened 2022-02-28
add launch config

elahehrashedi:
feature enhancement: https://github.com/microsoft/vscode-cpptools/issues/8930


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -545,6 +554,36 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                type: config.type,
+                request: config.request,
+                detail: config.detail ? config.detail : localize("pre.Launch.Task", "preLaunchTask: {0}", config.preLaunchTask),
+                preLaunchTask: config.preLaunchTask,
+                existing: TaskConfigStatus.configured
+            }));
+
+            // Remove the detected configs that are already configured once in launch.json.
+            const dedupExistingConfigs: vscode.DebugConfiguration[] = configs.filter(detectedConfig => {
+                let isAlreadyConfigured: boolean = false;
```
bobbrow: You don't need a temporary variable for this.  Just return `false` in the for loop or `true` at the end of the function.

But an even simpler, more JavaScript-y solution would be to do something like this for the filter function:
```
return !existingConfigs?.some(config => config.name === detectedConfig.name && ...);
```

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -257,16 +257,9 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
-            for (const taskJson of configuredBuildTasks) {
-                if ((taskDetected.definition.label as string) === (taskJson.definition.label as string)) {
-                    isAlreadyConfigured = true;
-                    break;
-                }
-            }
-            return !isAlreadyConfigured;
-        });
+        const dedupDetectedBuildTasks: CppBuildTask[] = DebugConfigurationProvider.detectedBuildTasks.filter(taskDetected =>
+            (configuredBuildTasks ? !configuredBuildTasks.some(taskJson => (taskJson.definition.label === taskDetected.definition.label)) : true));
```
bobbrow: you should be able to simplify this to not need the `? : true` by just returning `!configuredBuildTasks?.some(...)` instead. (`?.` replaces the `.`)

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -545,6 +547,25 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                detail: config.detail ? config.detail : localize("pre.Launch.Task", "preLaunchTask: {0}", config.preLaunchTask),
+                preLaunchTask: config.preLaunchTask,
+                existing: TaskConfigStatus.configured
+            }));
+
+            const areEqual = (config1: vscode.DebugConfiguration, config2: vscode.DebugConfiguration): boolean =>
+                (config1.name === config2.name && config1.preLaunchTask === config2.preLaunchTask
+                && config1.type === config2.type && config1.request === config2.request);
+            // Remove the detected configs that are already configured once in launch.json.
+            const dedupExistingConfigs: vscode.DebugConfiguration[] = configs.filter(detectedConfig => existingConfigs ? !existingConfigs.some(config => areEqual(config, detectedConfig)) : true);
```
bobbrow: you should be able to simplify this to not need the `? : true` by just returning `!existingConfigs?.some(...)` instead. (`?.` replaces the `.`)

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -526,6 +519,15 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         }
     }
 
+    public async getLaunchConfigs(folder: vscode.WorkspaceFolder | undefined): Promise<vscode.WorkspaceConfiguration[] | undefined> {
+        let configs: vscode.WorkspaceConfiguration[] | undefined = vscode.workspace.getConfiguration('launch', folder).get('configurations');
+        if (!configs) {
+            return undefined;
+        }
+        configs = configs.filter(config => (config.name && config.type === DebuggerType.cppvsdbg || config.type === DebuggerType.cppdbg) && config.request === "launch");
```
michelleangela:
```
 (config.name && config.type === DebuggerType.cppvsdbg || config.type === DebuggerType.cppdbg) && config.request === "launch"
```

The parenthesis `( )` may need clean up. Is this filter to check for:

- config name exits AND
- config request is launch AND
- `(`config type is either cppvsdg OR cppdbg`)`  ?

elahehrashedi: right, I cleaned it up

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: this `existingConfigs `could be undefined, I added an if statement and then simplified this.

sean-mcmanus (approved):

michelleangela (approved):

Extension/src/Debugger/configurationProvider.ts
bobbrow: > this `existingConfigs `could be undefined, I added an if statement and then simplified this.

Yes, that is what the `?.` does.  If the variable is undefined, it will return undefined instead of trying to call the `some` function.  When you add the `!` in front of it, it turns into `true`, so that case is handled.

bobbrow: What you've done with the if statement also works.

PR #8969, opened 2022-03-04
Change to run detached.

sean-mcmanus:
This is required in order for cpptools to be guaranteed to shutdown cleanly (all destructors run and threads exited), because it may be blocked on cpptools-srv updating. In particular, temp files might not be properly deleted.

This change requires cpptools-side changes as well.


bobbrow:
Why would we be blocked on cpptools-srv updating?  We should terminate all of those processes immediately since they don't save any state that we care about (I thought we did this already).  There is no need to wait for them to finish.

sean-mcmanus:
> Why would we be blocked on cpptools-srv updating? We should terminate all of those processes immediately since they don't save any state that we care about (I thought we did this already). There is no need to wait for them to finish.

I'm guessing it could lead to a corrupt ipch file, right?

Colengms (approved):

Issue #8963, opened 2022-03-02
Inconstent handling of shell escaping in compiler args (i.e. Between Windows and Linux)


Colengms:
Currently, when executing a process (i.e. a compiler query) on Windows, shell escaping in args is supported (due to how the process is invoked using a full command line).  On Linux/Mac, the same scenario does not support shell escaping (as args are passed as an array when processes are invoked via API).

A good solution may be to optionally support shell escaping on both Windows and Linux/Mac based on a bool arg, and to audit the code paths involving command lines and compiler arg arrays to be explicit about where shell escaping is supported/unsupported or expected/unexpected.

Other issues involving shell escaping:

https://github.com/microsoft/vscode-cpptools/issues/8962
https://github.com/microsoft/vscode-cpptools/issues/8649
https://github.com/microsoft/vscode-cpptools/issues/6773

All of these should probably be addressed at the same time to avoid fixes causing regressions in other scenarios.

Colengms:
Apparently, most of the 'shell' processing on Windows is actually done by the executable being invoked.  Some good information on how that process works is here: https://daviddeley.com/autohotkey/parameters/parameters.htm#WINPASS

Unfortunately, that complicates things.  Technically, we don't know how to remove shell escaping from a command line and turn it into an array of arguments on Windows, without knowing how the executable was built.  Though, using the MS C/C++ parameter passing rules would probably work for most (if not all) compilers.

Issue #8649, opened 2022-01-10
`arguments` field in `compile_commands.json` should be interpreted as not including shell escaping (?)


rtbo:
Bug type: Language Service

## Describe the bug
- OS and Version: **Linux 5.15.13-arch1-1**
- VS Code Version: **1.63.2**
- C/C++ Extension Version: **1.7.1**

When `compile_commands.json` have argument to define a string literal from the command line such as `"-DIDENT=\"Some string\""`, VS Code should parse this as `#define IDENT "Some string"` but instead omits the quotes and parse as `#define IDENT Some string`.

A real world case is when using gettext in a large application with multiple modules, there is the following macro which expects a `PACKAGE` definition, which is typically given by the build system on the command line.
```cpp
#define _(Text)  dgettext (PACKAGE,Text)
```
So for every internationalized string in the code, the bug is hit.
(BTW the only problem I have with this extension which works otherwise incredibly well with `compile_commands.json` :+1:)

## Steps to reproduce

**main.cpp**
```cpp
#include <iostream>

int main() {
    std::cout << HELLO_BUG << std::endl;
}
```

**compile_commands.json**
```json
[
    {
        "arguments": [
            "/usr/bin/g++",
            "-DHELLO_BUG=\"Hello, Bug!\"",
            "main.cc",
            "-o",
            "test"
        ],
        "directory": "/home/remi/dev/test/repro_define_bug",
        "file": "main.cpp"
    }
]
```

**.vscode/c_cpp_properties.json**
```json
{
    "configurations": [
        {
            "name": "Linux",
            "compilerPath": "/usr/bin/g++",
            "compileCommands": "${workspaceFolder}/compile_commands.json",
            "intelliSenseMode": "linux-gcc-x64"
        }
    ],
    "version": 4
}
```

![image](https://user-images.githubusercontent.com/3889105/148842077-03ea17b2-f963-4221-8a9e-e0917f8d4a15.png)

## Logs

- Logs from running `C/C++: Log Diagnostics` from the VS Code command palette
```
-------- Diagnostics - 10/01/2022, 22:30:40
Version: 1.7.1
Current Configuration:
{
    "name": "Linux",
    "compilerPath": "/usr/bin/g++",
    "compileCommands": "${workspaceFolder}/compile_commands.json",
    "intelliSenseMode": "linux-gcc-x64",
    "compilerPathIsExplicit": true,
    "cStandardIsExplicit": false,
[...]

PR #8970, opened 2022-03-04
Address shell escaping issues

Colengms:
Addresses:
https://github.com/microsoft/vscode-cpptools/issues/8649
https://github.com/microsoft/vscode-cpptools/issues/8963
https://github.com/microsoft/vscode-cpptools/issues/6773

There is a corresponding PR for the native component.  That change adds the `cpptools_wordexp` executable.  Due to how we package VSIX's, leveraging the native `wordexp` API directly in TypeScript was problematic.  Instead, I built a small executable that calls the function and returns the result as JSON text.  It can be bundled like existing native binaries.

With this change, only command lines and command line fragments may contain shell variables and escaping. Any arrays of args (internally, or when specified as an array by the user) are assumed to not contain shell escaping. The matches the spec'ed behavior of compile_commands.json files. It also makes some sense, as a command line is a shell concept, which arrays of args (such as argv) are not.

Adds a `legacyCompilerArgsBehavior` setting that may help users dependent on the original behavior to transition to the new behavior. The old behavior was rather unusual on Linux, where some arguments were effectively assumed to contain escaping and other were not.

Updates to a new cpptools API.  Old providers get the original behavior.  Providers that use the new API version will be expected to pass compilerArgs with no shell escaping present.  A new field has been added for `compileFragments`.  Providers such as CMake Tools can use this to pass along existing fragments for cpptools to parse, so they don't need to their own code to parse them correctly.


Extension/src/main.ts
```diff
@@ -100,6 +100,7 @@ async function makeBinariesExecutable(): Promise<void> {
         const commonBinaries: string[] = [
             "./bin/cpptools",
             "./bin/cpptools-srv",
+            "./bin/cpptools_wordexp",
```
bobbrow: Can you use a hyphen in the binary name instead of an underscore for consistency with the other binaries we ship?

Extension/src/LanguageServer/client.ts
```diff
@@ -2758,28 +2773,40 @@ export class DefaultClient implements Client {
...
-            configurations: configurations,
+            configurations: [],
             currentConfiguration: this.configuration.CurrentConfigurationIndex,
             workspaceFolderUri: this.RootPath,
             isReady: true
         };
-        // Separate compiler path and args before sending to language client
-        params.configurations.forEach((c: configs.Configuration) => {
+        const settings: CppSettings = new CppSettings(this.RootUri);
+        // Clone each entry, as we make modifcations before sending it, and don't
```
sean-mcmanus: modifications

sean-mcmanus (approved):

Extension/package.nls.json
```diff
@@ -210,6 +210,8 @@
...
     "c_cpp.configuration.filesExcludeWhen.markdownDescription": { "message": "Additional check on the siblings of a matching file. Use `$(basename)` as variable for the matching file name.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.configuration.debugger.useBacktickCommandSubstitution.markdownDescription": { "message": "If `true`, debugger shell command substitution will use obsolete backtick ``(`)``.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.legacyCompilerArgsBehavior.markdownDescription": "Enable pre-v1.10.0 behavior for how shell escaping is handled in compiler arg settings. Shell escaping is no longer expected or supported by default in arg arrays after v1.9.7.",
```
michelleangela: Might be more clear to "starting in v1.10.0." instead of saying "after v1.9.7" so users will just look for the actual version when it changed instead of looking for the next version. Or does this mean including in v1.9.7 and after?

michelleangela (approved):

Issue #8974, opened 2022-03-05
Active file is reset with no custom configuration when reconfigured by `updateCustomConfigurations`


stroan:
Bug type: Language Service

- OS and Version: Windows 10 or 11 host w/ Ubuntu 20.04 in local devcontainer
- VS Code Version: 1.65.0
- C/C++ Extension Version: 1.8.4
- Other extensions you installed (and if the issue persists after disabling them): ms-vscode.cmake-tools
- OS of remote machine: Ubuntu 20.04 in devcontainer

I am having an issue where after using cmake-tools for a short period (changing target, regenerating build files, running and debugging, closing and opening the dev environment), vscode starts reporting the following on a random file:

```
#include errors detected based on information provided by the configurationProvider setting. Squiggles are disabled for this translation unit 
cannot open source file "X.h"
```

This was also reported here I think: https://github.com/microsoft/vscode-cmake-tools/issues/1621, but I believe the bug lies in cpptools. I've attached the extension debug logs from the point where the bug starts: [c_cpp.log.txt](https://github.com/microsoft/vscode-cpptools/files/8189639/c_cpp.log.txt)

Changing cmake settings triggers `updateCustomConfigurations` in the language server client, which calls `provideCustomConfiguration` for all tracked files. The logs show the custom configurations received for all the opened files, except one. Having debugged the ms-vscode.cmake-tools side the request never arrives for the configuration for the broken file. On the cpptools side, all the tracked files do have `provideCustomConfiguration` invoked against them. The file that's broken ends up erroring out on this case: https://github.com/microsoft/vscode-cpptools/blob/5f888eef3062dd2f0c053ab27067b154abd89394/Extension/src/LanguageServer/client.ts#L1974. The QueryTranslationUnitSourceResult reports no candidates, and so it never requests the custom configuration from the provider. I haven't found any series of actions that gets the editor out of this state.

**Steps to reproduce**
1. Launch the following in a devcontainer: https://github.com/stroan/cpptoolsbug
2. Open a few of the source files.
3. Cause updateCustomConfigurations to be triggered by cmake-tools a bunch. Build, regenerate cmake files. switch build target, switch variant, run the tests with the ctest button. disconnect from the container and reconnect. I've had this repro on multiple machines everytime, after a few minutes of poking.
4. Observe that one of the files starts reporting an error on the first include.

**Expected behavior**
It doesn't break

**Code sample and logs**
Code sample: https://github.com/stroan/cpptoolsbug
[...]

PR #8984, opened 2022-03-08
Fix issue with `updateCustomConfigurations` failing to refresh config of active file

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/8974
(Requires native side change as well)

This change adds a field to `QueryTranslationUnitSourceParams` to cause it to ignore existing custom configurations or TU's when requesting a source file path (potentially associated with a header) to request a custom configuration for.  (Normally, existance of either of those would imply it's unnecessary to request new custom configuration for the file).

This change works around a race condition in which clearing of custom configurations causes a reset of the active TU.  If that reset completes too quickly, the subsequent `didChangeCustomConfigurations` for that file gets ignored due to the TU already existing.

We will continue to clear all custom configurations in `updateCustomConfigurations`, not try to replace them, as it's possible the configuration provider may no longer provide configurations for some files.  The new param is currently only used to work around this issue.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -295,6 +295,7 @@ interface CompileCommandsPaths extends WorkspaceFolderParams {
 
 interface QueryTranslationUnitSourceParams extends WorkspaceFolderParams {
     uri: string;
+    ignoreExisting: boolean;
 }
 
 interface QueryTranslationUnitSourceResult {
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

sean-mcmanus:
Was there a reason for the different naming? ignoreExisting versus replaceExisting?

michelleangela (approved):

Colengms:
> Was there a reason for the different naming? ignoreExisting versus replaceExisting?

In the call to `provideCustomConfiguration`, the boolean indicates whether an existing TU or custom configuration should be replaced/updated with a new configuration, if available.  In the call to `QueryTranslationUnitSource`, the boolean indicates whether an existing TU or custom configuration should be ignored when determining the TU source file.  They bools are related, but their purposes are slightly different between the two functions.

PR #8989, opened 2022-03-08
make extension pack workspace kind

sandy081:
Make c++ extension pack workspace kind by adding dummy main and activation event

### DIFF
--- a/ExtensionPack/package.json
+++ b/ExtensionPack/package.json
@@ -16,9 +16,10 @@
     "categories": [
         "Extension Packs"
     ],
-    "extensionKind": [
-        "workspace"
+    "activationEvents": [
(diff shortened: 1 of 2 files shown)


ExtensionPack/package.json
```diff
@@ -16,9 +16,10 @@
     "categories": [
         "Extension Packs"
     ],
-    "extensionKind": [
-        "workspace"
+    "activationEvents": [
+        "onCommand:ms-vscode.cpptools-extension-pack.dummyCommand"
```
bobbrow, line 20: I would prefer not to have the word "dummy" in our repo. It could flag policheck at some point.  Just call it "command" or "absentCommand" or something.

bobbrow (changes requested):

bobbrow:
replacing with #8991

PR #8991, opened 2022-03-08
update extension pack to be hidden from web environment

bobbrow:
replaces #8989


ExtensionPack/CHANGELOG.md
```diff
@@ -1,25 +1,9 @@
 # C/C++ Extension Pack for Visual Studio Code Change Log
 
-## Version 1.1.0: November 2021
+## Version 1.2.0
 ### List of extensions:
 * C/C++
 * C/C++ Themes
-* CMake
```
bobbrow, line 7: `CMake` is transitively still included, but that's a detail of `CMake Tools` at this point. We may switch it to an alternate LSP extension.

sean-mcmanus (approved):

PR #8995, opened 2022-03-09

csigs:


Colengms:
There are a lot of know issues with loc strings that need to be manually resolved before committing.  Please don't commit this with those unaddressed.  Thanks!

Issue #9068, opened 2022-03-22
C_Cpp.loggingLevel set to "7" doesn't work with VS Code Insiders 1.66.0


sean-mcmanus:
It works with VS Code 1.65.2.

I've filed a VS Code bug at https://github.com/microsoft/vscode/issues/145772, but they might want us to make a change to fix it.

sean-mcmanus:
It no longer repros with today's Insiders.

PR #9069, opened 2022-03-22
1.9.7 changelog

sean-mcmanus:


michelleangela (approved):

sean-mcmanus:
I don't know yet if we want to add some change/fix for https://github.com/microsoft/vscode-cpptools/issues/9068 -- a regression caused by a VS Code Insiders change.

michelleangela:
> I don't know yet if we want to add some change/fix for #9068 -- a regression caused by a VS Code Insiders change. It would be a TypeScript change.

According to VS Code March 2022 plan, it looks like they're code complete. https://github.com/microsoft/vscode/issues/145549. So they may release insiders soon or early next month. 

If we're not releasing 1.10 until after the next VS Code release and the fix for this is low risk and quick, perhaps address it in 1.9.7.

sean-mcmanus:
> > I don't know yet if we want to add some change/fix for #9068 -- a regression caused by a VS Code Insiders change. It would be a TypeScript change.
> 
> According to VS Code March 2022 plan, it looks like they're code complete. [microsoft/vscode#145549](https://github.com/microsoft/vscode/issues/145549). So they may release insiders soon or early next month.
> 
> If we're not releasing 1.10 until after the next VS Code release and the fix for this is low risk and quick, perhaps address it in 1.9.7.

I don't think it would block 1.9.7 -- if we need users to use loggingLevel 7 or equivalent we could have them switch to 1.10.0.

Issue #8952, opened 2022-03-01
Codeanalysis with clang-tidy fails when run from VSCode but not from commandline


hriekehof:
Bug type: Language Service

**Describe the bug**
- OS and Version: macOS Monterery 12.2.1 (ARM) 
- VS Code Version: 1.64.2
- C/C++ Extension Version: 1.9.1 pre-release

I want to run code analysis with `clang-tidy` on my codebase for which i have a large `compile_commands.json`. Around 100MB with 12000 entries. Also the compile command itself is quite long. Many defined symbols mostly. If count the lines in the C++ Extension output with loglevel="Information" there are 547 lines.

For all files i have tested i get a `clang-diagnostic-error` when i run the `Run Code Analysis on Active File` command. The errors are not shown on the extension output so i ran the same command in the terminal which shows alot of errors.
```bash
/Users/hriekehof/.vscode/extensions/ms-vscode.cpptools-1.9.1/bin/../LLVM/bin/clang-tidy \
--checks="clang-analyzer-*" \
--header-filter=".*" \
--extra-arg=-ferror-limit=0 \
--quiet \
/Users/hriekehof/src/repo/shared/someFile.cpp \
-- \
-std=c++14 \
-xobjective-c++ \
-Wno-pragma-pack \
...
```
 Unfortunately i cannot share the full output since the project is closed source, but i can share the first error printed:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__hash_table:123:76: error: call to '__libcpp_clz' is ambiguous [clang-diagnostic-error]
    return __n < 2 ? __n : (size_t(1) << (numeric_limits<size_t>::digits - __libcpp_clz(__n-1))); 
```
The list of errors in the system headers goes on a while. To try around a little i installed a separate clang toolchain via `brew install llvm` and tried this version of `clang-tidy` instead of the one which comes with VSCode. 
```bash
/opt/homebrew/opt/llvm/bin/clang-tidy \
--checks="clang-analyzer-*" \
--header-filter=".*" \
--extra-arg=-ferror-limit=0 \
--quiet \
/Users/hriekehof/src/repo/shared/someFile.cpp \
-- \
-std=c++14 \
-xobjective-c++ \
-Wno-pragma-pack \
...
```
[...]

Issue #8740, opened 2022-01-29
Switch to using -p for clang-tidy when configured by compileCommands


devzeb:
Bug type: Language Service

**Describe the bug**
- OS and Version: Windows 10 Pro 20H2
- VS Code Version: 1.63.2
- C/C++ Extension Version: v1.8.1 Prerelease

In my project I'm using the [cmsis_compiler.h](https://github.com/ARM-software/CMSIS_5/blob/def6f800f95661eb3451d317f7d0dde504f6020d/CMSIS/Core_A/Include/cmsis_compiler.h) header which checks for the compiler used.

A minimal example of the problematic header looks like this:
```c++
//error_header.h
#ifndef TEST_HEADER_
#define TEST_HEADER_

#if defined ( __GNUC__ )
  //include some stuff
#else
  #error not compiling with gcc
#endif

#endif
```
I'm compiling with the GNU arm-none-eabi toolchain, so `__GNUC__` is defined.

When I include this header in a source file, some warnings, which were previously shown in the editor, are now not shown anymore. See the following two pictures:

`//test.cpp`
![grafik](https://user-images.githubusercontent.com/12809051/151653887-41e21f35-0e93-46ab-89fe-22fc24283678.png)
![grafik](https://user-images.githubusercontent.com/12809051/151653906-cd5a5c2a-600a-49d3-9085-f9537527dc16.png)

However, if i manually invoke clang-tidy from the command line (same arguments as set in setting.json), all warnings are shown like in the first picture:

```
clang-tidy -p D:\code\display\build d:\code\display\common\utilities\tools\src\test.cpp <

6 warnings generated.
D:/code/display/common/utilities/tools/src/test.cpp:4:9: warning: variable 'a' is not initialized [cppcoreguidelines-init-variables]
    int a;
        ^
          = 0
D:/code/display/common/utilities/tools/src/test.cpp:5:9: warning: Value stored to 'b' during its initialization is never read [clang-analyzer-deadcode.DeadStores]
    int b = a + 2;
        ^   ~~~~~
D:/code/display/common/utilities/tools/src/test.cpp:5:9: note: Value stored to 'b' during its initialization is never read
    int b = a + 2;
        ^   ~~~~~
D:/code/display/common/utilities/tools/src/test.cpp:5:9: warning: unused variable 'b' [clang-diagnostic-unused-variable]
    int b = a + 2;
        ^
[...]

PR #9104, opened 2022-03-28
Add clangTidy.useBuildPath setting

sean-mcmanus:
For https://github.com/microsoft/vscode-cpptools/issues/8740 , https://github.com/microsoft/vscode-cpptools/issues/8952  .


Extension/package.json
```diff
@@ -461,6 +461,12 @@
           "markdownDescription": "%c_cpp.configuration.codeAnalysis.clangTidy.args.markdownDescription%",
           "scope": "resource"
         },
+        "C_Cpp.codeAnalysis.clangTidy.useDashP": {
```
bobbrow: I'm having a hard time with the name.  Can it be `useBuildPath` (the full name of the argument) instead?  The description doesn't need to change.

bobbrow: Or maybe `useCompileCommandsPath` if you don't like `useBuildPath`?

sean-mcmanus: Sure, that is fine, which should I use?

bobbrow: Whichever you like better.  I don't think I have a strong preference.

sean-mcmanus: On potential downside of using "useBuildPath" is that there's a possible confusion if users don't read the description because it still reads/uses the build path if compileCommands is set, but it's read by IntelliSense code. But I'm not sure if that will actually lead to any real confusion.

Colengms (approved):

sean-mcmanus:
Don't check this in until the cpptools side is in...

PR #9129, opened 2022-04-02
Replace substr with substring.

sean-mcmanus:
I noticed VS Code marks substr as deperecated:
image

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1600,8 +1600,12 @@ export class CppProperties {
             // Get env text
             let envText: string = "";
             const envStart: number = curText.search(/\"env\"\s*:\s*\{/);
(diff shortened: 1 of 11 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1601,7 +1601,7 @@ export class CppProperties {
             let envText: string = "";
             const envStart: number = curText.search(/\"env\"\s*:\s*\{/);
             const envEnd: number = envStart === -1 ? -1 : curText.indexOf("},", envStart);
-            envText = curText.substr(envStart, envEnd);
+            envText = curText.substr(envStart, envEnd); // TODO: Replace with substring.
```
michelleangela: It looks like this code is checking if `env` exists. If it does, then get the offset for `env`. If not, then there's no need to do a substring on curText to get envText, so the substring statement can be placed in a conditional block.

michelleangela (approved):

PR #9158, opened 2022-04-08
Convert "general extension" and "feature request" templates from markdown to yaml

michelleangela:
This converts the "general extension" and "feature request" templates to yaml, which uses input fields.

Instructions for converting markdown to yaml:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#converting-a-markdown-issue-template-to-a-yaml-issue-form-template

Schema
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#markdown


.github/ISSUE_TEMPLATE/general-extension.yml
```diff
@@ -0,0 +1,58 @@
+name: Bug Report - General Extension
+description: Create a bug report for downloading, installing, or building the extension.
+body:
+- type: markdown
+  attributes:
+    value: |
+      ### Is there an existing issue for this?
+      Please search our [existing issues](https://github.com/microsoft/vscode-cpptools/issues) to see if an issue already exists for the bug you encountered.
+- type: textarea
```
WardenGnaw, line 9: Wouldn't this be better as multiple `input` instead of a `textarea`?

michelleangela: Thought about making these as individual `input`, but it will make the form look longer than it is and cumbersome. 

Example, there will be multiple large headings per each input. Whereas it would be easier to type and read all the environment info from `textarea`.
image
(image is from https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)

sean-mcmanus: Why are the other templates not also updated? Was that so we could get feedback first?

sean-mcmanus:
Why are the other templates not also updated? Was that so we could get feedback first?

michelleangela:
> Why are the other templates not also updated? Was that so we could get feedback first?

Yes, waiting on feedback on OneNote that was sent in the sync up team meeting.

sean-mcmanus (approved):

PR #9160, opened 2022-04-09
Handle "external" include directories

stravant:
MSVC supports "external" include directories in the form "-external:I"
rather than the conventional "-I", and some tools generate such includes
in compile_commands.json exports.

See: https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics

The current language server does not understand the external include
directories, so use the replacement mechanism to rewrite the external
include directories into conventional ones that it can understand.

### DIFF
--- a/Extension/bin/common.json
+++ b/Extension/bin/common.json
@@ -18,6 +18,10 @@
       "replace": "-I\n$1"
     },
     {
+      "match": "^[-/]external:I(.*)",
+      "replace": "-I\n$1"
+    },
+    {
       "match": "^/D(.*)",
       "replace": "-D$1"
     },


Colengms:
Hi @stravant .  Thanks for helping out!  

> The current language server does not understand the external include directories

Could you open an issue with a repro that fails when they are used?  There is support in the native binary for `-external:I  <path>` and `/external:I <path>`.  It should be processed identically to `-I`.  The native binary component of the extension is currently closed-source.

sean-mcmanus:
Yeah, so this change shouldn't be necessary, right? Do we have a repro for the issue?

Colengms:
Hi @stravant .  I'm going to close this PR for now.  If you are seeing a problem, please open an issue for us, and we'll be happy to take a look.

Issue #7506, opened 2021-05-10
Add support for `-std=c++2b` (C++23) and `-std=gnu++2b` GCC compiler args


Colengms:
GCC has added experimental support for `-std=c++2b` and `-std=gnu++2b` compiler args.

Opening this issue to track supporting these when seen passed to GCC, and adding a new cppStandard value for C++23.

https://gcc.gnu.org/projects/cxx-status.html

Colengms:
I'm not sure if there is anything we need to do here, until we have C++23 specific support.  We should already be falling back to C++20 if we see an unrecognized value in a `-std=` arg.

cjwijtmans:
Can we add C++23 now? What is EDG? Or at least support c++latest. I dislike getting linting errors while the compiler works fine. Although it seems to be doing fine now at the moment when i add c++23 in the cmake config globally, it did not like it when it was added as a property on a executable.

sean-mcmanus:
We have some C++23 support in our compiler now, so we could add the new language standards.

cjwijtmans:
I just realized operator<=> is c++20 and not c++23 so i am guessing it is seeing c++23 in the cmake generated projects and falling back to c++17 even though the C_cpp.default.cppStandard is set to c++20 in my user settings.json. Shouls i open a new issue to maybe fallback to cppStandard first and then extension default? Or am i wrong in this one.

cjwijtmans:
I appreciate it 🙏🏻. I hope its as simple as adding c++23 to the list.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

cjwijtmans:
I do not use `cpp_properties.json` i use `C_cpp.x` in user `settings.json`. i believe in more cross platform support for the project and relying on the user to set up the environment correctly on his own. Therefore `.vscode` is blacklisted in `git`. Configuration provider is set to`cmake`. This also means `cpp_properties.json` is not useful to me.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

sean-mcmanus:
This should be fixed with 1.10.2 (pre-release).

PR #9177, opened 2022-04-12
Add support for c++23 language standard arg

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/7506

There is a corresponding PR on the native side as well.


michelleangela (approved):

sean-mcmanus (changes requested):
This should be value="c++23", right?

Extension/ui/settings.html
```diff
@@ -598,12 +598,14 @@
         <div class="section-text" data-loc-id="cpp.standard.description">The version of the C++ language standard to use for IntelliSense. Note: GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense will emulate the equivalent C++ standard version.</div>
         <div>
             <select name="inputValue" id="cppStandard" class="select-default">
+                <option value="c++20">c++23</option>
```
michelleangela: What Sean said, didn't see `value` didn't get updated.

michelleangela (changes requested):

michelleangela (approved):

sean-mcmanus (approved):

sean-mcmanus (changes requested):
We need to update the MSC_VER stuff. See for example https://github.com/microsoft/vscode-cpptools/blob/main/Extension/bin/windows.msvc.x64.json#L7

sean-mcmanus (approved):

Issue #9083, opened 2022-03-24
Extension refuses to activate after update to 1.9.7


tomhughes:
Bug type: General

**Describe the bug**
- OS and Version: Fedora 35
- VS Code Version: code-1.65.2-1646927812.el7.x86_64
- C/C++ Extension Version: 1.9.7
- Other extensions you installed (and if the issue persists after disabling them): clangd
- A clear and concise description of what the bug is.

**To Reproduce**
*Please include code sample and `task.json` files.*
1. Try and start vscode on a C project
2. Notice that debugging is not available
3. Consult the console and find:

```
mainThreadExtensionService.ts:107 Activating extension 'ms-vscode.cpptools' failed: Do not activate the extension when IntelliSense is disabled..
```

**Expected behavior**

The extension continues to start as normal.

michelleangela:
@tomhughes 
Could you see if debugging will be activated when `C_Cpp.intelliSenseEngine` is set to true? There were recent changes to debugging that could have caused debugging to not be available when `C_Cpp.intelliSenseEngine` is set to false.

tomhughes:
Yes that allows cpptools to load, but then the clangd extension popups a warning that it is not compatible with handing Intellisense enabled (https://github.com/clangd/vscode-clangd/blob/78cfb5de1606e3130710e0a2691686d2c69fe6e8/src/extension.ts#L38).

FrankBlabu:
Might that be related to https://github.com/microsoft/vscode-cpptools/issues/9088 ?

michelleangela:
#9088 is an installation issue whereas this issue is about enabling the extension's debugger features while IntelliSense server is disabled.

junchao98:
I returned the cpptools version to 1.8.4 this problem is gone

sean-mcmanus:
This should be fixed with 1.10.2 (pre-release).

PR #9178, opened 2022-04-12
Fix exception when IntelliSenseEngine is Disabled

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/9083


Extension/src/LanguageServer/extension.ts
```diff
@@ -268,18 +268,18 @@ export async function activate(): Promise<void> {
...
-        sendActivationTelemetry();
-        const checkForConflictingExtensions: PersistentState<boolean> = new PersistentState<boolean>("CPP." + util.packageJson.version + ".checkForConflictingExtensions", true);
-        if (checkForConflictingExtensions.Value) {
-            checkForConflictingExtensions.Value = false;
-            const clangCommandAdapterActive: boolean = vscode.extensions.all.some((extension: vscode.Extension<any>, index: number, array: Readonly<vscode.Extension<any>[]>): boolean =>
-                extension.isActive && extension.id === "mitaki28.vscode-clang");
-            if (clangCommandAdapterActive) {
-                telemetry.logLanguageServerEvent("conflictingExtension");
-            }
+        return;
```
bobbrow: Should we move the cppbuild task provider code up above this?  (lines 314-322)

Extension/src/main.ts
```diff
@@ -62,20 +62,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
     }
+    const currentIntelliSenseEngineValue: string | undefined = settings.intelliSenseEngine;
+    disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
+        const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
+        if (!reloadMessageShown && settings.intelliSenseEngine !== currentIntelliSenseEngineValue) {
+            reloadMessageShown = true;
+            util.promptForReloadWindowDueToSettingsChange();
+        }
+    }));
     await LanguageServer.activate();
```
bobbrow: Maybe in the future we pull any non-language server stuff out of `LanguageServer.activate` and just don't call it if the intelliSenseEngine is disabled.  I think that's a bigger refactor than what we want to do now (pulling task provider out of LanguageServer).

Colengms: This didn't seem too big, so I went ahead and did so.

Extension/src/main.ts
```diff
@@ -58,27 +65,58 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
+                const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(config);
+                vscode.languages.setTextDocumentLanguage(doc, "jsonc");
             }
-        }));
+        }
     }
-    await LanguageServer.activate();
 
-    UpdateInsidersAccess();
+    taskProvider = vscode.tasks.registerTaskProvider(CppBuildTaskProvider.CppBuildScriptType, cppBuildTaskProvider);
```
bobbrow: cleanup: we have access to `disposables` here, so instead of tracking this in its own variable, we can add it to the `disposables` array.

bobbrow (approved):

PR #9202, opened 2022-04-19
Add code analysis code action fixes

sean-mcmanus:
image

There are 3 codeAnalysis.codeAction.show* options for configuring the available options -- we might want to change some of the defaults, not sure yet.

If users want fixes on a particular file or files, they can just run code analysis on those specific files.

References to "code analysis" in the code actions could possibly be changed to "clang-tidy".


bobbrow:
My preferred order of menu items.  I think it's better to have the impact of changes start with the narrowest change at the top  and go increasingly more broad.  The secondary sort (fix/hide/disable) goes from best outcome "fix" to worst outcome "disable the feature".  Documentation makes sense to be last.

Fix this ... problem
Fix all ... problems
Fix all code analysis problems
Hide all ... problems
Hide all code analysis problems
Disable all ... problems
Show documentation for ...

sean-mcmanus:
> My preferred order of menu items. I think it's better to have the impact of changes start with the narrowest change at the top and go increasingly more broad (in case of mis-clicks). The secondary sort (fix/hide/disable) goes from best outcome "fix" to worst outcome "disable the feature". Documentation makes sense to be last.
> 
> Fix this ... problem Fix all ... problems Fix all code analysis problems Hide all ... problems Hide all code analysis problems Disable all ... problems Show documentation for ...

Sure, I can do that. That's the order I had originally. The reason I had changed it to "all" first was that it seemed like those would be the most commonly used.

Currently the "fix all" is marked isPreferred -- did you want me to change the "Fix this" to be isPreferred or have no isPreferred?

You want "Remove" changed to "Hide"?

bobbrow:
> Currently the "fix all" is marked isPreferred

What is "isPreferred"?

> You want "Remove" changed to "Hide"?

Yes, when I read it, I was thinking that "remove" felt a little ambiguous (suppress vs disable vs hide).  It also got me thinking about the potential for us to add "suppressions" in the code as a potential fixit option in the future.

sean-mcmanus:
> > Currently the "fix all" is marked isPreferred
> 
> What is "isPreferred"?
> 
> > You want "Remove" changed to "Hide"?
> 
> Yes, when I read it, I was thinking that "remove" felt a little ambiguous (suppress vs disable vs hide). It also got me thinking about the potential for us to add "suppressions" in the code as a potential fixit option in the future.

Yeah, I have a TODO? in the code for "Suppress".

Extension/src/LanguageServer/client.ts
```diff
@@ -64,12 +64,76 @@ let debugChannel: vscode.OutputChannel;
...
+export function rangeEquals(range1: vscode.Range | Range, range2: vscode.Range | Range): boolean {
+    return range1.start.line === range2.start.line && range1.start.character === range2.start.character &&
+    range1.end.line === range2.end.line && range1.end.character === range2.end.character;
+}
+
+export function vscodeLocation(cpptoolsLocation: Location): vscode.Location {
+    return new vscode.Location(vscode.Uri.parse(cpptoolsLocation.uri), vscodeRange(cpptoolsLocation.range));
+}
+
+export function vscodeTextEdits(cpptoolsTextEdits: TextEdit[]): vscode.TextEdit[] {
```
Colengms: Could these functions be moved into common.ts?  Also, could they be named to reflect the action they perform?  Maybe something like `toVscodeTextEdits` or `makeVscodeTextEdits` ?

sean-mcmanus:
I don't think this is ready for code review yet -- I just added code reviewers in case anyone wanted to do a preview code review.
1. Still working on some bug fixes/testing, e.g. some data structures aren't being updated correctly after a fix is applied...not sure how much work is left yet.
2. Might improve the code organization....thinking about moving the code analysis stuff to a separate file. Probably will add more comments.

sean-mcmanus:
@bobbrow I re-worded "Hide" to "Clear" which is the original term we currently use with 1.9.7. That seems like the most easily understood term for a temporary removal and not a permanent suppression or disabling. I also enabled Disable by default and added a showClear enum option that defaults to "AllAndAllType" (not sure yet if that is the best default or not).

Extension/package.json
```diff
@@ -73,7 +73,7 @@
     "onCommand:C_Cpp.RunCodeAnalysisOnActiveFile",
     "onCommand:C_Cpp.RunCodeAnalysisOnOpenFiles",
     "onCommand:C_Cpp.RunCodeAnalysisOnAllFiles",
-    "onCommand:C_Cpp.ClearCodeAnalysisSquiggles",
+    "onCommand:C_Cpp.RemoveAllCodeAnalysisProblems",
```
sean-mcmanus, line 76: The external command is "Clear" but internally it's referred to as "Remove" -- doesn't seem worth renaming.

Extension/package.json
```diff
@@ -473,6 +503,7 @@
             "enum": [
               "*",
               "abseil-*",
+              "abseil-cleanup-ctad",
```
sean-mcmanus, line 506: These are new checks from clang 14...some might be missed ones from clang 13 too.

sean-mcmanus:
I'm thinking about potentially adding a warning message to the quick fix description text for fix types that are known to generate code that doesn't compile (i.e. garbage edits, which are still useful in most cases, but require manual editing to fix), either in this PR or a follow-up.

Extension/src/LanguageServer/client.ts
```diff
@@ -179,45 +266,328 @@ function publishDiagnostics(params: PublishDiagnosticsParams): void {
     clientCollection.timeTelemetryCollector.setUpdateRangeTime(realUri);
 }
 
-function publishCodeAnalysisDiagnostics(params: PublishDiagnosticsParams): void {
+// Rebuild codeAnalysisCodeToFixes and codeAnalysisAllFixes.fixAllCodeActions.
+function rebuildCodeAnalysisCodeAndAllFixes(): void {
```
Colengms, line 249: Maybe it would make sense to move some of this code into a new `codeAnalysis.ts` file?

Extension/src/LanguageServer/extension.ts
```diff
@@ -594,8 +600,55 @@ async function onRunCodeAnalysisOnAllFiles(): Promise<void> {
...
+async function onRemoveAllCodeAnalysisProblems(): Promise<void> {
+    getActiveClient().handleRemoveAllCodeAnalysisProblems();
+}
+
+async function onRemoveCodeAnalysisProblems(refreshSquigglesOnSave: boolean, identifiersAndUris: CodeAnalysisDiagnosticIdentifiersAndUri[]): Promise<void> {
+    getActiveClient().handleRemoveCodeAnalysisProblems(refreshSquigglesOnSave, identifiersAndUris);
+}
+
+// Needed due to https://github.com/microsoft/vscode/issues/148723 .
+const codeActionAbortedString: string = localize('code.action.aborted', "The code action fix was skipped because it was generated from an older document version.");
```
Colengms: Should this refer to a "code analysis fix" or a "quick fix"?  It's unclear to me what the ideal terminology would be.  I see both "code action" and "quick fix" referred to in the VS Code UI, but here the context seems to be specific code analysis.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: I'll check...

Extension/src/LanguageServer/client.ts
sean-mcmanus: Potentially -- but I didn't want to move existing code analysis code because it would make the PR hard to diff/review.

I'm definitely in favor of moving more stuff out of client.ts.

Should I do that in this PR or a follow up one?

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I'm not sure. I think "fix" is understandable (when preceded by "code action"), and I'm not sure how "code action quick fix" would be more understandable. I know VS Code has a "quick fix" type that the code action has. I could just use "quick fix" if we wanted the shortest amount of text.

Colengms: How about `code analysis fix`?  Also, the user might not be aware of the process of 'generating' code actions or of document versions.  How about "The fix could not be applied because the document has changed" ?

Extension/src/LanguageServer/client.ts
Colengms: A later PR seems fine.

sean-mcmanus: I had started refactoring to a codeAnalysis.ts, but there were too many changes involved so I abandoned that (since the diff would be harder to see).

michelleangela: Yeah, it would be easier to follow changes of a refactor in a separate PR. I would refactor first before adding/fixing features.

Extension/src/LanguageServer/extension.ts
michelleangela: "The code analysis fix could not be applied because the document has changed" is simple and clear. It would be helpful to add "code analysis" to let the user know what type of fix failed to apply.

sean-mcmanus: Mentioning "code analysis" (or "code action") seems redundant/obvious, because the user would have had to have just clicked the `"Fix this <clang-tidy type> message"`. However, I'm not opposed to being more explicit in the message.

Extension/package.nls.json
```diff
@@ -43,6 +43,13 @@
...
     "c_cpp.configuration.codeAnalysis.runAutomatically.markdownDescription": { "message":  "If `true`, code analysis will run automatically on a file after it is opened or saved.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showDisable.markdownDescription": { "message": "If `true`, the 'Disable' quick fix will be shown when available (the next time code analysis is run, if this setting was disabled). When used, it adds the warning code to the `C_Cpp.codeAnalysis.clangTidy.checks.disabled` setting.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
michelleangela: It's hard to understand the first sentence because there is "`Disable` quick fix" and "(the next time code analysis is run, if this setting was disabled)". To which does the first `Disable` refer to, could this be more clear if the actual setting ID was used instead of `Disable` quick fix?

For the second disable part, does this mean, the setting `c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showDisable` will be active the next time code analysis is run if `c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showDisable` was previously disabled?

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, that sounds good. If I find any high priority issues that should be fixed I can do them in PR separate from the refactor and deal with any potential merge conflicts, depending on how long the refactor takes, etc.

Extension/src/LanguageServer/extension.ts
michelleangela: I vote for explicit message just in case.

Extension/package.nls.json
sean-mcmanus: Yeah, I'm confused by it too. The option is for showing the "Disable ..." code action, but it requires that code analysis be re-run before the setting takes effect, and it would only take effect if the setting is set to true. And when invoked it'll modify the "*.disabled" setting mentioned. I'll re-word the description.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Sure

Extension/src/LanguageServer/client.ts
sean-mcmanus: Done

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Done

Extension/package.nls.json
sean-mcmanus: Done

michelleangela (approved):

Colengms (approved):

michelleangela (approved):

PR #9205, opened 2022-04-19
check for failed tasks

### LINKED ISSUE #8828: Debugging shouldn't continue if there's a build error
With the cppbuild task, add a build error and do F5 or the Build/Run button.

Ideally, it would not continue to the debug/run stages if a build error occurred. Not sure if that's possible.

In particular, the build error in the Terminal gets covered up when it switches to the Debug Console pane, so it's easy to miss.

### LINKED ISSUE #9060: launch occurs (running previous executable file) even when build finishes with errors
Steps to reproduce the problem:

1. open an empty folder and open it with vscode
2. write a C++ hello world program with an error
3. press F5 key to debug
4. VSCode automatically creates .vscode folder
5. Build finished with error(s), but launch still occurs

the cause of the problem probably lies at the "type" key in "tasks.json" file that VSCode automatically created. "type" should be either "process" or "shell", but VSCode sets it with initial value "cppbuild"

after changing value of "type" key from "cppbuild" to "process", launch fails as expected.

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/9060
bug fix: https://github.com/microsoft/vscode-cpptools/issues/8828

### DIFF
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -418,7 +418,7 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
             let error: string = "";
             let stdout: string = "";
             let stderr: string = "";
-            const result: number = await new Promise<number>(resolve => {
+            const spawnResult: number = await new Promise<number>(resolve => {
                 if (child) {
                     child.on('error', err => {
                         splitWriteEmitter(err.message);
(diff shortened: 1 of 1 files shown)


michelleangela (dismissed):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -418,7 +418,8 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
             let error: string = "";
             let stdout: string = "";
             let stderr: string = "";
-            const result: number = await new Promise<number>(resolve => {
+            let buildTerminated: boolean = false;
+            await new Promise<number>(resolve => {
```
bobbrow: If you no longer use the number returned by `resolve(-1)` or `resolve(0)`, then this should be a `Promise<void>`.  But I don't see anything significantly different\*\* in this PR from what you're already doing.

\*\*except... (see my next comment)

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -439,30 +440,38 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
                         this.writeEmitter.fire(this.endOfLine);
                         if (result === null) {
                             this.writeEmitter.fire(localize("build.run.terminated", "Build run was terminated.") + this.endOfLine);
+                            buildTerminated = true;
                             resolve(-1);
```
bobbrow, line 429: Is the real problem that `resolve(-1)` doesn't exit the function and we immediately send `resolve(0)` right after this?  What happens if you just add a `return;` after line 444?  Maybe the reason this change works is because it now ignores the -1 and 0 returned in the failed build case.

(It's possible you might also need to do something to deal with multiple `resolve`s being called if both the `error` block and the `close` block can execute.)

bobbrow (dismissed):
Requesting changes since this PR was approved already, but I'm not confident it's the right way to go.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: In new changes, I kept the result of the promise and checked for when it is -1.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -427,29 +427,37 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
...
             });
-            this.printBuildSummary(error, stdout, stderr);
+            let result: number = this.printBuildSummary(error, stdout, stderr);
+            if (spawnResult === -1) {
+                result = -1;
+            }
             this.closeEmitter.fire(result);
         } catch {
             this.closeEmitter.fire(-1);
+            return -1;
```
bobbrow: Remove this line.

bobbrow (approved):

Issue #8493, opened 2021-12-06
Use `/Fe<file>` (single arg) syntax when generating `tasks.json` to support older versions of `cl.exe`


toninlg:
Bug type: Language Service

**Describe the bug**
- OS and Version: Windows 10 20H2
- VS Code Version: 1.62.2
- C/C++ Extension Version: 1.7.1
- I installed Windows SDK 7.1 on Windows 10. I launch the developer command prompt, typing cl gives "Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86" and start code from the prompt. Whatever the project is single or multi projects, cl.exe is not detected.
- If I set the compilerpath manually in c_cpp_properties.json, the includepath is not set. If I manually set the includepath, the build failed in terminal: 
```
Starting build...
cl.exe /Zi /EHsc /nologo /Fe: C:\Home\Public\test\main.exe C:\Home\Public\test\main.c
cl : Command line warning D9024 : 
unrecognized source file type 'C:\Home\Public\test\main.exe', object file assumed


main.c

LINK : fatal error LNK1181: cannot open input file 'C:\Home\Public\test\main.exe'

Build finished successfully.
```

**Expected behavior**
Correct detection of cl.exe and paths. Successful build and coherent output message, why is the build finished successfully?

- Configurations in `c_cpp_properties.json` after manual settings of compilerpath and includepath
```
{
    "configurations": [
        {
            "name": "Win32",
            "includePath": [
                "${workspaceFolder}/**",
                "C:\\Program Files (x86)\\Microsoft Visual Studio 10.0\\VC\\include"
            ],
            "defines": [
                "_DEBUG",
                "UNICODE",
                "_UNICODE"
            ],
            "cStandard": "c89",
            "cppStandard": "c++98",
            "intelliSenseMode": "windows-msvc-x86",
            "compilerPath": "C:\\Program Files (x86)\\Microsoft Visual Studio 10.0\\VC\\bin\\cl.exe"
        }
    ],
    "version": 4
}
```

- generated tasks.json
```
{
			"type": "cppbuild",
			"label": "C/C++: cl.exe build active file",
			"command": "cl.exe",
			"args": [
				"/Zi",
				"/EHsc",
				"/nologo",
				"/Fe:",
				"${fileDirname}\\${fileBasenameNoExtension}.exe",
				"${file}"
			],
			"options": {
				"cwd": "${fileDirname}"
			},
[...]

PR #9211, opened 2022-04-20
fail the tasks, if there is a fatal link error

elahehrashedi:
patially mentioned in issue: https://github.com/microsoft/vscode-cpptools/issues/8493


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -445,7 +445,7 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
 
     private printBuildSummary(error: string, stdout: string, stderr: string): number {
         if (error || (!stdout && stderr && stderr.includes("error")) ||
-            (stdout && stdout.includes("error C"))) { // cl.exe compiler errors
+            (stdout && (stdout.includes("error C") || stdout.includes("LINK : fatal error")))) { // cl.exe compiler errors
```
bobbrow, line 452: I'm not sure we'll be able to reliably cover all possible failure strings (including localized compilers) here.  Fixing the Promise to return the result instead of 0 (which you have done) and checking it here would be the more robust solution.  Can you pass in the error code to this function and check it as well?  You may be able to remove the other string checks.

sean-mcmanus: This section of code is for cl.exe errors and cl.exe messages aren't localized, right? Your suggestion might still be a better approach, I haven't checked.

elahehrashedi: This checks for both the result value and the stdout string

bobbrow: > This checks for both the result value and the stdout string

I don't see it.  `printBuildSummary` only accepts an error string, stdout, and stderr.  It also needs to look at the value of `spawnResult` from line 409.

bobbrow (changes requested):
Please check the `spawnResult` in `printBuildSummary`

Extension/src/LanguageServer/cppBuildTaskProvider.ts
elahehrashedi: that is right, I only considered it in firing the results.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -433,17 +433,19 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
...
+            const result: number = this.printBuildSummary(error, stdout, stderr, spawnResult);
             this.closeEmitter.fire(result);
         } catch {
             this.closeEmitter.fire(-1);
         }
     }
 
-    private printBuildSummary(error: string, stdout: string, stderr: string): number {
+    private printBuildSummary(error: string, stdout: string, stderr: string, spawnResult: number): number {
+        if (spawnResult === -1) {
```
bobbrow: Failure is defined as any non-zero result.  line 431 could have returned `1` or `100` or any other number.  The only time it indicates success is when it is `0`.

bobbrow (changes requested):

bobbrow (approved):

Issue #9100, opened 2022-03-28
refactor "Run > Add Configuration" to generate launch.json file


samival:
Bug type: Debugger

**Describe the bug**
- OS and Version: Windows 10 / Linux Mint 19 / Ubuntu 20
- VS Code Version: 1.65.2
- C/C++ Extension Version: probably all version between 1.9.2 to 1.9.6 if not earlier
- Other extensions you installed (and if the issue persists after disabling them): No other extension installed.
- Compiler/Debugger: On windows: gcc and gdb using Mingw / On Linux: gcc and gdb.

- A clear and concise description of what the bug is:
Usually when you press F5 you get prompted with options for generating default tasks.json and launch.json
but with the recent updates, you no longer get launch.json file generated.

**To Reproduce**
*Please include a code sample and `launch.json` configuration.*
1. create an empty folder
2. open folder with VSCode.
3. create a simple .c file such as:
`int main(void) { return 0; }`
(or any valid C code of your choice)
4. press F5
5. choose "C++ GDB/LLDB"  when prompted
6. choose "C/C++ gcc ..." or "C/C++ gcc.exe ..." when prompted
7. result: tasks.json is generated but launch.json is not. (sometimes you get an error and sometimes it just runs/debugs somehow without a launch.json file)

Additionally it appears generating tasks.json using ctrl+alt+b combination is also broken:
1. follow steps 1-3 above
2. press Ctrl+Alt+B
3. choose "C/C++ gcc ..." or "C/C++ gcc.exe ..." when prompted
4. no tasks.json is generated no executable file is created. terminal output is something like:
```
> Executing task: C/C++: g++ build active file <

Starting build...
C:\usr\bin\g++ -fdiagnostics-color=always -g ...\hello.c -o ...\hello.exe
spawn C:\WINDOWS\system32\cmd.exe ENOENTBuild finished with error(s).
The terminal process failed to launch (exit code: -1).

Terminal will be reused by tasks, press any key to close it.
```

I've yet to find an old version of the C/C++ extension that works. I've tested on Two different computers and two virtual machines. On my Windows machine it worked for a while so I thought it was a problem specifically with Linux but than the extension on Windows got updated and I started getting the same behavior. To make sure it wasn't a problem with cache/config/temp files I've deleted them and even tested on a  zipped version with a portable setup  (with "./data/tmp" structure inside VSCode's folder). 

** possible workarounds **
[...]

PR #9226, opened 2022-04-22
add "add configuration" feature (ready)

elahehrashedi:
Find the feature in "Right Click" and the "gear button" on the editor menu.
feature request: https://github.com/microsoft/vscode-cpptools/issues/9100

image

image


bobbrow:
We talked about it but just wanted to post here as well.  We should not need a new icon for this because the configure option should not be "sticky" like the others.  I recommend just running the command and then leaving the last selection (run or debug) as the remembered action for the button.

Extension/package.json
```diff
@@ -2752,6 +2752,12 @@
         "title": "%c_cpp.command.BuildAndRunFile.title%",
         "category": "C/C++",
         "icon": "$(run)"
+      },
+      {
+        "command": "C_Cpp.AddConfiguration",
```
bobbrow: Can we call it AddDebugConfiguration to be more specific?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,102 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(launchJsonPath);
+        if (doc) {
+            vscode.window.showTextDocument(doc);
+        }
+    }
+
+    public async addConfiguration(textEditor: vscode.TextEditor): Promise<void> {
+        const folder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(textEditor.document.uri);
+        if (!folder) {
+            vscode.window.showWarningMessage(localize("add.configuration.not.available.for.single.file", "Add configuration is not available for single file."));
```
bobbrow: should we add it to the settings, or disable the ability to set this for single file?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,102 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            return;
+        }
+        // Write preLaunchTask into tasks.json file.
+        if (selectedConfig.preLaunchTask) {
+            await cppBuildTaskProvider.writeBuildTask(selectedConfig.preLaunchTask);
+        }
+
+        // Write debug configuraion in launch.json file.
+        await this.writeDebugConfig(selectedConfig, folder);
+
```
bobbrow, line 660: nit: delete empty line

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,102 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        if (this.isClConfiguration(selectedConfig.name) && this.showErrorIfClNotAvailable(selectedConfig.name)) {
+            return;
+        }
+
+        // Keep track of the entry point where the debug has been selected, for telemetry purposes.
+        selectedConfig.debuggerEvent = DebuggerEvent.launchPlayButton;
+        // startDebugging will trigger a call to resolveDebugConfiguration.
+        await vscode.debug.startDebugging(folder, selectedConfig, {noDebug: !debugModeOn});
+
```
bobbrow: nit: delete empty line

jureid:
> We talked about it but just wanted to post here as well. We should not need a new icon for this because the configure option should not be "sticky" like the others. I recommend just running the command and then leaving the last selection (run or debug) as the remembered action for the button.

+1

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,102 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         return false;
     }
 
+    private getLaunchJsonPath(): string | undefined {
+        return util.getJsonPath("launch.json");
+    }
+
+    public getRawLaunchJson(): Promise<any> {
+        const path: string | undefined = this.getLaunchJsonPath();
+        return util.getRawJson(path);
```
bobbrow, line 602: I believe this deletes any comments in the `launch.json` file.  I would think we should try to preserve the comments.  Preserving formatting would be nice too, but using the VSCode setting for tab size should be ok.

elahehrashedi:
@bobbrow  Here are some thoughts about the adding the `Add debug configuration` to the `Run and Debug` button:
This button is meant to keep the last selection, as far as I searched, there is no functionality to make one of the buttons not stick. 
If any other extension adds a command to this button, it will be shown in the drop-down list too. for example, as you can see the `Code Runner` extension has added the `Run Code` option to the menu too. In that sense, I think adding `Add debug configuration` to this button is not compatible with the other commands in this list.
image


1. One option is to add it as a separate button as below and then it won't be always in the way of debugging. the downside is that we may need to point it out to the users somehow.
image

2. the second option is to add it to the drop-down list. This is easier for the user to find it. the downside is that it will stick if it is chosen.
image

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: I will open an issue to track keeping the comments in .json files https://github.com/microsoft/vscode-cpptools/issues/9236. 
Vs Code removes the comments in tasks.json as well, https://github.com/microsoft/vscode/issues/29453

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: I disabled the feature from the single-mode file. 
The command is not available from the button and from right-click,
but if the user uses the command palette to run "C/C++: add debug configuration", they will see the warning.

Extension/package.json
```diff
@@ -2752,7 +2751,12 @@
         "title": "%c_cpp.command.BuildAndRunFile.title%",
         "category": "C/C++",
         "icon": "$(run)"
-      }
+      },
+      {
+        "command": "C_Cpp.AddDebugConfiguration",
+        "title": "%c_cpp.command.AddDebugConfiguration.title%",
+        "category": "C/C++",
+        "icon": "$(add)"      }
```
michelleangela: Place ending `}` in newline.

Extension/src/Debugger/configurationProvider.ts
michelleangela: Could you add a comment to the code here that references the VS Code issue 29453?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,100 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+    }
+
+    public async writeDebugConfig(config: vscode.DebugConfiguration, folder?: vscode.WorkspaceFolder): Promise<void> {
+        const settings: OtherSettings = new OtherSettings();
```
michelleangela: Move `const settings: OtherSettings = new OtherSettings();` to line 364 just before it is used?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -593,13 +593,100 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        await this.writeDebugConfig(selectedConfig, folder);
+
+    }
+
     public async buildAndRun(textEditor: vscode.TextEditor): Promise<void> {
         // Turn off the debug mode.
         return this.buildAndDebug(textEditor, false);
     }
 
     public async buildAndDebug(textEditor: vscode.TextEditor, debugModeOn: boolean = true): Promise<void> {
```
michelleangela, line 668: `addDebugConfiguration` and `buildAndDebug` have some duplicate code. Place duplicate code in helper function to reuse in both`addDebugConfiguration` and `buildAndDebug`?

Extension/src/Debugger/extension.ts
```diff
@@ -39,6 +41,13 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
     const debugProvider: DebugConfigurationProvider = new DebugConfigurationProvider(assetProvider, DebuggerType.all);
     disposables.push(vscode.commands.registerTextEditorCommand("C_Cpp.BuildAndDebugFile", async (textEditor: vscode.TextEditor, edit: vscode.TextEditorEdit, ...args: any[]) => { await debugProvider.buildAndDebug(textEditor); }));
     disposables.push(vscode.commands.registerTextEditorCommand("C_Cpp.BuildAndRunFile", async (textEditor: vscode.TextEditor, edit: vscode.TextEditorEdit, ...args: any[]) => { await debugProvider.buildAndRun(textEditor); }));
+    disposables.push(vscode.commands.registerTextEditorCommand("C_Cpp.AddDebugConfiguration", async (textEditor: vscode.TextEditor, edit: vscode.TextEditorEdit, ...args: any[]) => {
+        const folder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(textEditor.document.uri);
+        if (!folder) {
+            vscode.window.showWarningMessage(localize("add.configuration.not.available.for.single.file", "Add configuration is not available for single file."));
```
michelleangela: Update `"Add configuration is not available for single file."` to "Add debug configuration" to reflect UI name of command?

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: the telemetry part was not duplicate so I kept it there, but the checkng for ClConfiguration was duplicate and I moved it to the higher function.

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: @Colengc currently we are not preserving the comments, right?

michelleangela:
Will wait to sign off after final decision for button/command is made.

michelleangela:
Will wait to sign off after final decision for button/command is made.

bobbrow (approved):

PR #9264, opened 2022-05-02
add compiler path to debug configs details

elahehrashedi:
Add more details to debug configs:

image

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -53,12 +53,13 @@ const addDetailToConfigs = (items: MenuItem[]): MenuItem[] => {
                 break;
             }
             default : {
-                item.detail = "";
                 break;
             }
         }
-        // Add the compiler path of the preLaunchTask to the description of the debug configuration.
-        item.detail = item.detail + (item.configuration.taskDetail ? " (" + item.configuration.taskDetail + ")" : undefined);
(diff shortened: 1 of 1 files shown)


Colengms:
Is it possible to use an additional line?  If not, perhaps we could omit the "compiler:" portion, so more of the path is visible?

bobbrow (approved):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -37,11 +37,29 @@ export enum TaskConfigStatus {
...
+                item.detail = localize("detected.task", "Detected Task");
+                break;
+            }
+            default : {
+                break;
+            }
+        }
+        if (item.configuration.taskDetail) {
+            // Add the compiler path of the preLaunchTask to the description of the debug configuration.
+            item.detail = (item.detail ? item.detail : "") + " (" + item.configuration.taskDetail + ")";
```
sean-mcmanus: Can this be simplified to `(item.detail ?? "")`?

bobbrow:
> Is it possible to use an additional line? If not, perhaps we could omit the "compiler:" portion, so more of the path is visible?

I don't believe it is possible to have a 3rd line.

sean-mcmanus (changes requested):
.

sean-mcmanus (approved):

PR #9310, opened 2022-05-13
Avoid calling wordexp for Mac 10.15 or older.

### LINKED ISSUE #9309: With cpptools 1.10.2, use of std::shared_mutex causes crash on Mac 10.11 (El Capitan) or earlier
It's because it calls wordexp/wordfree which aren't available on those earlier OS's. I'm not sure what the fallback behavior is in this case. IntelliSense might fail to configure correctly.

The workaround would be to downgrade to 1.9.8.

Officially, we don't support Mac 10.11, but in this case the regression can be avoided.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9309

### DIFF
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -253,6 +253,8 @@ export class CppSettings extends Settings {
                     }
                 }
             }
+        } else if (this.forceLegacyCompilerArgs) {
+            return true;
         }
         return super.Section.get<boolean>("legacyCompilerArgsBehavior");
     }
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -239,7 +239,23 @@ export class CppSettings extends Settings {
     public get defaultCustomConfigurationVariables(): { [key: string]: string } | undefined { return super.Section.get<{ [key: string]: string }>("default.customConfigurationVariables"); }
     public get useBacktickCommandSubstitution(): boolean | undefined { return super.Section.get<boolean>("debugger.useBacktickCommandSubstitution"); }
     public get codeFolding(): boolean { return super.Section.get<string>("codeFolding") === "Enabled"; }
-    public get legacyCompilerArgsBehavior(): boolean | undefined  { return super.Section.get<boolean>("legacyCompilerArgsBehavior"); }
+    private forceLegacyCompilerArgs: boolean | undefined;
+    public get legacyCompilerArgsBehavior(): boolean | undefined  {
+        if (this.forceLegacyCompilerArgs === undefined) {
+            this.forceLegacyCompilerArgs = false;
+            if (os.platform() === "darwin") {
+                const releaseParts: string[] = os.release().split(".");
+                if (releaseParts.length >= 1) {
+                    // wordexp doesn't work for older Mac OS's.
+                    if (parseInt(releaseParts[0]) < 16) {
+                        this.forceLegacyCompilerArgs = true;
+                        return true;
+                    }
+                }
+            }
+        }
+        return super.Section.get<boolean>("legacyCompilerArgsBehavior");
+    }
 
     public get enhancedColorization(): boolean {
         return super.Section.get<string>("enhancedColorization") === "Enabled"
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -245,7 +245,8 @@ export class CppSettings extends Settings {
             this.forceLegacyCompilerArgs = false;
             if (os.platform() === "darwin") {
                 const releaseParts: string[] = os.release().split(".");
-                // wordexp doesn't work for Mac OS 10.15 (Darwin 15.x) or older.
+                // wordexp doesn't work for Mac OS 10.11 (Darwin 15.x) or older,
+                // given our minimum build target of Mac OS 10.12.
                 if (releaseParts.length >= 1 && parseInt(releaseParts[0]) < 16) {
                     this.forceLegacyCompilerArgs = true;
                     return true;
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/settings.ts
```diff
@@ -239,7 +239,23 @@ export class CppSettings extends Settings {
...
     public get codeFolding(): boolean { return super.Section.get<string>("codeFolding") === "Enabled"; }
-    public get legacyCompilerArgsBehavior(): boolean | undefined  { return super.Section.get<boolean>("legacyCompilerArgsBehavior"); }
+    private forceLegacyCompilerArgs: boolean | undefined;
+    public get legacyCompilerArgsBehavior(): boolean | undefined  {
+        if (this.forceLegacyCompilerArgs === undefined) {
+            this.forceLegacyCompilerArgs = false;
+            if (os.platform() === "darwin") {
+                const releaseParts: string[] = os.release().split(".");
+                if (releaseParts.length >= 1) {
+                    // wordexp doesn't work for older Mac OS's.
```
michelleangela: Nit: put in comment explicit Mac version.

Extension/src/LanguageServer/settings.ts
```diff
@@ -239,7 +239,23 @@ export class CppSettings extends Settings {
...
+                    if (parseInt(releaseParts[0]) < 16) {
```
michelleangela: combine `releaseParts.length >= 1` and `parseInt(releaseParts[0]) < 16`?

Extension/src/LanguageServer/settings.ts
sean-mcmanus: I didn't think a comment was necessary because the code below is `< 16`.

Extension/src/LanguageServer/settings.ts
sean-mcmanus: Yeah, I thought about that...not sure if it matters.

michelleangela: Less nesting of conditionals.

sean-mcmanus: Yeah, but it's a very small nesting and it makes it more clear what the comment applies to versus using an && -- however, I could change it if you want.

Extension/src/LanguageServer/settings.ts
michelleangela: For devs who aren't familiar with Mac versions, adding version in comment makes things clear without needing to figure out where 16 is being compared to. Also, since comment mentions "older" I was expecting version number.

Don't mean to be nit picker, but it's nice to have clear comments/code :).

Extension/src/LanguageServer/settings.ts
michelleangela: Either way, it's coding style.

sean-mcmanus: Done

michelleangela (approved):

Colengms:
Are you sure wordexp doesn't work for Mac OS 10.15 (Darwin 15.x) or older?  Could you please cite your sources of information?  Doing a quick search, I see references to these functions being available as far back as 10.9.5 (presumably a lot further).   Since they are documented to work (and referred to by threads on the internet), perhaps this needs further investigation?

The macOS documentation refers to this function, in BSD-derived docs dated from 2002.  https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/wordexp.3.html .

I see people asking questions on stackoverflow about wordexp, over a decade ago: https://stackoverflow.com/questions/5367378/wordexp-followed-by-strcpy-exc-bad-access-sharedlibrary-apply-load-rules-all

sean-mcmanus:
> > Are you sure wordexp doesn't work for Mac OS 10.15 (Darwin 15.x) or older? Could you please cite your sources of information? Doing a quick search, I see references to these functions being available as far back as 10.9.5 (presumably a lot further). Since they are documented to work (and referred to by threads on the internet), perhaps this needs further investigation?
> > The macOS documentation refers to this function, in BSD-derived docs dated from 2002. https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/wordexp.3.html .
> > I see people asking questions on stackoverflow about wordexp, over a decade ago: https://stackoverflow.com/questions/5367378/wordexp-followed-by-strcpy-exc-bad-access-sharedlibrary-apply-load-rules-all
> 
> We're getting DLL load failure telemetry for cpptools-wordexp on Mac 15.6 (and only that OS version). Do you know of another potential cause/fix?

Our Mac build target is 10.12 or newer, and Darwin 15.6 is 10.11, so there could be some DLL dependency issue with the implementation such that the 10.12 targetting DLL requires a DLL dependency that is missing on 10.11.

sean-mcmanus:
> Our Mac build target is 10.12 or newer, and Darwin 15.6 is 10.11, so there could be some DLL dependency issue with the implementation such that the 10.12 targetting DLL requires a DLL dependency that is missing on 10.11.

We could potentially fix the issue by decreasing the build target to 10.11 for cpptools-wordexp (not sure if that's possible yet).

sean-mcmanus:
> > Our Mac build target is 10.12 or newer, and Darwin 15.6 is 10.11, so there could be some DLL dependency issue with the implementation such that the 10.12 targetting DLL requires a DLL dependency that is missing on 10.11.
> 
> We could potentially fix the issue by decreasing the build target to 10.11 for cpptools-wordexp (not sure if that's possible yet).

It looks like we'd have to alter our build process for cpptools-wordexp to have a lower build target (cpptools itself won't build with the lower build target), but I don't think that's worth the effort, i.e. seems better to just avoid calling wordexp, since we don't officially support Mac 10.11 anyway.

sean-mcmanus:
Seems the Mac 10.15 situation might need more investigation...

PR #9313, opened 2022-05-13
Remove [Log] editor.wordWrap "off" default.

sean-mcmanus:
I believe this was added for the C/C++: Peek References console, but that doesn't appear to be used any more, and it causes build output to not be visible when running CMake Tools to build.


michelleangela (approved):

sean-mcmanus:
@bobbrow Did you still have some concern with this?

bobbrow:
> @bobbrow Did you still have some concern with this?

No. You can proceed.  I will probably just add this setting to my User settings.

Issue #5845, opened 2020-07-28
Parameter Name Hints shown inline


ghost:
Now that we have doxygen comment supported to help code readability, it would be also useful to have a CLion-like name hint to be shown inline all the time (just like named parameters)

So that instead of this
![image](https://user-images.githubusercontent.com/53231928/88621427-a517e800-d0ca-11ea-974e-5578688cb26b.png)

we get this instead

![image](https://user-images.githubusercontent.com/53231928/88621418-97626280-d0ca-11ea-951b-1e955589245c.png)

sean-mcmanus:
I've filed a feature request on VS at https://developercommunity.visualstudio.com/idea/1130438/c-intellisense-inline-parameter-name-hints.html , although it's possible we would still need to do some more work to setup VS Code specific code lens providers if they decide to implement the feature.

Diegorro98:
The Inlay Hint provider API for VS Code has been released. Any updates on this?

sean-mcmanus:
We haven't looked into it yet. Seems like we'd have to query our database for all function symbol parameter names for the entire document, which we don't currently only do as part of signature help.

sean-mcmanus:
Parameter hints (aka inlay hints) are added with 1.11.0 (pre-release). The relevant settings to change are
```json
    "C_Cpp.inlayHints.autoDeclarationTypes.enabled": true,
    "C_Cpp.inlayHints.parameterNames.enabled": true,
    "C_Cpp.inlayHints.referenceOperator.enabled": true
```
assuming you want all those enabled or a subset.

Another setting we added is
```json
"C_Cpp.inlayHints.parameterNames.suppressWhenArgumentContainsName": true
```

Our next 1.11.1 release may change the defaults like
```json
    "[cpp]": {
        "editor.inlayHints.fontSize": 11
    },
    "[c]": {
        "editor.inlayHints.fontSize": 11
    },
```
to match VS and to avoid the confusing the inlay text too much with existing text, particular for the "auto" case.

One thing that may not be obvious is that the inlay hint for "auto" is intended to replace all the potential existing type modifiers and it is not just adding additional info that the auto hides, so any potentially confusing or redundant existing const *& modifier text is intended to be ignored in favor of the more complete inlay hint type info.

PR #9392, opened 2022-06-02
Add InlayHintsProvider and Display hints

michelleangela:
For https://github.com/microsoft/vscode-cpptools/issues/5845.

### This feature will display inlay hints for:
- Auto declaration types.
- Parameter names.
- Reference operator `&` for parameters passed by non-const reference.

### Tasks:
- [x] Gets inlay hints data from language server
- [x] Displays inlay hints
- [x] [separate PR](https://github.com/microsoft/vscode-cpptools/pull/9460): add and support inlay hints settings specific to extension.

## Example
Without inlay hints
image

With inlay hints for auto type and parameters.
image

### DIFF
--- a/Extension/src/LanguageServer/Providers/inlayHintProvider.ts
+++ b/Extension/src/LanguageServer/Providers/inlayHintProvider.ts
@@ -27,9 +27,11 @@ interface GetInlayHintsResult {
     inlayHints: CppInlayHint[];
 }
 
-// TODO: either use separate lists or filter
-type InlayHintsCacheEntry = { FileVersion: number, 
-    TypeHints: vscode.InlayHint[], ParameterHints: vscode.InlayHint[] }
+type InlayHintsCacheEntry = {
+    FileVersion: number;
+    TypeHints: vscode.InlayHint[];
+    ParameterHints: vscode.InlayHint[];
+};
 
 const GetInlayHintsRequest: RequestType<GetInlayHintsParams, GetInlayHintsResult, void, void> =
     new RequestType<GetInlayHintsParams, GetInlayHintsResult, void, void>('cpptools/getInlayHints');
@@ -45,8 +47,8 @@ export class InlayHintsProvider implements vscode.InlayHintsProvider {
         this.onDidChangeInlayHints = this.onDidChangeInlayHintsEvent.event;
     }
 
-    public async provideInlayHints(document: vscode.TextDocument, range: vscode.Range, token: vscode.CancellationToken)
-        : Promise<vscode.InlayHint[] | undefined> {
+    public async provideInlayHints(document: vscode.TextDocument, range: vscode.Range, token: vscode.CancellationToken):
+    Promise<vscode.InlayHint[] | undefined> {
         await this.client.awaitUntilLanguageClientReady();
         const uriString: string = document.uri.toString();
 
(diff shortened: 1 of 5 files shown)


Extension/src/LanguageServer/Providers/inlayHintProvider.ts
```diff
@@ -0,0 +1,112 @@
...
+        const cacheEntry: InlayHintsCacheEntry | undefined = this.cache.get(uriString);
+        if (cacheEntry && cacheEntry.FileVersion === document.version) {
+            return this.getHintsBasedOnSettings(cacheEntry);
+        }
+
+        // Get new results from the language server
+        const params: GetInlayHintsParams = { uri: uriString };
+        const inlayHintsResult: GetInlayHintsResult = await this.client.languageClient.sendRequest(GetInlayHintsRequest, params, token);
+        if (inlayHintsResult.canceled || (inlayHintsResult.fileVersion !== openFileVersions.get(uriString))) {
+            throw new vscode.CancellationError();
```
Colengms: Looking at code in the vscode repo, it's unclear to me whether `CancellationError` would be handled for `provideInlayHints`.  `CancellationError` was added specifically for semantic tokens, and didn't work properly for me with `FoldingRangeProvider`.  https://github.com/microsoft/vscode/issues/125799  .   In that issue, aeschli mentioned:

> CancellationError is currently only handled for semantic token provider

It might be necessary to return undefined and possibly trigger `onDidChangeInlayHints` if returning undefined does not result in a new request.  (You'll probably get a new request if VS Code cancelled the token, but might not get a new request due to a file version change.  You may need to check.)

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+    Type = 0,
+    Parameter = 1,
+}
+
+interface CppInlayHint {
+    position: Position;
+    label: string;
+    inlayHintKind: InlayHintKind;
+    isValueRef: boolean;
+    // hasParamName: boolean;
```
sean-mcmanus, line 23: Why is this commented out? Should there be a comment explaining it?

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+    public onDidChangeInlayHints?: vscode.Event<void>;
+    private cache: Map<string, InlayHintsCacheEntry> = new Map<string, InlayHintsCacheEntry>();
+
+    constructor(client: DefaultClient) {
+        this.client = client;
+        this.onDidChangeInlayHints = this.onDidChangeInlayHintsEvent.event;
+    }
+
+    public async provideInlayHints(document: vscode.TextDocument, range: vscode.Range, token: vscode.CancellationToken):
+    Promise<vscode.InlayHint[] | undefined> {
```
sean-mcmanus: Indentation

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+                vscode.InlayHintKind.Parameter);
+            inlayHint.paddingRight = true;
+            resolvedHints.push(inlayHint);
+        });
+        return resolvedHints;
+    }
+
+    private createCacheEntry(inlayHintsResults: GetInlayHintsResult): InlayHintsCacheEntry {
+        const typeHints: vscode.InlayHint[] = [];
+        inlayHintsResults.inlayHints.forEach((h: CppInlayHint) => {
```
sean-mcmanus: Would a for loop be better? I think forEach is not recommended to use, although we some of our existing code does that still.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
michelleangela: Lint expects 4 spaces. I think it's because it's part of function signature. It's 8 spaces for indentation for function content.

sean-mcmanus: Oh, that is unexpected.

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
michelleangela: It's pending change for next PR.

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
michelleangela: Changed to `for...of` loop.

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
bobbrow: Can you put the full signature on one line instead?

sean-mcmanus: 2 lines seems better, the line was long, right?

sean-mcmanus (approved):

Extension/src/LanguageServer/Providers/inlayHintProvider.ts
bobbrow: I've seen longer lines in the TypeScript codebase and they keep signatures to one line.  I don't have a strong opinion here, but the indenting does look weird.  My recommendation is to either put it all on one line, or move the `:` to the second line to make it more obvious you are putting the return type there (assuming the linter is ok with that).

PR #9402, opened 2022-06-03
Code analysis refactor

sean-mcmanus:
This can wait for 1.11.0.

I was going to refactor more the of the common/util stuff, but I undid those changes so that it was more limited.


Extension/src/LanguageServer/client.ts
```diff
@@ -3513,38 +3072,11 @@ export class DefaultClient implements Client {
...
-                    break;
-                }
-                if (!removed) {
-                    newDiagnostics.push(diagnostic);
-                }
-            }
-            diagnosticsCollectionCodeAnalysis.set(uri, newDiagnostics);
+        if (removeCodeAnalysisProblems(identifiersAndUris)) {
+            // Need to notify the language client of the removed diagnostics so it doesn't re-send them.
+            this.languageClient.sendNotification(RemoveCodeAnalysisProblemsNotification, {
```
michelleangela, line 3079: Can this "sendNotification" be done from codeAnalysis.ts?

Extension/src/LanguageServer/codeAnalysis.ts
```diff
@@ -0,0 +1,501 @@
...
+import { CppSourceStr } from './extension';
+import { CppSettings } from './settings';
+import * as nls from 'vscode-nls';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+let diagnosticsCollectionCodeAnalysis: vscode.DiagnosticCollection;
+
+export function RegisterCodeAnalysisNotifications(languageClient: LanguageClient): void {
```
michelleangela, line 20: Good idea on `RegisterCodeAnalysisNotifications`. I'll do this for InlayHints.

Extension/src/LanguageServer/codeAnalysis.ts
```diff
@@ -0,0 +1,501 @@
...
+
+// Used to handle fix invalidation after an edit,
+// i.e. it's set to undefined if it becomes invalid.
+interface CodeActionWorkspaceEdit {
+    workspaceEdit?: vscode.WorkspaceEdit;
+}
+
+interface CodeActionPerUriInfo {
+    // These two arrays have the same length, i.e. index i of identifiers
+    // is usee to index into workspaceEdits to get the corresponding edit.
```
michelleangela: misspelling "usee".

Suggestion to use a spell checker extension. (e.g. https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, but it's unclear to me which way is better, i.e. should the client code send the messages directly or should other code use a reference to the client to send messages themselves. We have code that does it both ways. My goal was just to remove the bulk of the code analysis code from client.ts and not necessarily to get rid of all of it in cases like this where there is interaction with the client.

michelleangela (approved):

PR #9403, opened 2022-06-04
Fix resolve variable with non-string.

### LINKED ISSUE #9302: Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: TypeError: ret.replace is not a function
Bug type: Language Service

**Describe the bug**
- OS and Version: Windows 10
- VS Code Version: 1.67.1
- C/C++ Extension Version: 1.9.8
- If using SSH remote, specify OS of remote machine:

When starting VSCode with C++ source, the following pops up:

Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: TypeError: ret.replace is not a function

**Steps to reproduce**

**Expected behavior**

**Code sample and logs**

- Code sample

- Configurations in `c_cpp_properties.json`

- Logs from running `C/C++: Log Diagnostics` from the VS Code command palette

- Logs from [the language server logging](https://code.visualstudio.com/docs/cpp/enable-logging-cpp#_enable-logging-for-the-language-server)

**Screenshots**

**Additional context**

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9302


Extension/src/common.ts
```diff
@@ -342,6 +342,13 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
         return "";
     }
 
+    // jsonc parser may assign a non-string object to a string.
```
bobbrow, line 345: Do you know which parse call results in the malformed object?  We shouldn't ever assume that a JSON file we parse is in the format we expect.  The data has to be validated before it can be used.  I'm worried we have an unvalidated object somewhere in the code still.

sean-mcmanus: It's not a malformed object. It repros whenever a non-string is used. The parsing code just returns an 'any' and we blindly assign it to what we expect, i.e. a string. This fix essentially checks for that and converts it to a string.

Extension/src/common.ts
```diff
@@ -342,6 +342,13 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
...
+    if (typeof(input) !== "string") {
+        const inputAny: any = input;
+        input = inputAny.toString();
```
Colengms, line 349: Have you tried this with an explicit `undefined`?  You might need to check the input for null before calling toString().

Colengms (changes requested):

Extension/src/common.ts
bobbrow: line 341 should guard against `undefined`

Extension/src/common.ts
bobbrow: Malformed may not be the correct terminology.  I was assuming you parsed the json into a defined interface instead of `any`.  I'm still surprised that TypeScript would let you pass an `any` for `string` without so much as a warning.  Are we doing a type assertion before calling this function?  That's what I'm trying to point out.  That we should be doing validation somewhere instead of a type assertion.

Extension/src/common.ts
Colengms: Thanks, that was outside of the diff.

Colengms (approved):

Extension/src/common.ts
bobbrow: The thing that I don't want to see throughout the code is this:
```ts
function foo(bar: string) {
  if (typeof(bar) !== "string") {
    ...
  }
}
```
If we can't rely on the variable/argument typings, then TypeScript isn't serving us well.

Extension/src/common.ts
```diff
@@ -342,6 +342,13 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
         return "";
     }
 
+    // jsonc parser may assign a non-string object to a string.
+    if (typeof(input) !== "string") {
```
bobbrow: (I think we have a `isString` utility function in common.ts)

Extension/src/common.ts
sean-mcmanus: Yeah, I was looking into potentially doing the validation earlier on when the parse is done, but doing it here seems just as well...and I figured we could move the check later if we find a reason to.

Also, the jsonc library has updated to not use "any" any more, however, switching to the new API has breaking changes and so switching to that would require a significantly more amount of work, but probably recommended in the future.

sean-mcmanus: I was considering adding a comment about that, but I decided not to (seems like the comment should go at the jsonc calls, but there are multiple uses, so I wasn't sure where to add the comment).

bobbrow: If you check in this fix, please open a new issue to fix the place where we did a type assertion instead of a validation.

sean-mcmanus: I filed https://github.com/microsoft/vscode-cpptools/issues/9414

PR #9418, opened 2022-06-08
Add deploySteps and variables to cppdbg

xisui-MSFT:
Add deploySteps and variables to cppdbg. Spec available upon request.

There will be follow up PRs, which implement deploy in multiroot scenarios, and process picker for gdb extended-remote, separately.


Extension/package.nls.json
```diff
@@ -294,6 +294,26 @@
     "c_cpp.debuggers.hardwareBreakpoints.description": "Explicit control of hardware breakpoint behavior for remote targets.",
     "c_cpp.debuggers.hardwareBreakpoints.require.description": "If true, always use hardware breakpoints. Defaults to false.",
     "c_cpp.debuggers.hardwareBreakpoints.limit.description": "Optional limit on the number of available hardware breakpoints to use. Only enforced when \"require\" is true and \"limit\" is greater than 0. Defaults to 0.",
+    "c_cpp.debuggers.variables.description": "Varaibles for recursive substitution in this launch configuration. Each variable may be referred to in others.",
```
WardenGnaw:
```suggestion
    "c_cpp.debuggers.variables.description": "Variables for recursive substitution in this launch configuration. Each variable may be referred to in others.",
```

Extension/package.nls.json
```diff
@@ -294,6 +294,26 @@
...
+    "c_cpp.debuggers.variables.properties.description": "Varaible for recursive substitution in this launch configuration. The value may referred to other variables.",
```
WardenGnaw:
```suggestion
    "c_cpp.debuggers.variables.properties.description": "Variable for recursive substitution in this launch configuration. The value may referred to other variables.",
```

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -295,6 +311,12 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
             // logger.showOutputChannel();
         }
 
+        // Run deploy steps
+        const deploySucceeded: boolean = await this.deploySteps(config, folder, token);
+        if (!deploySucceeded || token?.isCancellationRequested) {
+            return;
```
WardenGnaw: returning `null` will cause a pop-up notification 

```suggestion
            return undefined;
```

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -295,6 +311,12 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
             // logger.showOutputChannel();
         }
 
+        // Run deploy steps
+        const deploySucceeded: boolean = await this.deploySteps(config, folder, token);
```
WardenGnaw: I would wrap lines 315-318 around this check:
```
if (!config.deploySteps || config.deploySteps.length === 0) {
```

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -864,6 +886,118 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        const vars: ExpansionVars = config.variables ? config.variables : {};
+        vars.workspaceFolder = folderPath || '{workspaceFolder}';
+        vars.workspaceFolderBasename = folderPath ? path.basename(folderPath) : '{workspaceFolderBasename}';
+        const expansionOptions: ExpansionOptions = { vars, recursive: true };
+        return expandAllStrings(config, expansionOptions);
+    }
+
+    // Returns true when ALL steps succeed
+    private async deploySteps(config: vscode.DebugConfiguration, folder: vscode.WorkspaceFolder | undefined, cancellationToken?: vscode.CancellationToken): Promise<boolean> {
+        if (!config.deploySteps || config.deploySteps.length === 0) {
```
WardenGnaw: Move this condition to caller of this function.

Extension/src/SSH/commandInteractors.ts
```diff
@@ -0,0 +1,369 @@
...
+            return Number.isNaN(num) ? undefined : num;
+        } catch (err) {
+            return undefined;
+        }
+    }
+
+    return 0;
+}
+
+/**
```
WardenGnaw, line 181: Can you expand on this function's comment?

Extension/tools/GenerateOptionsSchema.ts
```diff
@@ -95,10 +91,15 @@ function replaceReferences(definitions: any, objects: any): any {
             objects[key].properties = updateDefaults(objects[key].properties, objects[key].default);
         }
 
-        // Recursively replace references if the array has objects in items.
+        // Items in array contains ref
         if (objects[key].hasOwnProperty('type') && objects[key].type === "array" && objects[key].items !== null && objects[key].items.hasOwnProperty('$ref')) {
             objects[key].items = refReplace(definitions, objects[key].items);
         }
+
+        // Recursively replace references if the array has objects taht contains ref in items.
```
WardenGnaw:
```suggestion
        // Recursively replace references if the array has objects that contains ref in items.
```

Extension/package.nls.json
WardenGnaw: Also leave an example of how it should look in the description or set an example using `default` with
```
"default": {
   "<variable-name>": "<variable-value>"
}
```

Extension/src/expand.ts
```diff
@@ -0,0 +1,133 @@
+/* --------------------------------------------------------------------------------------------
```
WardenGnaw, line 1: Can we get some tests covering methods in this file?

WardenGnaw (changes requested):

WardenGnaw:
In this PR, could you provide an example `launch.json` with `variables` and `deploy` and give a quick explanation of what would happen with those configuration fields set (or a GIF of it if it is working end to end.)

Extension/src/SSH/sshCommandRunner.ts
```diff
@@ -0,0 +1,518 @@
...
+                        ? vscode.Uri.file(os.homedir() || 'c:\\')
+                        : vscode.Uri.file(os.homedir() || '/')),
+                name: nickname,
+                shellPath: getShellPath(terminalIsWindows),
+                shellArgs,
+                hideFromUser: true
+            };
+
+            let terminalDataHandlingQueue: Promise<void> = Promise.resolve();
+            terminalListener = systemInteractor.onDidWriteTerminalData(async e => {
```
Tyriar, line 423: You should dispose of this listener as soon as you don't need it (eg. `if (args.showLoginTerminal)`). This is because if there exists a single one of these listeners, _all_ data for _all_ terminals gets sent to _all_ extension hosts 😱

Tyriar (changes requested):

Extension/src/SSH/sshCommandRunner.ts
xisui-MSFT: It's disposed in clean(), at line 266. There is one case where done is called but clean isn't, which is when continueOn is specified. But in this case, the command will eventually be cancelled or the terminal will be closed, and the listeners will be disposed.

xisui-MSFT:
> In this PR, could you provide an example `launch.json` with `variables` and `deploy` and give a quick explanation of what would happen with those configuration fields set (or a GIF of it if it is working end to end.)

Sample launch.json:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug Pi1",
            "type": "cppdbg",
            "request": "launch",
            "variables": {
                "user": "pi",
                "host": "${input:targetIP}",
                "port": "22100",
                "program": "${command:cmake.launchTargetPath}",
                "targetDir": "/home/pi",
                "targetFilePath": "${targetDir}/${command:cmake.launchTargetFilename}"
            },
            "cwd": "${workspaceFolder}",
            "program": "${program}",
            "stopAtEntry": true,
            "MIMode": "gdb",
            "miDebuggerPath": "/usr/bin/gdb-multiarch",
            "miDebuggerServerAddress": "${host}:${port}",
            "deploySteps": [
                {
                    "type": "scp",
                    "files": "${program}",
                    "host": {
                        "user": "${user}",
                        "hostName": "${host}"
                    },
                    "targetDir": "${targetDir}"
                },
                {
                    "type": "ssh",
                    "host": {
                        "user": "${user}",
                        "hostName": "${host}"
                    },
                    "command": "chmod +x ${targetFilePath}",
                },
                {
                    "type": "ssh",
                    "host": {
                        "user": "${user}",
                        "hostName": "${host}"
                    },
                    "command": "gdbserver :${port} ${targetFilePath}",
                    "continueOn": "Listening on port",
                    "debug": true
                },
                {
                    "type": "ssh",
                    "host": {
                        "user": "${user}",
                        "hostName": "${host}"
                    },
                    "command": "${targetFilePath}",
                    "debug": false
                }
            ],
            "setupCommands": [
                {
                    "ignoreFailures": true
                }
            ]
        }
    ],
    "inputs": [
        {
            "type": "promptString",
            "id": "targetIP",
            "description": "Enter the IP address of the target device"
        }
    ]
}
```

Stepps:
1. Expand all variables in this launch config that are not resolved by VS Code.
2. Run deploy steps. ask for password for scp/ssh steps. The authentication method is the same as using ssh in the terminal. For example, if key based authentication is used, then no authentication will be asked.
  - Copy binary to the device using SCP.
  - chom +x in case there are any permission issue.
  - Start gdbserver when starting debugging; start the program only when starting without debugging.
3. Start debugging.

GIF:
deploy

Extension/src/Debugger/configurationProvider.ts
xisui-MSFT: `return;` is equivalent to  `return undefined;`. Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return

> Syntax: return [expression];
expression
The expression whose value is to be returned. If omitted, undefined is returned instead.

Extension/src/SSH/commands.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (jumpHosts && jumpHosts.length > 0) {
+        args.push('-J', jumpHosts.map(getFullHostAddress).join(','));
+    }
+    if (host.port) {
+        // upper case P
+        args.push('-P', `${host.port}`);
+    }
+    args.push(files.map(uri => `"${uri.fsPath}"`).join(' '), `${getFullHostAddressNoPort(host)}:${targetDir}`);
+
+    return runSshTerminalCommandWithLogin(host, { systemInteractor: defaultSystemInteractor, nickname: 'scp', command: `scp ${args.join(' ')}`, token: cancellationToken });
```
WardenGnaw: Does this require the user to have `scp` in their `PATH`?

For `pipeTransport` we allow users to provide the full path using `pipeProgram`.

I can see this being an issue on Windows 7 and Windows 10 if they do not have OpenSSH installed. 
Do we want to add a way for users to customize the path to the programs?

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: I would still prefer it to be displayed to keep with the other `return`s in this method.

Extension/src/SSH/commands.ts
xisui-MSFT: This feature already requires a rather recent version of VS Code because of the proposed API we are using. I would be surprised if the users don't have ssh/scp on their path. I think it's safe enough to see if we receive any related feedbacks. However, if you think this is a concerning issue, I think adding an entry to user/machine settings makes more sense, since launch.json is supposed to be shared across a team.

Extension/src/Debugger/configurationProvider.ts
xisui-MSFT: ok

Extension/src/SSH/commandInteractors.ts
xisui-MSFT: What would you like to see in this comment?

WardenGnaw: Can you please reword the comment for the function to capture what its doing?
The current comment is a bit hard to follow. I had to read the code-flow first before understanding this comment. 
```
/* 
* 's password:
 * or
 * Password:
 * not
 * 's old password:
 * 's new password:
 */
```

Extension/src/SSH/commands.ts
WardenGnaw: VS Code still has support for Windows 8, so those users may still have to point to an external path for SSH / SCP

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -864,6 +886,118 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            return true;
+        }
+
+        let succeeded: boolean = true;
+        const deployStart: number = new Date().getTime();
+
+        STEPS: for (const step of config.deploySteps) {
```
bobbrow: Instead of using this strange syntax for breaking out of the `for` instead of the `switch`, can you move this whole for loop into a new function and have the `break STEPS;` become `return succeeded;` instead?  I believe that's the only piece of state that needs to be returned out of this block.

Extension/src/SSH/commandInteractors.ts
```diff
@@ -0,0 +1,369 @@
...
+ * 's password:
+ * or
+ * Password:
+ * not
+ * 's old password:
+ * 's new password:
+ */
+function getPasswordPrompt(
```
bobbrow: can you please put function signatures on a single line?

bobbrow: (multiple occurrences in this file)

Extension/src/SSH/commands.ts
xisui-MSFT: We may need to support rsync (probably others?) in the future. Adding one setting entry for each path seems quite awkward. @bobbrow Any suggestions?

bobbrow: I don't find that awkward.  The property can be optional for those who want to set an explicit path.  There is already precedent for it with `pipeProgram`.

xisui-MSFT: @bobbrow I prefer adding it to user/machine settings than launch.json, since launch.json should be shared in a team, and each team member may install them at different locations.

bobbrow: How is it different from the `${targetDir}` variable in your example?  Team members may prefer different output sites as well.  The way to make it work for cross-team configurations is to use environment variables.  My opinion is that we shouldn't add settings to the C++ extension that are essentially just environment variables for launch configurations.

xisui-MSFT: TargetDir is usually fixed, just like default install dir for software, especially for embedded devices which are re-imaged all the time. Also I think adding to env var is hardly different from adding to path... So how about not adding this at all and just assume they are on path? :D

bobbrow: It's usually a good idea to provide a way to override the path to a given tool, but I don't have a strong opinion about it.  If @WardenGnaw doesn't object, you can leave it out and see if you get feedback from developers who use the feature.

I only have a strong opinion favoring adding a property to the launch configuration instead of adding a new setting.

WardenGnaw: Discussed offline. I would prefer a path override defined.

WardenGnaw (approved):
:shipit:

Extension/package.nls.json
```diff
@@ -298,6 +298,28 @@
     "c_cpp.debuggers.hardwareBreakpoints.description": "Explicit control of hardware breakpoint behavior for remote targets.",
     "c_cpp.debuggers.hardwareBreakpoints.require.description": "If true, always use hardware breakpoints. Defaults to false.",
     "c_cpp.debuggers.hardwareBreakpoints.limit.description": "Optional limit on the number of available hardware breakpoints to use. Only enforced when \"require\" is true and \"limit\" is greater than 0. Defaults to 0.",
+    "c_cpp.debuggers.variables.description": "Variables for recursive substitution in this launch configuration. Each variable may be referred to in others.",
+    "c_cpp.debuggers.variables.properties.description": "Variable for recursive substitution in this launch configuration. The value may referred to other variables.",
```
jogo-: Missing word: may referred -> may **be** referred

Extension/package.nls.json
```diff
@@ -298,6 +298,28 @@
...
+    "c_cpp.debuggers.host.description": "Host information.",
+    "c_cpp.debuggers.host.user.description": "User logging into the host.",
+    "c_cpp.debuggers.host.hostName.description": "Host name.",
+    "c_cpp.debuggers.host.port.description": "SSH port on the host. Default is 22.",
+    "c_cpp.debuggers.host.jumpHost.description": "Connect to the target host by first making an connection to the jump hosts.",
```
jogo-: Typo: a**n** connection -> a connection

sean-mcmanus (approved):

PR #9439, opened 2022-06-14
Remove legacyCompilerArgsBehavior deprecation.

sean-mcmanus:
The setting is hard to discover, i.e. hidden, if it's deprecated (no autocomplete or visibility in the settings UI).

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2670,8 +2670,7 @@
           "type": "boolean",
           "default": false,
           "markdownDescription": "%c_cpp.configuration.legacyCompilerArgsBehavior.markdownDescription%",
-          "scope": "resource",
-          "deprecationMessage": "%c_cpp.configuration.legacyCompilerArgsBehavior.deprecationMessage%"
+          "scope": "resource"
         }
       }
     },


michelleangela (approved):

Colengms (changes requested):
This was intentionally deprecated.  We didn't want to advertise this feature.  The intention was only to refer people to this setting if they reported an issue it might address.  We don't want people using the old behavior unless they absolutely need it.

bobbrow:
> This was intentionally deprecated. We didn't want to advertise this feature. The intention was only to refer people to this setting if they reported an issue it might address. We don't want people using the old behavior unless they absolutely need it.

@Colengms It's impossible to find it in the case that they do need it though.  We can deprecate it next release.  I'd like to get telemetry on if anyone's using it in the meantime.

Colengms:
The idea was, if someone reports an issue, we would provide them a line of text to add to `settings.json`.  If they paste the setting in, it squiggles to indicate it's deprecated, but hover still provides documentation and the value still auto-completes.  Ideally, no-one should need to use this setting.

sean-mcmanus:
Plan is to re-added deprecation for 1.11.

PR #9442, opened 2022-06-14
Update cmake language package in README

hofbi:
VS Code is recommending that the language support extension bundled with the cmake-tools package has been updated (see https://github.com/microsoft/vscode-cmake-tools/pull/2597). This PR updates the reference in the README from the old to the recommended cmake language extentions.

### DIFF
--- a/ExtensionPack/README.md
+++ b/ExtensionPack/README.md
@@ -3,7 +3,7 @@
 This extension pack includes a set of popular extensions for C++ development in Visual Studio Code:
 * [C/C++](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools)
 * [C/C++ Themes](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools-themes)
-* [CMake](https://marketplace.visualstudio.com/items?itemName=twxs.cmake)
+* [CMake](https://marketplace.visualstudio.com/items?itemName=josetr.cmake-language-support-vscode)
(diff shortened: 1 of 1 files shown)


ExtensionPack/README.md
```diff
@@ -3,7 +3,7 @@
 This extension pack includes a set of popular extensions for C++ development in Visual Studio Code:
 * [C/C++](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools)
 * [C/C++ Themes](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools-themes)
-* [CMake](https://marketplace.visualstudio.com/items?itemName=twxs.cmake)
+* [CMake](https://marketplace.visualstudio.com/items?itemName=josetr.cmake-language-support-vscode)
```
bobbrow: Thank you for the PR.  Can you also update the extension name in the hyperlink to `CMake Language Support`?

bobbrow (changes requested):

bobbrow:
@hofbi Thanks for doing this for us.  The bad news is that I reviewed your PR prior to reviewing feedback from the CMake Tools community about this change, and we're actually going to revert the dependency back to the old extension and push an update imminently as a result.

bobbrow:
I'm not sure what you mean.  We updated it earlier this year.

image

hofbi:
What I mean is the CMake package that I was actually updating with my PR. This CMake package has its last commit from October 2019.

Issue #9446, opened 2022-06-14
Unable to start preLaunchTask when "C_Cpp.intelliSenseEngine": "Disabled"


fonqL:
### Environment

- OS and version: Windows 10
- VS Code: 1.65.0
- C/C++ extension: 1.10.5
- GDB / LLDB version: gdb 11.2

### Bug Summary and Steps to Reproduce

Bug Summary:

I'm using clangd for code completion and also expect to use C/C++ for debug, so I enable the C/C++ plugin and disable the C/C++ intellisense engine. After pressing F5 vscode reports an error says the .exe program doesn't exist. But my preLaunchTask works well and generates .exe when I either disable C/C++ plugin or enable the "C_Cpp.intelliSenseEngine". 
Disable clangd or reboot system can't solve the problem. 
Also, if I set preLaunchTask with a non-existent task, it also report .exe doesn't exist instead of reporting "Could not find the task ...".
Why? and what should I do?

Steps to reproduce:
1. Config launch.json and tasks.json, and set preLaunchTask to build task.
2. Enable C/C++ and set "C_Cpp.intelliSenseEngine": "Disabled"
3. Press F5
4. vscode reports an error says the .exe program doesn't exist.

### Debugger Configurations

```shell
launch.json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "(gdb)launch",
            "type": "cppdbg",
            "request": "launch",
            "program": "${fileDirname}\\${fileBasenameNoExtension}.exe",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${fileDirname}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "miDebuggerPath": "D:\\MinGW\\bin\\gdb.exe",
            "setupCommands": [
                {
                    "description": "Enable pretty-printing for gdb",
                    "text": "-enable-pretty-printing",
                    "ignoreFailures": true
                }
            ],
            "preLaunchTask": "build"
        }
    ]
}

tasks.json
{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "process",
            "label": "build",
            "command": "g++",
            "args": [
                "-g",
                "${file}",
                "-o",
                "${fileDirname}\\${fileBasenameNoExtension}.exe",
                "-m64",
                "-Wall",
                "-Wextra",
                "-Wshadow",
                "-std=c++14",
                "-static-libgcc",
                "-fexec-charset=GBK",
                "-DFread",
[...]

PR #9448, opened 2022-06-14
resolve tasks when IntelliSense is disabled

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cpptools/issues/9446

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -113,7 +113,6 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
      * resolveDebugConfigurationWithSubstitutedVariables will be automatically called after this function.
      */
     async resolveDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: CppDebugConfiguration, token?: vscode.CancellationToken): Promise<CppDebugConfiguration | null | undefined> {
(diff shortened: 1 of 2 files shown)


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -148,6 +147,18 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
         if (config.preLaunchTask) {
             config.configSource = this.getDebugConfigSource(config, folder);
+            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === "Disabled";
+            // Run the build task if IntelliSense is disabled.
+            if (isIntelliSenseDisabled) {
+                try {
+                    await cppBuildTaskProvider.runBuildTask(config.preLaunchTask);
```
bobbrow, line 154: I think we should check if `config.preLaunchTask` has a value first.  That's what `resolvePreLaunchTask` does.

bobbrow: Nevermind. It's on line 148.

sean-mcmanus:
Should we get this into 1.10.6?

sean-mcmanus (approved):

sean-mcmanus:
> Should we get this into 1.10.6?

It seems like we should.

bobbrow:
> Should we get this into 1.10.6?

If it's not too much trouble, this seems like a pretty safe change.

sean-mcmanus (approved):

PR #9460, opened 2022-06-16
Add settings for inlay hints feature

### LINKED ISSUE #5845: Parameter Name Hints shown inline

michelleangela:
image


Extension/package.nls.json
```diff
@@ -166,6 +166,10 @@
...
+    "c_cpp.configuration.inlayHints.autoDeclarationTypes.enabled.description": { "message": "Enable/disable inlay hints for deduced type when `auto` is used in a declaration.", "comment": ["Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."]},
+    "c_cpp.configuration.inlayHints.parameterNames.enabled.description": "Enable/disable inlay hints for parameter names at call sites.",
+    "c_cpp.configuration.inlayHints.parameterNames.addReferenceOperator.description": { "message": "Add the reference operator `&` to hints when a parameter is passed by non-const reference. Note: this setting depends on `C_Cpp.inlayHints.parameterNames.enabled` to be enabled.", "comment": ["Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."]},
```
bobbrow: Rewording suggestion:
"Note: this setting takes effect only when \`C_Cpp.inlayHints.parameterNames.enabled\` is \`true\`."

Applies to this line and the line below.

Extension/package.nls.json
```diff
@@ -166,6 +166,10 @@
...
     "c_cpp.configuration.inactiveRegionForegroundColor.description": "Controls the font coloring of inactive preprocessor blocks. Input is in the form a hexadecimal color code or a valid Theme Color. If not set, this defaults to the syntax coloring scheme of the editor. This setting only applies when inactive region dimming is enabled.",
     "c_cpp.configuration.inactiveRegionBackgroundColor.description": "Controls the background coloring of inactive preprocessor blocks. Input is in the form a hexadecimal color code or a valid Theme Color. If not set, this defaults to transparent. This setting only applies when inactive region dimming is enabled.",
+    "c_cpp.configuration.inlayHints.autoDeclarationTypes.enabled.description": { "message": "Enable/disable inlay hints for deduced type when `auto` is used in a declaration.", "comment": ["Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."]},
```
bobbrow: In the UI, there will be a checkbox for boolean settings, so I think having "enable/disable" is unnecessary in the description.  Consider replacing "enable/disable" with "show".

Applies to this line and the line below.

Extension/package.nls.json
bobbrow: I'm also wondering if we need the Note at all because of the way the settings are grouped.  For example, we don't caveat the `clang_format*` or `vcFormat` settings by saying that the `formatting` setting needs to be set a certain way.

Extension/package.json
```diff
@@ -2211,6 +2211,30 @@
           "description": "%c_cpp.configuration.inactiveRegionBackgroundColor.description%",
           "scope": "resource"
         },
+        "C_Cpp.inlayHints.autoDeclarationTypes.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "%c_cpp.configuration.inlayHints.autoDeclarationTypes.enabled.description%",
```
sean-mcmanus: This needs to be "markdownDescription" and not "description".

sean-mcmanus: Same for the other settings that use markdown.

Extension/package.nls.json
sean-mcmanus: Is "this" supposed to be capital?

sean-mcmanus (changes requested):

Extension/package.nls.json
bobbrow: I understand JS/TS settings use "enable/disable" and we modeled these after theirs, but from a UX perspective, I think it is incorrect.

Extension/package.nls.json
michelleangela: Descriptions changed.

bobbrow:
The updates to the settings look good.  I'm signed off on that.  I did not review the rest of the code.

The only last thing to consider is whether you want to add a code sample to the description like the TS/JS settings have.  I think it would be a nice addition if you have time today.

michelleangela:
> The updates to the settings look good. I'm signed off on that. I did not review the rest of the code.
> 
> The only last thing to consider is whether you want to add a code sample to the description like the TS/JS settings have. I think it would be a nice addition if you have time today.

I was able to add it. I just figured out how. Needed to add the "cpp" specifier to ```.

sean-mcmanus (approved):

PR #9480, opened 2022-06-20
Reduce default C++ inlay hint font size to match VS and potentially reduce confusion

sean-mcmanus:
The smaller font size makes it more obvious that the inlay hint is not actual editor text, and hopefully users will realize that the type data shown is the complete type data (existing type data should be ignored) and not intended to be mentally combined with the existing type data on the left, e.g. in the example below, it's not a pointer to a const, but a const pointer to a non-const int:

image
This doesn't affect inlay hints for other languages.

UPDATE: Size 11 seemed slightly too small and the VS UI is in flux, so it seems better to go with 12, which matches the 90% used by editor.codeLensFontSize in VS Code.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -4034,17 +4034,20 @@
       "[cpp]": {
         "editor.wordBasedSuggestions": false,
         "editor.suggest.insertMode": "replace",
-        "editor.semanticHighlighting.enabled": true
+        "editor.semanticHighlighting.enabled": true,
+        "editor.inlayHints.fontSize": 11
       },
       "[cuda-cpp]": {
         "editor.wordBasedSuggestions": false,
         "editor.suggest.insertMode": "replace",
-        "editor.semanticHighlighting.enabled": true
+        "editor.semanticHighlighting.enabled": true,
+        "editor.inlayHints.fontSize": 11
       },
       "[c]": {
         "editor.wordBasedSuggestions": false,
         "editor.suggest.insertMode": "replace",
-        "editor.semanticHighlighting.enabled": true
+        "editor.semanticHighlighting.enabled": true,
+        "editor.inlayHints.fontSize": 11
       }
     },
     "semanticTokenTypes": [
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -4035,19 +4035,19 @@
         "editor.wordBasedSuggestions": false,
         "editor.suggest.insertMode": "replace",
         "editor.semanticHighlighting.enabled": true,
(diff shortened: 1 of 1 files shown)


Extension/package.json
```diff
@@ -4034,17 +4034,20 @@
       "[cpp]": {
         "editor.wordBasedSuggestions": false,
         "editor.suggest.insertMode": "replace",
-        "editor.semanticHighlighting.enabled": true
+        "editor.semanticHighlighting.enabled": true,
+        "editor.inlayHints.fontSize": 11
```
RaulPerez1: If you have nothing (like it is right now) is it inheriting from a default in VS Code? If so the better fix would be to have that fixed at that level so that you get consistent behavior across extensions.

RaulPerez1: Also meant to say, something that can be done separately, for now this change does what we want.

RaulPerez1 (approved):

Extension/package.json
bobbrow: Is this guaranteed to give us what we want?  What if the normal font size is already 11?

They should have done it like the code lens setting:
image

sean-mcmanus: Yeah, it sets it to 11 regardless of the editor.fontSize...not sure if that matters enough? Users who change the editor.fontSize can change the editor.inlayHints.fontSize for [cpp] etc. too. Did you have another suggestion and/or do you think I should raise some request on VS Code?

bobbrow: I sent a message to isidorn, but I think it would be good to open an issue as well.  If you do that, I'll forward the issue to him.

bobbrow: It's ok to ship this for now, I think, since we can undo it later.

sean-mcmanus: This is for 1.10.1 and not 1.10.0 so we can potentially get feedback both ways.

Also, the default editor.fontSize is 14, so VS is rendering the inlay hint at 80% of the editor font (my original screenshot comparison was comparing it with the incorrect default of 12). 

I've filed a feature request on VS at https://github.com/microsoft/vscode/issues/152690

sean-mcmanus: And FYI the difference in pixel height for the "c" of "const" is 8/9/10 pixels for 11/12/14 font sizes.

sean-mcmanus: FYI, I changed it from 11 to 12 -- I've updated the description and screenshots to explain why.

sean-mcmanus: I created a feature request on VS Code at https://github.com/microsoft/vscode/issues/152690

Issue #75, opened 2016-06-17
Allow breakpoints in other native languages source files


TurkeyMan:
C/C++ plugin integrates native debugging. I'm working on a mixed C++ and D project, but the IDE won't let me place breakpoints in `.d` files. I can step into those languages as you expect from GDB.
I suspect this may also be a problem for other native languages (rust, obj-c, etc?).

jhasse:
I have the same problem with Fortran files (`.f`, `.F90`).

TurkeyMan:
I just wanna ask if there's any movement on this? It's really holding up my work, and it seems like it should be a trivial fix?

jacdavis:
VSCode debugger extensions have to declare what files they allow breakpoints to be set in. This exists in the package.json file for the extension. If you add file extensions for the languages you care about in this section of ~/.vscode/extensions/ms-vscode.cpptools-0.9.2/package.json:
 "debuggers": [
      {
        "type": "cppdbg",
        "label": "C++ (GDB/LLDB)",
        "enableBreakpointsFor": {
          "languageIds": [
            "cpp",
            "c",
            ".h",
            ".cc",
            ".cxx",
            ".hpp",
            ".hh",
            ".hxx"
          ]
        },

Do things work better? Note that your milage may vary here, because you mentioned, we really only test with C/C++ with this extension.

jhasse:
I've added `.F90`, but still wasn't able to add breakpoints in .F90 files. I've disabled, reloaded, enabled, reloaded the C/C++ extension. Do I need to do anything else for these changes to take effect?

bobbrow:
package.json is not the correct location for this.  Have you tried manually adding an association to files.associations in your settings.json file? (File -> Preferences -> Settings)

```json
{
  "files.associations": {
    "*.F90": "cpp",
    "*.d": "cpp"
  }
}
```
We can't override the default file set that maps to "c" and "cpp" languageIds in our extension, so if these extensions are already mapped by VSCode, it might not work.  VSCode owns the mapping of file extensions to languageIds.

KytoDragon:
You can circumvent this with the following:
Got to File -> Preferences -> Settings -> Debug ->debug.allowBreakpointsEverywhere = true

Also #440 is a duplicate of this.

escape0707:
Wow, I'm surprised to see that this issue has been unsolved for so long. 

[...]

PR #9484, opened 2022-06-21
Contribute breakpoints for the "rust" language

chuckries:
Since rust code can be debugged with this extension, contribute breakpoints for the "rust" language such that the workaround of "allow breakpoints in all files" is no longer required.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3953,6 +3953,9 @@
       },
       {
         "language": "cuda"
+      },
+      {
+        "language": "rust"
       }
     ],
     "jsonValidation": [
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2871,7 +2871,8 @@
         "languages": [
           "c",
           "cpp",
-          "cuda-cpp"
+          "cuda-cpp",
+          "rust"
         ],
         "_aiKeyComment": "Ignore 'Property aiKey is not allowed'. See https://github.com/microsoft/vscode/issues/76493",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
@@ -3549,7 +3550,8 @@
         "languages": [
           "c",
           "cpp",
-          "cuda-cpp"
+          "cuda-cpp",
+          "rust"
         ],
         "_aiKeyComment": "Ignore 'Property aiKey is not allowed'. See https://github.com/microsoft/vscode/issues/76493",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",


chuckries:
fixes #75 for rust only

WardenGnaw:
@gregg-miskelly 

Do we want to set `contributes.debugger.languages: [ ..., "rust"]` to both debuggers?

chuckries:
I wasn't sure about "languages" as we don't offer any language features for rust? just bp support.

I tried disabling the rust-analyzer extension and turned off "bps can be set in any file" and was still able to set bps in .rs files.

chuckries:
@WardenGnaw @gregg-miskelly hm. It looks like, even if we contribute bp's and debuggers for "rust", we don't define the rust language or specify that it is .rs files (an extension like rust-analyzer does). Therefore, if you have cpptools installed, but not rust analyzer, you still can't set breakpoints in .rs files.

If you *do* have both installed, you can set bps in .rs files w/o modifying settings. Does this seem like what we want?

WardenGnaw:
Just adding here as a note to update the docs:
https://code.visualstudio.com/docs/languages/rust#_using-rust-analyzer-debug

image

sean-mcmanus (approved):

gregg-miskelly:
> Do we want to set `contributes.debugger.languages: [ ..., "rust"]` to both debuggers?

I think so.

gregg-miskelly (approved):

WardenGnaw:
Also add ` "onLanguage:rust"`, so the extension activates when selecting a rust file.

https://github.com/microsoft/vscode-cpptools/blob/4a85594360e7d0832469a1fd85edb96ff53a2e48/Extension/package.json#L50

gregg-miskelly:
> Also add ` "onLanguage:rust"`, so the extension activates when selecting a rust file.
> 
> https://github.com/microsoft/vscode-cpptools/blob/4a85594360e7d0832469a1fd85edb96ff53a2e48/Extension/package.json#L50

Is there a reason we need to? I didn't see any issues without this.

chuckries:
@gregg-miskelly at least from my experience, the debugger wouldn't activate if only viewing rust files, which prevented C++ from showing up as a launch.json template source. I'd need to test again to confirm.

chuckries:
@gregg-miskelly selecting Run and Debug with a .rs file active: 
image

selecting it with a .cpp file active:
image

I'm not actually sure what the correct fix is here. I thought it was an issue of extension activation but it appears based on the active editor. C++ doesn't show up for the .rs file even after opening a .cpp file.

chuckries:
@gregg-miskelly @WardenGnaw even if I can get the C++ options to pop up from Run and Debug and I select C++ (Windows), then select Default Configuration, I get this:
image

I'm not sure how far we want to go at this point, given that a lot of this starts to get C++ specific. It'd be nice to get a default config working for Rust, but I feel like its probably non trivial to stand up and support?

WardenGnaw:
I believe this is due to the changes of `Run and Debug` by @elahehrashedi .

A lot of the config generation is done by the C++ language service now.

chuckries:
@gregg-miskelly I'm curious how you started debugging rust if at all? Did you manually create a launch.json? Were you just using the inline hints provided by rust-analyzer?

gregg-miskelly:
> @gregg-miskelly at least from my experience, the debugger wouldn't activate if only viewing rust files, which prevented C++ from showing up as a launch.json template source. I'd need to test again to confirm.

That is what I saw too. But I wasn't sure if we would want the C++ templates to show up or not. If we do want it, I wonder if a better option is to activate on `onDebugInitialConfigurations`, which is what the C# extension does (see [here](https://github.com/OmniSharp/omnisharp-vscode/blob/a9d0b56f35e0433d72673deffbdf6e5063f87313/package.json#L582)).

The reason we might NOT want the C++ templates to show up is that the rust-analyze extension does offer their own templates through the 'Rust Analyzer: Debug' command.

gregg-miskelly:
> @gregg-miskelly I'm curious how you started debugging rust if at all? Did you manually create a launch.json? Were you just using the inline hints provided by rust-analyzer?

I tried it three different ways --
1. I manually created a launch.json before I read the docs.
2. I used 'Rust Analyzer: Debug' to generate a launch.json for me. That was fine, except it didn't add `"preLaunchTask": "rust: cargo build"` like I wanted.
3. I used 'Rust Analyzer: Debug' to debug without a launch.json

WardenGnaw (approved):

sean-mcmanus:
This change is available in 1.11.2 (pre-release): https://github.com/microsoft/vscode-cpptools/releases

sean-mcmanus:
It's actually in 1.10.8 (release) too: https://github.com/microsoft/vscode-cpptools/releases/tag/v1.10.8

PR #9500, opened 2022-06-24
Add setting to hide leading underscores for inlay hints.

### LINKED ISSUE #9485: Add setting to remove "_" from the start of inlay hint parameters
Certain libraries, such as msvc headers have parameters that start with "_" -- we should just check for that and remove it, since it's not helpful to show in the inlay hint.

image

It affects gcc too:
image

And/or we could also file a bug on the STL to try to get them to remove the _ from all their parameter names, but I believe our own code base sometimes has parameters that start with `_` and others as well.

UPDATE: Seems like system libraries are required to use the underscores in the parameter names to avoid naming conflicts with user code (`#defines`).

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9485


Extension/src/LanguageServer/Providers/inlayHintProvider.ts
```diff
@@ -101,6 +101,11 @@ export class InlayHintsProvider implements vscode.InlayHintsProvider {
             let paramHintLabel: string = "";
             if (settings.inlayHintsParameterNames) {
                 paramHintLabel = (settings.inlayHintsParameterNamesSuppressName && hint.hasParamName) ? "" : hint.label;
+                if (paramHintLabel !== "" && settings.inlayHintsParameterNamesHideLeadingUnderscores) {
+                    while (paramHintLabel.startsWith('_')) {
```
bobbrow: Would it be better to find the index to the first non-underscore character and just do a single `substring` from there?  I know you showed an example of double underscores in some gcc libraries.

sean-mcmanus: I figure it doesn't matter, right? Most cases should be 1 or 2 underscores.

sean-mcmanus: I could change it if you want.

bobbrow: I think it's a good idea.  It also would make it easier to protect against a (probably rare) case that the argument is just named `_` or `__`. (i.e. don't do the substr if there are no non-underscore characters).

sean-mcmanus: Okay, good catch.

Colengms (approved):

PR #9506, opened 2022-06-24
Add setting to show auto inlay hints on the left.

### LINKED ISSUE #9478: Add setting to display "auto" inlay hints on the right of the variable
...which causes selection of the identifier to select the inlay hint too (https://github.com/microsoft/vscode/issues/152672)

The current design of the inlay hint is to describe the full type of the variable (i.e. the existing text of the type is supposed to be ignored) and not just adding info on what the "auto" is omitting (i.e. the existing type text and the inlay hint are not expected to be mentally merged together currently).

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9478


sean-mcmanus:
@Colengms Can you approve this? The C/C++ side changes got checked in already.

Colengms (approved):

sean-mcmanus:
Also, we could potentially change the default setting value later based on customer feedback and/or discussions with the VS PM team.

PR #9514, opened 2022-06-28
Change C_Cpp.debugShortcut to resource scope

sean-mcmanus:
Enables a potential fix for https://github.com/platformio/platformio-vscode-ide/issues/3229 .


bobbrow:
This will disable it for all [c] files if that Platform.io extension is installed though.  Is that what we want?

sean-mcmanus:
> @bobbrow That would disable it by default -- users could add and modify the setting themselves to re-enable it. Another option would be to make it a per-workspace or per-workspace folder setting.

Changing it to "language-override" is the "least permissive" solution, so we can change it to "window" or "resource" scope later on without breaking people if we find there is sufficient reason to.

Colengms (approved):

bobbrow:
What I mean is that if the platform.io extension so much as activates (even in a workspace that does not have a platform.io project), no [c] files in that workspace will have the run/debug button visible by default.

sean-mcmanus:
> What I mean is that if the platform.io extension so much as activates (even in a workspace that does not have a platform.io project), no [c] files in that workspace will have the run/debug button visible by default.

Yeah, that seems fine to me if that's what the PlatformIO extension wants -- users should be able to override it with
```json
"[c]": {
"C_Cpp.debugShortcut": true
}
```
if they really want.

bobbrow:
I know users can override it. I'm just saying that we risk getting more issues opened in our repository as a result.  I don't know how big their user base is because they generate c_cpp_properties.json instead of using the config provider, but maybe we could hide the button based on some other information they provide.  I'm more just hoping that overriding the setting in their package.json would be a temporary fix as opposed to a long term one.

sean-mcmanus:
> I know users can override it. I'm just saying that we risk getting more issues opened in our repository as a result. I don't know how big their user base is because they generate c_cpp_properties.json instead of using the config provider, but maybe we could hide the button based on some other information they provide. I'm more just hoping that overriding the setting in their package.json would be a temporary fix as opposed to a long term one.

Yeah, it could be a temporary fix if we had some other mechanism to signal that a workspace should not show the debug buttons, potentially a new property in c_cpp_properties.json, but then if they switched to being a configuration provider then that wouldn't work (unless the configuration provider API also added a proprerty), so I'm hesitant to do that.

sean-mcmanus:
The change didn't work for some unknown reason. Also, I found out that language-overridable is actually a subset of "resource" scope, so I just changed it to resource scope, so I'm checking if that works instead.

sean-mcmanus:
Don't check this in until I can verify if it works -- it may not.

sean-mcmanus:
image

I've verified that it works -- extensions can override the default and users can change it explicitly.

sean-mcmanus:
@bobbrow Also, extension can already override our default settings even with the pre-existing "application" scope -- VS Code just shows it's not allowed in the package.json (not sure what effect that has though) and users can only override it in the application scope, so this change just got rid of the warning shown in the package.json for extension developers and enables users to change it on a per-workspace (folder) basis.

PR #9517, opened 2022-06-29
Update changelog for 1.11.1

sean-mcmanus:


sean-mcmanus:
@Colengms @bobbrow I'm checking this in now, but feel free to do a follow up PR to fix any issues with it.

PR #9524, opened 2022-06-29
Update to latest vscode-languageclient

Colengms:


Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -74,7 +74,7 @@ export function createProtocolFilter(clients: ClientCollection): Middleware {
                 // didOpen), and first becomes visible.
             }
         },
-        didChange: (textDocumentChangeEvent, sendMessage) => {
+        didChange: async (textDocumentChangeEvent, sendMessage) => {
```
bobbrow, line 77: Functions should only be marked `async` if they use `await` somewhere in the body.  Does this change the return type from `void` to `Promise<void>` without an `await`?

Colengms: Yes, this changes the return type from `void` to a `Promise<void>` automatically, and resolves the compilation error.  Alternatively, I could return a resolved `Promise<void>`, but this seems cleaner.

sean-mcmanus:
What milestone/release is this targetting?

bobbrow:
This can wait.

sean-mcmanus (approved):

PR #9544, opened 2022-07-06
Fix cl.exe being excluded for .c files and c/cpp caching.

sean-mcmanus:
1. Fix cl.exe build tasks not showing for .c files.
2. Fix .c build tasks being cached for .cpp files (and vice versa).

I hit these bugs while investigating https://github.com/microsoft/vscode-cpptools/discussions/9535 .

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -482,13 +482,13 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         if (fileIsCpp) {
             if (!DebugConfigurationProvider.detectedCppBuildTasks || DebugConfigurationProvider.detectedCppBuildTasks.length === 0) {
                 DebugConfigurationProvider.detectedCppBuildTasks = await cppBuildTaskProvider.getTasks(true);
-                DebugConfigurationProvider.detectedBuildTasks = DebugConfigurationProvider.detectedCppBuildTasks;
             }
+            DebugConfigurationProvider.detectedBuildTasks = DebugConfigurationProvider.detectedCppBuildTasks;
         } else {
             if (!DebugConfigurationProvider.detectedCBuildTasks || DebugConfigurationProvider.detectedCBuildTasks.length === 0) {
                 DebugConfigurationProvider.detectedCBuildTasks = await cppBuildTaskProvider.getTasks(true);
-                DebugConfigurationProvider.detectedBuildTasks = DebugConfigurationProvider.detectedCBuildTasks;
             }
+            DebugConfigurationProvider.detectedBuildTasks = DebugConfigurationProvider.detectedCBuildTasks;
         }
     }
 
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -38,6 +38,8 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     private assetProvider: IConfigurationAssetProvider;
     // Keep a list of tasks detected by cppBuildTaskProvider.
     private static detectedBuildTasks: CppBuildTask[] = [];
+    private static detectedCppBuildTasks: CppBuildTask[] = [];
+    private static detectedCBuildTasks: CppBuildTask[] = [];
     protected static recentBuildTaskLabel: string;
 
     public constructor(assetProvider: IConfigurationAssetProvider, type: DebuggerType) {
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -122,10 +122,19 @@ export class CppBuildTaskProvider implements TaskProvider {
...
         let knownCompilers: configs.KnownCompiler[] | undefined = await activeClient.getKnownCompilers();
         if (knownCompilers) {
-            const compiler_condition: (info: configs.KnownCompiler) => boolean = info => ((fileIsCpp && !info.isC) || (fileIsC && info.isC)) &&
-                (!isCompilerValid || (!!userCompilerPathAndArgs &&
-                (path.basename(info.path) !== userCompilerPathAndArgs.compilerName))) &&
-                (!isWindows || !info.path.startsWith("/")); // TODO: Add WSL compiler support.
+            const compiler_condition: (info: configs.KnownCompiler) => boolean = info =>
+                (
+                    // Filter out c compilers for cpp files and vice versa, except for cl.exe, which handles both.
+                    path.basename(info.path) === "cl.exe" ||
```
sean-mcmanus, line 128: This check fixes cl.exe not appearing for .c files.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -449,8 +451,44 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     }
 
     private async loadDetectedTasks(): Promise<void> {
-        if (!DebugConfigurationProvider.detectedBuildTasks || DebugConfigurationProvider.detectedBuildTasks.length === 0) {
-            DebugConfigurationProvider.detectedBuildTasks = await cppBuildTaskProvider.getTasks(true);
+        const editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
```
sean-mcmanus, line 454: These checks are taken from the start of getTask -- I didn't see a good way to refactor it so I didn't bother.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -449,8 +451,44 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+
+        // Don't offer tasks if the active file's extension is not a recognized C/C++ extension.
+        const fileIsCpp: boolean = util.isCppFile(editor.document.uri);
+        const fileIsC: boolean = util.isCFile(editor.document.uri);
+        if (!(fileIsCpp || fileIsC)) {
+            DebugConfigurationProvider.detectedBuildTasks = emptyTasks;
+            return;
+        }
+
+        if (fileIsCpp) {
```
sean-mcmanus, line 482: This fixes the same detectedBuildTasks cache being used for c and cpp.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -38,6 +38,8 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     private assetProvider: IConfigurationAssetProvider;
     // Keep a list of tasks detected by cppBuildTaskProvider.
     private static detectedBuildTasks: CppBuildTask[] = [];
```
Colengms, line 40: Is this 3rd array still needed?

Colengms (changes requested):

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: Yeah, that's the one that is actually used. The c or cpp cache gets copied into it depending on the current file's type.

Colengms (approved):

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: One way is not to keep a cache and always load the tasks. It may help when a new compiler is being installed in the middle of a session. I added this cache so that we don't keep getting the tasks over and over again, but now I am not sure that it was the right decision. Which way you think is better? to keep the cache or to always load the tasks?

sean-mcmanus: I don't know. Using caches seems fine to me -- I see the cache get hit 2 times for every debug invocation and it's possible the list of compilers could be big, so it's hard to say how long that could potentially add to debug startup time. Installing a new compiler in the middle of a sessions seems pretty rare, and a user who did that and didn't see the new compiler added would probably guess they might need to reload VS Code.

Issue #9505, opened 2022-06-24
Support attach process on a remote target with remote/extended-remote mode


fbs2016:
### Feature Request

Now cpptools support attach local process, but don't support attach a process with remote target, both "target remote" and "target extended-remote" mode.

Not Pipe transport, the target don't support ssh connection, just like launch, we can connect to target with miDebuggerServerAddress and connect to gdbserver directly.
From gdb command, attach function just one command after target connection, like "attach xxx".

Refer to "Native Debug", it supports it now.

fbs2016:
https://github.com/microsoft/vscode-cpptools/issues/4166

fbs2016:
https://github.com/microsoft/vscode-cpptools/issues/78

sean-mcmanus:
Is this resolved with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.11.3 ? I'm not sure why there are 4 issues tracking the same issue. Let us know if something isn't fully resolved.

fbs2016:
@sean-mcmanus  Thanks.  I try the new option "useExtendedRemote" it works for extended-remote launch. But it still has some issue with attach function.

xisui-MSFT:
@fbs2016 Can you explain your scenario in more detail?

Based on your description, seem like your scenario is similar to:
1. Start gdbserver on the target, via "gdbserver <hostName>:<port> --multi"
2. In launch.json, set miDebuggerServerAddress to <hostName>:<port> and useExtendedRemote to true. If you don't specify process ID, a process picker will be shown. This is equivalent to `gdb` -> `target extended-remote <hostname>:<port>` -> `attach <PID>`.

This is supported though. What are the differences between this and your case?

fbs2016:
@xisui-MSFT  
Thanks
We have 2 ways to start gdbserver, gdbserver --multi or --once:port exeFile on remote target.

Next we want to attach the running process from VSC cpptools extention on local host. How to configure it? The processId in launch.json is process id on localhost but I want to attach the processs on remote target.

From gdb client, the command is simple just like you described : gdb-> target remote/extended-remote targetIp:gdbserverPort->attach pId in target

I can't find a right way to run it from VSC.

xisui-MSFT:
@fbs2016 If you are using latest cpptools, you don't need to provide process ID anymore. A quick pick menu will be shown with processes on the target machine. Note that you'll need to use `--multi` for this since gdb need to connect once to fetch the processes.

[...]

Issue #9195, opened 2022-04-16
cppdbg `miDebuggerServerAddress` parameter generates warning


zfields:
Using `  "miDebuggerServerAddress": "localhost:3333"` seems to have the same effect as setting `target remote :3333` in the `postRemoteConnectCommands`, which generates the following warning:

```none
openocd: Warn : Prefer GDB command "target extended-remote 3333" instead of "target remote 3333"
```

However, setting `target extended-remote :3333` in the `postRemoteConnectCommands` and removing `miDebuggerServerAddress` eliminates the warning.

Is this an oversight, or is it by design?

I'm using an ESP32 WROOM-32D and the ESP Prog debugger, as well as the `openocd` and `gdb` provided by the latest version of ESP-IDF.

*'launch.json'*
```json
{
	"name": "ESP-Prog Debug (C/C++)",
	"type": "cppdbg",
	"request": "launch",
	"program": "${workspaceRoot}/Radar/build/esp_prog_vscode_debug.elf",
	"args": [],
	"cwd": "${workspaceRoot}/Radar",
	"environment": [
		{
			"name": "PATH",
			"value": "${config:esp_toolchain_paths}:${env:PATH}"
		}
	],
	"externalConsole": false,
	"MIMode": "gdb",
	"miDebuggerPath": "${config:esp_gdb_path}/xtensa-esp32-elf-gdb",
	"debugServerPath": "${config:esp_openocd_path}/openocd",
	"debugServerArgs": "-f interface/ftdi/esp32_devkitj_v1.cfg -f target/esp32.cfg -c \"esp32.cpu1 configure -rtos FreeRTOS\"",
	"serverStarted": "Listening on port [0-9]+ for gdb connections",
	"filterStderr": true,
	"filterStdout": false,
	"launchCompleteCommand": "None",
	"postRemoteConnectCommands": [
		{
			"description": "Target Remote Device on Port 3333",
			"text": "target extended-remote :3333",
			"ignoreFailures": false
		},
		{
			"description": "Respect Hardware Limitations (as prescribed by Espressif)",
			"text": "set remote hardware-watchpoint-limit 2",
			"ignoreFailures": false
		},
		{
			"description": "Hard Reset and Immediately Halt",
			"text": "monitor reset halt",
			"ignoreFailures": false
		},
		{
			"description": "Flush Internal Register Cache",
			"text": "flushregs",
			"ignoreFailures": false
		},
		{
[...]

Issue #8497, opened 2021-12-07
Progress picker not showing the remote progresses


xisui-MSFT:
Bug type: Debugger

**Describe the bug**
- OS and Version: Windows 10, 19044.1348
- VS Code Version: 1.62.3
- C/C++ Extension Version: 1.7.1
- Other extensions you installed (and if the issue persists after disabling them):
- A clear and concise description of what the bug is.
When specifying a remote address in launch.json, like `"miDebuggerServerAddress": "192.168.35.2:6000"`, and setting `requestType` to `attach`, the progress picker shown is for local processes. We need the remote processes instead (with a command like `info os processes` for gdb).

**To Reproduce**
1. Setup a GDB server on a remote machine
2. Add a local launch file, that specifies `miDebuggerServerAddress`, with `requestType` set to `attach`.
3. Start debugging. Expecting to see a progress picker for a remote machine,

WardenGnaw:
For now, the remote process picker does not work with `miDebugServerAddress`, you will need to use `pipeTransport` for the extension to communicate with the machine.

xisui-MSFT:
@WardenGnaw Looks like pipeTransport requires a pipe program https://code.visualstudio.com/docs/cpp/pipe-transport, but we are using GDB on the local machine. Sample command:
```
gdb \
<binary-on-host-with debug-info\> \
   -ex "set sysroot ${SYSROOT}" \
   -ex "set solib-search-path ${SYSROOT}" \
   -ex "set debug-file-directory ${SYSROOT}" \
   -ex "target extended-remote 192.168.35.2:6000" \
   -ex "info os processes"
```
This command (`info os processes`) prints the processes on the remote machine. Also I think it's just a gdb server on the remote machine so ssh probably won't work.

Any suggestions?

WardenGnaw:
Unfortunately, you will have to supply `processId` manually. 

The VS Code extension does not talk to GDB directly and the process picker can only work with a connection to the terminal so it can use `ps` to get process information.

xisui-MSFT:
Is it possible to modify MIEngine somehow to have an option to get processes via GDB instead?

WardenGnaw:
[...]

PR #9569, opened 2022-07-11
Extended remote support

xisui-MSFT:
Extended remote support #9491 #9505 #8497 #9195
Also show process picker automatically when the process ID is not provided in launch.json


xisui-MSFT:
Forgot to run linter...

sean-mcmanus:
What release is this for?

xisui-MSFT:
> What release is this for?

Extended remote was added to MIEngine several months ago: https://github.com/microsoft/MIEngine/pull/1260, and this doesn't have any other dependencies. So ideally we would like to target the earliest release we can achieve.

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -167,6 +177,58 @@ export class RemoteAttachPicker {
...
+    private parseProcessesFromInfoOsProcesses(processList: string): AttachItem[] {
+        const lines: string[] = processList?.split('\n');
+        if (!lines?.length) {
+            return [];
+        }
+
+        const processes: AttachItem[] = [];
+        for (const line of lines) {
+            const trimmedLine: string = line.trim();
+            if (!trimmedLine.endsWith('?') && !trimmedLine.includes('/usr/bin/gdbserver')) {
```
WardenGnaw: Will there be an issue if gdbserver is not in /usr/bin/ but in another path?

WardenGnaw (approved):

Extension/src/Debugger/attachToProcess.ts
xisui-MSFT: There won't be an issue but gdbserver will show up in the process picker.

That said... People probably will use gdbserver to debug another gdbserver process... So I'll remove this check.

xisui-MSFT:
> @xisui-MSFT We tend to use back-tick quoting in our settings descriptions (which render as markdown) to clarify for translators that they shouldn't translate specific words (such as `true` and `false`, which are the values required in JSON settings). You might consider making this clear in your strings. I see your existing use of true/false getting localized by some translators.

Can you give an example? I saw formats like {0}, so all relevant strings in this PR uses this format.

sean-mcmanus:
> > @xisui-MSFT We tend to use back-tick quoting in our settings descriptions (which render as markdown) to clarify for translators that they shouldn't translate specific words (such as `true` and `false`, which are the values required in JSON settings). You might consider making this clear in your strings. I see your existing use of true/false getting localized by some translators.
> 
> Can you give an example? I saw formats like {0}, so all relevant strings in this PR uses this format.

In https://github.com/microsoft/vscode-cpptools/blob/main/Extension/package.nls.json .

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -39,48 +39,58 @@ export class RemoteAttachPicker {
...
+            const argList: string = RemoteAttachPicker.createArgumentList(pipeArgs);
 
-        const argList: string = RemoteAttachPicker.createArgumentList(pipeArgs);
+            const pipeCmd: string = `"${pipeProgram}" ${argList}`;
 
-        const pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            processes = await this.getRemoteOSAndProcesses(pipeCmd);
+        } else if (!pipeTransport && useExtendedRemote) {
+            if (!miDebuggerPath || !miDebuggerServerAddress) {
+                throw new Error(localize("debugger.path.and.server.address.requried", "{0} in debug configuration requires {1} and {2}", "useExtendedRemote", "miDebuggerPath", "miDebuggerServerAddress"));
```
jogo-: Typo: requ**ri**ed -> requ**ir**ed

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -167,6 +177,58 @@ export class RemoteAttachPicker {
...
+
+    /**
+    Format:
+    pid      usr      command     cores
+    1        ?
+    2        ?
+    3                 /usr/bin/sample 0
+    4        root     /usr/bin/gdbserver --multi :6000 0
+
+    Returns aa array AttachItems, each one contains a label of "<user   >command", and a pid.
```
jogo-: Typo: a**a** -> a**n**

Extension/src/Debugger/attachToProcess.ts
```diff
@@ -167,6 +177,58 @@ export class RemoteAttachPicker {
...
+    Unfortunately becuase the format of each line is not fixed, and everything except pid is optional, it's hard
```
jogo-: Typo: bec**ua**se -> bec**au**se

michelleangela (approved):

sean-mcmanus:
@xisui-MSFT Is this going to be checked in? We're planning to release 1.11.3 on Monday. Does the changelog need to be updated?

sean-mcmanus (approved):

PR #9581, opened 2022-07-14
Show a better error message when a debugger is not installed

### LINKED ISSUE #8836: Misleading error message when GDB is not installed
Bug type: Debugger

If I don't have GDB installed (yet), and try to use the 'build and debug active file' feature (on Linux, with g++), I get an error message: 

image

However, all I have at this point is `tasks.json` (not `launch.json`, which it refs to), and there is no reference to `miDebuggerPath` in it.

It would be useful if this error message provided enough information to determine that the issue is that GDB needs to be installed.

ChaseKnowlden:
Resolves: #8836

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -857,7 +857,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
     public async buildAndDebug(textEditor: vscode.TextEditor, debugModeOn: boolean = true): Promise<void> {
         let folder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(textEditor.document.uri);
-        const err: boolean = false;
+        let err: boolean = false;
 
         const selectedConfig: CppDebugConfiguration | undefined = await this.selectConfiguration(textEditor);
         if (!selectedConfig) {
@@ -876,6 +876,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         selectedConfig.debugType = debugModeOn ? DebugType.debug : DebugType.run;
         // startDebugging will trigger a call to resolveDebugConfiguration.
         if (err) {
+            err = true;
             throw new Error(localize("debugger.not.installed", "The debugger you're using is not installed. Please install the debugger first."));
         } else {
             await vscode.debug.startDebugging(folder, selectedConfig, { noDebug: !debugModeOn });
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
@WardenGnaw Can you code review this? We plan to do a release on Monday.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -874,6 +884,10 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         selectedConfig.debugType = debugModeOn ? DebugType.debug : DebugType.run;
         // startDebugging will trigger a call to resolveDebugConfiguration.
         await vscode.debug.startDebugging(folder, selectedConfig, { noDebug: !debugModeOn });
+
+        if (error) {
```
WardenGnaw: When is `error` re-evaluated?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -779,6 +779,14 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         return util.getRawJson(path);
     }
 
+    private showUnifiedErrorMessage(error: any): string {
+        let message: string = error.message;
+        if (error.stack) {
+            message = localize('launch.error.stack', "The stack trace was: {0}", error.stack);
```
WardenGnaw: We should also include the original `error.message` with the stacktrace. I think you wanted to append this.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -874,6 +884,10 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         selectedConfig.debugType = debugModeOn ? DebugType.debug : DebugType.run;
         // startDebugging will trigger a call to resolveDebugConfiguration.
         await vscode.debug.startDebugging(folder, selectedConfig, { noDebug: !debugModeOn });
+
+        if (error) {
+            this.showUnifiedErrorMessage(errorMessage);
```
WardenGnaw: `showUnifiedErrorMessage` looks like it takes an object. However, you are passing in a string and trying to use `errorMessage.message` and `errorMessage.stack`. 

Can you define the class you expect instead of using `any`?

WardenGnaw (changes requested):

Extension/src/Debugger/configurationProvider.ts
ChaseKnowlden: after it starts debugging.

Extension/src/Debugger/configurationProvider.ts
ChaseKnowlden: Sounds good!

Extension/src/Debugger/configurationProvider.ts
ChaseKnowlden: I'll try.

Extension/src/Debugger/configurationProvider.ts
WardenGnaw: `error` shouldn't magically be assigned in this method.

Did you want to get the boolean from `startDebugging`?

```
error = await vscode.debug.startDebugging(folder, selectedConfig, { noDebug: !debugModeOn });
```

Method definition:
>  startDebugging(folder: vscode.WorkspaceFolder | undefined, nameOrConfig: string | vscode.DebugConfiguration, options: vscode.DebugSessionOptions): Promise<boolean>

ChaseKnowlden:
Closing

PR #9622, opened 2022-07-21
Add inlay hint tests, fix setting update of autoDeclarationTypes.showOnLeft

michelleangela:
- Fix setting update of `autoDeclarationTypes.showOnLeft`.
- Add inlay hint tests to E2E intelliSense tests. (The E2E intelliSense tests only run once a week internally)

### DIFF
--- a/Extension/src/LanguageServer/Providers/inlayHintProvider.ts
+++ b/Extension/src/LanguageServer/Providers/inlayHintProvider.ts
@@ -35,7 +35,7 @@ interface GetInlayHintsResult {
 
 type InlayHintsCacheEntry = {
     FileVersion: number;
-    TypeHints: vscode.InlayHint[];
+    TypeHints: CppInlayHint[];
     ParameterHints: CppInlayHint[];
 };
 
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
@michelleangela Is the test passing or failing? I see the last run failing on Linux:
  1) [Inlay hints test]
       [Inlay Hints - parameter names]:

      AssertionError [ERR_ASSERTION]: Incorrect number of results.
      + expected - actual

also, one my changes caused the Windows E2E binary test to fail so the typescript tests aren't running -- I'm not sure why there's a difference test output with the E2E version versus our primary test pass, i.e. it seems like they should have the same results and I haven't seen what a difference is yet.

sean-mcmanus:
Also -- why is the root test task showing as a pass when the inlay hint test fails?

sean-mcmanus:
If you want to me to get the Windows E2E test to pass right away you or I could potentially add a fallback baseline for the failing tests until I can figure out what is causing the difference (not sure when that will be yet...hopefully next week).

sean-mcmanus:
> If you want to me to get the Windows E2E test to pass right away you or I could potentially add a fallback baseline for the failing tests until I can figure out what is causing the difference (not sure when that will be yet...hopefully next week).

Oh, I believe the test failure is caused by a difference between the use of git clone and a non-git clone checkout leading to \r\n versus \n differences -- I think I can just add a fallback to handle both cases.

michelleangela:
> @michelleangela Is the test passing or failing? I see the last run failing on Linux:
> 
> 1. [Inlay hints test]
>    [Inlay Hints - parameter names]:
>    AssertionError [ERR_ASSERTION]: Incorrect number of results.
>    
>    * expected - actual
>      -20
>      +16
> 
> also, one of my changes caused the Windows E2E binary test to fail so the Windows typescript tests aren't running -- I'm not sure why there's a difference test output with the E2E version versus our primary test pass, i.e. it seems like they should have the same results and I haven't seen what a difference is yet.

The E2E test on Linux should show fail for that one test case not passing. The test case "[Inlay Hints - parameter name]" is supposed to fail until the issue for split parameter is resolved.

sean-mcmanus (approved):

PR #9624, opened 2022-07-22
Fix inlay hints Chinese translate mistakes.

kouhe3:
fix : 
- [x] Improper translation of 'suppress'
- [x] Should not translate `const`, `auto index` , `swap`


sean-mcmanus (approved):

michelleangela:
Do we need to file an internal bug to the loc team for this so they can correct it on their source code? Otherwises the next time they provide updates to the loc files, this will get reverted.

sean-mcmanus:
> Do we need to file an internal bug to the loc team for this so they can correct it on their source control? Otherwise, the next time they provide updates to the loc files, this will get reverted.

Yes, @Colengms usually handles that.

michelleangela (approved):

Colengms:
> > Do we need to file an internal bug to the loc team for this so they can correct it on their source control? Otherwise, the next time they provide updates to the loc files, this will get reverted.
> 
> Yes, @Colengms usually handles that.

This can go in.  I will open bugs for this on my next pass through the loc strings (as I'll find more issues then and open a bug per language).

PR #9633, opened 2022-07-23
Add tag parsing feature

### LINKED ISSUE #9570: Add some C_Cpp.tagParseAsC setting to enable C mode tag parsing
### Feature Request

Currently we tag parse all files as C++, but this is buggy when a C identifier matches a C++ keyword. We could use add some setting to switch it to tag parse as C instead, which would be useful for C projects. Not sure if this "C-ness" of a project should be generalized though, beyond tag parsing, e.g. C_Cpp.useCMode. It's possible this may also cause unexpected issues with tag parsing headers that contain C++ (potentially new crash regressions).

There are multiple previous issues related to this. VS has Won't Fixed the issue, but I don't think they have as many C users compared to C++.

ChaseKnowlden:
Adds a new C++ feature.

Resolve: #9570


Extension/src/LanguageServer/client.ts
```diff
@@ -1414,6 +1414,9 @@ export class DefaultClient implements Client {
                 references: {
                     ...vscode.workspace.getConfiguration("C_Cpp.references", this.RootUri)
                 },
+                tagParseAsC: {
```
sean-mcmanus, line 1417: This is not required. This would only be used if "tagParseAsC" had additional settings on the right of a ".". Also, the setting isn't being passed to the language server during initialization.

sean-mcmanus (changes requested):
This feature can't be implemented without changes to our closed source code to handle the new tagParseAsC setting. Until we make those changes, this PR just adds a setting that isn't used.

ChaseKnowlden:
I don't have access to the code!

ChaseKnowlden:
Gotta close anyway because of that comment.

Issue #9614, opened 2022-07-21
Show a more informative error message when trying to use IntelliSense-related commands when IntelliSense is Disabled.


bferguson3:
### Environment

- OS and version: Mac M1 Monterey 12.3
- VS Code: 1.69
- C/C++ extension: 1.10.8
- OS and version of remote machine (if applicable): n/a

### Bug Summary and Steps to Reproduce

Bug Summary:
As was reported several times before, out of nowhere I am getting the 'command 'C_Cpp.ResetDatabase' not found' error. This same style of error occurs for every C/C++ extension command. 

Steps to reproduce:
1. Delete all configuration and cache files for VS Code *everywhere* 
2. Install newest VS Code and newest C/C++ extension
3. Install recommended files for Carbon 
refer to https://github.com/carbon-language/carbon-lang/blob/trunk/.vscode/extensions.json
clangd required a configuration change. This likely broke c/cpp extension.
4. Open a C/C++ file and attempt to run the command Reset Intellisense Database 
5. Observe the error occurs

### Other Extensions

https://github.com/carbon-language/carbon-lang/blob/trunk/.vscode/extensions.json

### Additional Information

This issue persists after a reboot and after deleting every configuration file I can find

Colengms:
Hi @bferguson3 .  When  you installed the clangd extension, it would have prompted you with something like the following:

![image](https://user-images.githubusercontent.com/49173979/180317085-5487b59f-ddc9-461c-8a3e-6873588626b6.png)

Assuming you clicked on `Disable IntelliSense`, that would have disabled the majority of the functions of the C/C++ Extension.  As a result, many of the C/C++ Extension commands involving IntelliSense will fail and trigger the error dialog you referred to.

We can use this issue to track displaying a more informative error message in this scenario.  Though, the non-functionality of those features is intentional when IntelliSense is disabled.

bferguson3:
Hi Colen,

Thanks for the reply. Yes, that was the message I saw.

Please re-read the ticket. I have done a complete removal of all
extensions, VS Code itself, and cache subdirectories. The issue persists.

When you get a chance to look more closely let me know.

sean-mcmanus:
@bferguson3 Do you have the C_Cpp.intelliSenseEngine set to "Disabled" in your settings?

Colengms:
[...]

PR #9640, opened 2022-07-26
Display warning for commands when IntelliSenseEngine is Disabled

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/9614

This is implemented in such a way that registering a new command still requires adding only 1 line, and it's not necessary to maintain two separate lists of commands.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -38,6 +38,7 @@ let clients: ClientCollection;
 let activeDocument: string;
 let ui: UI;
 const disposables: vscode.Disposable[] = [];
+const commandDisposables: vscode.Disposable[] = [];
 let languageConfigurations: vscode.Disposable[] = [];
 let intervalTimer: NodeJS.Timer;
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -383,53 +384,66 @@ function onInterval(): void {
...
+
+function onDisabledCommand(): void {
+    const message: string = localize(
+        {
+            key: "on.disabled.command",
+            comment: [
+                "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
+            ]
+        },
+        "IntelliSense related commands cannot be executed when `C_Cpp.intelliSenseEngine` is set to `Disabled`");
```
sean-mcmanus: IntelliSense-related

sean-mcmanus: I think it should end with a period.

sean-mcmanus (approved):

PR #9654, opened 2022-07-27
Cache and reuse SSH passwords in the current session

xisui-MSFT:
- Cache password when one is entered.
- Reuse this password if the user match in later commands.
- If we get a password prompt for the same user and hostName in the same commands, we likely got a wrong password, so ask for it again.
- Cached passwords are persisted via SecretStorage in extension context.
- Show a progress notification when running deploy steps to avoid confusion during waiting.


Extension/src/SSH/commandInteractors.ts
```diff
@@ -4,7 +4,14 @@
...
 import * as vscode from 'vscode';
-import { stripEscapeSequences, isWindows, escapeStringForRegex } from '../common';
+import { stripEscapeSequences, isWindows, escapeStringForRegex, ISshHostInfo, getFullHostAddress, extensionContext } from '../common';
+
+/**
+ * Users' passwords already used for auth.
+ * If a user's pass is already used and yet we still get the same prompt, we probably got a wrong password.
+ * Needs to be reset for each command.
+ */
+export const userUsedPasswords: Set<string> = new Set<string>();
```
WardenGnaw: nit: Naming suggests this a set of passwords. But it is a collection of users who have inputted passwords before.

I would name it `knownPasswordUsers` and clean-up the description to be:
```
The users that have already inputted a password for auth.
```

WardenGnaw (approved):

Extension/src/SSH/commandInteractors.ts
xisui-MSFT: Ah I see. I'll name it "autoFilledPasswordForUsers". Since this isn't for users that has their passwords cached, but for users we inputted passwords automatically.

WardenGnaw (approved):

sean-mcmanus (approved):

Issue #9643, opened 2022-07-26
No colors in task terminal with clang


skarasov:
Issue Type: <b>Bug</b>

Simple c code.
Run build task with clang or cl.
In task terminal no colored output, but in ordinary terminal all fine.
![b](https://user-images.githubusercontent.com/50782570/181044305-c3c691c6-208d-4d9e-8d74-1e3ed35c03b6.jpg)

refer:
https://github.com/microsoft/vscode/issues/155444

Extension version: 1.11.4
VS Code version: Code 1.69.2 (3b889b090b5ad5793f524b5d1d39fda662b96a2a, 2022-07-18T16:12:52.460Z)
OS version: Windows_NT x64 10.0.19044
Restricted Mode: No

sean-mcmanus:
The issue doesn't repro for me. What task type are you using in your task.json? Are you using arg " -fdiagnostics-color=always"?

skarasov:
No color only with clang.exe and cl.exe. With gcc.exe all fine.

task.json:
```
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "cppbuild",
			"label": "C/C++: clang.exe build active file",
			"command": "C:\\Program Files\\LLVM\\bin\\clang.exe",
			"args": [
				"-fdiagnostics-color=always",
				"-g",
				"${file}",
				"-o",
				"${fileDirname}\\${fileBasenameNoExtension}.exe"
			],
			"options": {
				"cwd": "${fileDirname}"
			},
			"problemMatcher": [
				"$gcc"
			],
			"group": {
				"isDefault": true,
				"kind": "build"
			},
			"detail": "compiler: \"C:\\Program Files\\LLVM\\bin\\clang.exe\""
		}
	]
}
```

Colengms:
Hi @skarasov .  

The root of this problem has to do with how gcc and clang detect whether or not they are running within a shell terminal.  If they detect they are in a shell terminal, they enable colorization.  We could consider it a VS Code bug that the terminal used for tasks does not look like a shell terminal to them.  I suspect it has something to do with how VS Code is redirecting and scanning the output of the task, circumventing the compiler's check.  We can work around this issue with gcc by forcing colorization to be enabled using `-fdiagnostics-color=always`.

After some experimentation, it looks like we can convince clang to also enable colorization using:

```
                "-fcolor-diagnostics",
                "-fansi-escape-codes",
```

[...]

PR #9655, opened 2022-07-27
Enable colorized output in tasks for clang

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/9643

### DIFF
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -178,7 +178,12 @@ export class CppBuildTaskProvider implements TaskProvider {
             const taskLabel: string = ((appendSourceToName && !compilerPathBase.startsWith(ext.configPrefix)) ?
                 ext.configPrefix : "") + compilerPathBase + " " + localize("build_active_file", "build active file");
             const programName: string = util.defaultExePath();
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -178,7 +178,12 @@ export class CppBuildTaskProvider implements TaskProvider {
             const taskLabel: string = ((appendSourceToName && !compilerPathBase.startsWith(ext.configPrefix)) ?
                 ext.configPrefix : "") + compilerPathBase + " " + localize("build_active_file", "build active file");
             const programName: string = util.defaultExePath();
-            let args: string[] = isCl ? ['/Zi', '/EHsc', '/nologo', `/Fe${programName}`, '${file}'] : ['-fdiagnostics-color=always', '-g', '${file}', '-o', programName];
+            const isClang: boolean = !isCl && compilerPathBase.toLowerCase().includes("clang");
```
Colengms, line 181: This is not comprehensive.  On macOS, there are symbolic links from `gcc` that point to clang.  Perhaps we could indicate whether a compiler is gcc, clang or cl in the list of compilers returned from the native side, as the native side can determine which is clang and gcc based on defines returned when querying them.

sean-mcmanus (approved):

michelleangela:
Will this only work if language server is not disabled?

PR #9661, opened 2022-07-29
Code analysis fix formatting.

### LINKED ISSUE #9322: Add auto-formatting support for clang-tidy fixes
Some clang-tidy fix types such as for readability-braces-around-statements generates fixes that need to be formatted in order to give acceptable results (otherwise, } will have no indentation). Potentially, we just run the currently configured formatting on the lines that are changed.

One workaround is to add `--format-style=` to C_Cpp.codeAnalysis.clangTidy.args if clang-format style formatting is desired. Or add FormatStyle to a .clang-tidy file. UPDATE: The workaround doesn't work.

UPDATE: I noticed with readability-else-after-return it is desirable to format the entire range, but the fix only applies it to the start/end lines...not sure if there's a way to improve that.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9322 .

Code action fixes are applied and then formatting is applied. The first undo undoes any formatting if any was done, otherwise it undoes the fix. If formatting was done, a 2nd undo will undo the fix. The operations apply across all files that have fixes generated for them.


sean-mcmanus:
Not 100% ready yet, still fixing bugs.

sean-mcmanus:
I'm investigating if there's a way to combine multiple "onChange" documentFormat requests into a single edit so they can be undone with a single undo instead of 1 per file.

sean-mcmanus:
I need to set tabSize and insertSpaces somehow when I invoke the format command...

Extension/package.nls.json
```diff
@@ -50,6 +50,7 @@
     "c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showClear.AllOnly.description": "Show only the 'Clear all' code action (or 'Clear all <type>' if there is only one type or 'Clear this' if there is only one problem).",
     "c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showClear.AllAndAllType.description": "Show the 'Clear all' code action (if there are multiple problem types) and the 'Clear all <type>' code action (or 'Clear this' if there is only one problem for the <type>)",
     "c_cpp.configuration.codeAnalysis.clangTidy.codeAction.showClear.AllAndAllTypeAndThis.description": "Show the 'Clear all' (if there are multiple problem types), 'Clear all <type>' (if there are multiple problems for the <type>), and 'Clear this' code actions",
+    "c_cpp.configuration.codeAnalysis.clangTidy.codeAction.formatFixes.markdownDescription": { "message": "If `true`, formatting will be run on the lines changed by 'Fix' code actions.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
michelleangela, line 53: Is "Fix" supposed to be code? It uses ' instead of `. Was this intended?

Extension/src/LanguageServer/Providers/documentFormattingEditProvider.ts
```diff
@@ -16,6 +16,34 @@ export class DocumentFormattingEditProvider implements vscode.DocumentFormatting
...
+                    }
+                    if (editor.options.tabSize && typeof editor.options.tabSize === "number") {
+                        options.tabSize = editor.options.tabSize;
+                        tabSizeSet = true;
+                    }
+                    break;
+                }
+            };
+
+            if (!insertSpacesSet || !tabSizeSet) {
```
michelleangela, line 37: Why do the two conditionals below need to be around the check "if (!insertSpacesSet || !tabSizeSet)" and then there are individual checks for each? It looks like unnecessary double checking.

Extension/package.nls.json
sean-mcmanus: It's supposed to be '' , it's not code but a command and we use '' for commands...follows what VS Code does.

Colengms (approved):

PR #9734, opened 2022-08-12
Rename doxygen settings

Colengms:


michelleangela (approved):

Extension/package.json
```diff
@@ -2333,21 +2333,21 @@
           "markdownDescription": "%c_cpp.configuration.simplifyStructuredComments.markdownDescription%",
           "scope": "application"
         },
-        "C_Cpp.autocompleteDoxygenComment":{
-          "type":"boolean",
-          "default":true,
-          "description": "%c_cpp.configuration.autocompleteDoxygenComment.description%",
+        "C_Cpp.generateDoxygenCommentsOnTyping":{
```
bobbrow: Should we do a category for doxygen settings?

e.g. `C_Cpp.doxygen.generateOnType`, `C_Cpp.doxygen.commentStyle`

bobbrow (changes requested):

Extension/package.json
sean-mcmanus: I like that idea. Makes it easier to find settings. We do that style for other recent features like codeAnalysis and vcFormat.

sean-mcmanus: but FYI, I recall there is some spot in the settings update code that would need to change to handle the "." correctly

sean-mcmanus: ...it may not matter now with 2 settings, but it could be more of a settings organization issue if we add more doxygen-related settings in the future.

Extension/package.json
```diff
@@ -2333,22 +2333,22 @@
           "markdownDescription": "%c_cpp.configuration.simplifyStructuredComments.markdownDescription%",
           "scope": "application"
         },
-        "C_Cpp.autocompleteDoxygenComment":{
-          "type":"boolean",
-          "default":true,
-          "description": "%c_cpp.configuration.autocompleteDoxygenComment.description%",
+        "C_Cpp.doxygen.generateCommentsOnTyping":{
```
bobbrow: onType is a more or less common setting/command name (e.g. formatOnType) which is why I suggested `generateOnType`.  I was hoping for shorter names, and I don't think "comments" is needed because 'doxygen' is synonymous with 'comments'

bobbrow: Though I suppose I somewhat contradict that point by suggesting commentStyle.  So maybe I just like it shorter in this case and not the other. 🤷‍♂️

sean-mcmanus (approved):

bobbrow (approved):

Extension/package.nls.json
```diff
@@ -185,8 +185,8 @@
...
     "c_cpp.configuration.commentContinuationPatterns.items.anyof.string.markdownDescription": { "message": "The pattern that begins a multiline or single line comment block. The continuation pattern defaults to ` * ` for multiline comment blocks or this string for single line comment blocks.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
-    "c_cpp.configuration.autocompleteDoxygenComment.description":  "Controls whether to automatically insert the Doxygen comment after typing the chosen comment style.",
-    "c_cpp.configuration.generatedDoxygenCommentStyle.description": "The string of characters used as the starting line of the Doxygen comment.",
+    "c_cpp.configuration.doxygen.generateOnType.description":  "Controls whether to automatically insert the Doxygen comment after typing the chosen comment style.",
```
bobbrow, line 187: oh weird, these descriptions split the descriptions of commentContinuationPatterns

PR #9771, opened 2022-08-23
Fix debug with IntelliSense engine disabled when there's no existing launch.json

### LINKED ISSUE #9762: I can't run/Debug C program when I disable c/c++ intellisense for clangd.
### Environment

- OS and version: Windows11 x64
- VS Code: 1.70
- C/C++ extension: 1.20
- GDB / LLDB version: 12.1

Bug Summary:

Steps to reproduce:
1. Install clangd for intellisense, it asked me to disable c/c++ intellisense or there will be conflict.
2. Disable c/c++ intelli sense Engine.
3. Click 'Run and Debug'
4. Then a error 'Default config not found in provideDebugConfigurations()' popup on the bottom-right.

### Debugger Configurations

```shell
None.
There problem is just 'cannot generate tasks.json'.
```

### Debugger Logs

```shell
None.
```

### Other Extensions

_No response_

### Additional Information

Code_jTkyUHBcTD

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9762

I just moved the code into the "if" blocks...GitHub's diffing isn't handling it well.

Regression compared to 1.8.4. Follow up to https://github.com/microsoft/vscode-cpptools/pull/8876  .

The experience after this fix isn't the same as 1.8.4 because we don't create the launch.json automatically, but the user can use the UI options to generate the launch.json, i.e. click the Open 'launch.json' button option in the dialog box that appears, and then using the Add Configuration button in the bottom right.

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -383,106 +383,107 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
             (!configuredBuildTasks.some(taskJson => (taskJson.definition.label === taskDetected.definition.label))));
         buildTasks = buildTasks.concat(configuredBuildTasks, dedupDetectedBuildTasks);
 
-        if (buildTasks.length === 0) {
-            return [];
-        }
-
         // Filter out build tasks that don't match the currently selected debug configuration type.
-        buildTasks = buildTasks.filter((task: CppBuildTask) => {
-            const command: string = task.definition.command as string;
-            if (!command) {
+        if (buildTasks.length !== 0) {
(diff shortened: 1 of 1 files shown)


bobbrow:
> I just moved the code into the "if" blocks...GitHub's diffing isn't handling it well.

Use the "hide whitespace" option and it looks better.

image

sean-mcmanus:
> > I just moved the code into the "if" blocks...GitHub's diffing isn't handling it well.
> 
> Use the "hide whitespace" option and it looks better.
> 
> image

Cool, thanks -- I didn't know about that.

Colengms (approved):

Issue #9786, opened 2022-08-25
Update clang-format settings to support file: 


EcotechBrad:
### Feature Request

in Clang-format v14 you can specify a file path using `-style=file:<path-to-file>` 
however, variable expansions do not work in setting  "C_Cpp.clang_format_style"

i.e something like "C_Cpp.clang_format_style": "file:${workspaceFolder}/tools/.clang-format" is not expanded, and returns 

Formatting failed:
"C:\Program Files\LLVM\bin\clang-format.exe" -style=file:${workspacefolder}/tools/.clang-format

if someone could point me to where variables like clangFormatPath are expanded i could open a PR myself, 

i've already tried  adding `util.resolveVariables` in client.ts mimicking `settings_clangFormatPath.push(util.resolveVariables(setting.clangFormatPath, this.AdditionalEnvironment));`
```
  commit: 5d736cc6c8e07be89b1b79a2ceb124243304a434 file client.ts line 1125
  settings_clangFormatStyle.push(util.resolveVariables(setting.clangFormatStyle, this.AdditionalEnvironment)); 
```
however no luck.

sean-mcmanus:
Thanks for reporting this. The resolving of the ${workspaceFolder} is done in our closed source code -- it should be fixed in 1.13.0.

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.13.0 .

PR #9788, opened 2022-08-25
Update clang_format_style description.

sean-mcmanus:
Related to https://github.com/microsoft/vscode-cpptools/issues/9786


michelleangela:
What's the reasoning for changing the setting scope to "machine-overridable"?

sean-mcmanus:
@michelleangela Because it can have a file path, but I've reverted that change, because it may not always contain a path, in which case users may want to use the User scope setting still.

Colengms (approved):

Extension/package.nls.json
```diff
@@ -152,7 +152,7 @@
...
+    "c_cpp.configuration.clang_format_style.markdownDescription": { "message": "Coding style, currently supports: `Visual Studio`, `LLVM`, `Google`, `Chromium`, `Mozilla`, `WebKit`, `Microsoft`, `GNU`. Use `file` to load the style from a `.clang-format` file in the current or parent directory, or use `file:<path>/.clang-format` to reference a specific path. Use `{key: value, ...}` to set specific parameters. For example, the `Visual Studio` style is similar to: `{ BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Allman, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All, FixNamespaceComments: false }`.", "comment": [ "Words 'key' and 'value' in '{key: value, ...}' and the word 'path' in 'file:<path>/.clang-format' should be translated, but all other markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
bobbrow, line 155: `file:<path>/.clang-format` seems wrong.  Shouldn't it be `file://<path>/.clang-format` to avoid confusion?  And do we accept Uri's here now too?

sean-mcmanus: @bobbrow No -- the syntax is from clang-format itself, see https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormat.html. I don't know what you mean by "do we accept Uri's here now too". The setting value is passed directly to clang-format, so whatever clang-format accepts would work -- I'm following the documentation details/terminology provided by the clang-format docs and what I tested, i.e. they don't mention if "file://" uri's are accepted.

bobbrow: I see.  I wasn't expecting `file:<some_path>` to be the syntax.  It looked like a file URI that was missing the `//`.  Disregard my comment.

michelleangela (approved):

PR #9809, opened 2022-08-29
add caseSensitiveFileSupport

### LINKED ISSUE #1994: Need a setting to force case-sensitive folder support
### System

Windows 10, Version 1703 (OS Build 15063.966)
VSCode 1.23.1
Cpptools 0.17.0

### Description

I'm seeing issues with workspaces on drives mapped using the built-in Windows NFS client (i.e., the one installed from "Windows Features -> Services for NFS -> Client for NFS").

With case sensitivity enabled (mount option `casesensitive=yes`), the extension logs an error: `Unable to retrieve file system information for <path>. error = -1` for paths in `browse.path` that reside on the NFS drive. Tooltips and navigation shortcuts (e.g., "Go to Definition" and "Switch Header/Source") subsequently fail in this configuration because the extension cannot find my files.

With case sensitivity disabled (`casesensitive=no`), the error goes away and indexing seems to work, but any files opened using navigation shortcuts have their paths converted to ALL CAPS and you can end up with duplicate tabs (one with normal casing and one with all caps) or duplicate search results if the file in question was already opened through the Explorer pane.

#### Sample mount output
```
>mount -o casesensitive=yes <servername>:/share x:
x: is now successfully connected to <servername>:/share

>mount

Local    Remote                                 Properties
-------------------------------------------------------------------------------
x:       \\<servername>\share                   UID=<uid>, GID=<gid>
                                                rsize=131072, wsize=131072
                                                mount=soft, timeout=6.4
                                                retry=1, locking=yes
                                                fileaccess=755, lang=ANSI
                                                casesensitive=yes
                                                sec=sys
```

#### Sample Cpptools log
```
IntelliSense Engine = Default.
The extension will use the Tag Parser for IntelliSense when #includes don't resolve.
Autocomplete is enabled.
Error squiggles are enabled.
File exclude: **/.git
File exclude: **/.svn
File exclude: **/.hg
File exclude: **/CVS
File exclude: **/.DS_Store
File exclude: **/.vscode
Search exclude: **/node_modules
Search exclude: **/bower_components
Search exclude: **/.vscode
Code browsing service initialized
  Unable to retrieve file system information for X:\vscode-cpp-test. error = -1
  Folder: C:/TOOLS/MINGW64/LIB/GCC/X86_64-W64-MINGW32/5.3.0/INCLUDE/ will be indexed
  Folder: C:/TOOLS/MINGW64/LIB/GCC/X86_64-W64-MINGW32/5.3.0/INCLUDE-FIXED/ will be indexed
  Folder: C:/TOOLS/MINGW64/X86_64-W64-MINGW32/INCLUDE/ will be indexed
Discovering files...
  Processing folder (recursive): C:/TOOLS/MINGW64/LIB/GCC/X86_64-W64-MINGW32/5.3.0/INCLUDE/
  Processing folder (recursive): C:/TOOLS/MINGW64/LIB/GCC/X86_64-W64-MINGW32/5.3.0/INCLUDE-FIXED/
  Processing folder (recursive): C:/TOOLS/MINGW64/X86_64-W64-MINGW32/INCLUDE/
  Discovering files: 2283 file(s) processed
  0 file(s) removed from database
Done discovering files.
Parsing remaining files...
  Parsing: 0 files(s) processed
Done parsing remaining files.
```

#### Sample Cpptools configuration
```json
{
    "configurations": [
        {
            "name": "MinGW",
            "intelliSenseMode": "clang-x64",
            "compilerPath": "C:/tools/mingw64/bin/gcc.exe",
            "includePath": [
                "${workspaceFolder}",
                "${workspaceFolder}/foo",
                "${workspaceFolder}/bar/baz"
            ],
            "defines": [],
            "browse": {
                "path": [
                    "${workspaceFolder}"
                    ],
                "limitSymbolsToIncludedHeaders": true,
                "databaseFilename": ""
            },
            "cStandard": "c11",
            "cppStandard": "c++17"        
        }
    ],
    "version": 4
}
```

browntarik:
Fixes https://github.com/microsoft/vscode-cpptools/issues/1994


Extension/src/LanguageServer/settings.ts
```diff
@@ -203,6 +226,7 @@ export class CppSettings extends Settings {
     public get clangFormatSortIncludes(): string | undefined { return super.Section.get<string>("clang_format_sortIncludes"); }
     public get experimentalFeatures(): string | undefined { return super.Section.get<string>("experimentalFeatures"); }
     public get suggestSnippets(): boolean | undefined { return super.Section.get<boolean>("suggestSnippets"); }
+    public get caseSensitiveFileSupport(): string | undefined { return super.Section.get<string>("caseSensitiveFileSupport"); }
```
bobbrow: Probably want to put a comment here stating that the value could be incorrect if it was set in a workspace folder.

Extension/src/LanguageServer/settings.ts
```diff
@@ -73,6 +73,29 @@ class Settings {
         }
         return undefined;
     }
+
+    public getWorkspaceFolderValue<T>(section: string): T | undefined {
```
bobbrow: After our iterations, I think this function has now become obsolete.  I believe it will behave the same as the official `get<T>` API now.

Extension/src/LanguageServer/settingsTracker.ts
```diff
@@ -38,12 +39,14 @@ export class SettingsTracker {
     private collectSettings(filter: FilterFunction): { [key: string]: string } {
         const settingsResourceScope: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("C_Cpp", this.resource);
         const settingsNonScoped: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("C_Cpp");
-        const selectCorrectlyScopedSettings = (rawSetting: any): vscode.WorkspaceConfiguration =>
+        const caseSensitiveFileSupportLastState: PersistentWorkspaceState<string> = new PersistentWorkspaceState<string>("CPP.isFilePathHandlingCaseSensitivePersistent", "default");
```
bobbrow: >isFilePathHandlingCaseSensitivePersistent [](http://example.com/codeflow?start=126&length=41)

The name of the state variable should be updated to reflect the new setting name.

'CPP.currentCaseSensitiveFileSupport' or something like that.

Extension/src/LanguageServer/client.ts
```diff
@@ -947,6 +949,9 @@ export class DefaultClient implements Client {
     }
 
     private createLanguageClient(allClients: ClientCollection): LanguageClient {
+        // do we need operating system?
+        const isFilePathHandlingCaseSensitivePersistent: PersistentWorkspaceState<string> = new PersistentWorkspaceState<string>("CPP.isFilePathHandlingCaseSensitivePersistent", "default");
+        let detectFilePathChange: boolean = false;
```
bobbrow: A better name for this might be 'resetDatabase' since it reflects what will happen when it's true.

Extension/src/common.ts
```diff
@@ -888,6 +888,16 @@ export function promptReloadWindow(message: string): void {
     });
 }
 
+export function promptForIgnoreSettingsChange(): void {
+    promptIgnoreChange(localize("ignore.workspace.for.changes", "Setting change is ignored at folder level"));
```
bobbrow: Since there's only one setting that can trigger this right now, it might be more useful to the customer to tell them specifically which setting is ignored.

Extension/package.json
```diff
@@ -2652,6 +2652,16 @@
           "markdownDescription": "%c_cpp.configuration.suggestSnippets.markdownDescription%",
           "scope": "resource"
         },
+        "C_Cpp.caseSensitiveFileSupport": {
+          "type": "string",
+          "enum": [
+            "default",
```
bobbrow, line 2668: Do we have a follow-up task to change the PascalCase enum values to camelCase for 1.13?

bobbrow: I'm seeing some potential complications with the rename...

```
  "when": "config.C_Cpp.intelliSenseEngine == 'Default'
```

Extension/src/LanguageServer/settings.ts
browntarik: Not relevant due to scope change

Extension/src/LanguageServer/settings.ts
browntarik: Not relevant due to scope change, will remove these

Extension/src/common.ts
browntarik: Not relevant due to scope change, has been removed

Extension/package.json
browntarik: This doesn't affect our feature, does it?

bobbrow: It doesn't affect this PR, but we want to make sure we're tracking the work.  Can you open an issue for this and put it in 1.13?

browntarik: Will do!

Extension/package.nls.json
```diff
@@ -219,6 +219,7 @@
...
     "c_cpp.configuration.experimentalFeatures.description": "Controls whether \"experimental\" features are usable.",
     "c_cpp.configuration.suggestSnippets.markdownDescription": { "message": "If `true`, snippets are provided by the language server.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.caseSensitiveFileSupport.markdownDescription": { "message": "If set to `default`, the extension will follow case sensitivity of the file system, otherwise case sensitivity will be enforced if set to `enabled`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
```
Colengms: How about:

If set to `default`, the file system of the workspace is assumed to be case insensitive on Windows and case sensitive on macOS or Linux.  If set to `enabled`, the file system of the workspace is assumed to be case sensitive on Windows.

Extension/src/LanguageServer/client.ts
```diff
@@ -1201,6 +1207,13 @@ export class DefaultClient implements Client {
             localizedStrings.push(lookupString(i));
         }
 
+        if (workspaceSettings.caseSensitiveFileSupport !== currentCaseSensitiveFileSupport.Value) {
+            resetDatabase = true;
+            if (workspaceSettings.caseSensitiveFileSupport) {
```
Colengms: Is this `IF` needed?  Shouldn't `currentCaseSensitiveFileSupport.Value` be updated every time we see that it changed?   As it is now, if the user has transitioned it from `true` to `false` (while our extension was not active, so we did not intercept it via a didChangeSettings event), then this will fail to update it to `false`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1566,6 +1581,11 @@ export class DefaultClient implements Client {
                     if (changedSettings["legacyCompilerArgsBehavior"]) {
                         this.configuration.handleConfigurationChange();
                     }
+
+                    if (changedSettings["caseSensitiveFileSupport"]) {
+                        util.promptForReloadWindowDueToSettingsChange();
```
Colengms, line 1581: Maybe only display this on Windows for now, as changing the setting between default and enabled has no effect on other platforms.

Extension/src/LanguageServer/persistentState.ts
```diff
@@ -51,7 +51,7 @@ export class PersistentState<T> extends PersistentStateBase<T> {
         super(key, defaultValue, util.extensionContext ? util.extensionContext.globalState : undefined);
     }
 }
-
+// make setting a persistent workspace state and pass memento as a setting
```
Colengms: Did you intend to keep this comment?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -947,6 +948,9 @@ export class DefaultClient implements Client {
     }
 
     private createLanguageClient(allClients: ClientCollection): LanguageClient {
+        // do we need operating system?
```
bobbrow: Remove this comment

Extension/src/LanguageServer/settings.ts
```diff
@@ -162,7 +162,7 @@ export class CppSettings extends Settings {
         return path;
     }
 
-    public get maxConcurrentThreads(): number | undefined | null { return super.Section.get<number | null>("maxConcurrentThreads"); }    public get maxMemory(): number | undefined | null { return super.Section.get<number | null>("maxMemory"); }
+    public get maxConcurrentThreads(): number | undefined | null { return super.Section.get<number | null>("maxConcurrentThreads"); } public get maxMemory(): number | undefined | null { return super.Section.get<number | null>("maxMemory"); }
```
bobbrow: Not sure what happened on this line, but can you split it and put the `maxMemory` getter on its own line?

Extension/src/LanguageServer/client.ts
browntarik: How would we detect a change in that scenario?

Extension/src/LanguageServer/persistentState.ts
browntarik: no removed it

Extension/src/LanguageServer/client.ts
bobbrow: I don't think @Colengms means to update it every time it changes (e.g. didChangeConfiguration). I believe he just means to update it when we create the language client (here).

And `workspaceSettings.caseSensitiveFileSupport` isn't a boolean, so this check will always return `true` anyway.

browntarik: Oh I understand, it was originally there because the typescript complained about `workspaceSettings.caseSensitiveFileSupport` potentially being undefined. So the check is needed, unless its ok for me to removed the undefined type from the setting

bobbrow: I thought the value we get from VSCode's `get<T>` API should never be undefined because we have a default value in `package.json` for this setting.  But if it can be undefined, then I understand why we'd want to keep the `if`.

Colengms: Be careful when using just the variable name itself as a boolean to detect if it's defined or not, in TypeScript.  That syntax will return false if it's defined but has a value of 0 or false.  Instead, always be explicit and use `var === undefined` or `var !== undefined`, unless your intention is to assume the variable is 0/false if undefined.

Bob is right that we should never receive an 'undefined' value for the setting, due to having a default in `package.json`.  However, the API is capable of returning `undefined`, so we need to check for that to make TypeScript happy and let us compile.

IMHO, it would be a good idea to resolve this setting into a boolean immediately after reading it, so you don't need to deal with it being a string or potentially undefined, after reading it from the settings class.

Extension/src/LanguageServer/client.ts
```diff
@@ -995,6 +998,7 @@ export class DefaultClient implements Client {
         const settings_dimInactiveRegions: boolean[] = [];
         const settings_enhancedColorization: string[] = [];
         const settings_suggestSnippets: (boolean | undefined)[] = [];
+        const settings_caseSensitiveFileSupport: (string | undefined)[] = [];
```
sean-mcmanus: I don't think this variable is used any more.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
browntarik: Changed the setting to a boolean so we only deal with the enum val when we call the setting in settings.ts

Extension/src/LanguageServer/client.ts
browntarik: you're right its removed

Extension/src/LanguageServer/settings.ts
```diff
@@ -253,7 +254,9 @@ export class CppSettings extends Settings {
     public get defaultCustomConfigurationVariables(): { [key: string]: string } | undefined { return super.Section.get<{ [key: string]: string }>("default.customConfigurationVariables"); }
     public get useBacktickCommandSubstitution(): boolean | undefined { return super.Section.get<boolean>("debugger.useBacktickCommandSubstitution"); }
     public get codeFolding(): boolean { return super.Section.get<string>("codeFolding") === "Enabled"; }
-    public get legacyCompilerArgsBehavior(): boolean | undefined  { return super.Section.get<boolean>("legacyCompilerArgsBehavior"); }
+    public get caseSensitiveFileSupport(): boolean { return super.Section.get<string>("caseSensitiveFileSupport") === "enabled"; }
```
bobbrow: If we're doing this as a boolean, then I think we should force it to true for linux/mac.  Basicially, `!util.isWindows() || setting === 'enabled'`

bobbrow: ... because I didn't see any code on the native side to deal with this.

Extension/src/LanguageServer/client.ts
```diff
@@ -1209,8 +1217,10 @@ export class DefaultClient implements Client {
             ],
             initializationOptions: {
                 freeMemory: os.freemem() / 1048576,
+                resetFilePathHandling: resetDatabase,
```
bobbrow: 'resetFilePathHandling' sounds like it's left over from the old naming ideas.  Should this be 'resetDatabase'?

Extension/src/LanguageServer/settings.ts
browntarik: Sounds good

bobbrow (approved):

Colengms (approved):

sean-mcmanus (approved):

sean-mcmanus:
Don't check this in until the cpptools side gets checked in.

michelleangela (approved):

PR #9813, opened 2022-08-30
check if the compiler path and debugger path exist.

### LINKED ISSUE #8836: Misleading error message when GDB is not installed

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cpptools/issues/8836
Do not create a debug configuration for invalid compiler paths.


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -410,6 +410,15 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
                 const definition: CppBuildTaskDefinition = task.definition as CppBuildTaskDefinition;
                 const compilerPath: string = definition.command;
+                // Filter out the tasks that has an invalid compiler path.
+                const compilerPathExists: boolean = path.isAbsolute(compilerPath) ?
+                    // Absolute path, just check if it exists
+                    await util.checkFileExists(compilerPath) :
+                    // Non-absolute. Check on $PATH
+                    ((await util.whichAsync(compilerPath)) !== undefined);
+                if (!compilerPathExists) {
+                    console.log("Unable to find {0}. {1} task is ignored.", compilerPath, definition.label);
```
sean-mcmanus: Is this the correct way to report the error? The string isn't localized. I see other code using stuff like `logger.getOutputChannelLogger().showErrorMessage(localize(`. I think our existing uses of console.log are also incorrect/unideal, but we haven't gotten around to improving it.

sean-mcmanus (approved):

elahehrashedi:
@Colengc Here I stopped showing the debug configurations in which the debugger is invalid in the drop down. And the user will see the warning in the output channel to know that the debug config was being ignored. Is this a good approach?

elahehrashedi:
@sean-mcmanus does this message look good?
('debugger.path.not.exists', "Unable to find the {0} debugger in this path: {1}. The debug configuration for {2} is ignored.", `\"${debuggerName}\"`, `\"${debuggerPath}\"`, compilerName)

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -467,22 +476,23 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
-                        if (isWindows) {
-                            newConfig.miDebuggerPath = debuggerPath;
-                            return resolve(newConfig);
-                        } else {
-                            fs.stat(debuggerPath, (err, stats: fs.Stats) => {
-                                if (!err && stats && stats.isFile()) {
-                                    newConfig.miDebuggerPath = debuggerPath;
-                                } else {
-                                    newConfig.miDebuggerPath = path.join("/usr", "bin", debuggerName);
+                        if (!await util.checkFileExists(debuggerPath)) {
```
michelleangela: Could you clarify why it checks `await util.checkFileExists(debuggerPath)` as false at line 479 and then at line 481 it checks it again as true when there doesn't seem to be a change to `debuggerPath` that is being verified? 

It seems conditional block 479 could be refactored?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -467,22 +476,23 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                            if (!isWindows) {
+                                const newDebuggerPath: string = path.join("/usr", "bin", debuggerName);
+                                if (await util.checkFileExists(debuggerPath)) {
```
sean-mcmanus: It seems like this should be `if (await util.checkFileExists(newDebuggerPath)) {`

sean-mcmanus (changes requested):

sean-mcmanus:
> @sean-mcmanus does this message look good? ('debugger.path.not.exists', "Unable to find the {0} debugger in this path: {1}. The debug configuration for {2} is ignored.", `\"${debuggerName}\"`, `\"${debuggerPath}\"`, compilerName)

Sure.

Extension/src/Debugger/configurationProvider.ts
elahehrashedi: line 481 had to use `newDebuggerPath`, it is fixed now.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -467,22 +480,29 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                            // Check if debuggerName exists on $PATH
+                            newConfig.miDebuggerPath = debuggerName;
+                            return resolve(newConfig);
                         } else {
-                            fs.stat(debuggerPath, (err, stats: fs.Stats) => {
-                                if (!err && stats && stats.isFile()) {
-                                    newConfig.miDebuggerPath = debuggerPath;
-                                } else {
-                                    newConfig.miDebuggerPath = path.join("/usr", "bin", debuggerName);
+                            if (!isWindows) {
```
michelleangela: Nit: `if (!isWindows)` can be part of the `else` on line 492.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -467,22 +480,29 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                                const newDebuggerPath: string = path.join("/usr", "bin", debuggerName);
```
michelleangela: Nit: recommend to rename `newDebuggerPath` to `usrDebuggerPath` to more represent the actual path.

michelleangela (approved):

Extension/src/Debugger/configurationProvider.ts
elahehrashedi:
```C++
                        } else {
                            if (!isWindows) {
                                const usrDebuggerPath: string = path.join("/usr", "bin", debuggerName);
                                if (await util.checkFileExists(usrDebuggerPath)) {
                                    newConfig.miDebuggerPath = usrDebuggerPath;
                                    return resolve(newConfig);
                                }
                            }
                            logger.getOutputChannelLogger().appendLine(localize('debugger.path.not.exists', "Unable to find the {0} debugger. The debug configuration for {2} is ignored.", `\"${debuggerName}\"`, compilerName));
                            return resolve(undefined);
                        }
```

The else has two parts, `iswindows `and `!inswindows`.
I simplify it .

michelleangela (approved):

michelleangela (approved):

michelleangela (approved):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -408,9 +408,18 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         // Generating a task is async, therefore we must *await* *all* map(task => config) Promises to resolve.
         let configs: CppDebugConfiguration[] = [];
         if (buildTasks.length !== 0) {
-            configs = await Promise.all(buildTasks.map<Promise<CppDebugConfiguration>>(async task => {
+            configs = (await Promise.all(buildTasks.map<Promise<CppDebugConfiguration|undefined>>(async task => {
```
sean-mcmanus: I think we normally surround the | with spaces.

sean-mcmanus (changes requested):

sean-mcmanus (approved):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -439,20 +448,24 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
                 const isCl: boolean = compilerName === "cl.exe";
                 newConfig.cwd = isWindows && !isCl && !process.env.PATH?.includes(path.dirname(compilerPath)) ? path.dirname(compilerPath) : "${fileDirname}";
 
-                return new Promise<CppDebugConfiguration>(resolve => {
+                return new Promise<CppDebugConfiguration|undefined>(async resolve => {
```
sean-mcmanus: Can you add spaces around | here too?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #9822, opened 2022-08-31
Revert "Switch to new key format since AIF is retired"

bobbrow:
Reverts microsoft/vscode-cpptools#9769

All of our telemetry disappeared since 1.12.1.


bobbrow:
We'll need to repackage 1.12.3 into 1.12.4 with this change and re-release

michelleangela (approved):

PR #9828, opened 2022-09-01
Update HTTP Proxy to fix telemetry

lramos15:
This should fix the lack of telemetry caused by the new key.

After a deep dive into the C++ extension code it appears that the older versions of `https-proxy-agent` would rewrite the `https` module in a way that was non standard to what the node docs would have. Therefore the use of the https module within `@vscode/extension-telemetry` was throwing an error when it was in fact syntactically correct. This updates the module and I now see data flowing correctly via the new key into the extension data tables. I still recommend testing all other use cases of the `https-proxy-agent`.

Ref: https://github.com/microsoft/vscode/issues/93167#issuecomment-604886404

TS Config Update Ref: https://github.com/TooTallNate/node-agent-base/issues/56

cc @Colengms  @bobbrow

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2920,7 +2920,7 @@
           "rust"
         ],
         "_aiKeyComment": "Ignore 'Property aiKey is not allowed'. See https://github.com/microsoft/vscode/issues/76493",
-        "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
+        "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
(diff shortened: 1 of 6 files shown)


sean-mcmanus (approved):

bobbrow:
Actually, I don't think we use those download functions anymore.  Can we just remove https-proxy-agent and the now unused functions that call it?

@Colengms @sean-mcmanus

Colengms:
> Actually, I don't think we use those download functions anymore. Can we just remove https-proxy-agent and the now unused functions that call it?

Shall we abandon this PR, and go with *[this](https://github.com/microsoft/vscode-cpptools/pull/9833)* one instead?

PR #9833, opened 2022-09-01
Remove HttpsProxyAgent

Colengms:
Removes unused functions, removing our dependency on https-proxy-agent.

Also restores the new aiKey values.


Extension/yarn.lock
```diff
@@ -700,20 +700,20 @@ acorn@^8.5.0:
...
-    es6-promisify "^5.0.0"
-
 agent-base@6:
   version "6.0.2"
   resolved "https://registry.yarnpkg.com/agent-base/-/agent-base-6.0.2.tgz#49fff58577cfee3f37176feab4c22e00f86d7f77"
   integrity sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==
   dependencies:
     debug "4"
 
+agent-base@^4.3.0:
```
bobbrow: This needs to be removed.

bobbrow (changes requested):
What about the tsconfig changes from the other PR? Do we need them?

sean-mcmanus (approved):

Extension/CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # C/C++ for Visual Studio Code Changelog
 
+## Version 1.12.5: September 1, 2022
+### Other
+* Restore changes to telemetry key. [PR #9833](https://github.com/microsoft/vscode-cpptools/pull/9822)
```
sean-mcmanus: 9822->9833

Extension/yarn.lock
lramos15: This is still here because I believe removing `https-proxy-agent` was forgotten as there is `http-proxy-agent` and `https-proxy-agent` in devDependencies and dependencies respectively

lramos15:
> What about the tsconfig changes from the other PR? Do we need them?

Those were only needed to cause the newer versions of `https-proxy-agent` and `http-proxy-agent` to compile. If both are removed they are not needed

bobbrow (approved):

Issue #9836, opened 2022-09-01
Don't always show SSH Targets View


xisui-MSFT:
### Feature Request

#9760 Always shows an SSH Targets View in the debug panel, which may be annoying for users who are not using this feature. 

It's hard to determine if users use or don't use it. We can probably show it when a related command is invoked, but not being able to see a list of devices before invoking commands seems pretty awkward and confusing...

Andrew suggests that we can check 'host' in launch.json. It's a good idea, with the only drawback being it requires listening to configuration change event, which fires on all settings changes.

Also open to other suggestions.

sean-mcmanus:
Available in 1.13.1: https://github.com/microsoft/vscode-cpptools/releases/tag/v1.13.1

PR #9856, opened 2022-09-07
Show SSH Targets View Conditionally

xisui-MSFT:
As described in #9836, to check if SSH Targets View is needed, we check if any of the SSH Targets commands are used in tasks.json, launch.json, or the workspace file. And therefore, we listen to changes in workspace file if exists, and add and change events in tasks.json and launch.json in all folders.

Don't hide the view in the current session. The SSH targets features are likely used later if they are used before even after deleted; and if they are used.

Personally, I don't like this method, especially after the changes I've made, since there are a lot of listeners and file reads added to make the one-click collapse behavior look better, which IMO isn't too bad in the first place. So I'm still open to ideas on this.


xisui-MSFT:
Targeting 1.13.1+

WardenGnaw:
Can we not use the VS Code API's `vscode.window.workspace.createFileSystemWatcher` or is `chokidar` preferred over that. 

https://code.visualstudio.com/api/references/vscode-api#workspace

xisui-MSFT:
> Can we use the VS Code API's `vscode.window.workspace.createFileSystemWatcher` or is `chokidar` preferred over that.
> 
> https://code.visualstudio.com/api/references/vscode-api#workspace

It can't monitor 'add' event.

xisui-MSFT:
> Can we use the VS Code API's `vscode.window.workspace.createFileSystemWatcher` or is `chokidar` preferred over that.
> 
> https://code.visualstudio.com/api/references/vscode-api#workspace

When I firstly visited file watchers available in VS Code, they don't support 'add' event. Now looks like they support it. I'll try to switch to this API.

Extension/package.json
```diff
@@ -2761,6 +2761,17 @@
           "default": false,
           "markdownDescription": "%c_cpp.configuration.legacyCompilerArgsBehavior.markdownDescription%",
           "scope": "resource"
+        },
+        "C_Cpp.sshTargetsView": {
+          "type": "string",
+          "enum": [
+            "Enabled",
```
bobbrow: Tarik is in the process of changing enum values to camelCase.  Can you change the values to `enabled`, `disabled`, `auto`?

Extension/src/SSH/TargetsView/common.ts
```diff
@@ -36,5 +36,5 @@ export class LabelLeafNode implements BaseNode {
     }
 }
 
-export const refreshCppSshTargetsView: string = 'C_Cpp.refreshCppSshTargetsView';
-export const addSshTarget: string = 'C_Cpp.addSshTarget';
+export const cmd_refreshCppSshTargetsView: string = 'C_Cpp.refreshCppSshTargetsView';
```
bobbrow: We would prefer not to have `_` in variable names.

Extension/src/LanguageServer/settings.ts
```diff
@@ -564,6 +564,10 @@ export class CppSettings extends Settings {
             && vscode.workspace.getConfiguration("workbench").get<string>("colorTheme") !== "Default High Contrast";
     }
 
+    public get sshTargetsView(): string {
+        return super.Section.get<string>("sshTargetsView") || 'Auto';
```
bobbrow: lower case `auto` (when you update the enum values)

Extension/src/Debugger/extension.ts
```diff
@@ -79,38 +81,72 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
...
-        fileWatchers.push(chokidar.watch(sshConfig, {ignoreInitial: true})
-            .on('add', () => vscode.commands.executeCommand(refreshCppSshTargetsView))
-            .on('change', () => vscode.commands.executeCommand(refreshCppSshTargetsView))
-            .on('unlink', () => vscode.commands.executeCommand(refreshCppSshTargetsView)));
+
+    // Decide if we should show the SSH Targets View.
+    sshTargetsViewSetting = (new CppSettings()).sshTargetsView;
+    if (sshTargetsViewSetting === 'Enabled') {
+        // Don't wait
+        enableSshTargetsView().then(() => {}, () => {});
```
bobbrow: Do we need an empty `then`?

Extension/src/Debugger/extension.ts
```diff
@@ -79,38 +81,72 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
...
     }
-    vscode.commands.executeCommand('setContext', 'showCppSshTargetsView', true);
+
+    disposables.push(vscode.workspace.onDidChangeConfiguration(async (e: vscode.ConfigurationChangeEvent) => {
+        if (e.affectsConfiguration('C_Cpp.sshTargetsView')) {
+            sshTargetsViewSetting = (new CppSettings()).sshTargetsView;
+            if (sshTargetsViewSetting === 'Enabled') {
+                await enableSshTargetsView();
+            } else {
+                await vscode.commands.executeCommand('setContext', 'enableCppSshTargetsView', false);
```
bobbrow: If it's `auto` doesn't this disable it?

Extension/src/Debugger/extension.ts
```diff
@@ -79,38 +81,72 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
...
     disposables.push(sshTargetsProvider);
-    for (const sshConfig of getSshConfigurationFiles()) {
-        fileWatchers.push(chokidar.watch(sshConfig, {ignoreInitial: true})
-            .on('add', () => vscode.commands.executeCommand(refreshCppSshTargetsView))
-            .on('change', () => vscode.commands.executeCommand(refreshCppSshTargetsView))
-            .on('unlink', () => vscode.commands.executeCommand(refreshCppSshTargetsView)));
+
+    // Decide if we should show the SSH Targets View.
+    sshTargetsViewSetting = (new CppSettings()).sshTargetsView;
+    if (sshTargetsViewSetting === 'Enabled') {
```
bobbrow: If it's auto, do we want to show it if there's an active ssh target?  I didn't see whether we save the active target in PersistentWorkspaceState.

Extension/package.json
xisui-MSFT: Sure

Extension/src/Debugger/extension.ts
xisui-MSFT: Ah right, good point

Extension/src/Debugger/extension.ts
xisui-MSFT: Sure we can do that

Extension/src/Debugger/extension.ts
xisui-MSFT: Looks like there's no lint rule to force await in cpptools. I can remove it if you prefer.

xisui-MSFT:
> > Can we use the VS Code API's vscode.window.workspace.createFileSystemWatcher or is chokidar preferred over that.

> > https://code.visualstudio.com/api/references/vscode-api#workspace

> When I firstly visited file watchers available in VS Code few years ago, they didn't support 'add' event. Now looks like they support it. I'll try to switch to this API.

@WardenGnaw Turns out createFileSystemWatcher can only watch files opened in the current workspace/folder. I'll add chokidar back

Extension/src/Debugger/extension.ts
bobbrow: the linter rule requires `void` in front of an unawaited promise.  You can add that if you want.  Min bar is to remove the empty `then`

xisui-MSFT: Already removed, and don't see lint errors ;)

sean-mcmanus:
> > > Can we use the VS Code API's vscode.window.workspace.createFileSystemWatcher or is chokidar preferred over that.
> 
> > > https://code.visualstudio.com/api/references/vscode-api#workspace
> 
> > When I firstly visited file watchers available in VS Code few years ago, they didn't support 'add' event. Now looks like they support it. I'll try to switch to this API.
> 
> @WardenGnaw Turns out createFileSystemWatcher can only watch files opened in the current workspace/folder. I'll add chokidar back

FYI, our language server code uses fs.watch, but there are some known issues with that, so chokidar might be better.

WardenGnaw (approved):

bobbrow (approved):

Extension/package.json
```diff
@@ -2761,6 +2761,17 @@
           "default": false,
           "markdownDescription": "%c_cpp.configuration.legacyCompilerArgsBehavior.markdownDescription%",
           "scope": "resource"
+        },
+        "C_Cpp.sshTargetsView": {
+          "type": "string",
+          "enum": [
+            "enabled",
+            "disabled",
+            "auto"
```
sean-mcmanus: I think "auto" should be "default" to be consistent with the other setting enums. Or is there are reason for choosing "auto" over "default"?

Extension/package.json
```diff
@@ -2761,6 +2761,17 @@
...
+          ],
+          "default": "auto",
+          "description": "%c_cpp.configuration.sshTargetsView.description%",
+          "scope": "resource"
```
sean-mcmanus: Is this the right scope? i.e. does it make sense for 2 different workspace folders to have different values? Maybe it should be "application" or "window" scope? It would be window scope only if it makes sense to change it per workspace.

Extension/src/LanguageServer/settings.ts
```diff
@@ -564,6 +564,10 @@ export class CppSettings extends Settings {
             && vscode.workspace.getConfiguration("workbench").get<string>("colorTheme") !== "Default High Contrast";
     }
 
+    public get sshTargetsView(): string {
+        return super.Section.get<string>("sshTargetsView") || 'auto';
```
sean-mcmanus: I think the `||` should be `??` if the intention is for `auto` to be used when the get returns null or undefined. I think practically it's just a style change though.

sean-mcmanus (changes requested):

Extension/package.json
xisui-MSFT: Right, because user probably want to use this feature in one folder/workspace but not in another.

Extension/src/LanguageServer/settings.ts
xisui-MSFT: Sure

Extension/package.json
xisui-MSFT: Personally I prefer "auto", since "default" is less descriptive and requires more explanation, but if "default" is a convention, I'm ok with it.

sean-mcmanus (approved):

Extension/package.json
sean-mcmanus: But "resource" scope is for when you want 2 different folders in a multi-root workspace to have different values be used, which is unusual for a UI setting, normally it would be "window" scope.

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #9862, opened 2022-09-08
change to camelCase for Default, Enabled, Disabled

browntarik:
change casing to reflect changes made in adding the case sensitive file support setting


Extension/package.json
```diff
@@ -2198,9 +2198,9 @@
...
           "type": "string",
           "enum": [
-            "Enabled",
-            "Disabled",
-            "EnabledIfIncludesResolve"
+            "enabled",
+            "disabled",
+            "enabledIfIncludesResolve"
           ],
           "default": "EnabledIfIncludesResolve",
```
bobbrow: update this value

Extension/package.json
```diff
@@ -2174,18 +2174,18 @@
...
         },
         "C_Cpp.autocomplete": {
           "type": "string",
           "enum": [
-            "Default",
-            "Disabled"
+            "default",
+            "disabled"
           ],
           "default": "Default",
```
bobbrow: update this value

Extension/package.json
```diff
@@ -5092,17 +5092,17 @@
       "editor/context": [
         {
           "command": "C_Cpp.SwitchHeaderSource",
-          "when": "config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cuda-cpp'",
+          "when": "config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cuda-cpp'",
```
bobbrow: Is this comparison case sensitive?  If it is, then we need to check both 'Disabled' and 'disabled' for now.

Extension/package.json
```diff
@@ -5092,17 +5092,17 @@
...
           "group": "custom1@1"
         },
         {
           "command": "workbench.action.gotoSymbol",
-          "when": "config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cuda-cpp'",
+          "when": "config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cuda-cpp'",
```
bobbrow: same

Extension/package.json
```diff
@@ -5092,17 +5092,17 @@
...
           "group": "custom1@2"
         },
         {
           "command": "workbench.action.showAllSymbols",
-          "when": "config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cuda-cpp'",
+          "when": "config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'disabled' && editorLangId == 'cuda-cpp'",
```
bobbrow: same

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -166,7 +166,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
         if (config.preLaunchTask) {
             config.configSource = this.getDebugConfigSource(config, folder);
-            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === "Disabled";
+            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === "disabled";
```
bobbrow, line 169: You'll need to force the value we have to lowercase before doing the comparison.

Extension/src/LanguageServer/client.ts
```diff
@@ -897,7 +897,7 @@ export class DefaultClient implements Client {
                         this.disposables.push(vscode.languages.registerDocumentSymbolProvider(this.documentSelector, new DocumentSymbolProvider(this), undefined));
                         this.disposables.push(vscode.languages.registerCodeActionsProvider(this.documentSelector, new CodeActionProvider(this), undefined));
                         const settings: CppSettings = new CppSettings();
-                        if (settings.formattingEngine !== "Disabled") {
+                        if (settings.formattingEngine !== "disabled") {
```
bobbrow, line 900: You'll need to force the value we have to lowercase before doing the comparison.

Extension/src/LanguageServer/client.ts
```diff
@@ -1526,7 +1526,7 @@ export class DefaultClient implements Client {
                     const settings: CppSettings = new CppSettings();
                     if (changedSettings["formatting"]) {
                         const folderSettings: CppSettings = new CppSettings(this.RootUri);
-                        if (folderSettings.formattingEngine !== "Disabled") {
+                        if (folderSettings.formattingEngine !== "disabled") {
```
bobbrow, line 1529: You'll need to force the value we have to lowercase before doing the comparison.

Extension/src/LanguageServer/client.ts
```diff
@@ -1962,7 +1962,7 @@ export class DefaultClient implements Client {
                     return;
                 }
                 const settings: CppSettings = new CppSettings(this.RootUri);
-                if (settings.configurationWarnings === "Enabled" && !this.isExternalHeader(docUri) && !vscode.debug.activeDebugSession) {
+                if (settings.configurationWarnings === "enabled" && !this.isExternalHeader(docUri) && !vscode.debug.activeDebugSession) {
```
bobbrow: You'll need to force the value we have to lowercase before doing the comparison.

bobbrow (changes requested):
A couple of 'default' values in the package.json need to be updated.

When doing string comparisons, you will need to lowercase the value before checking against the new lowercase value because some people might still have a PascalCase value for the setting.  I didn't mark all the places that need to be updated.

Extension/src/LanguageServer/settings.ts
```diff
@@ -287,8 +287,8 @@ export class CppSettings extends Settings {
     }
 
     public get enhancedColorization(): boolean {
-        return super.Section.get<string>("enhancedColorization") === "Enabled"
-            && super.Section.get<string>("intelliSenseEngine") === "Default"
+        return super.Section.get<string>("enhancedColorization") === "enabled"
+            && super.Section.get<string>("intelliSenseEngine") === "default"
             && vscode.workspace.getConfiguration("workbench").get<string>("colorTheme") !== "Default High Contrast";
     }
```
bobbrow, line 294: note that there is also a reference in a string to 'Default' on line 756 of settings.ts that needs to be updated.

browntarik: do these comparisons need to be forced to lowercase as well?

browntarik: I see all comparison need to be converted, gotcha

Extension/src/LanguageServer/extension.ts
```diff
@@ -681,19 +681,19 @@ function onAddToIncludePath(path: string): void {
 function onEnableSquiggles(): void {
```
bobbrow, line 681: line 446 has a capitalized 'Disabled' in a string.

bobbrow: Sorry, GitHub won't let me place the comment on that line

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
         }
         let currentIntelliSenseEngineValue: string | undefined = settings.intelliSenseEngine;
         disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
             const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
             if (!reloadMessageShown && settings.intelliSenseEngine !== currentIntelliSenseEngineValue) {
-                if (currentIntelliSenseEngineValue === "Disabled") {
+                if (currentIntelliSenseEngineValue === "disabled") {
                     // If switching from disabled to enabled, we can continue activation.
                     currentIntelliSenseEngineValue = settings.intelliSenseEngine;
                     languageServiceDisabled = false;
```
bobbrow, line 101: line 144 has a capitalized 'Disabled'
(sorry, can't put the comment on that line directly)

Extension/src/LanguageServer/settings.ts
```diff
@@ -730,7 +730,7 @@ export class CppSettings extends Settings {
     // entries, or a .clang-format file, to determine which settings to use.
```
bobbrow, line 734: 'Default' is still capitalized on the line above

Extension/package.json
```diff
@@ -5092,27 +5092,27 @@
       "editor/context": [
         {
           "command": "C_Cpp.SwitchHeaderSource",
-          "when": "config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cuda-cpp'",
+          "when": "(config.C_Cpp.intelliSenseEngine != 'disabled' || config.C_Cpp.intelliSenseEngine != 'Disabled') && editorLangId == 'c' || (config.C_Cpp.intelliSenseEngine != 'disabled' || config.C_Cpp.intelliSenseEngine != 'Disabled') && editorLangId == 'cpp' || (config.C_Cpp.intelliSenseEngine != 'disabled' || config.C_Cpp.intelliSenseEngine != 'Disabled') && editorLangId == 'cuda-cpp'",
```
bobbrow: I don't think this condition is correct.  The part in parenthesis will always evaluate to `true`.  I think you need to use `&&` instead of `||` for the `default` vs `Default` check.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -166,7 +166,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
         if (config.preLaunchTask) {
             config.configSource = this.getDebugConfigSource(config, folder);
-            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === "Disabled";
+            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine?.toLowerCase.toString() === "disabled";
```
Colengms: I assume you wanted to call `toLowerCase()` here, and not call `toString()` on the `toLowerCase` function as if it were an object.

Extension/src/LanguageServer/client.ts
```diff
@@ -897,7 +897,7 @@ export class DefaultClient implements Client {
                         this.disposables.push(vscode.languages.registerDocumentSymbolProvider(this.documentSelector, new DocumentSymbolProvider(this), undefined));
                         this.disposables.push(vscode.languages.registerCodeActionsProvider(this.documentSelector, new CodeActionProvider(this), undefined));
                         const settings: CppSettings = new CppSettings();
-                        if (settings.formattingEngine !== "Disabled") {
+                        if (settings.formattingEngine?.toLowerCase.toString() !== "disabled") {
```
Colengms: Similar to previous comment.  I'm not sure what `toLowerCase.toString()` results in, but I assume it's not returning what you expect.

Extension/src/LanguageServer/client.ts
```diff
@@ -1526,7 +1526,7 @@ export class DefaultClient implements Client {
                     const settings: CppSettings = new CppSettings();
                     if (changedSettings["formatting"]) {
                         const folderSettings: CppSettings = new CppSettings(this.RootUri);
-                        if (folderSettings.formattingEngine !== "Disabled") {
+                        if (folderSettings.formattingEngine?.toLowerCase.toString() !== "disabled") {
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1962,7 +1962,7 @@ export class DefaultClient implements Client {
                     return;
                 }
                 const settings: CppSettings = new CppSettings(this.RootUri);
-                if (settings.configurationWarnings === "Enabled" && !this.isExternalHeader(docUri) && !vscode.debug.activeDebugSession) {
+                if (settings.configurationWarnings?.toLowerCase.toString() === "enabled" && !this.isExternalHeader(docUri) && !vscode.debug.activeDebugSession) {
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1979,7 +1979,7 @@ export class DefaultClient implements Client {
                     vscode.window.showInformationMessage(message, dismiss, disable).then(response => {
                         switch (response) {
                             case disable: {
-                                settings.toggleSetting("configurationWarnings", "Enabled", "Disabled");
+                                settings.toggleSetting("configurationWarnings", "enabled", "eisabled");
```
Colengms: `toggleSetting` will do a comparison against the first string.  You will need to update `toggleSetting` (in settings.ts) to do that comparison in a case insensitive way.  Also, "disabled" here is misspelled as "eisabled".

Extension/src/LanguageServer/customProviders.ts
```diff
@@ -158,7 +158,7 @@ export class CustomConfigurationProviderCollection {
     }
 
     public add(provider: CustomConfigurationProvider, version: Version): boolean {
-        if (new CppSettings(ext.getActiveClient().RootUri).intelliSenseEngine === "Disabled") {
+        if (new CppSettings(ext.getActiveClient().RootUri).intelliSenseEngine === "disabled") {
```
Colengms, line 161: Doesn't this need to be compared against in a case insensitive way?  Perhaps there is a `toLowerCase()` needed here?

Extension/src/LanguageServer/settings.ts
```diff
@@ -253,7 +253,7 @@ export class CppSettings extends Settings {
     public get defaultEnableConfigurationSquiggles(): boolean | undefined { return super.Section.get<boolean>("default.enableConfigurationSquiggles"); }
     public get defaultCustomConfigurationVariables(): { [key: string]: string } | undefined { return super.Section.get<{ [key: string]: string }>("default.customConfigurationVariables"); }
     public get useBacktickCommandSubstitution(): boolean | undefined { return super.Section.get<boolean>("debugger.useBacktickCommandSubstitution"); }
-    public get codeFolding(): boolean { return super.Section.get<string>("codeFolding") === "Enabled"; }
+    public get codeFolding(): boolean { return super.Section.get<string>("codeFolding")?.toLowerCase.toString() === "enabled"; }
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/settings.ts
```diff
@@ -287,8 +287,8 @@ export class CppSettings extends Settings {
     }
 
     public get enhancedColorization(): boolean {
-        return super.Section.get<string>("enhancedColorization") === "Enabled"
-            && super.Section.get<string>("intelliSenseEngine") === "Default"
+        return super.Section.get<string>("enhancedColorization")?.toLowerCase.toString() === "enabled"
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/settings.ts
```diff
@@ -287,8 +287,8 @@ export class CppSettings extends Settings {
...
+            && super.Section.get<string>("intelliSenseEngine")?.toLowerCase.toString() === "default"
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/settings.ts
```diff
@@ -560,7 +560,7 @@ export class CppSettings extends Settings {
 
     public get dimInactiveRegions(): boolean {
         return super.Section.get<boolean>("dimInactiveRegions") === true
-            && super.Section.get<string>("intelliSenseEngine") === "Default"
+            && super.Section.get<string>("intelliSenseEngine")?.toLowerCase.toString() === "default"
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/LanguageServer/settings.ts
```diff
@@ -730,7 +730,7 @@ export class CppSettings extends Settings {
     // entries, or a .clang-format file, to determine which settings to use.
     // This is intentionally not async to avoid races due to multiple entrancy.
     public useVcFormat(document: vscode.TextDocument): boolean {
-        if (this.formattingEngine !== "Default") {
+        if (this.formattingEngine?.toLowerCase.toString() !== "default") {
```
Colengms: `toLowerCase.toString()` should be `toLowerCase()`.

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
     // Read the setting and determine whether we should activate the language server prior to installing callbacks,
     // to ensure there is no potential race condition. LanguageServer.activate() is called near the end of this
     // function, to allow any further setup to occur here, prior to activation.
-    const isIntelliSenseEngineDisabled: boolean = settings.intelliSenseEngine === "Disabled";
+    const isIntelliSenseEngineDisabled: boolean = settings.intelliSenseEngine === "disabled";
```
Colengms, line 84: Needs `toLowerCase()` for comparison.

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
     const shouldActivateLanguageServer: boolean = (!isIntelliSenseEngineDisabled && !isOldMacOs);
 
     if (isOldMacOs) {
         languageServiceDisabled = true;
         vscode.window.showErrorMessage(localize("macos.version.deprecated", "Versions of the C/C++ extension more recent than {0} require at least macOS version {1}.", "1.9.8", "10.12"));
     } else {
-        if (settings.intelliSenseEngine === "Disabled") {
+        if (settings.intelliSenseEngine === "disabled") {
```
Colengms, line 91: Needs `toLowerCase()` for comparison.

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
             languageServiceDisabled = true;
         }
         let currentIntelliSenseEngineValue: string | undefined = settings.intelliSenseEngine;
         disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
             const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
             if (!reloadMessageShown && settings.intelliSenseEngine !== currentIntelliSenseEngineValue) {
-                if (currentIntelliSenseEngineValue === "Disabled") {
+                if (currentIntelliSenseEngineValue === "disabled") {
```
Colengms, line 98: Need `toLowerCase()` for comparison.

Colengms (changes requested):

Extension/package.json
sean-mcmanus: Are you sure this works? I recall the last time we weren't able to get `(a || b) && (c || d)` to be evaluated correctly by VS Code, i.e. it only worked with `a || b && c || d` style tests.

Extension/package.nls.json
```diff
@@ -227,7 +227,7 @@
...
-    "c_cpp.configuration.enhancedColorization.markdownDescription": { "message": "If enabled, code is colorized based on IntelliSense. This setting only applies if `#C_Cpp.intelliSenseEngine#` is set to `Default`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
+    "c_cpp.configuration.enhancedColorization.markdownDescription": { "message": "If enabled, code is colorized based on IntelliSense. This setting only applies if `#C_Cpp.intelliSenseEngine#` is set to `default`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.configuration.codeFolding.description": "If enabled, code folding ranges are provided by the language server.",
```
sean-mcmanus, line 231: I'm seeing `    "c_cpp.configuration.autocomplete.markdownDescription": { "message": "Controls the auto-completion provider. If `Disabled` `

sean-mcmanus (changes requested):

Extension/package.json
Colengms: It's an open issue: [Add support for parenthesis in "when" conditions #91473](https://github.com/microsoft/vscode/issues/91473)

https://code.visualstudio.com/api/references/when-clause-contexts

bobbrow: Do we even need parentheses since these two conditions need to be `&&`ed together anyway?

browntarik: I didn't realize there was an issue with that, I can just remove the parenthesis.

Extension/src/LanguageServer/settings.ts
```diff
@@ -560,13 +560,13 @@ export class CppSettings extends Settings {
...
         return super.Section.get<boolean>("dimInactiveRegions") === true
-            && super.Section.get<string>("intelliSenseEngine") === "Default"
+            && super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "default"
             && vscode.workspace.getConfiguration("workbench").get<string>("colorTheme") !== "Default High Contrast";
     }
 
     public toggleSetting(name: string, value1: string, value2: string): void {
         const value: string | undefined = super.Section.get<string>(name);
-        super.Section.update(name, value === value1 ? value2 : value1, getTarget());
+        super.Section.update(name, value === value1.toLowerCase() ? value2.toLowerCase() : value1.toLowerCase(), getTarget());
```
Colengms: I think this should be:

```
super.Section.update(name, value.toLowerCase() === value1.toLowerCase() ? value2 : value1, getTarget());
```
I assume the values should be lowercased for comparison but set to the values provided in the call to `toggleSettings`.

Extension/src/LanguageServer/customProviders.ts
Colengms: Alternatively, you could lower case these in the getters in `settings.ts` so they always return lower case strings.  Then all comparisons between settings originating from these getters would not need any conversions.  That would seem simpler, as we'd need to deal with potentially different casing in much fewer places (no other places?).

Also, for setters that are simply enabled/disabled, it would be most efficient if they were converted into bools by the getters in `settings.ts`, so could thereafter be evaluated as bools instead of needing to do any more expensive string conversions/comparisons.

Colengms (changes requested):

Extension/package.json
sean-mcmanus: I think it can be re-written, but you can't just remove the parens, because the && gets precedence.

bobbrow: What I commented on earlier is that:
```
(config.C_Cpp.intelliSenseEngine != 'disabled' || config.C_Cpp.intelliSenseEngine != 'Disabled')
```
should actually be
```
config.C_Cpp.intelliSenseEngine != 'disabled' && config.C_Cpp.intelliSenseEngine != 'Disabled'
```
So you don't need parentheses if the `||` separates all the `&&` conditions

sean-mcmanus: Ah, yeah, `config.C_Cpp.intelliSenseEngine != 'disabled' || config.C_Cpp.intelliSenseEngine != 'Disabled'` would always evaluate to true.

Colengms (approved):

sean-mcmanus (approved):

sean-mcmanus:
@bobbrow Can you re-review/approve this? It says you're requesting changes still.

Extension/package.json
```diff
@@ -5103,27 +5103,27 @@
...
+          "when": "config.C_Cpp.intelliSenseEngine != 'disabled' && config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'disabled' && config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'disabled' && config.C_Cpp.intelliSenseEngine != 'Disabled' && editorLangId == 'cuda-cpp'",
           "group": "custom1@3"
         },
         {
           "command": "C_Cpp.RunCodeAnalysisOnActiveFile",
-          "when": "config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cuda-cpp'",
+          "when": "config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'cuda-cpp'",
```
bobbrow: The condition got flipped incorrectly.  And since we can't do parenthesis, you'll have to have 6 pairs of `&&` (two checks per language)

```
"config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'c' ||
 config.C_Cpp.intelliSenseEngine == 'default' && editorLangId == 'c' ||
 config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cpp' ||
 config.C_Cpp.intelliSenseEngine == 'default' && editorLangId == 'cpp' ||
 config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cuda-cpp' ||
 config.C_Cpp.intelliSenseEngine == 'default' && editorLangId == 'cuda-cpp'"
```

Extension/package.json
```diff
@@ -5103,27 +5103,27 @@
...
           "group": "custom2@1"
         },
         {
           "command": "C_Cpp.RestartIntelliSenseForFile",
-          "when": "config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine == 'Default' && editorLangId == 'cuda-cpp'",
+          "when": "config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'c' || config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'cpp' || config.C_Cpp.intelliSenseEngine != 'default' && config.C_Cpp.intelliSenseEngine != 'Default' && editorLangId == 'cuda-cpp'",
```
bobbrow: Same thing here.  6 pairs of `&&` conditions where you check the engine `==`, not `!=`.

Extension/src/LanguageServer/settings.ts
```diff
@@ -202,15 +202,15 @@ export class CppSettings extends Settings {
     public get clangFormatStyle(): string | undefined { return super.Section.get<string>("clang_format_style"); }
     public get clangFormatFallbackStyle(): string | undefined { return super.Section.get<string>("clang_format_fallbackStyle"); }
     public get clangFormatSortIncludes(): string | undefined { return super.Section.get<string>("clang_format_sortIncludes"); }
-    public get experimentalFeatures(): string | undefined { return super.Section.get<string>("experimentalFeatures"); }
+    public get experimentalFeatures(): boolean | undefined { return super.Section.get<string>("experimentalFeatures")?.toLowerCase() === "enabled"; }
     public get suggestSnippets(): boolean | undefined { return super.Section.get<boolean>("suggestSnippets"); }
-    public get intelliSenseEngine(): string | undefined { return super.Section.get<string>("intelliSenseEngine"); }
-    public get intelliSenseEngineFallback(): string | undefined { return super.Section.get<string>("intelliSenseEngineFallback"); }
+    public get intelliSenseEngine(): boolean | undefined { return super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "enabled"; }
```
bobbrow: Making these settings booleans and not changing their names makes for some code that reads strangely:
```
if (intelliSenseEngine === false)
```
If you want to keep them as booleans, they should be renamed like `intelliSenseEngineEnabled` so you know what the true/false refers to.  Otherwise, keep them as strings and don't change the names (like `errorSquiggles`)

bobbrow (dismissed):

Extension/src/LanguageServer/settings.ts
browntarik: It turns out I need to roll back the changes made to intellisenseEngine because the options are not only enabled,disabled. I don't know why I missed this the first time

Colengms: While `intelliSenseEngine` appears to have been incorrectly converted to a bool, IMHO consistency is important.   Currently, the settings getters are consistently named to match the setting (with some inconsistency).  Whatever we do, we should intend to be consistent, or to make things more (not less) consistent.  There are lots of (all) existing bool settings that do not use the 'is' syntax.  If we want to use the "is" syntax consistently, we should apply it to the rest of code base as well.  IMHO, it seems that the other settings, with only "Enabled" and "Disabled" values, actually read intuitively as booleans.  I think it would be better to keep the setting names consistent.

bobbrow: I don't recall asking for 'is' at the beginning of these. Boolean naming is typically a predicate of some kind, never a noun. 'Is' is an acceptable way to start a predicate, but is not required. Many verbs work and there are several examples here.

With the exception of the `vcFormat.space` and `vcFormat.newLine` options where "insert" is implied to keep the name shorter, there are only a few boolean `get` properties here that don't follow that convention. The settings with enum values were named appropriately based on what they were (not booleans).  Had they been defined as booleans in package.json, they would have been given predicate names. If you want to change the type to boolean behind the scenes, that's fine, but it is best to fix the name too so that the code reads well.

This is my recommendation. You are welcome to accept or reject. I will not block the PR for this.

bobbrow:
I removed my block on this PR, but the 2 package.json `when` clauses need to be updated before this PR goes in.

Extension/src/LanguageServer/settings.ts
```diff
@@ -204,7 +204,7 @@ export class CppSettings extends Settings {
     public get clangFormatSortIncludes(): string | undefined { return super.Section.get<string>("clang_format_sortIncludes"); }
     public get experimentalFeatures(): boolean | undefined { return super.Section.get<string>("experimentalFeatures")?.toLowerCase() === "enabled"; }
     public get suggestSnippets(): boolean | undefined { return super.Section.get<boolean>("suggestSnippets"); }
-    public get intelliSenseEngine(): boolean | undefined { return super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "enabled"; }
+    public get intelliSenseEngine(): string | undefined { return super.Section.get<string>("intelliSenseEngine"); }
```
bobbrow: This should still return the lower-case version.
```
return super.Section.get<string>("intelliSenseEngine")?.toLowerCase();
```

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -166,7 +166,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
         if (config.preLaunchTask) {
             config.configSource = this.getDebugConfigSource(config, folder);
-            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === false;
+            const isIntelliSenseDisabled: boolean = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined).intelliSenseEngine === "Disabled";
```
bobbrow: All of these comparisons for intelliSenseMode should be against the lowercase version: 'disabled'

Extension/src/LanguageServer/customProviders.ts
```diff
@@ -158,7 +158,7 @@ export class CustomConfigurationProviderCollection {
     }
 
     public add(provider: CustomConfigurationProvider, version: Version): boolean {
-        if (new CppSettings(ext.getActiveClient().RootUri).intelliSenseEngine === false) {
+        if (new CppSettings(ext.getActiveClient().RootUri).intelliSenseEngine === "Disabled") {
```
bobbrow: "disabled"

Extension/src/LanguageServer/settings.ts
```diff
@@ -288,7 +288,7 @@ export class CppSettings extends Settings {
 
     public get enhancedColorization(): boolean {
         return super.Section.get<string>("enhancedColorization")?.toLowerCase() === "enabled"
-            && super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "default"
+            && super.Section.get<string>("intelliSenseEngine") === "Default"
```
bobbrow:
```
&& super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "default"
```

Extension/src/LanguageServer/settings.ts
```diff
@@ -560,7 +560,7 @@ export class CppSettings extends Settings {
 
     public get dimInactiveRegions(): boolean {
         return super.Section.get<boolean>("dimInactiveRegions") === true
-            && super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "default"
+            && super.Section.get<string>("intelliSenseEngine") === "Default"
```
bobbrow:
```
&& super.Section.get<string>("intelliSenseEngine")?.toLowerCase() === "default"
```

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
     // Read the setting and determine whether we should activate the language server prior to installing callbacks,
     // to ensure there is no potential race condition. LanguageServer.activate() is called near the end of this
     // function, to allow any further setup to occur here, prior to activation.
-    const isIntelliSenseEngineDisabled: boolean = settings.intelliSenseEngine === false;
+    const isIntelliSenseEngineDisabled: boolean = settings.intelliSenseEngine === "Disabled";
```
bobbrow: "disabled"

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
     const shouldActivateLanguageServer: boolean = (!isIntelliSenseEngineDisabled && !isOldMacOs);
 
     if (isOldMacOs) {
         languageServiceDisabled = true;
         vscode.window.showErrorMessage(localize("macos.version.deprecated", "Versions of the C/C++ extension more recent than {0} require at least macOS version {1}.", "1.9.8", "10.12"));
     } else {
-        if (settings.intelliSenseEngine === false) {
+        if (settings.intelliSenseEngine === "Disabled") {
```
bobbrow: "disabled"

Extension/src/main.ts
```diff
@@ -81,21 +81,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
...
             languageServiceDisabled = true;
         }
-        let currentIntelliSenseEngineValue: boolean | undefined = settings.intelliSenseEngine;
+        let currentIntelliSenseEngineValue: string | undefined = settings.intelliSenseEngine;
         disposables.push(vscode.workspace.onDidChangeConfiguration(() => {
             const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
             if (!reloadMessageShown && settings.intelliSenseEngine !== currentIntelliSenseEngineValue) {
-                if (currentIntelliSenseEngineValue === false) {
+                if (currentIntelliSenseEngineValue === "Disabled") {
```
bobbrow: "disabled"

Extension/src/main.ts
```diff
@@ -141,7 +141,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
         LanguageServer.registerCommands(false);
         // The check here for isIntelliSenseEngineDisabled avoids logging
         // the message on old Macs that we've already displayed a warning for.
-        log(localize("intellisense.disabled", "intelliSenseEngine is disabled"));
+        log(localize("intellisense.disabled", "intelliSenseEngine is Disabled"));
```
bobbrow: disabled

bobbrow (changes requested):

Extension/src/main.ts
browntarik: oh I misunderstood, changing these back

Extension/src/Debugger/configurationProvider.ts
bobbrow: missed this one, but I pushed a change with the fix.

bobbrow (approved):

michelleangela (approved):

Extension/package.json
```diff
@@ -2158,11 +2158,11 @@
         "C_Cpp.intelliSenseEngine": {
           "type": "string",
           "enum": [
-            "Default",
+            "default",
             "Tag Parser",
```
sean-mcmanus, line 2162: What about the "Tag Parser" enum? That seems inconsistent now.

Extension/package.json
```diff
@@ -1625,16 +1625,16 @@
...
             "vcFormat",
-            "Default",
-            "Disabled"
+            "default",
+            "disabled"
           ],
           "markdownEnumDescriptions": [
             "%c_cpp.configuration.formatting.clangFormat.markdownDescription%",
             "%c_cpp.configuration.formatting.vcFormat.markdownDescription%",
             "%c_cpp.configuration.formatting.Default.markdownDescription%",
```
sean-mcmanus, line 1634: Maybe these string for the enum descriptions should be updated too?

Extension/package.json
sean-mcmanus: should it be "tagParser"? The C++ side might need updating.

sean-mcmanus (changes requested):
I made some comments

sean-mcmanus (approved):

PR #9874, opened 2022-09-13
Update IntelliSense strings.

sean-mcmanus:
It seems like some of these strings might not be translated ideally, e.g. "void" is translated in some cases when it seems like it shouldn't, since it's a C/C++ keyword. We can fix those issues later.


sean-mcmanus:
I'm checking this in now because the cpptools side changes got in already.

Issue #9837, opened 2022-09-02
Browse database (.BROWSE.VC.DB) goes big unexpectedly | Find All Reference slow


mhtvsSFrpHdE:
# Update
Temporary solution and cause of problem found, see this reply:  
https://github.com/microsoft/vscode-cpptools/issues/9837#issuecomment-1247146528

### Environment

- OS and Version: Windows 10 Pro 21H1 19043.1889 Experience Pack 120.2212.4180.0
- VS Code Version: 1.71.0 (user setup) Commit:`784b0177c56c607789f9638da7b6bf3230d47a8c`
- C/C++ Extension Version: v1.12.4
- Other extensions you installed (and if the issue persists after disabling them):  `ms-vscode.cpptools-extension-pack`, `ms-python.python`, `TabNine.tabnine-vscode`, `shardulm94.trailing-spaces`
- If using SSH remote, specify OS of remote machine:  No
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).

Hey guys, I know the days of carefully optimizing software are over,  
but I did surprised how "by design" can effect my PC.  
I wrote a VSCode cpptools performance guide for my friend and wish to share to you,  
so you may identify some problem in current design.

The original document is [here](https://github.com/mhtvsSFrpHdE/prefetch/tree/master/qpp/prefetch#warning-vscode-cpptools), but you don't need to open it at all.

### Bugged `.BROWSE.VC.DB`

The one contains include information, VSCode will parse include path to this database,  
then use them to analysis C++ source file.

However, when using giant library like `Qt` and `boost`, VSCode can take a century to iterate through file system,  
and save 70000+ file information to `.BROWSE.VC.DB`.  
Until it was finished, CPU and hard disk never gets a rest.

This is not the worst, everytime a VSCode close/open, cpptools may drop entire database and create from zero.  
I collect some information about when the database can lost:

- Close VSCode during "Parsing workspase files"  
   This can happen if you added new library to include path recently.
- You did a database backup and restore  
   But when backup, `.BROWSE.VC.DB-shm` and `.BROWSE.VC.DB-wal` exist on disk.  
   Even if VSCode is safe shutdown(not parsing), sometimes the two file can still exist.  
   Do not restore these two small files, and delete the two exist before start VSCode,  
   or cpptools can refuse to accept the restore.
[...]

PR #9881, opened 2022-09-15
Fix browse database goes big unexpectedly (#9837)

mhtvsSFrpHdE:
I have no any experience in JavaScript and TypeScript, these code syntax are come from my imagination so beware.

After code, I don't even know how to compile and run.

Just try anything I can do to speed up a problem to be solved.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -918,6 +918,9 @@ export class CppProperties {
             }
 
             configuration.browse.limitSymbolsToIncludedHeaders = this.updateConfigurationStringOrBoolean(configuration.browse.limitSymbolsToIncludedHeaders, settings.defaultLimitSymbolsToIncludedHeaders, env);
+            if (configuration.browse.limitSymbolsToIncludedHeaders !== false) {
+                configuration.browse.limitSymbolsToIncludedHeaders = true;
+            }
             configuration.browse.databaseFilename = this.updateConfigurationString(configuration.browse.databaseFilename, settings.defaultDatabaseFilename, env);
 
             if (i === this.CurrentConfigurationIndex) {
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -909,14 +909,9 @@ export class CppProperties {
             if (!configuration.browse.path) {
                 if (settings.defaultBrowsePath) {
                     configuration.browse.path = settings.defaultBrowsePath;
-                } else if (configuration.includePath) {
-                    // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
-                    configuration.browse.path = configuration.includePath.slice(0);
-                    if (configuration.includePath.findIndex((value: string, index: number) =>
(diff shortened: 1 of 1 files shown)


mhtvsSFrpHdE:
```
configuration.browse.limitSymbolsToIncludedHeaders = this.updateConfigurationStringOrBoolean(configuration.browse.limitSymbolsToIncludedHeaders, settings.defaultLimitSymbolsToIncludedHeaders, env);
```

That means, we have no detection on, if a boolean value is `undefined` or `NULL`?  
I assume this one should have a default value of `true`, but it will be ignored if this value not found in user config.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -909,20 +909,18 @@ export class CppProperties {
...
                     configuration.browse.path = settings.defaultBrowsePath;
-                } else if (configuration.includePath) {
-                    // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
-                    configuration.browse.path = configuration.includePath.slice(0);
-                    if (configuration.includePath.findIndex((value: string, index: number) =>
-                        !!value.match(/^\$\{(workspaceRoot|workspaceFolder)\}(\\\*{0,2}|\/\*{0,2})?$/g)) === -1
-                    ) {
-                        configuration.browse.path.push("${workspaceFolder}");
-                    }
+                } else {
```
sean-mcmanus, line 912: It is intentional to add the includePath paths to the browse.path as the default behavior. It is expected that users who don't want those paths in it will override the default.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -909,20 +909,18 @@ export class CppProperties {
...
+                    // If the user doesn't set browse.path, make sure ${workspaceFolder} is in there though...
+                    configuration.browse.path = ["${workspaceFolder}"];
                 }
             } else {
                 configuration.browse.path = this.updateConfigurationPathsArray(configuration.browse.path, settings.defaultBrowsePath, env);
             }
 
             configuration.browse.limitSymbolsToIncludedHeaders = this.updateConfigurationStringOrBoolean(configuration.browse.limitSymbolsToIncludedHeaders, settings.defaultLimitSymbolsToIncludedHeaders, env);
+            if (configuration.browse.limitSymbolsToIncludedHeaders !== false) {
+                configuration.browse.limitSymbolsToIncludedHeaders = true;
```
sean-mcmanus, line 922: This should already be defaulting to `true`. The problem you may be seeing is that our include iteration code only checks the filenames and not the complete paths that is used when combined with the includePath, i.e. `#include <string.h>` would include symbols for all "string.h" files found and not the one that is chosen based on the includePath.

sean-mcmanus (changes requested):

sean-mcmanus:
We have some docs on contributing at https://github.com/microsoft/vscode-cpptools/blob/main/CONTRIBUTING.md . FYI, the code for cpptools/cpptools-srv is closed source -- it's where all the parsing is done.

If you have ideas on how we can avoid a large database let us know and we could potentially make the change ourselves.

PR #9895, opened 2022-09-19
SSH terminal for targets

xisui-MSFT:
SSH terminal for targets:
sshterminal


Extension/src/Debugger/extension.ts
```diff
@@ -130,6 +131,12 @@ export function dispose(): void {
     disposables.forEach(d => d.dispose());
 }
 
+function sshTerminal(node: TargetLeafNode): void {
+    const terminal: vscode.Terminal = vscode.window.createTerminal(`SSH: ${node.name}`);
+    terminal.sendText(`ssh ${node.name}`);
```
merlynomsft: To consider: Can 'name' contain any characters that we'd need to quote or escape?

merlynomsft (approved):

Extension/src/Debugger/extension.ts
xisui-MSFT: Strings are escaped automatically.

xisui-MSFT: Did some tests
- Quotes in names are not supported at all in SSH config, so nothing will show up in our view.
- Everything after spaces in names are omitted

So we only need to add quotes around node.name to make the corner cases work.

olegkhar (approved):

Extension/package.nls.json
```diff
@@ -36,6 +36,7 @@
     "c_cpp.command.selectSshTarget.title": "Select SSH target",
     "c_cpp.command.activeSshTarget.title": "Get the active SSH target",
     "c_cpp.command.refreshCppSshTargetsView.title": "Refresh",
+    "c_cpp.command.sshTerminal.title": "Connect to this target in a new terminal",
```
sean-mcmanus: Should "target" be "SSH target"? I'm concerned the translators may not know how to translate it consistently with the other references to "SSH target" in the other strings. Maybe a "comment" to translators could be added instead.

Extension/package.json
```diff
@@ -5073,12 +5079,17 @@
...
           "when": "viewItem == CppSshTargetsView.targetLeafRemovable || viewItem == CppSshTargetsView.targetLeafRemovableCanSetActive",
-          "group": "inline@1"
+          "group": "inline@2"
         },
         {
           "command": "C_Cpp.setActiveSshTarget",
           "when": "viewItem == CppSshTargetsView.targetLeafCanSetActive || viewItem == CppSshTargetsView.targetLeafRemovableCanSetActive",
           "group": "inline@0"
+        },
+        {
```
sean-mcmanus, line 5084: Would it be better to sort these by the inline@ number?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

sean-mcmanus:
@xisui-MSFT You want this checked in for 1.13.1?

xisui-MSFT:
> @xisui-MSFT You want this checked in for 1.13.1?

Sure if that fits the release schedule.

Issue #9899, opened 2022-09-20
Strange "Failed to read file C:\ProgramData\ssh\ssh_config." in logging


sean-mcmanus:
What is this error? Can we get rid of it?

bobbrow:
Sorry, I was thinking about how CMake Tools does logging.  In cpptools, you should check the value of the c_cpp.loggingLevel before logging to the output channel.  These messages should only show if it's set to "Debug"

PR #9901, opened 2022-09-20
Add an SSH logging channel

xisui-MSFT:
#9899


Extension/src/logger.ts
```diff
@@ -96,6 +97,13 @@ export function getDiagnosticsChannel(): vscode.OutputChannel {
     return diagnosticsChannel;
 }
 
+export function getSshChannel(): vscode.OutputChannel {
+    if (!sshChannel) {
+        sshChannel = vscode.window.createOutputChannel(localize("c.cpp.ssh.channel", "{0}: SSH", "Cpptools"));
```
sean-mcmanus, line 102: Can this be "C/C++ SSH" to match the naming style of the other logging windows?

sean-mcmanus (changes requested):

sean-mcmanus:
So is the Failed to read SSH message "by design"?

bobbrow:
> So is the Failed to read ssh_config message "by design"?

I don't think we should be showing failure messages by default for this.  If the user is not configuring SSH, they should not see any of these messages.

Extension/src/logger.ts
xisui-MSFT: I would like it to match SSH targets view (Cpptools: SSH Targets), since as we discussed earlier, this feature can be used for other languages.

xisui-MSFT:
> > So is the Failed to read ssh_config message "by design"?
> 
> I don't think we should be showing failure messages by default for this. If the user is not explicitly configuring SSH, they should not see any of these messages.

Yeah so I changed this to only output that message for warning+ logging level. But since we are not reading other files currently, So I'll change it to ignore the error when err.code === 'ENOENT'. Might cause some confusion but chance is low.

sean-mcmanus (approved):

Issue #9930, opened 2022-09-29
SSH config parsing exception blocks extension from activating


xisui-MSFT:
### Environment

- OS and version:
- VS Code:
- C/C++ extension:
- OS and version of remote machine (if applicable):

### Bug Summary and Steps to Reproduce

Bug Summary:

Steps to reproduce:
1. In this environment...
2. With this config...
3. Do '...'
4. See error...

xisui-MSFT:
It's not a regression.

PR #9933, opened 2022-09-29
Don't throw when parsing SSH config

xisui-MSFT:
ssh-config throws when fail to parse SSH config, this blocks the extension from activating since we parse the config files on activation to initialize SSH features.

#9930


Colengms (approved):

sean-mcmanus:
@WardenGnaw Did you want to code review this?

@xisui-MSFT Is this good for 1.13.2 Wednesday?

PR #9950, opened 2022-10-04
Fix issue with Generate Doxygen Comment context menu

Colengms:
I'm not sure if this is due to the JSON refactoring, or a recent regression in VS Code.

VS Code appears to be delivering a Uri object as the argument to our `GenerateDoxygenComment` command when invoked via the editor context menu, despite that we have not indicated anywhere that this should be provided as an argument.  I can't find documentation that indicates why this arg is being added.  I'm reasonably sure this was not being passed in previous versions of VS Code, but it's possible it was and was simply not causing an issue prior to recent refactoring for JSON changes.

This change updates our handling of the command to ignore the Uri, and not confuse it with the arg we intentionally pass for Code Actions.  It's unclear to me why this arg would be automatically passed by VS Code, as right-clicking to display the context menu does set `vscode.window.activeTextEditor`.  If there is some reason why using `vscode.window.activeTextEditor` is insufficient, this code will need to be changed (as that may imply it also cannot rely on the current selection/cursor to identify where to do the insert).

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -682,7 +682,7 @@ export interface Client {
     handleConfigurationEditUICommand(viewColumn?: vscode.ViewColumn): void;
     handleAddToIncludePathCommand(path: string): void;
     handleGoToDirectiveInGroup(next: boolean): Promise<void>;
-    handleGenerateDoxygenComment(codeActionArguments: DoxygenCodeActionCommandArguments | undefined): Promise<void>;
+    handleGenerateDoxygenComment(args: DoxygenCodeActionCommandArguments | vscode.Uri | undefined): Promise<void>;
     handleCheckForCompiler(): Promise<void>;
     handleRunCodeAnalysisOnActiveFile(): Promise<void>;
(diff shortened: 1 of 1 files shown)


browntarik (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -682,7 +682,7 @@ export interface Client {
     handleConfigurationEditUICommand(viewColumn?: vscode.ViewColumn): void;
     handleAddToIncludePathCommand(path: string): void;
     handleGoToDirectiveInGroup(next: boolean): Promise<void>;
-    handleGenerateDoxygenComment(codeActionArguments: DoxygenCodeActionCommandArguments | undefined): Promise<void>;
+    handleGenerateDoxygenComment(args: DoxygenCodeActionCommandArguments | vscode.Uri | undefined): Promise<void>;
```
bobbrow, line 685: in the case where the uri is passed, do we not get the true args in a second parameter?

Colengms: The Uri is passed when using the context menu.  The DoxygenCodeActionCommandArguments are passed when a CodeAction is used.  There is no scenario in which both are passed.

PR #9955, opened 2022-10-04
Add a setting to select which Doxygen sections appear in the hover tooltip

### LINKED ISSUE #8525: Customizable Doxygen tags for Quick Info
Type: Feature Request

Could you please implement @details and eventually also @note and @warning in quick info.
If it cannot be a separate section, at least show the text after that of @brief.

Another suggestion : implement @n to force line break

Many thanks.

browntarik:
Add a setting that allows users to select which Doxygen tags appear in simplified structured view. This setting only allows the selection of Doxygen Section Header tags.

Fixes: https://github.com/microsoft/vscode-cpptools/issues/8525


Extension/package.json
```diff
@@ -2361,6 +2361,37 @@
...
+              "result",
+              "return",
+              "since",
+              "test",
+              "todo",
+              "version",
+              "warning"
+              ],
+          "description": "%c_cpp.configuration.doxygen.selectDoxygenTags.description%",
+          "scope": "resource"
```
bobbrow, line 2401: This should have a default value that matches the current set of hard-coded tags in the extension today.

Something like:
```
  "default": [
    "brief",
    "tparam",
    "param",
    "return",
    "retval",
    "deprecated",
    "exception"
  ]
```

Extension/package.json
```diff
@@ -2361,6 +2361,37 @@
           "description": "%c_cpp.configuration.doxygen.generatedStyle.description%",
           "scope": "resource"
         },
+        "C_Cpp.doxygen.selectDoxygenTags": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
```
bobbrow, line 2372: The tags that are currently hard-coded in the language server should be available in this enum.  It looks like some of them are missing.  The user should have the ability to remove any of the default values if they want.

Extension/src/nativeStrings.json
```diff
@@ -178,15 +178,35 @@
...
+    "invariant_label": "Invariant:",
     "file_label": "File:",
+    "note_label": "Note:",
     "parameters_label": "Parameters:",
+    "pre_label": "Precondition:",
+    "post_label": "Postcondition:",
+    "remark_label": "Remark:",
+    "remarks_label": "Remarks:",
+    "result_label": "Result:",
+    "return_label": "Return:",
```
Colengms, line 202: Our loc team tends to have a lot of difficulty with strings in our area, as they aren't coders and aren't generally familiar with the concepts involved. I'd recommend providing a hint for any word even remotely ambiguous.  Maybe add hints for `Exceptions` and `Return`.

Extension/package.json
sean-mcmanus: I think we should add some defaults though, such as "warning", possibly others if we have a reason to believe they'd make a better default.

Extension/package.json
```diff
@@ -2373,21 +2373,26 @@
...
+              "returns",
+              "retval",
               "since",
+              "tparam",
               "test",
               "todo",
               "version",
               "warning"
               ],
+          "default": [ "brief", "param", "tparam", "returns", "retval", "warning", "note" ],
```
bobbrow: "brief" is missing from the enum list.
"tparam" comes before "param"
"deprecated" and "exceptions" should be in this list somewhere after "retval"

Extension/package.json
```diff
@@ -2361,6 +2361,42 @@
           "description": "%c_cpp.configuration.doxygen.generatedStyle.description%",
           "scope": "resource"
         },
+        "C_Cpp.doxygen.selectDoxygenTags": {
```
bobbrow: Sorry I missed this earlier.  I think we need a better name for the setting.  Maybe `C_Cpp.doxygen.sectionTags`

Extension/package.json
```diff
@@ -2361,6 +2361,42 @@
...
+              "retval",
+              "since",
+              "tparam",
+              "test",
+              "todo",
+              "version",
+              "warning"
+              ],
+          "default": [ "brief", "param", "tparam", "returns", "retval", "warning", "note" ],
+          "description": "%c_cpp.configuration.doxygen.selectDoxygenTags.description%",
```
bobbrow: You need to add an entry in `package.nls.json` for this.

Extension/src/LanguageServer/settings.ts
```diff
@@ -386,7 +387,7 @@ export class CppSettings extends Settings {
     public get useBacktickCommandSubstitution(): boolean | undefined { return super.Section.get<boolean>("debugger.useBacktickCommandSubstitution"); }
     public get codeFolding(): boolean { return super.Section.get<string>("codeFolding")?.toLowerCase() === "enabled"; }
     public get caseSensitiveFileSupport(): boolean { return !isWindows() || super.Section.get<string>("caseSensitiveFileSupport") === "enabled" ; }
-
+    public get doxygenSelectDoxygenTags(): string[] | undefined {return super.Section.get<string[]>("doxygen.selectDoxygenTags"); }
```
sean-mcmanus: Space between { and return

michelleangela:
Could you add a screenshot of the new settings to the description of this PR?

michelleangela (approved):

Extension/package.json
```diff
@@ -2364,6 +2364,42 @@
           "description": "%c_cpp.configuration.doxygen.generatedStyle.description%",
           "scope": "resource"
         },
+        "C_Cpp.doxygen.sectionTags": {
+          "type": "array",
+          "default": ["tparam","param", "returns", "retval", "exceptions", "deprecated", "warning", "note" ],
```
sean-mcmanus: Add a space after `,` -- you should be able to run the Format Document command.

sean-mcmanus: Can you add "attention" to the defaults?

"result" and "return" are the same as "returns" -- should those be added too?

Extension/src/LanguageServer/settings.ts
```diff
@@ -387,7 +388,7 @@ export class CppSettings extends Settings {
     public get useBacktickCommandSubstitution(): boolean | undefined { return super.Section.get<boolean>("debugger.useBacktickCommandSubstitution"); }
     public get codeFolding(): boolean { return super.Section.get<string>("codeFolding")?.toLowerCase() === "enabled"; }
     public get caseSensitiveFileSupport(): boolean { return !isWindows() || super.Section.get<string>("caseSensitiveFileSupport") === "enabled" ; }
-
+    public get doxygenSectionTags(): string[] | undefined { return super.Section.get<string[]>("doxygen.selectDoxygenTags"); }
```
sean-mcmanus: selectDoxygenTags seems incorrect -- should it be sectionTags?

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
```diff
@@ -178,15 +178,38 @@
         "hint": "Refers to C++ template specializations. This is part of a description indicating there are N additional specializations of a function.  %d is replaced with that number."
     },
     "expands_to": "Expands to:",
+    "attention_label": "Attention:",
```
sean-mcmanus, line 181: @Colengms Should these be added to the end so it doesn't trigger re-translations for all the strings after this?

Extension/package.json
browntarik: I cannot add a flag for return in order to recognize it as a default the same way we do returns because it is a keyword in cpp. In order to keep consistency rather than trying to find a workaround I just removed it since we already support "returns" which has the same functionality.

sean-mcmanus: I don't understand. Why exactly can't you add "return"? The user's code might use "return".

sean-mcmanus (commented):
How do I cancel my review?

sean-mcmanus (approved):
I'm approving, but only because I'll be OOF and it won't let me cancel my review.

PR #9989, opened 2022-10-12
Add #include <c++> highlighting

benmcmorran:
Based on this tweet https://twitter.com/timur_audio/status/1172558628987031553

includecpp

Changes that should probably be made before this ships:

- Limit highlighting to C++ file types
- Rate limit how often we scan for `#include <c++>` in the file. Although it currently only ever scans the first 4k of the file, it does the scan on every keystroke.

### DIFF
--- a/Extension/src/includeCpp.ts
+++ b/Extension/src/includeCpp.ts
@@ -1,0 +1,88 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


sean-mcmanus (changes requested):
I don't think this should get checked in. Our product doesn't have any easter eggs and neither does VS Code itself. Does VS 2022 have this? We shouldn't be doing something just because CLion does it. Also, the Microsoft fwlink is sending users to a non-Microsoft site: https://www.includecpp.org/ -- that is potentially a security risk. VS Code or users may trust the Microsoft domain but not the other one.

benmcmorran:
I'm fine with this going either way and understand that there may be policy concerns with adding this to the product. There was an ask after my demo to open a draft PR so that you all could look at the changes.

bobbrow:
Yeah, sorry @benmcmorran, we won't be taking this into the product unless I hear different from management.

PR #10007, opened 2022-10-14
Disable Generate Doxygen Comment from context menu if intelliSenseEngine is "disabled"

sean-mcmanus:


Colengms (changes requested):
If the file is applicable, the extension should already be activated or in the process of activating.  Internally, they should be waiting for the language client to be ready (i.e. `notifyWhenLanguageClientReady`), IF needed.  If the client is not activated yet, checks that rely on it (i.e. `config.C_Cpp.intelliSenseEngine == 'Default'`) should fail, so inapplicable items should not be getting displayed.  If there is an issue here with commands not running when run too early, or being enabled when they should not, that would be a bug.  I don't think we should removing things from the menu to avoid investigating and fixing that.

I see no reason we should be hiding anything from the menu.  Commands _cause_ activation.  If it were not the intention to invoke commands prior to being activated, commands would not cause activation.

sean-mcmanus:
Sure, I can change it to just check the intelliSenseEngine setting.

Colengms (approved):

PR #10026, opened 2022-10-17
Update handleconfiguration() calls to only update cpp_properties.json when a document is saved not just edited 

### LINKED ISSUE #9185: Modifications to `c_cpp_properties.json` should not be applied until saved
Currently, modifications to `c_cpp_properties.json` go live immediately as the file is being modified.  If parsing all configurations requires some time, this can result in a backlog of requests sent to the native process, delaying subsequent IntelliSense operations.  Changes probably shouldn't go live until the file is saved.

Showing error squiggles in `c_cpp_properties.json` is currently dependent on this behavior, so that would need to be updated to parse the live view of the document instead.

browntarik:
Fixes https://github.com/microsoft/vscode-cpptools/issues/9185


Extension/src/LanguageServer/configurations.ts
```diff
@@ -241,23 +241,12 @@ export class CppProperties {
...
-                this.handleConfigurationChange();
-            }
+            this.handleConfigurationChange();
         });
 
-        vscode.workspace.onDidChangeTextDocument((e: vscode.TextDocumentChangeEvent) => {
+        vscode.workspace.onDidChangeTextDocument((e) => {
             if (e.document.uri.fsPath === settingsPath) {
-                this.handleConfigurationChange();
+                this.handleSquiggles();
```
Colengms, line 250: Could you add the other change we discussed, to call `handleSquiggles()` as the file becomes visible?  And, could you avoid calling `handleSquiggles()` here if this document is not within the list of `vscode.window.visibleTextEditors`?

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
Colengms: Don't use onDidChangeActiveTextEditor.  There can only be one active editor, but there can be multiple that are visible.  You'll want to use onDidChangeVisibleTextEditors (which returns a list of the currently visible documents), keep track of whether the settings file is visible, and call handleSquiggles when it transitions from not-visible to visible.

browntarik: right I totally forgot about the need to iterate through the editors

Colengms: Also, you could you add a check here to onDidChangeTextDocument, to avoid calling handleSquiggles if the document is not visible?

browntarik: isn't this redundant? You wouldn't be able to change the json file unless it was visible

sean-mcmanus: @browntarik The user might use the JSON UI editor?

Colengms: Right, the user might edit the config in the config UI, or sync down an updated c_cpp_properties.json via a git pull.  The idea I'm advocating is to simply (and only) update squiggles when the file is modified AND visible, or when it becomes visible.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -241,26 +242,25 @@ export class CppProperties {
...
-            if (e.document.uri.fsPath === settingsPath) {
-                this.handleConfigurationChange();
+        vscode.workspace.onDidChangeTextDocument((e) => {
+            if (e.document.uri.fsPath === settingsPath && isCurrentWindow) {
+                this.handleSquiggles();
             }
         });
 
+        vscode.window.onDidChangeVisibleTextEditors((editors) => {
+            isCurrentWindow = false;
```
Colengms: `isCurrentWindow` is a bit vague.  How about `isCppPropertiesJsonVisible`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -227,6 +227,7 @@ export class CppProperties {
         }
 
         const settingsPath: string = path.join(this.configFolder, this.configurationGlobPattern);
+        let isCurrentWindow: boolean = false;
```
Colengms: You'll want to put this in the CppProperties class, rather than as a local variable of the CompilerDefaults setter.  (Which begs the question, Why are we setting up file watchers, etc., in a setter for CompilerDefaults ?  But no need to fix that here...)

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -241,26 +242,25 @@ export class CppProperties {
...
+                this.handleSquiggles();
             }
         });
 
+        vscode.window.onDidChangeVisibleTextEditors((editors) => {
+            this.isCppPropertiesJsonVisible = false;
+            editors.forEach(editor => {
+                if (editor.document.uri.fsPath  === settingsPath) {
+                    this.isCppPropertiesJsonVisible = true;
+                    this.handleSquiggles();
```
Colengms: This code will always update squiggles whenever cpp_properties is visible, regardless of whether that was the document the the visiblity of changed.  Try this: Before clearing, `this.isCppPropertiesJsonVisible` to false, use a local variable (i.e. `wasVisible`) to set aside the original value.  Then, only call `handleSquiggles` if `wasVisible` was previously false.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -241,26 +243,26 @@ export class CppProperties {
...
-            if (e.document.uri.fsPath === settingsPath) {
-                this.handleConfigurationChange();
+        vscode.workspace.onDidChangeTextDocument((e) => {
+            if (e.document.uri.fsPath === settingsPath && this.isCppPropertiesJsonVisible) {
+                this.handleSquiggles();
             }
         });
 
+        vscode.window.onDidChangeVisibleTextEditors((editors) => {
+            wasVisible = this.isCppPropertiesJsonVisible;
```
Colengms: `wasVisible` is only referenced (written, then read) in this lambda function, so it's scope/lifespan can be limited to just this function.  Could you declare the variable here, instead of as a class member?  (Otherwise, the changes look good to me.)

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -241,26 +242,27 @@ export class CppProperties {
...
+            let wasVisible: boolean = false;
+            wasVisible = this.isCppPropertiesJsonVisible;
+            this.isCppPropertiesJsonVisible = false;
+            editors.forEach(editor => {
+                if (editor.document.uri.fsPath  === settingsPath && !wasVisible) {
+                    this.isCppPropertiesJsonVisible = true;
+                    this.handleSquiggles();
+                }
+            });
+        });
```
Colengms, line 265: A couple of problems remaining in this function.

```
let wasVisible: boolean = false;
wasVisible = this.isCppPropertiesJsonVisible;
```

Can be:

```
let wasVisible: boolean = this.isCppPropertiesJsonVisible;
```

And because you're only setting `isCppPropertiesJsonVisible` (back) to true if `!wasVisible`, you'll wind up leaving it as false whenever some other document becomes visible (even if cpp_properties is still also visible).

I'd suggest:

```
if (editor.document.uri.fsPath  === settingsPath) {
    this.isCppPropertiesJsonVisible = true;
    if (!wasVisible)
        this.handleSquiggles();
}
```

Colengms (changes requested):

Colengms (approved):

Issue #7610, opened 2021-05-27
Updates to compile_commands.json are not handled if specified using a relative path


Colengms:
If I specify `compileCommands` like so:
`"compileCommands": "build/compile_commands.json"`

.. it seems to work as expected, however changes to the file are not (always?) handled properly.

If I specify it like so:
`"compileCommands":"${workspaceFolder}/build/compile_commands.json"`

.. it appears to work as expected, including when changes are made.

It looks like we're resolving this path to within the workspace directory in some cases, but not in others.  If the current working directory is different, the file is looked for there.  This should either default to always resolving as relative to the workspace folder if not a full path (after variables like `${workspaceFolder}` are expanded).  Or, it should result in a squiggle in `c_cpp_properties.json`.

We had added some additional checking for changes to compile_commands.json, to address user reports that changes were not always being handled properly.  This is likely the actual cause, and that additional handling could potentially be removed.

michelleangela:
When the file path property `compileCommands` in the `c_cpp_properties.json` file (IntelliSense configuration) is set with a relative path, extension does not resolve path to the correct absolute path.

Examples of relative path cases for a workspace folder under path `Home/projects/myFolder` or the value of `${workspaceFolder}` is `Home/projects/myFolder` .


**Example path value:** `"${workspaceFolder}/build/compile_commands.json"` 
**Should resolve to:**  `"Home/projects/myFolder/build/compile_commands.json"` 

**Example path value:** ` "some_subfolder/compile_commands.json"`
**Should resolve to:**  `"Home/projects/myFolder/some_subfolder/compile_commands.json"` 

**Example path value:**` "compile_commands.json"`
**Should resolve to:**  `"Home/projects/myFolder/compile_commands.json"`

PR #10045, opened 2022-10-21
Prepending workspace path to relative paths from CppProperties

Bernardin-MS:
Issue: #7610


sean-mcmanus:
@Bernardin-MS The ESLint extension in VS Code can help detect the Linter issues locally. You can also run "yarn run lint".

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1381,6 +1381,11 @@ export class CppProperties {
             result = result.replace(/\*/g, "");
         }
 
+        // Make sure all paths result to an absolute path
+        if (!path.isAbsolute(result)) {
```
sean-mcmanus: Some existing callers of resolvePath appear to expect that the result of calling resolvePath may be a relative path, so this may cause a regression in those cases (not sure). Can you add another bool parameter to enable the resolving a relative path to an absolute one so this can be enabled in the one call?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: You could also modify the pre-existing callers of resolvePath that appear to expect a relative path in some cases (and verify they still work as expected), but it seems like that would be more work.

sean-mcmanus: Or maybe not make this change in resolvePath and instead do it outside of it?

Colengms: @sean-mcmanus I think you're right that we have other callers that check if the path is relative and adjust it if so.  But I think that logic is not well thought-out and should all be consolidated, presumably into resolvePath.  We cannot rely on relative paths, because we cannot rely on the current working directory, as it may be changed by an API call, etc. (We can't expect it to point to be a particular workspace in a multiroot workspace).  I think we should remove all of the other handling of relative paths, in favor of resolving relative paths into absolutely paths early - as we process incoming user-specified path values.  That way, from that point on, we can assume we're working with absolute paths everywhere, and don't have to deal with them in multiple places.

sean-mcmanus: @Colengms Yeah, that sounds good to me (but more work, more code paths to verify don't regress, as mentioned previously)...not sure if getting the fix in earlier is worth it and refactoring later though.

Bernardin-MS: @sean-mcmanus I looked through all the references for resolvePath() and verified it wouldn't harm anything if it only returned absolute paths

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1377,6 +1377,11 @@ export class CppProperties {
...
+            result = path.join(this.rootUri?.fsPath as string, result);
```
sean-mcmanus: This will fail with "'The "path" argument must be of type string. Received undefined" if this.rootUri is undefined. I'm not sure if that's possible but can you safeguard against that?

UPDATE: I mean I know this.rootUri is possible since that happens when no workspace folder is opened, but I wasn't sure if this code path was currently hittable in that case, but since this resolvePath is supposed to be general-purpose we should safeguard against any future use in that scenario.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
Colengms: I think this block needs to be moved up into the block that checks `if (this.rootUri)`.

Bernardin-MS: I think it'd be better to keep it at the bottom just to make sure

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1377,6 +1377,13 @@ export class CppProperties {
             result = result.replace(/\*/g, "");
         }
 
+        // Make sure all paths result to an absolute path
+        if (!path.isAbsolute(result)) {
```
sean-mcmanus: Can you combine the "if"? i.e. `if (!path.isAbsolute(result) && this.rootUri) {`?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #10056, opened 2022-10-24
Fix theme image paths

Colengms:
These had been updated to use relative paths.  It looks like VSCE automatically resolves them to repo URLs.  Unfortunately, that logic assumes the VSIX is being built from the root of the repo, which Themes is not.  So, it got the paths wrong.

I've only added `Themes/` to the paths VSCE generated.  So, the approach (i.e. using the github path, but not a raw content path) is consistent with what seems to be the default behavior for in-repo references.

We should probably move the Themes into their own repo.

### DIFF
--- a/Themes/README.md
+++ b/Themes/README.md
@@ -10,19 +10,19 @@ VS Code has since provided an API for semantic colorization. The C/C++ Extension
 
 Light Theme
 
-![Light Theme 2019 example](./assets/light.png)
+![Light Theme 2019 example](https://github.com/Microsoft/vscode-cpptools/raw/HEAD/Themes/assets/light.png)
 
 Dark Theme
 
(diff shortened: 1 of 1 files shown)


bobbrow:
There was another PR before the last one that changed the URL domain.  https://github.com/microsoft/vscode-cpptools/pull/7977 Should we use that instead?

bobbrow (approved):

Colengms:
> There was another PR before the last one that changed the URL domain. #7977 Should we use that instead?

I'm hesitant to diverge from the approach that VSCE uses (when it replaces relative paths with repo paths).  So, I'd like to go ahead with this change, and follow up if a change is still necessary to avoid broken image links in the marketplace.  If there is still an issue with broken image links in the marketplace, that would indicate a bug in VSCE as well.

PR #10066, opened 2022-10-26
Remove legacy WSL support in favor of the WSL extension

Colengms:
Removes our legacy direct support for WSL, in favor of the WSL extension (formerly the "Remote - WSL" extension).

There are still references to WSL in debugger related code.  My understanding from chatting with Andrew is that those are unrelated.


sean-mcmanus (approved):

gamazind:
Hello. I see that **Remote-SSH** stopped to suggest function names in IntelliSense too. I don`t see any debug logs too from C++ extension too. I compared stable versions 1.12.4 and 1.13.6. 
@Colengms, @sean-mcmanus could you please recheck it?

sean-mcmanus:
@gamazind It sounds like you may be hitting issue https://github.com/microsoft/vscode-cpptools/issues/10224, although that bug also repros with 1.12.14.

Issue #9681, opened 2022-08-03
Update VSIX release process to build TypeScript component on each native platform instead of all from the same host


Colengms:
Currently, cpptools' TypeScript component does not have any platform-dependent dependencies.  The Release pipeline for the cpptools VSIX builds the TypeScript component for each platform from the same (Windows) host.  That prevents us from introducing any platform-dependent dependencies, as they would be built for Windows and not the actual target platform.

This is preventing us from using the following dependencies, requiring us to work around these needs with non-ideal solutions:

 - wordexp (requiring us to deploy our own `cpptools-wordexp` utility)
 - node-pty (requiring us to add arguments to gcc and clang build tasks to force output colorization)

This issue is tracking updating our Release pipeline to build the TypeScript component separately on each target platform.

I believe this is blocked until Azure DevOps can provide ARM64 Mac support (or we can get an ARM64 Mac added back into our pool).

sean-mcmanus:
What about Windows/Linux ARM and Alpine?

Colengms:
@sean-mcmanus We have the ability to build for ARM and ARM64 Linux (using Docker and process emulation), but you may be right about needing ARM64  Windows support in the lab in addition to ARM64 Mac support.

`node-gyp` looks like it might provide a cross-compile solution, but I assume that would require target toolsets on the host.

Colengms:
This could be easily addressed once Mac M1 and Linux ARM64 VMs are available for Azure Pipelines.  Bugs tracking those are here:
Mac M1: https://github.com/actions/runner-images/issues/2187
Linux ARM64: https://github.com/actions/runner-images/issues/5631

Colengms:
Using `node-gyp` it looks like platform-dependent node modules can be configured to do cross-compilation, by specifying a specific compiler toolchain and tweaks to build args.  That should allow us to build for all architectures, provided just x64 hosts for Windows, Linux and Mac.

Colengms:
It looks like this can be accomplished using each target OS and does not require that the host be the same as the target architecture. We should be able to pass in an arg to yarn install (for node-gpy) specifying the target architecture for native components.

PR #10070, opened 2022-10-26
Fix issue with call to cpptools-wordexp with invalid arguments

Colengms:
There are 2 issues here.
 1. When command line arguments contain invalid shell escaping/content, `cpptools-wordexp` will (intentionally) exit with an error code.  This is translated into an exception by `execFileSync`, which wasn't being caught.
 2. On Alpine linux, despite explicit use of `shell: false` (which is also the default), `execFileSync` will perform shell parsing on arguments (making the invocation of `cpptools-wordexp` redundant).  This appears to be a bug in node or Alpine itself.  This does not occur on non-Alpine linux distros.  This also causes an exception to be thrown from `execFileSync` when the command line args contain invalid shell escaping/content.

Issue 1 is fixed by catching the exception, which also works around the symptoms of 2.  In both cases, the compilerPath will not be usable.

This does not address incorrect command line parsing results on Alpine, as shell parsing is occurring twice.  (This only impacts when a user pastes a full command line into `compilerPath` instead of just the compiler path).  This would be addressed by #9681 , as that would allow us to call `wordexp` directly.

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -1012,12 +1012,12 @@ function extractArgs(argsString: string): string[] {
         }
         return result;
     } else {
-        const wordexpResult: any = child_process.execFileSync(getExtensionFilePath("bin/cpptools-wordexp"), [argsString]);
-        if (wordexpResult === undefined) {
-            return [];
-        }
-        const jsonText: string = wordexpResult.toString();
         try {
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

sean-mcmanus:
Did you want this in 1.13.3 or 1.13.4?

Issue #10054, opened 2022-10-24
Additional problem matchers - ARMCC, IAR, Keil


morsisko:
### Feature Request

Hello,

I use few different compilers to compile C programs (ARMCC, IAR, Keil) those compilers outputs warnings and errors in different format than GCC compiler. Currently there is only problem matcher for GCC compiler, so if you compile your code with those I've listed above, erorrs and warnings won't be present in "Problems" window.

If I made pull request with problem matchers for those compilers, would you consider merging it to the main branch, or should I rather made my own vsc extension with those matchers?

Colengms:
Hi @morsisko .  Yes, we would be open to a contribution of problem matchers.  :)

sean-mcmanus:
Your fix is available with 1.13.4 (pre-release): https://github.com/microsoft/vscode-cpptools/releases/tag/v1.13.4

morsisko:
Thanks for the information

PR #10085, opened 2022-10-30
Added IAR problem matcher (#10054)

morsisko:
This pull requests adds IAR problem matcher.

Warning example that is generated by IAR compiler:

`"D:\iar_test\main.c",6  Warning[Pe177]: variable "c" was declared but never referenced`

There is one thing that needs to be said (not sure if it is issue or not). By default for some reasons IAR splits the error/warning message over multiple lines, if the message is too long. In such case problem matcher will catch only the first part of the message, for example:
Code_2022-10-30_23-45-24

However, there is an option that can be passed to IAR - `--no_wrap_diagnostics` in such situation the problem matcher works without any problem.
Code_2022-10-31_00-49-50

I tried to use the multiline problem matcher, however this doesn't look good in my opinion, plus from what I see the IAR message may be split over multiple lines (more than two), so it would be hard or even impossible to implement such problem matcher.

Code_2022-10-30_23-46-12

In my opinion best idea is to leave it as in my pull request. If someone want to get the whole message, they should use the `--no_wrap_diagnostics` IAR option.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -280,6 +280,20 @@
           "severity": 4,
           "message": 5
         }
+      },
+      {
+        "name": "iar",
+        "source": "iar",
+        "owner": "cpptools",
+        "fileLocation" : "absolute",
+        "pattern": {
+          "regexp": "^\"(.*?)\",(\\d+)\\s+(?:[Ff]atal\\s+)?([Ww]arning|[Ee]rror)\\[(\\w+\\d+)\\]:\\s+(.*)$",
+          "file": 1,
+          "line": 2,
+          "severity": 3,
+          "code": 4,
+          "message": 5
+        }
       }
     ],
     "configuration": [


Extension/package.json
```diff
@@ -280,6 +280,19 @@
...
           "message": 5
         }
+      },
+      {
+        "name": "iar",
+        "source": "iar",
+        "owner": "cpptools",
+        "fileLocation" : "absolute",
+        "pattern": {
+          "regexp": "^\"(.*?)\",(\\d+)\\s+(?:[Ff]atal\\s+)?([Ww]arning|[Ee]rror)(\\[\\w+\\d+]:\\s+.*)$",
```
sean-mcmanus: Seems like `]` should be `\\]`?

sean-mcmanus (changes requested):

Extension/package.json
sean-mcmanus: `(\\[\\w+\\d+]:\\s+.*)` can be `\\[(\\w+\\d+)\\]:\\s+(.*)` if you use

```json
"code": 4,
"message": 5
```

to enable displaying the error code separately from the message.

See the C++/cl.exe./mscompile problemMatcher in vscode as an example: https://github.com/microsoft/vscode/blob/c07ff2eb1d8d024accf1f5deba8b6bce384d4cb6/src/vs/workbench/contrib/tasks/common/problemMatcher.ts#L1209

sean-mcmanus (changes requested):

Extension/package.json
morsisko: Thanks, you are right

morsisko: Done

morsisko:
Requested changes has been implemented

sean-mcmanus (approved):

Issue #177, opened 2016-08-29
"Go to definition" command show wrong file.


sandy081:
_From @magpier84 on August 28, 2016 11:57_
- VSCode Version: 1.4.0
- OS Version: OSX 10.11.2
- C/C++: 0.8.1

Steps to Reproduce:
1. Select header file of some class in workspace file browser.
2. Go to definition of some its methods.
3. Cpp file is opened but its path isn't relative to root of the workspace. In "Open Editors" section it is displayed as path from root of the drive (cut on left side: "t/subfolder/file.cpp") with hint  "/Users/admin/project/subfolder/file.cpp". The hint of tab shows "/Users/admin/project/subfolder/file.cpp"
4. Breakpoints can not be set if file was opened in above described way.
5. Two instances of the same file can be shown in tabs.

_Copied from original issue: Microsoft/vscode#11064_

jgoshi:
This is fixed in the upcoming release of cpptools (version 0.9.0) and should be available in the next week or so. If this still repros for you at that time please reopen the bug.

PR #10101, opened 2022-11-01
Added ARMCC/Keil problem matcher (#10054)

morsisko:
This pull requests adds ARMCC (the compiler used by Keil) problem matcher.

Warning example that is generated by ARM compiler:

`"main.c", line 8: Warning:  #177-D: variable "c"  was declared but never referenced`

Worth to note:

There are two version of ARM Compiler. Arm Compiler 5 (also known as ARMCC) and Arm Compiler 6 (also known as ARMClang). The latter is often confused with the first one, this is why I propose to name this matcher "armcc5".

PS. From what i Know the Arm Compiler 6 uses the same warning/error output format as GCC, so I don't think there is a need for additional matcher.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -280,6 +280,23 @@
           "severity": 4,
           "message": 5
         }
+      },
+      {
+        "name": "armcc5",
+        "source": "armcc5",
+        "owner": "cpptools",
+        "fileLocation": [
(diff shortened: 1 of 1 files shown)


morsisko:

sean-mcmanus (approved):

Issue #664, opened 2017-04-26
Add the ability to generate definitions from declarations and vice versa


benjaminjack:
I spend a lot of time typing out lists of arguments (or copying and pasting them) when I'm defining functions that have already been declared in a header file. It would be nice if cpptools could automatically generate a skeleton definition of a function that has already been declared, or at least provide a shortcut to auto-fill the argument list.

BrianSipple:
I was going to suggest something very similar. I'd also be interested in a feature to create a header- and source-file pair with a single task invocation. This would be similar to XCode, for example, which generates both a `.h` and `.cpp` file when the user selects an option to create a new class file.

sean-mcmanus:
@bobbrow I believe this is the most used feature in VS that we are missing (used more often than Find All References).

As mentioned in https://github.com/Microsoft/vscode-cpptools/issues/3056, we should also support the "opposite" case of adding the declaration to the header from an existing definition.

bobbrow:
Yeah, I have seen this feedback a few times in our survey responses.  It is on my radar.

daranday:
Would love to have this feature!

davidbitton:
Upvote!

exocode:
Is this something you are looking for?: 

It is able to preserve existing methods and inclusions

![2019-06-03 12 55 16](https://user-images.githubusercontent.com/1597621/58797001-165f2680-85ff-11e9-9b97-46e02c5c2547.gif)


https://github.com/reignofwebber/vscode-c-cpp-definition-generator

davidbitton:
@exocode yep! your work?

yannickbt64:
Can't wait to have this implemented. Such a basic feature, yet very useful.

ArcticLampyrid:
Can't wait to use it.
Do we have any plan to implement it?

sean-mcmanus:
@1354092549 Yes, but we don't have it on our schedule yet.

rafaelchen:
it would be really useful feature :)

PoetaKodu:
I would love that!

exocode:
@davidbitton sorry, no.. some other genius did that for all of us  :-)

dbsxdbsx:
Just  upVote it, and it woud be better to have it with funcitonality like "changing signature".

Flamefire:
Very related: Move implementation to source file (possibly creating it) which not only generates the signature but also moves the implementation from header to source.

[...]

PR #10123, opened 2022-11-08
Create declaration or definition feature

michelleangela:
Add the feature to create a declaration or definition for a function.
Issue ref: #664 

**Current known bugs/limitations:**
- file creation (for example creating a corresponding source file for a header file) is not fully working.
- Created declaration or definition is not formatted.

**Create Definition**
Example: functions that are declared but have no definitions will show a hint on hover that no definition is found. Clicking on the declared function will offer a code action to create a definition for it.

Hint on hover
image

Code action
image

Definition created when code action is executed
image

**Create Declaration**
Example: class member function is defined in a source file but is not declared in the class's header file will offer a code action to create a declaration for it when clicked on.

Code action
image

Declaration created when code action is executed
image

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -168,8 +168,6 @@ function publishRefactorDiagnostics(params: PublishRefactorDiagnosticsParams): v
 
     const fileUri: vscode.Uri = vscode.Uri.parse(params.uri);
     diagnosticsCollectionRefactor.set(fileUri, newDiagnostics);
-
-    clients.timeTelemetryCollector.setUpdateRangeTime(fileUri);
 }
 
 interface WorkspaceFolderParams {
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -67,6 +67,7 @@ const languageClientCrashTimes: number[] = [];
 let pendingTask: util.BlockingTask<any> | undefined;
 let compilerDefaults: configs.CompilerDefaults;
 let diagnosticsCollectionIntelliSense: vscode.DiagnosticCollection;
+let diagnosticsCollectionRefactor: vscode.DiagnosticCollection;
 
 let workspaceDisposables: vscode.Disposable[] = [];
 export let workspaceReferences: refs.ReferencesManager;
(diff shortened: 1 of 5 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -247,11 +280,38 @@ interface IntelliSenseDiagnostic {
...
+export interface CreateDeclarationOrDefinitionParams {
+    uri: string;
+    range: Range;
+}
+
+export interface CreateDeclarationOrDefinitionResult {
+    changes: { [key: string]: any[] };
+}
+
+// interface TextEdit {
```
sean-mcmanus: Should this commented out code be removed?

Extension/src/LanguageServer/client.ts
```diff
@@ -2903,6 +2967,60 @@ export class DefaultClient implements Client {
         this.handleRemoveCodeAnalysisProblems(false, identifiersAndUris);
     }
 
+    public async handleCreateDeclarationOrDefinition(): Promise<void> {
+        let range: vscode.Range | undefined;
+        let uri: vscode.Uri| undefined;
```
sean-mcmanus: space before `|`. There might be a linter check that can report on this?

Extension/src/LanguageServer/client.ts
```diff
@@ -2903,6 +2967,60 @@ export class DefaultClient implements Client {
...
+        // range is based on the cursor position.
+        const editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+        if (editor) {
+            uri = editor.document.uri;
+            if (editor.selection.isEmpty) {
+                range = new vscode.Range(editor.selection.active, editor.selection.active);
```
sean-mcmanus, line 2975: I'm confused by this. If the selection is empty is there supposed to be an active one? You may want to consider using editor.selection.start, editor.selection.end (or something like that instead of active).

Extension/src/LanguageServer/client.ts
```diff
@@ -2903,6 +2967,60 @@ export class DefaultClient implements Client {
...
+                range = new vscode.Range(editor.selection.anchor, editor.selection.active);
+            }
+        }
+
+        if (uri && range) {
+            const params: CreateDeclarationOrDefinitionParams = {
+                uri: uri.toString(),
+                range: {
+                    start: {
+                        character: range.start.character,
```
sean-mcmanus, line 2988: Can you run the formatter on this added code. The `}` isn't formatted correctly.

Extension/src/nativeStrings.json
```diff
@@ -264,5 +264,133 @@
...
+        "text": "Events",
+        "hint": "'Events' are a C++ event."
+    },
+    "cm_functions": {
+        "text": "Functions",
+        "hint": "'Functions' are a C++ function."
+    },
+    "cm_importdirectives": {
+        "text": "Import directives",
+        "hint": "'Import directives' are a C++ ."
```
sean-mcmanus: These hints don't seem correct. Can they be reworded?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
michelleangela: If `isEmpty` is true, then `start` and `end` position are equal or it means there is an active one but there's no range, just one position. So either `start` and `end` position works or `active` works for this case.

Extension/src/nativeStrings.json
michelleangela: Those hints were place holders. I don't think these are used on the UI, but the internal API requests for them. Pending on what to do with these.

bobbrow (dismissed):

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -144,6 +145,33 @@ function publishIntelliSenseDiagnostics(params: PublishIntelliSenseDiagnosticsPa
...
+            }
+        }
+
+        newDiagnostics.push(diagnostic);
+    });
+
+    const fileUri: vscode.Uri = vscode.Uri.parse(params.uri);
+    diagnosticsCollectionRefactor.set(fileUri, newDiagnostics);
+
+    clients.timeTelemetryCollector.setUpdateRangeTime(fileUri);
```
Colengms: I believe this was used for telemetry in `publishIntelliSenseDiagnotics`, to determine when the first set of squiggles is received, and probably shouldn't be duplicated here.

Colengms (changes requested):

Colengms (approved):

PR #10126, opened 2022-11-08
Move status bar icons into the Language Status

### LINKED ISSUE #8405: Contribute to the language status
Hi VS Code PM here 👋 

Currently the C++ extension contributes individual items to the status bar. On my macOS I see the `Mac` item in the status bar. I also see some progress items like the occasional flame icon (pictures).
I suggest to use the [language status bar](https://code.visualstudio.com/updates/v1_61#_new-javascript-and-typescript-language-status-item), Typescript is already using this (picture attached).

This way all the C++ status bar items will come under a common item making the whole status bar experience cleaner for our users.

The status bar items which are actually progress our current advise is to use the `withProgress`-API. With that you announce a long running thing to VS Code. Our current thinking is to add a new progress location which is the extension’s language status. With that we can render something on the composite item as well as on the item itself. This is being discussed [here](https://github.com/microsoft/vscode/issues/129037#issuecomment-957360329), please chime in.

fyi @bobbrow @jureid @jrieken @egamma

Screenshot 2021-11-15 at 16 46 07

Screenshot 2021-11-16 at 11 34 26

Screenshot 2021-11-16 at 11 36 10

Bernardin-MS:
Fixes https://github.com/microsoft/vscode-cpptools/issues/8405


Extension/src/LanguageServer/ui.ts
```diff
@@ -89,31 +86,96 @@ export class UI {
...
+        }
+        this.configStatusBarItem = vscode.languages.createLanguageStatusItem("c.cpp.configuration.tooltip", this.documentSelector);
+        this.configStatusBarItem.name = localize("c.cpp.configuration.tooltip", "C/C++ Configuration");
+        // TODO: Confirm title and tooltip localization
+        this.configStatusBarItem.command = {
+            command: "C_Cpp.ConfigurationSelect",
+            title: this.configStatusBarItem.name as string,
+            tooltip: this.configStatusBarItem.name as string
+        };
+
```
sean-mcmanus: Blank line

Extension/src/LanguageServer/ui.ts
```diff
@@ -89,31 +86,96 @@ export class UI {
...
-        this.browseEngineStatusBarItem.tooltip = (this.isParsingFiles ? `${this.parsingFilesTooltip} | ` : "") + label;
+
+        this.ensureBrowseEnginerStatus();
+        if (this.browseEngineStatusBarItem) {
+            this.browseEngineStatusBarItem.detail = (this.isParsingFiles ? `${this.parsingFilesTooltip} | ` : "") + label;
+        }
     }
 
     private setIsParsingWorkspace(val: boolean): void {
+
```
sean-mcmanus: Can you remove the blank lines at the beginning of function definitions?

Extension/src/LanguageServer/ui.ts
```diff
@@ -233,27 +334,34 @@ export class UI {
...
         }
     }
 
     private set ShowConfiguration(show: boolean) {
-        if (show) {
-            this.configStatusBarItem.show();
-        } else {
-            this.configStatusBarItem.hide();
-        }
+        // Not needed. Goes away when not needed due to documentFilters
```
sean-mcmanus: Can ShowConfiguration be deleted?

Extension/src/LanguageServer/ui.ts
```diff
@@ -432,9 +540,9 @@ export class UI {
 
     public dispose(): void {
         this.configStatusBarItem.dispose();
-        this.browseEngineStatusBarItem.dispose();
-        this.intelliSenseStatusBarItem.dispose();
-        this.referencesStatusBarItem.dispose();
+        if (this.browseEngineStatusBarItem) {this.browseEngineStatusBarItem.dispose(); }
```
sean-mcmanus: Can you use `this.browseEngineStatusBarItem?.dispose();` instead?

Extension/src/LanguageServer/ui.ts
```diff
@@ -196,20 +275,35 @@ export class UI {
...
+                this.referencesStatusBarItem.text = "";
+                if (this.referencesStatusBarItem.command) {
+                    this.referencesStatusBarItem.command.title = "";
+                    this.referencesStatusBarItem.command.tooltip = "";
+                }
+                this.ShowReferencesIcon = false;
+            } else {
+                this.referencesStatusBarItem.text = "$(search)";
+                if (this.referencesStatusBarItem.command) {
+                    this.referencesStatusBarItem.command.title = "Results";
```
sean-mcmanus: Should this be localized?

Extension/src/LanguageServer/ui.ts
```diff
@@ -36,10 +36,18 @@ interface ConfigurationStatus {
...
-    private browseEngineStatusBarItem: vscode.StatusBarItem;
-    private intelliSenseStatusBarItem: vscode.StatusBarItem;
-    private referencesStatusBarItem: vscode.StatusBarItem;
+    private configStatusBarItem!: vscode.LanguageStatusItem;
+    private browseEngineStatusBarItem!: vscode.LanguageStatusItem | undefined;
+    private intelliSenseStatusBarItem!: vscode.LanguageStatusItem | undefined;
+    private referencesStatusBarItem!: vscode.LanguageStatusItem | undefined;
+    // This is a duplicate of what's in client.ts
+    // TODO: Confirm whether the orignal can be reused here
+    private documentSelector: vscode.DocumentFilter[] = [
```
sean-mcmanus: Yeah, I see this used in 2 places in client.ts -- I'm thinking it could go in some common/util file as an exported variable. We want to try to reduce the size of client.ts and it doesn't seem directly related to the client so it probably shouldn't go there.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -36,10 +36,18 @@ interface ConfigurationStatus {
 }
 
 export class UI {
-    private configStatusBarItem: vscode.StatusBarItem;
-    private browseEngineStatusBarItem: vscode.StatusBarItem;
-    private intelliSenseStatusBarItem: vscode.StatusBarItem;
-    private referencesStatusBarItem: vscode.StatusBarItem;
+    private configStatusBarItem!: vscode.LanguageStatusItem;
```
sean-mcmanus: What does `!:` mean?

Extension/src/LanguageServer/ui.ts
```diff
@@ -36,10 +36,18 @@ interface ConfigurationStatus {
...
+    private browseEngineStatusBarItem!: vscode.LanguageStatusItem | undefined;
```
sean-mcmanus: It seems like the `| undefined` could be replaced by `browseEngineStatusBarItem?:` but I'm not sure what the `!:` is as mentioned earlier.

Extension/src/LanguageServer/ui.ts
```diff
@@ -125,49 +187,66 @@ export class UI {
...
-        this.intelliSenseStatusBarItem.tooltip = (this.isUpdatingIntelliSense ? this.updatingIntelliSenseTooltip : "")
-            + (twoStatus ? " | " : "")
-            + (val ? this.runningCodeAnalysisTooltip : "");
-        this.intelliSenseStatusBarItem.command = val ? "C_Cpp.ShowCodeAnalysisCommands" : undefined;
+        if (this.intelliSenseStatusBarItem) {
+            this.intelliSenseStatusBarItem.text = showIcon ? "$(flame)" : "";
+            this.intelliSenseStatusBarItem.detail = (this.isUpdatingIntelliSense ? this.updatingIntelliSenseTooltip : "")
+                + (twoStatus ? " | " : "")
+                + (val ? this.runningCodeAnalysisTooltip : "");
+            this.intelliSenseStatusBarItem.command = val ? {command: "C_Cpp.ShowCodeAnalysisCommands", title: "Show Code Analysis"} : undefined;
```
sean-mcmanus: Should this be localized?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
Bernardin-MS: It basically means that variable can get away with not being initialized in the constructor

sean-mcmanus: @Bernardin-MS I still don't understand. Do you happen to know where that's documented online or what to search for?

Extension/src/LanguageServer/ui.ts
Bernardin-MS: I think so. Seems like it's only used for disappearing the status icon when the current viewed file type changes. I'm waiting for a second opinion

Extension/src/LanguageServer/ui.ts
Bernardin-MS: [Non-null assertion operator](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator)

Extension/src/LanguageServer/ui.ts
Bernardin-MS: Yeah you're right. It was from a previous idea that didn't work out

Extension/src/LanguageServer/ui.ts
Bernardin-MS: That gives a syntax error since `?` is meant for expressions apparently

bobbrow: Can you do `void this.browseEngineStatusBarItem?.dispose();`?  The 'void' might convert it to an expression that the linter won't complain about.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: The non-null assertion operator is only used in expressions, but I found the similar feature at https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions .

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Oh...I'm okay with either approach....but if you're sticking with the `if (){` there should be space after the `{` -- I think running the formatter would fix it.

Extension/src/LanguageServer/ui.ts
```diff
@@ -89,31 +86,94 @@ export class UI {
...
-        if (val) {
-            this.browseEngineStatusBarItem.command = "C_Cpp.ShowParsingCommands";
-        } else {
-            this.browseEngineStatusBarItem.command = undefined;
+        this.ensureBrowseEnginerStatus();
+        if (this.browseEngineStatusBarItem) {
+            if (val) {
+                this.browseEngineStatusBarItem.command = {
+                    command: "C_Cpp.ShowParsingCommands",
+                    title: "Show Parsing Commands",
```
michelleangela: Do the title and tooltip for "C_Cpp.ShowParsingCommands" need to be localized?

Extension/src/LanguageServer/ui.ts
```diff
@@ -89,31 +86,94 @@ export class UI {
...
     }
 
     private setIsParsingWorkspacePausable(val: boolean): void {
-        if (val) {
-            this.browseEngineStatusBarItem.command = "C_Cpp.ShowParsingCommands";
-        } else {
-            this.browseEngineStatusBarItem.command = undefined;
+        this.ensureBrowseEnginerStatus();
+        if (this.browseEngineStatusBarItem) {
+            if (val) {
```
michelleangela: nit: small refactoring opportunity -- rename all "val" variables to something descriptive.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Previously, those command titles were not user visible, and we don't want them to be, so I'm guessing not.

Extension/src/LanguageServer/ui.ts
bobbrow: In short set/get functions, I think using something that means "value" is fine.  In C# you literally must use `value` in the `set` function since you don't get to name the variable.  The only alternative I see here would be `newValue` which isn't a significant improvement.

Extension/src/LanguageServer/ui.ts
Bernardin-MS: I think they do now that they are user facing

Extension/src/LanguageServer/ui.ts
```diff
@@ -36,122 +37,254 @@ interface ConfigurationStatus {
...
-        this.codeAnalysisProgram = "clang-tidy";
-        this.runningCodeAnalysisTooltip = localize(
-            { key: "running.analysis.tooltip", comment: [this.codeAnalysisTranslationHint] }, "Running {0}", this.codeAnalysisProgram);
-        this.codeAnalysisPausedTooltip = localize(
-            { key: "code.analysis.paused.tooltip", comment: [this.codeAnalysisTranslationHint] }, "{0} paused", this.codeAnalysisProgram);
     }
 
     private set ActiveConfig(label: string) {
-        this.configStatusBarItem.text = label;
+        this.configStatusBarItem.text = label ?? localize("configuration.notselected.text", "Configuation: Not selected");
```
jogo-: Typo: Config**ua**tion -> Configu**r**ation

mudanzasIPROM (approved):

Extension/src/LanguageServer/ui_old.ts
```diff
@@ -0,0 +1,442 @@
...
+        }
+        this.isRunningCodeAnalysis = val;
+        const showIcon: boolean = val || this.isUpdatingIntelliSense;
+        const twoStatus: boolean = val && this.isUpdatingIntelliSense;
+        this.ShowFlameIcon = showIcon;
+        this.intelliSenseStatusBarItem.text = showIcon ? "$(flame)" : "";
+        this.intelliSenseStatusBarItem.tooltip = (this.isUpdatingIntelliSense ? this.updatingIntelliSenseTooltip : "")
+            + (twoStatus ? " | " : "")
+            + (val ? this.runningCodeAnalysisTooltip : "");
+        this.intelliSenseStatusBarItem.command = val ? "C_Cpp.ShowCodeAnalysisCommands" : undefined;
```
sean-mcmanus: Is this supposed to be changed? It looks like this command got replaced by the idle/active ones. However, as mentioned elsewhere, I'm concerned about having the 2 different lists of commands, because the "idle" commands are still valid to run when non-idle, i.e. run code analysis on active file, switch file, run code analysis on active file (the new one, in parallel).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,551 @@
...
+            this.codeAnalysisProcessed = 0;
+        }
+        this.isRunningCodeAnalysis = val;
+        this.codeAnalysisStatusBarItem.busy = val;
+        this.codeAnalysisStatusBarItem.text = val ? "Code Analysis: Running" : `Code Analysis State: ${this.codeAnalysisCurrentState()}`;
+        this.codeAnalysisStatusBarItem.command = val ? {
+            command: "C_Cpp.ShowActiveCodeAnalysisCommands",
+            title: localize("c.cpp.codeanalysis.statusbar.showCodeAnalysisOptions", "Options")
+        } : {
+            command: "C_Cpp.ShowIdleCodeAnalysisCommands",
```
sean-mcmanus: As mentioned elsewhere, I'm concerned that hiding the "idle" code analysis commands when it's active is potentially confusing because the commands are still valid/usable while active (i.e. it'll just queue more work, potentially in parallel).

mudanzasIPROM (approved):

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: I've addressed it here the way I proposed
https://github.com/microsoft/vscode-cpptools/blob/42e26923401ff5ee5e73d678f709ae395613390d/Extension/src/LanguageServer/ui_new.ts#L483
https://github.com/microsoft/vscode-cpptools/blob/42e26923401ff5ee5e73d678f709ae395613390d/Extension/src/LanguageServer/client.ts#L2766

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,551 @@
...
+            this.referencesStatusBarItem.show();
+        } else {
+            this.referencesStatusBarItem.hide();
+        }
+    }
+
+    public activeDocumentChanged(): void {
+        const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+        if (activeEditor) {
+            // const isCpp: boolean = (activeEditor.document.uri.scheme === "file" && (activeEditor.document.languageId === "c" || activeEditor.document.languageId === "cpp" || activeEditor.document.languageId === "cuda-cpp"));
```
sean-mcmanus: This should be deleted.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,551 @@
...
+                    this.intelliSenseStatusBarItem.severity = vscode.LanguageStatusSeverity.Warning;
+                }
+            }, this.iconDelayTime);
+        }
+        this.intelliSenseStatusBarItem.command = {
+            command: "C_Cpp.RestartIntelliSenseForFile",
+            title: localize("rescan.intellisense.text", "Rescan"),
+            tooltip: localize("rescan.intellisense.tooltip", "Rescan IntelliSense")
+        };
+
```
sean-mcmanus: Extra newline

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,551 @@
...
+import { getCustomConfigProviders, CustomConfigurationProviderCollection, isSameProviderExtensionId } from './customProviders';
+import * as nls from 'vscode-nls';
+import { setTimeout } from 'timers';
+import { CppSettings } from './settings';
+import { UI } from './ui';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+// let ui: UI;
```
sean-mcmanus: Seems like this can be deleted.

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
I'm seeing a blank config during startup. Can you hide that until a configuration is visible?

image

sean-mcmanus (changes requested):
The Configuration status command says "Edit Configuration", but when hovered it shows "Select Configuration" and when clicked it shows options to do both. Should the UI just show something else like "Configuration Options"?

image

sean-mcmanus (changes requested):
I'm confused by the "(i) and check" icons on the far left. What are those supposed to mean? I don't see them change. What are the possible icons?

image

In particular, it seems inconsistent for the "Parsing workspace" status to show a "check" icon always when the IntelliSense status changes to an (i), and I don't understand why the Code Analysis status has an (i) when it's not running.

sean-mcmanus (changes requested):
It shows "IntelliSense: Ready" when intelliSenseEngine is set to "Tag Parser", which is misleading/confusing -- can you change that to say something like "IntelliSense: Tag Parser"?

sean-mcmanus (changes requested):
The command "Pause Workspace" seems unusual since you can't really pause a workspace -- can you change it to just "Pause" with a hover that shows "Pause Workspace Parsing"? Another possibility would be to show the full "Pause Workspace Parsing" in the command, but that might cause the UI too be bigger than we'd like.

image

sean-mcmanus (changes requested):
It doesn't show the number of files Discovered/Indexed and it doesn't show the number of files to parse and parsed % anymore. That is important in order to see the progress.

sean-mcmanus (changes requested):
I'm seeing all the rows of the status UI disappearing and reappearing (i.e. flickering) when there appears to be no visible change. I think the flickering would be fine if there was some change going on, but when there's no change is seems like a bug. Not sure if there's a way we can get that to stop or if it's a VS Code issue.

Bernardin-MS:
> I'm seeing a blank config during startup. Can you hide that until a configuration is visible?
> 
> image

I don't this so. But I can initialize it to a placeholder like "Loading configuration..."

Bernardin-MS:
> I'm confused by the "(i) and check" icons on the far left. What are those supposed to mean? I don't see them change. What are the possible icons?
> 
> image
> 
> In particular, it seems inconsistent for the "Parsing workspace" status to show a "check" icon always when the IntelliSense status changes to an (i), and I don't understand why the Code Analysis status has an (i) when it's not running.

It's an optional feature I was testing out to see how people felt about it. Those symbol in front appear when I set a "severity" for a status item

Bernardin-MS:
> It shows "IntelliSense: Ready" when intelliSenseEngine is set to "Tag Parser", which is misleading/confusing -- can you change that to say something like "IntelliSense: Tag Parser"?

I'm not sure how you mean. Is there a way to repro this on my end?

Bernardin-MS:
> I'm seeing all the rows of the status UI disappearing and reappearing (i.e. flickering) when there appears to be no visible change. I think the flickering would be fine if there was some change going on, but when there's no change is seems like a bug. Not sure if there's a way we can get that to stop or if it's a VS Code issue.

Yeah that happens because of how I'm updating the commands hover text. I'm still planning to look into a way to update it without having to completely replace the Command object each time

sean-mcmanus:
> > It shows "IntelliSense: Ready" when intelliSenseEngine is set to "Tag Parser", which is misleading/confusing -- can you change that to say something like "IntelliSense: Tag Parser"?
> 
> I'm not sure how you mean. Is there a way to repro this on my end?

Set C_Cpp.intelliSenseEngine to "Tag Parser" -- also in that scenario it would be a nice addition (not ship blocking) to add a command like "Enable IntelliSense" to encourage users to switch from Tag Parser mode to the real IntelliSense mode.

sean-mcmanus:
> It's an optional feature I was testing out to see how people felt about it. Those symbol in front appear when I set a "severity" for a status item

I think it might make sense if there are cases where a warning or error severity would be appropriate. The (i) information (?) severity might make sense for commands in progress that aren't finished and we reserved the "check" mark for status that is "done".

Bernardin-MS:
At the moment it doesn't seem like I'm able to update the hover text without the flyout flickering each time. We're currently using the hover text to show progress to the user for the tag parsing operation, so I don't think taking it out is an option either

Extension/src/LanguageServer/extension.ts
```diff
@@ -184,7 +186,7 @@ export async function activate(): Promise<void> {
 
     disposables.push(vscode.workspace.onDidChangeConfiguration(onDidChangeSettings));
     disposables.push(vscode.window.onDidChangeActiveTextEditor(onDidChangeActiveTextEditor));
-    ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
+    if (ui instanceof OldUI) { void ui?.activeDocumentChanged(); } // Handle already active documents (for non-cpp files that we don't register didOpen).
```
bobbrow: Please format these `if (ui instanceof OldUI)` lines. We don't normally put full blocks on a single line in TypeScript.

sean-mcmanus (changes requested):
I'm only seeing the old UI for some reason with the latest changes.

Extension/src/LanguageServer/ui.ts
```diff
@@ -438,9 +460,13 @@ export class UI {
...
 }
 
-export function getUI(): UI {
+export async function getUI(): Promise<UI> {
     if (!ui) {
-        ui = new UI();
+        const experimentationService: IExperimentationService | undefined = await telemetry.getExperimentationService();
+        if (experimentationService !== undefined) {
+            const useNewUI: boolean | undefined = experimentationService.getTreatmentVariable<boolean>("vscode", "splitUIUsers");
+            ui = useNewUI ? new NewUI() : new OldUI();
```
sean-mcmanus: Could we also check the C_Cpp.experimentalFeatures setting to enable users to opt-in to using this? Otherwise, it makes testing harder and users can't opt-in.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,528 @@
...
```
michelleangela: Remove commented out code?

Extension/src/LanguageServer/ui.ts
```diff
@@ -6,16 +6,35 @@
 
 import * as vscode from 'vscode';
 import { Client } from './client';
+import * as nls from 'vscode-nls';
+import { NewUI } from './ui_new';
+// import { OldUI } from './ui_old';
```
michelleangela: Remove commented out code?

Extension/src/LanguageServer/ui.ts
```diff
@@ -6,16 +6,35 @@
...
+import * as telemetry from '../telemetry';
+import { IExperimentationService } from 'tas-client';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 let ui: UI;
 
+export interface UI {
+    whichUI: boolean;
```
michelleangela: This name suggests it is expected to return some enum value or specific value instead of a boolean. Perhaps rename to something such as "isUsingNewUI" or "isNewUI".

Extension/src/LanguageServer/client.ts
```diff
@@ -2458,6 +2461,7 @@ export class DefaultClient implements Client {
             const compilerPathAndArgs: util.CompilerPathAndArgs =
                 util.extractCompilerPathAndArgs(!!settings.legacyCompilerArgsBehavior, c.compilerPath, c.compilerArgs);
             modifiedConfig.compilerPath = compilerPathAndArgs.compilerPath;
+            this.model.currentCompiler.setValueIfActive(modifiedConfig.compilerPath ?? "");
```
sean-mcmanus: This shouldn't be set for each configuration, right?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,528 @@
...
+        client.CodeAnalysisProcessedChanged(value => { this.setCodeAnalysisProcessed(value); });
+        client.CodeAnalysisTotalChanged(value => { this.setCodeAnalysisTotal(value); });
+        client.ReferencesCommandModeChanged(value => { this.ReferencesCommand = value; });
+        client.TagParserStatusChanged(value => { this.TagParseStatus = value; });
+        client.ActiveConfigChanged(value => { this.ActiveConfig = value; });
+    }
+
+    public async showConfigurations(configurationNames: string[]): Promise<number> {
+        const options: vscode.QuickPickOptions = {};
+        options.placeHolder = localize("select.a.configuration", "Select a Configuration...");
```
sean-mcmanus, line 395: Why is the hover "Select a Configuration" when the link shows "Edit Configuration"? I mentioned this earlier, but I'm not sure what the answer was.

sean-mcmanus (changes requested):
There's some state near the end (before Parsing Complete) when the Parsing label is blank. Can that be fixed? I'm not sure what it should say since I'm not sure what state it's in at that point. It's possible we should show some new state info that reflects what is going on...or maybe just show Parsing Complete.

image

sean-mcmanus (changes requested):
It doesn't show the % done for workspace parsing. UPDATE: I found that in the hover. I guess that's "by design", right?

Was there a reason to not show the % info in the main info? Does it cause an update problem?

sean-mcmanus (changes requested):
I'm seeing it show "Parsing Workspace" but with no Pause Workspace Parsing option available and hover shows it's "Discovering files" and not parsing.

image

I think it should just show "Discovering files: " or something equivalent, maybe the "Indexing Workspace", which is shown earlier on but then disappears too early.

Extension/src/LanguageServer/extension.ts
```diff
@@ -433,6 +436,73 @@ export function registerCommands(enabled: boolean): void {
...
+        () => {
+            logForUIExperiment("ParsingCommands");
+            onShowParsingCommands();
+        }
+        : onDisabledCommand));
+    // ----------------------------------------
+}
+
+async function logForUIExperiment(command: string): Promise<void> {
+    const whichUI: string = (await getUI()).whichUI ? "New UI" : "Old UI";
```
bobbrow: Just a convention, we don't usually put spaces in the event names.  And something like this would be better logged as a boolean (converted to string) in the payload instead of concatenating it into the event name.

Consider this:
```ts
  const newUI = (await getUI()).whichUI.toString();
  telemetry.logLanguageServerEvent(`experiment${command}`, { newUI });
```

Extension/src/LanguageServer/client.ts
Bernardin-MS: That was from a previous implementation. I'll remove it

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: I don't remember too. But it was based off the spec so I believe it had to do with some agreement from that

Bernardin-MS:
> There's some state near the end (before Parsing Complete) when the Parsing label is blank. Can that be fixed? I'm not sure what it should say since I'm not sure what state it's in at that point. It's possible we should show some new state info that reflects what is going on...or maybe just show Parsing Complete.
> 
> image

This was a small bug. I was clearing text in the wrong place

Bernardin-MS:
> It doesn't show the % done for workspace parsing. UPDATE: I found that in the hover. I guess that's "by design", right?
> 
> Was there a reason to not show the % info in the main info? Does it cause an update problem?

Yeah it's in the hover text by design. Putting it in the main area would have made the text too long. We're already pushing it a bit with show the 2 different types of parsing at the same time

Bernardin-MS:
> I'm seeing it show "Parsing Workspace" but with no Pause Workspace Parsing option available and hover shows it's "Discovering files" and not parsing.
> 
> image
> 
> I think it should just show "Discovering files: " or something equivalent, maybe the "Indexing Workspace", which is shown earlier on but then disappears too early.

This happens when the parsing gets done really fast

Extension/src/LanguageServer/extension.ts
```diff
@@ -301,10 +305,12 @@ export function onDidChangeActiveTextEditor(editor?: vscode.TextEditor): void {
         activeDocument = "";
     } else {
         activeDocument = editor.document.uri.toString();
-        clients.activeDocumentChanged(editor.document);
+        if (ui instanceof OldUI) { clients.activeDocumentChanged(editor.document); }
```
bobbrow: can you reformat this one too?

Extension/src/LanguageServer/extension.ts
```diff
@@ -433,6 +442,73 @@ export function registerCommands(enabled: boolean): void {
...
+            logForUIExperiment("ParsingCommands");
+            onShowParsingCommands();
+        }
+        : onDisabledCommand));
+    // ----------------------------------------
+}
+
+async function logForUIExperiment(command: string): Promise<void> {
+    const isNewUI: string = (await getUI()).isNewUI.toString();
+    telemetry.logLanguageServerEvent(`experiment${command}`, { ui: isNewUI });
```
bobbrow: Can you call the property `isNewUI` instead?  In telemetry it will show up as ui = true|false. It works, but it's missing some context and we may forget what true|false means.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -36,6 +36,13 @@ interface ConfigurationStatus {
     priority: ConfigurationPriority;
 }
 
+enum LanguageStatusPriroty {
```
bobbrow: misspelled "Priority"

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -80,27 +87,27 @@ export class NewUI implements UI {
     get isNewUI(): boolean { return true; };
 
     constructor() {
-        this.configStatusBarItem = vscode.languages.createLanguageStatusItem("c.cpp.configuration.tooltip", this.configDocumentSelector);
-        this.configStatusBarItem.name = localize("c.cpp.configuration.tooltip", "Select Configuration");
+        this.configStatusBarItem = vscode.languages.createLanguageStatusItem(`cpptools.status.${LanguageStatusPriroty.First}.configuration`, this.configDocumentSelector);
```
jogo-: Typo: LanguageStatusPri**ro**ty -> LanguageStatusPri**ori**ty

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -80,27 +87,27 @@ export class NewUI implements UI {
...
-        this.referencesStatusBarItem = vscode.window.createStatusBarItem("c.cpp.references.statusbar", vscode.StatusBarAlignment.Right, 901);
+        this.referencesStatusBarItem = vscode.window.createStatusBarItem(`c.cpp.references.statusbar`, vscode.StatusBarAlignment.Right, 901);
         this.referencesStatusBarItem.name = localize("c.cpp.references.statusbar", "C/C++ References Status");
         this.referencesStatusBarItem.tooltip = "";
         this.referencesStatusBarItem.command = "C_Cpp.ShowReferencesProgressUI_Telemetry";
         this.ShowReferencesIcon = false;
 
-        this.intelliSenseStatusBarItem = vscode.languages.createLanguageStatusItem("c.cpp.intellisense.statusbar", documentSelector);
-        this.intelliSenseStatusBarItem.name = localize("c.cpp.intellisense.statusbar", "C/C++ IntelliSense Status");
+        this.intelliSenseStatusBarItem = vscode.languages.createLanguageStatusItem(`cpptools.status.${LanguageStatusPriroty.Mid}.intellisense`, documentSelector);
```
jogo-: Typo: LanguageStatusPri**ro**ty -> LanguageStatusPri**ori**ty

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -80,27 +87,27 @@ export class NewUI implements UI {
...
+        this.intelliSenseStatusBarItem.name = localize("cpptools.status.intellisense", "C/C++ IntelliSense Status");
         this.intelliSenseStatusBarItem.text = this.idleIntelliSenseText;
 
-        this.browseEngineStatusBarItem = vscode.languages.createLanguageStatusItem("c.cpp.tagparser.statusbar", documentSelector);
-        this.browseEngineStatusBarItem.name = localize("c.cpp.tagparser.statusbar", "C/C++ Tag Parser Status");
+        this.browseEngineStatusBarItem = vscode.languages.createLanguageStatusItem(`cpptools.status.${LanguageStatusPriroty.Mid}.tagparser`, documentSelector);
```
jogo-: Typo: LanguageStatusPri**ro**ty -> LanguageStatusPri**ori**ty

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -80,27 +87,27 @@ export class NewUI implements UI {
...
+        this.browseEngineStatusBarItem.name = localize("cpptools.status..tagparser", "C/C++ Tag Parser Status");
```
jogo-: Typo: double dots ".." ?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -109,9 +116,9 @@ export class NewUI implements UI {
         };
         this.workspaceParsingStatus = this.workspaceParsingRunningText;
 
-        this.codeAnalysisStatusBarItem = vscode.languages.createLanguageStatusItem("c.cpp.codeanalysis.statusbar", documentSelector);
-        this.codeAnalysisStatusBarItem.name = localize("c.cpp.codeanalysis.statusbar", "C/C++ Code Analysis Status");
-        this.codeAnalysisStatusBarItem.text = `Code Analysis State: ${this.codeAnalysisCurrentState()}`;
+        this.codeAnalysisStatusBarItem = vscode.languages.createLanguageStatusItem(`cpptools.status.${LanguageStatusPriroty.Low}.codeanalysis`, documentSelector);
```
jogo-: Typo: LanguageStatusPri**ro**ty -> LanguageStatusPri**ori**ty

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+    private isRunningCodeAnalysis: boolean = false;
+    private isCodeAnalysisPaused: boolean = false;
+    private codeAnalysisProcessed: number = 0;
+    private codeAnalysisTotal: number = 0;
+    private readonly workspaceParsingRunningText: string = localize("running.tagparser.text", "Parsing Workspace");
+    private readonly workspaceParsingPausedText: string = localize("paused.tagparser.text", "Parsing Workspace: Paused");
+    private readonly workspaceParseingDoneText: string = localize("complete.tagparser.text", "Parsing Complete");
+    private workspaceParsingStatus: string = "";
+    private workspaceParsingProgress: string = "";
+    private readonly workspaceRescanText = localize("rescan.tagparse.text", "Rescan workspace");
```
michelleangela: Capitalize "workspace" in "Rescan workspace"?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+
+    private setIsRunningCodeAnalysis(val: boolean): void {
+        if (this.isRunningCodeAnalysis && !val) {
+            this.codeAnalysisTotal = 0;
+            this.codeAnalysisProcessed = 0;
+        }
+        this.isRunningCodeAnalysis = val;
+        this.codeAnalysisStatusBarItem.busy = val;
+        const activeText: string = this.isCodeAnalysisPaused ? this.codeAnalysisPausedText : this.codeAnalysisRunningText;
+        this.codeAnalysisStatusBarItem.text = val ? activeText : `Code Analysis Mode: ${this.codeAnalysisCurrentState()}`;
```
michelleangela: Is this supposed to be localized?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+        this.intelliSenseStatusBarItem.command = {
+            command: "C_Cpp.RestartIntelliSenseForFileUI_Telemetry",
+            title: localize("rescan.intellisense.text", "Rescan"),
+            tooltip: localize("rescan.intellisense.tooltip", "Rescan IntelliSense")
+        };
+    }
+
+    private codeAnalysisCurrentState(): string {
+        const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
+        const state: string = settings.codeAnalysisRunAutomatically && settings.clangTidyEnabled ? "Automatic" : "Manual";
```
michelleangela: `"Automatic" : "Manual";` are these supposed to be localized?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+            this.flameTimeout = setTimeout(() => {
+                if (this.intelliSenseStatusBarItem) {
+                    this.intelliSenseStatusBarItem.text = this.idleIntelliSenseText;
+                    this.intelliSenseStatusBarItem.detail = "";
+                }
+            }, this.iconDelayTime);
+        }
+        this.intelliSenseStatusBarItem.command = {
+            command: "C_Cpp.RestartIntelliSenseForFileUI_Telemetry",
+            title: localize("rescan.intellisense.text", "Rescan"),
```
michelleangela, line 274: "Rescan" should be localized.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+            tooltip: localize("rescan.intellisense.tooltip", "Rescan IntelliSense")
```
michelleangela, line 275: "Rescan IntelliSense" should be localized.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,530 @@
...
+        }
+        this.isRunningCodeAnalysis = val;
+        this.codeAnalysisStatusBarItem.busy = val;
+        const activeText: string = this.isCodeAnalysisPaused ? this.codeAnalysisPausedText : this.codeAnalysisRunningText;
+        this.codeAnalysisStatusBarItem.text = val ? activeText : `Code Analysis Mode: ${this.codeAnalysisCurrentState()}`;
+        this.codeAnalysisStatusBarItem.command = val ? {
+            command: "C_Cpp.ShowActiveCodeAnalysisCommandsUI_Telemetry",
+            title: localize("c.cpp.codeanalysis.statusbar.showCodeAnalysisOptions", "Options"),
+            // Make sure not to overwrite current progress
+            tooltip: this.codeAnalysisStatusBarItem.command?.tooltip ?? "Starting..."
```
michelleangela: "Starting..."

michelleangela (changes requested):

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: Which line? I don't see it

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: Which line? I don't see it

michelleangela (dismissed):

Extension/src/LanguageServer/ui_new.ts
michelleangela: I don't see it anymore either. Maybe I misread it initially.

Extension/src/LanguageServer/ui_new.ts
michelleangela: I don't see it anymore either. Maybe I misread it initially.

Extension/src/LanguageServer/extension.ts
```diff
@@ -184,7 +186,9 @@ export async function activate(): Promise<void> {
 
     disposables.push(vscode.workspace.onDidChangeConfiguration(onDidChangeSettings));
     disposables.push(vscode.window.onDidChangeActiveTextEditor(onDidChangeActiveTextEditor));
-    ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
+    if (ui instanceof OldUI) {
+        void ui?.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
```
sean-mcmanus: Why is the activeDocumentChanged code not run for the newUI? The intent of that code was to enable the configuration switching when c_cpp_properties.json was made active.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,535 @@
...
+    private configStatusBarItem: vscode.LanguageStatusItem;
+    private browseEngineStatusBarItem: vscode.LanguageStatusItem;
+    private intelliSenseStatusBarItem: vscode.LanguageStatusItem;
+    private referencesStatusBarItem: vscode.StatusBarItem;
+    private codeAnalysisStatusBarItem: vscode.LanguageStatusItem;
+    private configDocumentSelector: vscode.DocumentFilter[] = [
+        { scheme: 'file', language: 'c' },
+        { scheme: 'file', language: 'cpp' },
+        { scheme: 'file', language: 'cuda-cpp' },
+        { scheme: 'file', language: 'jsonc', pattern: '**/.vscode/*.json'},
```
sean-mcmanus, line 56: The .vscode/c_cpp_properties.json isn't getting picked up -- I think you need to add `        { scheme: 'file', language: 'json', pattern: '**/.vscode/*.json'},`. Or maybe there's some bug with us not associating that file with jsconc....

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Actually, there's a regression in that this code in activeDocumentChanged is no longer being hit:
```ts
            const isCpp: boolean = (activeEditor.document.uri.scheme === "file" && (activeEditor.document.languageId === "c" || activeEditor.document.languageId === "cpp" || activeEditor.document.languageId === "cuda-cpp"));

            let isCppPropertiesJson: boolean = false;
            if (activeEditor.document.languageId === "json" || activeEditor.document.languageId === "jsonc") {
                isCppPropertiesJson = activeEditor.document.fileName.endsWith("c_cpp_properties.json");
                if (isCppPropertiesJson) {
                    vscode.languages.setTextDocumentLanguage(activeEditor.document, "jsonc");
                }
            }
```
i.e. the c_cpp_properties.json should get associated with jsonc instead of json.

i.e. that code needs to be preserved with the new UI.

sean-mcmanus (changes requested):

sean-mcmanus:
The Tag Parser Initializing message is still being reported as Indexing with Indexing being reported as Parsing:
image
image

I thought that was going to be fixed?

sean-mcmanus:
Was "Parsing Open Files | Parsing Workspace" supposed to be collapsed to Parsing Workspace for V1 or is that for a follow-up?

Extension/src/LanguageServer/extension.ts
Bernardin-MS: In the new UI that is being handled by the documentSelector. Configuration has its own filter that includes json and jsonc files
https://github.com/microsoft/vscode-cpptools/blob/b522b62515a6117f3510d90d17bf3a582bfdefaf/Extension/src/LanguageServer/ui_new.ts#L52-59
```ts
    private configDocumentSelector: vscode.DocumentFilter[] = [
        { scheme: 'file', language: 'c' },
        { scheme: 'file', language: 'cpp' },
        { scheme: 'file', language: 'cuda-cpp' },
        { scheme: 'file', language: 'jsonc', pattern: '**/.vscode/*.json'},
        { scheme: 'file', language: 'jsonc', pattern: '**/*.code-workspace'},
        { scheme: 'output'}
    ];
```

bobbrow:
> The Tag Parser Initializing message is still being reported as Indexing with Indexing being reported as Parsing: image image
> 
> I thought that was going to be fixed?

From what I see in the code, "Indexing Workspace" is only ever present once per session.  I know that we can re-index if the database is reset, so there is something wrong here or we don't have enough data coming from the native side to make a determination.  Because this PR has been open for so long and this will be a small fix, I think we should just check this in now and then @Bernardin-MS needs to open a bug and fix it in a follow up PR today/tomorrow.  Will that work?

bobbrow:
> 

Actually, I think I identified the issue with @Bernardin-MS and we're looking at a solution.  With the old UI, we activate it when we see the first message from the tag parser (which happens to be "Discovering files")

sean-mcmanus (changes requested):
There's a bug in which "Parsing Open Files" doesn't go away from the status message after open file parsing is finished if the workspace parsing is paused. This appears to be a regression compared to the previous status bar. It goes away after the Resume is used though.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: @Bernardin-MS No, the code that associates c_cpp_properties.json is not being handled by the document selector, i.e. you need to keep that functionality in activeDocumentChanged, otherwise it's a regression bug and c_cpp_properties.json will get associated with json instead of jsonc, so users with existing comments will start getting warnings, i.e. you're referring to the other code in activeDocumentChanged, which is handled by the new document selector.

Bernardin-MS:
@sean-mcmanus after looking into it and a discussion with @bobbrow we're going to deal with showing the Indexing text in a later bug

michelleangela (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -174,7 +174,9 @@ export async function activate(): Promise<void> {
 
     console.log("starting language server");
     clients = new ClientCollection();
-    ui = getUI();
+    getActiveClient().awaitUntilLanguageClientReady().then(async () =>
```
Colengms: Is there a reason this needs to be deferred until after the language client is ready?  Looking at the implementation of getUI, couldn't you just await this here?  Also, updating the use of `ui` below to `ui?` is always going to fail, as `ui` will always be undefined there.  So, deferring `getUI` means that call to activeDocumentChanged will never occur.

Extension/src/LanguageServer/extension.ts
```diff
@@ -510,8 +582,8 @@ async function selectClient(): Promise<Client> {
     if (clients.Count === 1) {
         return clients.ActiveClient;
     } else {
-        const key: string = await ui.showWorkspaces(clients.Names);
-        if (key !== "") {
+        const key: string | undefined = await ui?.showWorkspaces(clients.Names);
```
Colengms: As mentioned previously, these `ui?`'s make me nervous, as these imply the behavior will change depending on execution timing, which is one of the worst types of bugs to investigate.  If we can await `getUI` in the first please, lets remove these `?`'s.

Extension/src/LanguageServer/ui.ts
```diff
@@ -57,19 +77,20 @@ export class UI {
...
     private runningCodeAnalysisTooltip: string = "";
     private codeAnalysisPausedTooltip: string = "";
+    get isNewUI(): boolean { return false; };
 
     constructor() {
         const configTooltip: string = localize("c.cpp.configuration.tooltip", "C/C++ Configuration");
         this.configStatusBarItem = vscode.window.createStatusBarItem("c.cpp.configuration.tooltip", vscode.StatusBarAlignment.Right, 0);
         this.configStatusBarItem.name = configTooltip;
-        this.configStatusBarItem.command = "C_Cpp.ConfigurationSelect";
+        this.configStatusBarItem.command = "C_Cpp.ConfigurationSelectUI_Telemetry";
```
Colengms, line 87: Does "_Telemetry" really need to be in the name of the command itself?  These strings are (potentially) exposed to users.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,548 @@
...
+            // Make sure not to overwrite current progress
+            tooltip: this.codeAnalysisStatusBarItem.command?.tooltip ?? localize("startup.codeanalysis.status", "Starting...")
+        } : {
+            command: "C_Cpp.ShowIdleCodeAnalysisCommandsUI_Telemetry",
+            title: localize("c.cpp.codeanalysis.statusbar.showRunNowOptions", "Run Now")
+        };
+    }
+
+    private updateCodeAnalysisTooltip(): void {
+        this.codeAnalysProgress = localize({ key: "running.analysis.processed.tooltip", comment: [this.codeAnalysisTranslationHint] }, "Running {0}: {1} / {2} ({3}%)", this.codeAnalysisProgram,
```
Colengms, line 309: Are you certain that it works to specifying the hint/comment using a variable here?  I know that the localizable string value itself MUST be a literal in calls to `localize`, due to how values are scraped from the code.  Could you run `yarn run translations-export` and check the generated files to ensure these hints are being captured correctly?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,548 @@
...
+        const showPrompt: () => Promise<ConfigurationStatus> = async () => {
+            const configured: boolean = await prompt();
+            return Promise.resolve({
+                priority: priority,
+                configured: configured
+            });
+        };
+
+        if (this.curConfigurationStatus) {
+            this.curConfigurationStatus = this.curConfigurationStatus.then(result => {
```
Colengms, line 522: We're trying to avoid use of `then` if not absolutely necessary, in favor of `async`/`await`.  Could you make this an async function and await here?  If you don't await `showConfigurationPrompt`, the effect should be the same.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -0,0 +1,548 @@
...
+        this.workspaceParsingStatus = this.workspaceParsingRunningText;
+
+        this.codeAnalysisStatusBarItem = vscode.languages.createLanguageStatusItem(`cpptools.status.${LanguageStatusPriority.Low}.codeanalysis`, documentSelector);
+        this.codeAnalysisStatusBarItem.name = localize("cpptools.status.codeanalysis", "C/C++ Code Analysis Status");
+        this.codeAnalysisStatusBarItem.text = `Code Analysis Mode: ${this.codeAnalysisCurrentMode()}`;
+        this.codeAnalysisStatusBarItem.command = {
+            command: "C_Cpp.ShowIdleCodeAnalysisCommandsUI_Telemetry",
+            title: localize("c.cpp.codeanalysis.statusbar.runNow", "Run Now")
+        };
+        // this.codeAnalysisStatusBarItem.severity = vscode.LanguageStatusSeverity.Information;
```
Colengms: Remove commented out code.

Colengms (changes requested):

Extension/src/LanguageServer/extension.ts
Bernardin-MS: It was necessary in order to for the experimentation API to get activated.

Bernardin-MS: In general, the need for a promise is because of the experiment so once that is done getUI will return to a normal function. Unless there is a way to use the experimentation api without a promise

Extension/src/LanguageServer/extension.ts
Bernardin-MS: I'm not sure how you mean

Extension/src/LanguageServer/ui.ts
Bernardin-MS: No the names can change, is it an issue to keep them the way they are?

Extension/src/LanguageServer/extension.ts
bobbrow: Colen is asking why we need to wait for the client to be ready before we can get the experimentation service.  Can we do this:
```
    ui = await getUI();
```
instead of:
```
    getActiveClient().awaitUntilLanguageClientReady().then(async () =>
        ui = await getUI()
    );
```

bobbrow: The experimentation service doesn't look like it has any dependencies on the client becoming active.

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: Great catch, i didn't notice. But this code is taken directly from the old ui implementation so it should work if it worked before.

Extension/src/LanguageServer/extension.ts
bobbrow: Also, if you set `ui` before creating `ClientCollection`, I think there is no risk to assuming the UI will never be `undefined` and the typing could be updated.

Bernardin-MS: My mistake, i got the error wrong. what I meant was without awaitUntilLanguageClientReady getUI() will be called multiple times before it can set ui. Which causes multiple constructor calls and resulting in a crash

bobbrow: That's probably because ClientCollection creates a Client whose constructor calls `getUI` asynchronously.  If you create the UI before the ClientCollection, I think that should resolve the problem.

Extension/src/LanguageServer/ui_new.ts
Bernardin-MS: This is current code, as discussed we'll address this separately.

Extension/src/LanguageServer/extension.ts
```diff
@@ -433,6 +438,73 @@ export function registerCommands(enabled: boolean): void {
...
+        () => {
+            logForUIExperiment("ParsingCommands");
+            onShowParsingCommands();
+        }
+        : onDisabledCommand));
+    // ----------------------------------------
+}
+
+async function logForUIExperiment(command: string): Promise<void> {
+    const isNewUI: string = (await getUI()).isNewUI.toString();
```
bobbrow: This could use the `ui` variable instead of `await getUI()`

Extension/src/LanguageServer/extension.ts
bobbrow: There appears to be a possible race in `getUI` where multiple callers could enter the `if (!ui)` branch, wait on the experimentation service promise, and then both try to create and set `ui`.  Changing the order of the calls as I suggest above mitigates that, but we may also want to follow up to this PR with a new one that fixes the reentrancy problem in `getUI`.

Bernardin-MS: Addressing it like how we discussed

Extension/src/LanguageServer/extension.ts
Bernardin-MS: Addressing it like how we discussed

Extension/src/LanguageServer/extension.ts
```diff
@@ -184,7 +184,7 @@ export async function activate(): Promise<void> {
 
     disposables.push(vscode.workspace.onDidChangeConfiguration(onDidChangeSettings));
     disposables.push(vscode.window.onDidChangeActiveTextEditor(onDidChangeActiveTextEditor));
-    ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
+    void ui.activeDocumentChanged(); // Handle already active documents (for non-cpp files that we don't register didOpen).
```
sean-mcmanus: Is this void needed?

sean-mcmanus: ...or why was it added?

sean-mcmanus: It doesn't appear needed to me since ui.activeDocumentChanged() is a void function already.

michelleangela (dismissed):

sean-mcmanus (approved):

sean-mcmanus (changes requested):
I'm seeing a regression in the Select Configuration command in which it isn't working after switching to a 2nd folder in a multi-root workspace.

Also, after switching to a different workspace folder, the configuration that is shown is not for the new workspace folder.

i.e. the status bar isn't updating in response to the workspace folder change (i.e. switching to a C/C++ file that belongs to another workspace folder).

sean-mcmanus (approved):

michelleangela (approved):

Colengms (approved):

PR #10138, opened 2022-11-11
Fixes include "header" issue with multi-root workspaces

### LINKED ISSUE #4983: Fully support relative paths in includePath
Right now, relative paths will sometimes work in the includePath, but with the new multiroot implementation it will probably fail, because it just uses the current working directory, which is probably incorrect for the multiroot scenario.

The original repro with relative paths is at https://github.com/microsoft/vscode-cpptools/issues/4976 .

The workaround is to manually add ${workspaceFolder} to the path.

Bernardin-MS:
Issue: #4983


Extension/src/LanguageServer/client.ts
```diff
@@ -2401,6 +2401,12 @@ export class DefaultClient implements Client {
             if (modifiedConfig.compileCommands) {
                 modifiedConfig.compileCommands = cppProperties.resolvePath(modifiedConfig.compileCommands, os.platform() === "win32");
             }
+
+            if (modifiedConfig.includePath) {
```
Colengms: I think (under) `updateServerOnFolderSettingsChange` is the right place for this fix (and to resolve all paths).  That function calls `updateConfigurationPathsArray` and a bunch of similar functions on each of the settings we process.  Those functions call `resolveAndSplit` on paths.  If you resolve relative to absolute paths there, it will apply to all settings paths that call it.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -2401,6 +2401,12 @@ export class DefaultClient implements Client {
...
+                const resolvedIncludePath: string[] = [];
+                modifiedConfig.includePath.forEach((path: string) => {resolvedIncludePath.push(cppProperties.resolvePath(path, os.platform() === "win32")); });
```
sean-mcmanus: This could be simplified to 
`modifiedConfig.includePath = modifiedConfig.includePath?.map((path: string) => cppProperties.resolvePath(path, os.platform() === "win32"));`

However, I don't know if this would still apply after following Colen's recommended changes.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
Bernardin-MS: They don't all call `resolveAndSplit` so that wouldn't really work

sean-mcmanus (dismissed):
I'm approving, but I'm not sure if this is what @Colengms was asking for or not so you should wait for his review.

Extension/src/LanguageServer/client.ts
Colengms: I don't think this is the right place to fix up these paths.  This is literally right before the info gets sent to the server, so none of the changes made here apply to use of those values on the TypeScript side.  I had added this bit regarding `modifiedConfig` to make some adjustments specifically for `compilerArgsLegacy`.  The right place to be fixing up strings to ensure variables are fully resolved, is in `updateServerOnFolderSettingsChange`.  That is where configuration paths are processed, multiple levels of defaults applied, variables expanded, etc..

Colengms (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -965,6 +965,29 @@ export class CppProperties {
                     this.lastCustomBrowseConfiguration.Value = undefined;
                 }
             }
+
+            /*
+             * Ensure all paths are absolute
+             */
+            if (configuration.macFrameworkPath) {
+                configuration.macFrameworkPath = configuration.macFrameworkPath.map((path: string) => this.resolvePath(path, os.platform() === "win32"));
```
michelleangela: Can the value of `os.platform() === "win32"` be saved and then reused in all the calls of `resolvePath`?

sean-mcmanus (dismissed):

sean-mcmanus (approved):

michelleangela (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -152,6 +152,7 @@ export class CppProperties {
     private lastCustomBrowseConfiguration: PersistentFolderState<WorkspaceBrowseConfiguration | undefined> | undefined;
     private lastCustomBrowseConfigurationProviderId: PersistentFolderState<string | undefined> | undefined;
     private lastCustomBrowseConfigurationProviderVersion: PersistentFolderState<Version> | undefined;
+    private isWin32 = os.platform() === "win32";
```
michelleangela: Does this save the value or does it always have to evaluate the comparison whenever `this.isWin32` is called?

sean-mcmanus: I think it saves the value (it doesn't appear to be a lambda function) -- but why is there no type? Can you add a type "boolean"? I would've expected a linter error at least.

sean-mcmanus (changes requested):
see my comment

sean-mcmanus (approved):

michelleangela (approved):

Colengms (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -965,6 +966,29 @@ export class CppProperties {
...
+            if (configuration.compileCommands) {
+                configuration.compileCommands = this.resolvePath(configuration.compileCommands, this.isWin32);
+            }
+
+            if (configuration.forcedInclude) {
+                configuration.forcedInclude = configuration.forcedInclude.map((path: string) => this.resolvePath(path, this.isWin32));
+            }
+
+            if (configuration.includePath) {
+                configuration.includePath = configuration.includePath.map((path: string) => this.resolvePath(path, this.isWin32));
```
sean-mcmanus, line 990: @Colengms @Bernardin-MS  This broke recursive includes because `${workspaceFolder}/**` has the `**` removed. I'm looking into a fix.

sean-mcmanus: I have a PR with a fix at https://github.com/microsoft/vscode-cpptools/pull/10337

PR #10145, opened 2022-11-14
Update IntelliSense localized strings and msvc versions.

sean-mcmanus:


sean-mcmanus:
I'm checking this in now since the cpptools side got updated (for 1.14.0).

Issue #10027, opened 2022-10-17
Compiler selection via QuickPick


michelleangela:
Add the ability to select a compiler to configure for IntelliSense if there is no compiler configured during first usage of the extension. Additionally, the extension will no longer automatically query a default compiler if one is found on a machine unless it is confirmed by the user to use the compiler.

sean-mcmanus:
@michelleangela The spec is an internal document (in progress). Are you saying we should add an external spec too?

michelleangela:
@sean-mcmanus, the specification for this issue would be for any user facing experience.

PR #10165, opened 2022-11-17
Add support for selecting a default compiler

browntarik:
**Addresses**:  https://github.com/microsoft/vscode-cpptools/issues/10027


Extension/src/LanguageServer/client.ts
```diff
@@ -513,6 +519,7 @@ const RemoveCodeAnalysisProblemsNotification: NotificationType<RemoveCodeAnalysi
 
 // Notifications from the server
 const ReloadWindowNotification: NotificationType<void> = new NotificationType<void>('cpptools/reloadWindow');
+const UpdateTrustedCompilersNotification: NotificationType<UpdateTrustedCompilerPathsResult> = new NotificationType<UpdateTrustedCompilerPathsResult>('cpptools/cpptools/updateTrustedCompilersList');
```
michelleangela: `cpptools/cpptools/`? or `cpptools/`?

Extension/src/LanguageServer/client.ts
```diff
@@ -807,6 +815,83 @@ export class DefaultClient implements Client {
...
     }
 
+    async promptSelectCompiler(command: boolean | undefined): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("selectCompiler.string", "Yes");
+        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+
+        if (compilerDefaults.compilerPath !== "") {
+            if (!command) {
+                const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", `The ${compilerDefaults.compilerPath} was found on this computer. Do you want to configure it for intellisense?`), confirmCompiler, selectCompiler);
```
michelleangela: intellisense to "IntelliSense"

Extension/src/LanguageServer/client.ts
```diff
@@ -807,6 +815,83 @@ export class DefaultClient implements Client {
...
+                }
+            } else {
+                var paths = compilerDefaults.knownCompilers.map(function (a) { return a.path; });
+                const index: number = await ui.showSelectCompiler(paths);
+                if (index < 0) {
+                }
+                util.updateTrustedCompilersList(paths[index])
+                const selectedCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);;
+                compilerDefaults = selectedCompilerDefaults;
+                clients.forEach(client => {
```
michelleangela:
```
                clients.forEach(client => {
                    if (client instanceof DefaultClient) {
                        const defaultClient: DefaultClient = <DefaultClient>client;
                        defaultClient.configuration.CompilerDefaults = compilerDefaults;
                        defaultClient.configuration.handleConfigurationChange();
                    }
                });
```

This repeating process could be moved out of each command check statement and placed at the end of the scope `if (!command) {`

Extension/src/LanguageServer/client.ts
```diff
@@ -3064,6 +3193,7 @@ class NullClient implements Client {
     activate(): void { }
     selectionChanged(selection: Range): void { }
     resetDatabase(): void { }
+    promptSelectCompiler(command: boolean | undefined): Promise<void> { return Promise.resolve(); }
```
Colengms: Is the `| undefined` needed?  If either true or false must be passed in, you don't need `| undefined`.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -857,9 +866,9 @@ export class CppProperties {
                 // compile_commands.json already specifies a compiler. compilerPath overrides the compile_commands.json compiler so
                 // don't set a default when compileCommands is in use.
                 configuration.compilerPath = this.updateConfigurationString(configuration.compilerPath, settings.defaultCompilerPath, env, true);
-                configuration.compilerPathIsExplicit = configuration.compilerPathIsExplicit || settings.defaultCompilerPath !== undefined;
                 if (configuration.compilerPath === undefined) {
-                    if (!!this.defaultCompilerPath) {
+                    configuration.compilerPathIsExplicit = configuration.compilerPathIsExplicit || settings.defaultCompilerPath !== undefined;
```
Colengms: Could you move this back up to before the IF statement?  `compilerPathIsExplicit` indicates whether the `compilerPath` field was set by the user (or got a value from settings.defaultCompilerPath, which was set by the user). 
 (When this is false, we hide warnings about failures to query a compiler they did not explicitly specify).  The prior call to `updateConfigurationString` could have updated this `compilerPath` to be equal to `settings.defaultCompilerPath`, so this line is updating `compilerPathIsExplicit` in case `settings.defaultCompilerPath` was used.

Extension/src/LanguageServer/client.ts
```diff
@@ -807,6 +815,83 @@ export class DefaultClient implements Client {
         };
     }
 
+    async promptSelectCompiler(command: boolean | undefined): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("selectCompiler.string", "Yes");
```
Colengms: Every call to localize must use a unique ID (the first arg).  Otherwise, only one of these strings will get localized.

Extension/src/LanguageServer/client.ts
```diff
@@ -807,6 +815,83 @@ export class DefaultClient implements Client {
...
+                clients.forEach(client => {
+                    if (client instanceof DefaultClient) {
+                        const defaultClient: DefaultClient = <DefaultClient>client;
+                        defaultClient.configuration.CompilerDefaults = compilerDefaults;
+                        defaultClient.configuration.handleConfigurationChange();
+                    }
+                });
+            }
+        } else {
+            vscode.window.showInformationMessage(localize("selectCompiler.message", "No compiler detected, please install a compiler")).then((value?: string) => {
```
Colengms: Is the lambda here intended to be empty?  If it doesn't need to do anything, then entire `then` can be removed.

Extension/src/LanguageServer/client.ts
```diff
@@ -893,12 +978,43 @@ export class DefaultClient implements Client {
...
+                        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+                        compilerPath.Value = [];
+                        const inputCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);
                         compilerDefaults = inputCompilerDefaults;
                         this.configuration.CompilerDefaults = compilerDefaults;
+                        this.configuration.setupConfigurations();
+                        if (!compilerDefaults.trustedCompilerFound) {
+                            this.notifyWhenLanguageClientReady(() => {
+                                // if there is a compilerPath in c_cpp_properties.json, make compiler trusted and query compiler
+                                clients.forEach(async client => {
```
Colengms: I'm not sure this is the right place for this.  Since the configuration logic will trust any explicitly set `compilerPath` it sees, these checks seem redundant.  It seems like we need a place where, after configurations have been processed for all clients, we can simply check if there is a trusted compiler yet, and prompt if there is not...  It looks like just be a matter of deferring this until all clients have been constructed.  Maybe the check to display the prompt could be moved into `activate` in extension.ts, right after `clients = new ClientCollection();`.

Colengms: It might also be worth setting a timer and not doing the check until after a few seconds, to give a custom configuration provider the opportunity to start up (which does not happen immediately).  When/if a custom configuration provider registeres with cpptools, it may provide a browse config with a (trustable) compiler path in it.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -723,6 +731,7 @@ export interface Client {
     selectionChanged(selection: Range): void;
     resetDatabase(): void;
     deactivate(): void;
+    promptSelectCompiler(command: boolean | undefined): Promise<void>;
```
Colengms: In the implementation, you don't have `| undefined`.  Could you remove that from here, and from around line 3336.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+        const index: number = await ui.showSelectCompiler(paths);
+        util.updateTrustedCompilersList(paths[index]);
+        if (index === paths.length - 1) {
+            paths[index] = "";
+        } else if (index === paths.length - 2) {
+            vscode.commands.executeCommand('vscode.open', "https://learn.microsoft.com/en-us/cpp/build/reference/compiling-a-c-cpp-program?view=msvc-170");
+            return;
+        } else if (index === paths.length - 3) {
+            const result: any = await vscode.window.showOpenDialog();
+            if (result?.toString()) {
```
Colengms: It's unclear to me what this is doing.  `showOpenDialog` returns `Uri[] | undefined`.  You should use that type instead of `any`.  Also, you're checking the result of calling `toString()` on a `Uri[]`.  Perhaps you want to check: `(result != undefined && result.length > 0)` ?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+                util.updateTrustedCompilersList(result?.toString());
```
Colengms: `updateTrustedCompilersList` takes a single string.  It looks like you're passing it a `Uri[]` converted to a string?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+                await vscode.workspace.getConfiguration("C_Cpp.default").update("compilerPath", result?.toString(), vscode.ConfigurationTarget.Global);
```
Colengms: Another instance of `result?.toString()`.  I'm guessing this isn't right.

Extension/src/common.ts
```diff
@@ -865,17 +866,30 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+        vscode.commands.executeCommand("workbench.action.reloadWindow");
+    }
+
+}
+
+export async function updateTrustedCompilersList(path: string): Promise<void> {
+    // detect duplicate paths
+    const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+    const compilerPaths: string[] = compilerPath.Value;
+    compilerPaths.forEach(compiler => {
```
Colengms: I'm not sure this `forEach` is accomplishing what you intended.  When you `return` from a lambda, it returns only from that lambda, not the calling function.  If your intention is to avoid this whole function if there is already a match for the compiler path, I'd suggest using a regular `for` loop.  That is also the only way to exit the loop early, as a 'forEach` will always visit all elements.

Extension/src/common.ts
```diff
@@ -865,17 +866,30 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+        if (compiler === path) {
+            return;
         }
     });
+    compilerPaths.push(path);
+    compilerPath.Value = compilerPaths;
```
Colengms: Is it necessary to assign to `compilerPath.Value` again, or would it be sufficient to modify the existing array within `compilerPath.Value` in place?

Extension/src/LanguageServer/client.ts
```diff
@@ -948,13 +1028,22 @@ export class DefaultClient implements Client {
...
                         // The configurations will not be sent to the language server until the default include paths and frameworks have been set.
                         // The event handlers must be set before this happens.
-                        const inputCompilerDefaults: configs.CompilerDefaults = await languageClient.sendRequest(QueryCompilerDefaultsRequest, {});
+                        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+                        const inputCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);
                         compilerDefaults = inputCompilerDefaults;
                         this.configuration.CompilerDefaults = compilerDefaults;
-
+                        this.configuration.setupConfigurations();
+                        clientCount++;
```
Colengms: The user may add a client later. For example, if the user adds 3 folders, one right after the other, this code will be hit 3 additional times, despite no changes having been made that would result in a trusted compiler being set.  You probably don't want to prompt the user 4 times.  I think you need an `I already did this check` boolean here, to prevent the prompt from appearing multiple times in that scenario.

Extension/src/LanguageServer/client.ts
```diff
@@ -948,13 +1028,22 @@ export class DefaultClient implements Client {
...
+                        // count number of clients, once all clients are configured, check for trusted compiler to display notification to user and add a short delay to account for config provider logic to finish
+                        if (clientCount === vscode.workspace.workspaceFolders?.length) {
+                            if (!compilerDefaults.trustedCompilerFound && !defaultCompilerFound) {
+                                // if there is no compilerPath in c_cpp_properties.json, prompt user to configure a compiler
+                                this.promptSelectCompiler(false);
+                                defaultCompilerFound = true;
```
Colengms: The call to `promptSelectCompiler` is not awaited.  That is actually good, as we don't want to stall the whole extension while waiting for the user to respond to that UI.  However, that also means we don't yet know if a default compiler was found.  Maybe `defaultCompilerFound` should not be set here?  Or, if your intention was to use this boolean as the `I already did this check` boolean from my other comment, I'd suggest naming it more concisely.

Extension/src/LanguageServer/client.ts
```diff
@@ -985,7 +1074,18 @@ export class DefaultClient implements Client {
                         // Listen for messages from the language server.
                         this.registerNotifications();
                     } else {
+                        // update all client configurations
                         this.configuration.CompilerDefaults = compilerDefaults;
+                        this.configuration.setupConfigurations();
```
Colengms: This code seems duplicated.  That can be an issue, if a bug is fixed in one place and not the other.  Could you reorganize this code so the duplication is not required?  In this case, it looks like some code that applies to both conditions can be moved to after the `else` block.

Extension/src/LanguageServer/extension.ts
```diff
@@ -174,6 +174,7 @@ export async function activate(): Promise<void> {
 
     console.log("starting language server");
     clients = new ClientCollection();
+    // set a timer that waits for a few seconds
```
Colengms: I think this comment can be removed?

Colengms (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
         return (selection) ? selection.index : -1;
     }
 
+    public async showSelectCompiler(paths: string[]): Promise<number> {
+        const options: vscode.QuickPickOptions = {};
+        options.placeHolder = localize("select.compile.commands", "Select a a compiler to configure for intellisense");
```
jogo-: Duplicate word: **a**
Wrong casing: **i**ntelli**s**ense -> **I**ntelli**S**ense

Extension/src/nativeStrings.json
```diff
@@ -177,6 +177,9 @@
         "text": "+%d specializations",
         "hint": "Refers to C++ template specializations. This is part of a description indicating there are N additional specializations of a function.  %d is replaced with that number."
     },
+    "compiler": {
+        "text": "Note: Intellisense not fully configured. Use the Select Default compiler command to finish configuration"
```
jogo-: Wrong casing: Intelli**s**ense -> Intelli**S**ense

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+                    util.updateTrustedCompilersList(compilerDefaults.compilerPath);
+                    await vscode.workspace.getConfiguration("C_Cpp.default").update("compilerPath", compilerDefaults.compilerPath, vscode.ConfigurationTarget.Global);
+                    const selectedCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);
+                    compilerDefaults = selectedCompilerDefaults;
+                    this.updateClientConfigurations();
+                } else if (value === selectCompiler) {
+                    this.handleCompilerQuickPick();
+                } else {
+                    const setCompiler: string = localize("setCompiler.string", "Set Compiler");
+                    const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", `You do not have a compiler configured. Unless you set your own configurations, Intellisense may not be functional`), selectCompiler);
```
jogo-: Wrong casing: Intelli**s**ense -> Intelli**S**ense

Extension/src/common.ts
browntarik: You have to assign it again due to how the persistent state class assigns new values. For example, compilerPath.Value.push() will not modify the existing array.

Extension/src/LanguageServer/client.ts
browntarik: This was meant to be the "I already checked" boolean. Do you have a recommendation for what it should be called?

Extension/src/LanguageServer/client.ts
browntarik: I don't understand, it looks like it should have to go before the isFirstClient check correct?

Extension/src/LanguageServer/client.ts
browntarik: Added a boolean for this in "defaultCompilerFound"

Extension/src/LanguageServer/client.ts
Colengms: Since you have some logic (the call to requestCompiler) that needs to happen first, it would seem to make more sense to move the common/duplicate code to after the else block.  (There is no reason the call to promptSelectCompiler can't happen after the call to this.registerNotifications.

Extension/src/LanguageServer/client.ts
Colengms: I guess my feedback is that the name of the variable is misleading.  Technically, a default compiler has not yet been determined until the user finishes selecting one.  Could you name it something more specific to it's purpose, such as `displayedSelectCompiler`.

Extension/src/LanguageServer/client.ts
Colengms: Maybe `displayedSelectCompiler`, `selectCompilerPromptDisplayed`, `promptSelectCompilerCalled`?

Extension/src/nativeStrings.json
sean-mcmanus: * Should "compiler" be a more specific label?
* "Intellisense" should be "IntelliSense".
* "not" should be "is not".
* "Select Default compiler" should be "Select Default Compiler".
* "configuration" should be "configuration.".

Extension/src/common.ts
```diff
@@ -865,17 +866,30 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+    }
+
+}
+
+export async function updateTrustedCompilersList(path: string): Promise<void> {
+    // detect duplicate paths
+    const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+    const compilerPaths: string[] = compilerPath.Value;
+    compilerPaths.forEach(compiler => {
+        if (compiler === path) {
```
sean-mcmanus: This forEach doesn't appear to do anything. It's just exiting from the lambda, right?

Extension/src/LanguageServer/ui.ts
sean-mcmanus: "intellisense" should be "IntelliSense" in all cases.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -205,18 +208,11 @@ export class CppProperties {
...
-        this.defaultCppStandard = compilerDefaults.cppStandard;
-        this.defaultIncludes = compilerDefaults.includes;
-        this.defaultFrameworks = compilerDefaults.frameworks;
-        this.defaultWindowsSdkVersion = compilerDefaults.windowsSdkVersion;
-        this.defaultIntelliSenseMode = compilerDefaults.intelliSenseMode;
+    public setupConfigurations(): void {
 
         // defaultPaths is only used when there isn't a c_cpp_properties.json, but we don't send the configuration changed event
         // to the language server until the default include paths and frameworks have been sent.
+
```
sean-mcmanus, line 217: Was this newline intentional? We generally don't add a newline to separate a comment from what it's describing.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+    }
+
+    public async handleCompilerQuickPick(): Promise<void> {
+        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+        const paths: string[] = compilerDefaults.knownCompilers.map(function (a: configs.KnownCompiler): string { return a.path; });
+        paths.push("Select another compiler on my machine");
+        paths.push("Help me install a compiler");
+        paths.push("Do not configure a compiler (not recommended)");
+        const index: number = await ui.showSelectCompiler(paths);
+        util.updateTrustedCompilersList(paths[index]);
```
Colengms: It looks like this is adding `path[index]` to the list of trusted compilers before checking if the user selected one of the compiler entries.  Should this maybe go into a final else?

Also, the function name is a bit vague.  Maybe `addTrustedCompiler` instead of `updateTrustedCompilersList` as that more accurately reflects what it does?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+        if (index === paths.length - 1) {
+            paths[index] = "";
+        } else if (index === paths.length - 2) {
+            vscode.commands.executeCommand('vscode.open', "https://learn.microsoft.com/en-us/cpp/build/reference/compiling-a-c-cpp-program?view=msvc-170");
```
Colengms: We should set up a fowarding URL for this, rather than hard-code a URL that may not live forever.  You may want to check with Bob for assistance with creating forwarding URLs.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,77 @@ export class DefaultClient implements Client {
...
+            return;
+        } else if (index === paths.length - 3) {
+            const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
+            if (result !== undefined && result.length > 0) {
+                util.updateTrustedCompilersList(result[0].fsPath);
+                await vscode.workspace.getConfiguration("C_Cpp.default").update("compilerPath", result[0].fsPath, vscode.ConfigurationTarget.Global);
```
Colengms: Could you move this (and other occurrences) into a `set` property in settings.ts.  As an example, there is something similar there for `filesAssociations`.

Extension/src/LanguageServer/client.ts
```diff
@@ -985,7 +1065,18 @@ export class DefaultClient implements Client {
                         // Listen for messages from the language server.
                         this.registerNotifications();
                     } else {
+                        // update all client configurations
                         this.configuration.CompilerDefaults = compilerDefaults;
+                        clientCount++;
```
Colengms: These 2 lines are also in common, and can be moved to after the else block.

Colengms (changes requested):

Extension/src/LanguageServer/settings.ts
```diff
@@ -964,6 +964,9 @@ export class OtherSettings {
     public set filesAssociations(value: any) {
         vscode.workspace.getConfiguration("files").update("associations", value, vscode.ConfigurationTarget.Workspace);
     }
+    public set defaultCompiler(value: any) {
+        vscode.workspace.getConfiguration("files").update("associations", value, vscode.ConfigurationTarget.Global);
```
Colengms: This is setting `files.associations`, and should be updated to set `C_Cpp.default.compilerPath`.  Also please avoid using `any` unless absolutely necessary.  TypeScript is strongly typed, and use of `any` bypasses that strong typing.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,94 @@ export class DefaultClient implements Client {
...
+        clients.forEach(client => {
+            if (client instanceof DefaultClient) {
+                const defaultClient: DefaultClient = <DefaultClient>client;
+                defaultClient.configuration.CompilerDefaults = compilerDefaults;
+                defaultClient.configuration.handleConfigurationChange();
+            }
+        });
+    }
+
+    public set defaultCompiler(value: any) {
```
Colengms: This function seems like it shouldn't exist.  It's the same as the setter in settings.ts.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,94 @@ export class DefaultClient implements Client {
...
+                await this.defaultCompiler(result[0].fsPath);
+                const selectedCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);
+                compilerDefaults = selectedCompilerDefaults;
+                this.updateClientConfigurations();
+                return;
+            }
+        } else {
+            util.addTrustedCompiler(paths[index]);
+        }
+        await vscode.workspace.getConfiguration("C_Cpp.default").update("compilerPath", paths[index], vscode.ConfigurationTarget.Global);
```
Colengms: Shouldn't this go through the setter in settings.ts?

Colengms: Also, if "Do not configure a compiler (not recommended)" is selected, does this result in a value of `""` being set in settings.json?  If so, that may not be desirable, as that has a special meaning (that compilerPath should be explicitly disabled).

Extension/src/LanguageServer/client.ts
```diff
@@ -985,8 +1081,19 @@ export class DefaultClient implements Client {
                         // Listen for messages from the language server.
                         this.registerNotifications();
                     } else {
+                        // update all client configurations
                         this.configuration.CompilerDefaults = compilerDefaults;
```
Colengms: This line is also the same in both blocks and can be moved to after the else block.

Colengms (changes requested):

Extension/src/LanguageServer/settings.ts
browntarik: Whenever I set the type to string, I get a "Type 'String' has no call signatures" error

Extension/src/LanguageServer/client.ts
browntarik: What should it be set to instead?

Colengms: If the intention is to prompt them again next time, then the default.compilerPath should not get set to anything.  If the intention is to never prompt them again and disable the use of a compiler in the fallback default configuration, then `""` is appropriate.  If the intention is for some other behavior, we may need to identify what that behavior is and figure out how to get it.  What should the behavior be?  :)

Extension/src/LanguageServer/settings.ts
Colengms: I could help you look into this, if you want.  Generally, we shouldn't use 'any' unless absolutely necessary.

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
...
+        options.placeHolder = localize("select.compile.commands", "Select a compiler to configure for IntelliSense");
+
+        const items: IndexableQuickPickItem[] = [];
+        for (let i: number = 0; i < paths.length; i++) {
+            let compiler: string | undefined;
+            let isCompiler: boolean = false;
+            if (paths[i].indexOf("\\") > 0) {
+                if (paths[i].split("\\").pop() !== undefined) {
+                    compiler = paths[i].split("\\").pop();
+                    isCompiler = true;
```
michelleangela: `isCompiler` needed or would it be sufficient to check if `compiler` was set?

Extension/src/common.ts
```diff
@@ -865,17 +866,31 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+    }
+
+}
+
+export async function addTrustedCompiler(path: string): Promise<void> {
+    // detect duplicate paths
+    const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+    const compilerPaths: string[] = compilerPath.Value;
+    let i: number = 0;
+    for (i = 0; i <= compilerPaths.length; i++) {
```
michelleangela: `(i = 0; i <= compilerPaths.length; i++)` scan `i` also be equal to length of the array?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,91 @@ export class DefaultClient implements Client {
...
+    }
+
+    async promptSelectCompiler(command: boolean): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
+        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+        const settings: OtherSettings = new OtherSettings();
+        if (compilerDefaults.compilerPath !== "") {
+            if (!command) {
+                const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", "The ${compilerDefaults.compilerPath} was found on this computer. Do you want to configure it for IntelliSense?"), confirmCompiler, selectCompiler);
```
michelleangela: Should the message say "The compiler <compiler_path> was found on this computer." Instead of "The <compiler_path> .... "

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,91 @@ export class DefaultClient implements Client {
...
+                    return;
+                default:
+                    throw new Error(localize("unexpected.os", "Unexpected OS"));
+            }
+        } else if (index === paths.length - 3) {
+            const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
+            if (result !== undefined && result.length > 0) {
+                util.addTrustedCompiler(result[0].fsPath);
+                settings.defaultCompiler = result[0].fsPath;
+                const selectedCompilerDefaults: configs.CompilerDefaults = await this.requestCompiler(compilerPath.Value);
```
michelleangela: can this be `compilerDefaults = await this.requestCompiler(compilerPath.Value);` to remove the extra variable `selectedCompilerDefaults`? Same for lines 929 and 945.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,91 @@ export class DefaultClient implements Client {
...
+                defaultClient.configuration.handleConfigurationChange();
+            }
+        });
+    }
+
+    public async handleCompilerQuickPick(): Promise<void> {
+        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
+        const settings: OtherSettings = new OtherSettings();
+        const paths: string[] = compilerDefaults.knownCompilers.map(function (a: configs.KnownCompiler): string { return a.path; });
+        paths.push("Select another compiler on my machine");
```
michelleangela: Are these strings localized?

michelleangela (changes requested):

Extension/src/LanguageServer/ui.ts
browntarik: This has to do with the way that I am adding extra options to the quick pick. There will non compiler values in the quick pick that allow users to do other things. The only way to make these visible in the quick pick is to make them a label so I have to distinguish whether the compiler variable is a compiler. Perhaps I will change the variable from "compiler" to "option"

Extension/src/common.ts
```diff
@@ -865,17 +866,31 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+    const value: string | undefined = await vscode.window.showInformationMessage(message, reload);
+    if (value === reload) {
+        vscode.commands.executeCommand("workbench.action.reloadWindow");
+    }
+
+}
+
+export async function addTrustedCompiler(path: string): Promise<void> {
+    // detect duplicate paths
+    const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
```
sean-mcmanus: Can you rename compilerPath to something else since it's a plural object? e.g. persistentCompilerPaths?

Extension/src/common.ts
```diff
@@ -865,17 +866,31 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+    const compilerPaths: string[] = compilerPath.Value;
+    let i: number = 0;
```
sean-mcmanus:
```typescript
if (compilerPaths.includes(path)) {
    return;
}
```

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -987,6 +1076,17 @@ export class DefaultClient implements Client {
                     } else {
                         this.configuration.CompilerDefaults = compilerDefaults;
                     }
+                    // update all client configurations
+                    this.configuration.setupConfigurations();
+                    clientCount++;
```
sean-mcmanus: Does this need to be reset? I'm concerned about what happens if cpptools crashes and it might recreate a client, not sure.

sean-mcmanus: Should this code be moved the ClientCollection constructor? The ClientColleciton is usually what deals with stuff related to multiple clients.

Extension/src/LanguageServer/client.ts
```diff
@@ -58,6 +58,7 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 let ui: UI;
 let timeStamp: number = 0;
 const configProviderTimeout: number = 2000;
+let clientCount: number = 0;
```
sean-mcmanus: Is this the same as this.clients.Count?

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
...
+                if (paths[i].split("\\").pop() !== undefined) {
+                    option = paths[i].split("\\").pop();
+                    isCompiler = true;
+                }
+            } else {
+                isCompiler = false;
+            }
+            if (option !== undefined && isCompiler) {
+                const path: string | undefined = paths[i].replace(option, "");
+                items.push({ label: option, description: "Found at " + path, index: i });
```
sean-mcmanus: This needs to be localized, right?

Extension/src/common.ts
```diff
@@ -865,17 +866,31 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
+export async function promptReloadWindow(message: string): Promise<void> {
     const reload: string = localize("reload.string", "Reload");
-    vscode.window.showInformationMessage(message, reload).then((value?: string) => {
-        if (value === reload) {
-            vscode.commands.executeCommand("workbench.action.reloadWindow");
+    const value: string | undefined = await vscode.window.showInformationMessage(message, reload);
+    if (value === reload) {
+        vscode.commands.executeCommand("workbench.action.reloadWindow");
+    }
+
```
sean-mcmanus: Blank extra newline

Extension/src/nativeStrings.json
```diff
@@ -177,6 +177,9 @@
         "text": "+%d specializations",
         "hint": "Refers to C++ template specializations. This is part of a description indicating there are N additional specializations of a function.  %d is replaced with that number."
     },
+    "compiler": {
+        "text": "Note: IntelliSense is not fully configured. Use the Select Default Compiler command to finish configuration."
```
sean-mcmanus: I believe this should be `'Select Default Compiler' command` with the `'` added since that's how VS Code's references commands in the settings UI.

Extension/src/LanguageServer/client.ts
browntarik: @Colengms could you give your input on this?

Extension/src/LanguageServer/client.ts
browntarik: I am not sure, does that attribute keep a running count of clients as they are initialized?

Colengms: Maybe rename it `initializedClientCount` to make that clear?

Extension/src/LanguageServer/client.ts
Colengms: @sean-mcmanus  The user should not be prompted to select a trusted compiler until after all processing of configurations, as any user-specified compiler would be implicitly trusted, making the prompt unnecessary.  So, this check needs to occur after all clients are created, async initialization of the native process is complete, and all configurations have been processed by the server.  We won't know about all user-specified compilers until the native process is up and running and has processed configurations for all workspace folders, as a compile_commands.json may be used, and those are processed entirely on the native side.  Can you suggest a better place to move this check to?  The current location may actually not be sufficient, as the native process may not yet have processed compile_commands.json files yet.  One solution to that might be to make didChangeCppProperties a request instead of a notification, so that we get a response only once it's done.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, so why can't it be moved to the end of the ClientCollection constuctor?

sean-mcmanus: Okay, I see now, the ClientCollection doesn't block for the clients to actually be ready.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, it should just be renamed so others aren't confused, like I was.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Also, it looks like it doesn't need to be reset due to "displayedSelectCompiler" being checked.

Extension/src/LanguageServer/client.ts
sean-mcmanus: in other words, the ClientCollection has a count of the clients already, but not whether they're finished initializing yet.

Extension/src/common.ts
```diff
@@ -865,17 +866,30 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
-            vscode.commands.executeCommand("workbench.action.reloadWindow");
+    const value: string | undefined = await vscode.window.showInformationMessage(message, reload);
+    if (value === reload) {
+        vscode.commands.executeCommand("workbench.action.reloadWindow");
+    }
+}
+
+export async function addTrustedCompiler(path: string): Promise<void> {
+    // detect duplicate paths
+    const  persistentCompilerPaths: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
```
sean-mcmanus: Double spaces

Extension/src/common.ts
sean-mcmanus: The loop should be removed (along with unused loop variable).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
...
+        for (let i: number = 0; i < paths.length; i++) {
+            let option: string | undefined;
+            let isCompiler: boolean = false;
+            if (paths[i].indexOf("\\") > 0) {
+                if (paths[i].split("\\").pop() !== undefined) {
+                    option = paths[i].split("\\").pop();
+                    isCompiler = true;
+                }
+            } else {
+                isCompiler = false;
```
michelleangela: Is there a need to set `isCompiler = false;` if the initial value is false and it will only get set to true on line 358?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,87 @@ export class DefaultClient implements Client {
...
+            if (client instanceof DefaultClient) {
+                const defaultClient: DefaultClient = <DefaultClient>client;
+                defaultClient.configuration.CompilerDefaults = compilerDefaults;
+                defaultClient.configuration.handleConfigurationChange();
+            }
+        });
+    }
+
+    public async handleCompilerQuickPick(): Promise<void> {
+        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
```
michelleangela: Should `compilerPath` be `compilerPaths`?

Extension/src/LanguageServer/client.ts
```diff
@@ -948,13 +1038,12 @@ export class DefaultClient implements Client {
 
                     if (isFirstClient) {
                         workspaceReferences = new refs.ReferencesManager(this);
-
                         // The configurations will not be sent to the language server until the default include paths and frameworks have been set.
                         // The event handlers must be set before this happens.
-                        const inputCompilerDefaults: configs.CompilerDefaults = await languageClient.sendRequest(QueryCompilerDefaultsRequest, {});
+                        const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);
```
michelleangela: Should `compilerPath` be `compilerPaths`?

michelleangela: If `const compilerPath: PersistentState<string[]> = new PersistentState<string[]>("CPP.trustedCompilerPaths", []);` is always called just before `requestCompiler`, can it be moved within `requestCompiler`?

Extension/src/LanguageServer/client.ts
```diff
@@ -987,6 +1076,17 @@ export class DefaultClient implements Client {
                     } else {
                         this.configuration.CompilerDefaults = compilerDefaults;
                     }
+                    // update all client configurations
+                    this.configuration.setupConfigurations();
+                    initializedClientCount++;
+                    // count number of clients, once all clients are configured, check for trusted compiler to display notification to user and add a short delay to account for config provider logic to finish
+                    if (initializedClientCount === vscode.workspace.workspaceFolders?.length) {
```
michelleangela: These two checks can be in one conditional statement.

Extension/src/common.ts
```diff
@@ -873,17 +874,27 @@ export async function renameAsync(oldName: string, newName: string): Promise<voi
...
-export function promptReloadWindow(message: string): void {
+export async function promptReloadWindow(message: string): Promise<void> {
     const reload: string = localize("reload.string", "Reload");
-    vscode.window.showInformationMessage(message, reload).then((value?: string) => {
-        if (value === reload) {
-            vscode.commands.executeCommand("workbench.action.reloadWindow");
-        }
-    });
+    const value: string | undefined = await vscode.window.showInformationMessage(message, reload);
+    if (value === reload) {
```
michelleangela, line 884: Does this need to be the exact string value or can it just be not undefined? Will it return other string values?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,87 @@ export class DefaultClient implements Client {
...
+                if (value === confirmCompiler) {
+                    util.addTrustedCompiler(compilerDefaults.compilerPath);
+                    settings.defaultCompiler = compilerDefaults.compilerPath;
+                    compilerDefaults = await this.requestCompiler(compilerPath.Value);
+                    this.updateClientConfigurations();
+                } else if (value === selectCompiler) {
+                    this.handleCompilerQuickPick();
+                } else {
+                    const setCompiler: string = localize("setCompiler.string", "Set Compiler");
+                    const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have a compiler configured. Unless you set your own configurations, IntelliSense may not be functional"), selectCompiler);
```
sean-mcmanus: I think this should end in a period, unless there's a reason not to?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,87 @@ export class DefaultClient implements Client {
...
+                    if (value === setCompiler) {
+                        this.handleCompilerQuickPick();
+                    }
+                }
+            } else {
+                this.handleCompilerQuickPick();
+            }
+        } else {
+            vscode.window.showInformationMessage(localize("noCompiler.message", "No compiler detected, please install a compiler"));
```
sean-mcmanus: I think this should end in a period, unless there's a reason not to?

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
...
+            }
+            if (option !== undefined && isCompiler) {
+                const path: string | undefined = paths[i].replace(option, "");
+                items.push({ label: option, description: localize("found.string", "Found at") + path, index: i });
```
sean-mcmanus: This results in `Found at/<path>`, etc. It should be something like `localize("found.string", "Found at {0}", path)`

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -527,6 +528,10 @@ function onResetDatabase(): void {
     clients.ActiveClient.resetDatabase();
 }
 
+function selectDefaultCompiler(): void {
+    clients.ActiveClient.promptSelectCompiler(true);
+}
+
 function onSelectConfiguration(): void {
     if (!isFolderOpen()) {
         vscode.window.showInformationMessage(localize("configuration.select.first", 'Open a folder first to select a configuration'));
```
sean-mcmanus: Can you add a period to this message too, even though it's a pre-existing issue?

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Should this dropdown quickpick show a "trusted" or "untrusted" label? Otherwise, I'm worried users might assume all the compilers in the list are "trustable" when they may not be, i.e. showing "untrusted" may emphasize that choosing it is an indication of their trusting it.

Extension/src/LanguageServer/ui.ts
```diff
@@ -344,6 +344,34 @@ export class UI {
         return (selection) ? selection.index : -1;
     }
 
+    public async showSelectCompiler(paths: string[]): Promise<number> {
+        const options: vscode.QuickPickOptions = {};
+        options.placeHolder = localize("select.compile.commands", "Select a compiler to configure for IntelliSense");
```
sean-mcmanus: Maybe something like `Select a compiler to configure for IntelliSense (that you trust to execute)`? Otherwise, it's not obvious that when it's chosen that our extension will assume it's trustable to execute.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Can/should a "current" label be added for the current default? That would make it easier to see what I'm switching away from using.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,87 @@ export class DefaultClient implements Client {
...
+        paths.push(localize("selectAnotherCompiler.string", "Select another compiler on my machine"));
+        paths.push(localize("installCompiler.string", "Help me install a compiler"));
+        paths.push(localize("noConfig.string", "Do not configure a compiler (not recommended)"));
+        const index: number = await ui.showSelectCompiler(paths);
+        if (index === paths.length - 1) {
+            paths[index] = "";
+        } else if (index === paths.length - 2) {
+            switch (os.platform()) {
+                case 'win32':
+                    vscode.commands.executeCommand('vscode.open', "https://go.microsoft.com/fwlink/?linkid=2217614");
```
sean-mcmanus, line 936: This opens the "Using GCC with MinGW" docs -- can/should we create a new doc page that is specific to installing compilers on Windows, which would include info on how to install the cl.exe compiler, the mingw compiler, the LLVM clang compilers, or maybe whatever other compilers we see high enough usage data for?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -103,6 +102,8 @@ export interface Browse {
 export interface KnownCompiler {
     path: string;
     isC: boolean;
+    isTrusted: boolean;
```
sean-mcmanus: Is this supposed to be used in the TypeScript?

sean-mcmanus: ...same as isCl. I'm not sure why it's being returned if they're not used. It seems like the info could theoretically be useful, but normally we wouldn't add it until we wanted to use it.

Extension/src/LanguageServer/client.ts
browntarik: It can't because it is possible that a trusted compiler has been added to the persistent state elsewhere on startup. If so we want that present in our first call to query default compiler

Extension/src/common.ts
browntarik: Can you give an example?

Extension/src/LanguageServer/ui.ts
browntarik: For now, there is no need to expose the "trusted" aspect of the feature to users. That can be added in a more robust feature implementation later that allows users to adjust that list.

browntarik: We previously did not worry about telling the user what "Default, Default" we fellback too. I am not sure we need to now.

Extension/src/LanguageServer/ui.ts
browntarik: This isn't meant to be exposed as a security feature but rather a configuration feature. From what I understand.

Extension/src/LanguageServer/client.ts
browntarik: This is a question for @AlexandraKemperMS

Extension/src/LanguageServer/configurations.ts
browntarik: Yes, it is, it is necessary to parse the properties that are sent in the json object received on the typescript side.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,87 @@ export class DefaultClient implements Client {
...
+            const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
+            if (result !== undefined && result.length > 0) {
+                util.addTrustedCompiler(result[0].fsPath);
+                settings.defaultCompiler = result[0].fsPath;
+                compilerDefaults = await this.requestCompiler(compilerPath.Value);
+                this.updateClientConfigurations();
+                return;
+            }
+        } else {
+            util.addTrustedCompiler(paths[index]);
```
sean-mcmanus: I'm seeing `null` get added a lot, but I'm not sure how that happened. When the user hits "escape" the index is -1, so the result is `undefined` which also gets added. I also see the empty string get added, not sure how. It seems like some checking should be added to prevent this persisted compiler list from having invalid data.

sean-mcmanus: I'm seeing the persistedCompilerPaths have inconsistently formatted paths, i.e. backslash and forward slash, lower case drive letter and upper case, all lower case letters versus actual case -- does that matter? I haven't checked how those strings are used.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -883,6 +890,9 @@ export class CppProperties {
                             configuration.macFrameworkPath = this.defaultFrameworks;
                         }
                     }
+                } else {
+                    // add compiler to list of trusted compilers
+                    util.addTrustedCompiler(configuration.compilerPath);
```
sean-mcmanus: It looks like it's adding null, and 'Select another compiler on my machine' at this point (and maybe other invalid values). The 'Select another compiler on my machine' case happens if the user hits Escape on the open file dialog box when selecting the compiler.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: But why do the properties exist on the object in the first place?

Extension/src/LanguageServer/ui.ts
sean-mcmanus: I think for V1 we should only show "trusted" compilers then. Otherwise, I think users might blame our extensions for presenting them an untrusted compiler in the list (i.e they just click it because we showed it to them for configuring IntelliSense). We could add untrusted compilers in a follow up PR if we can figure out a better way to present them.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Sure, I'm fine with this wording, but see my other comment (about only showing trusted compilers).

Extension/src/LanguageServer/client.ts
sean-mcmanus: Also, I don't think it's necessary for V1, but it'd be a nice follow up improvement in regards to docs and the compiler acquisition story.

Extension/src/LanguageServer/client.ts
```diff
@@ -987,6 +1076,17 @@ export class DefaultClient implements Client {
                     } else {
                         this.configuration.CompilerDefaults = compilerDefaults;
                     }
+                    // update all client configurations
+                    this.configuration.setupConfigurations();
```
Colengms, line 1109: I had commented here in the context of one of Sean's comments, but it looks like he resolved it.

This may be displaying the prompt too earlier.  If the user is using a `compile_commands.json` file, we will not know about those compilers used until after the configuration has been processed. 

We do process configurations in `setupConfigurations`, immediately before this.  However,   `ChangeCppPropertiesNotification` is a notification, so the TypeScript code does not wait for it to complete before proceeding.  A simple solution might be to change that message from a notification to request and await it, so that the TypeScript code does not proceed until the configurations have been processed and any discovered compilers in `compile_commands.json` reported to the TypeScript side.

Colengms (changes requested):

Extension/src/LanguageServer/ui.ts
browntarik: For now we can address this in a follow up PR as well

Extension/src/LanguageServer/configurations.ts
browntarik: I don't think I understand, are you asking why we keep track of whether a compiler is trusted or not?

Extension/src/LanguageServer/client.ts
browntarik: I don't believe this should matter due to canonicalization

Extension/src/common.ts
michelleangela: Example
```
// no need to check exact string value since there's only one set value?
// if it returns undefined does that mean user dismissed the message?
if (value !== undefined)  
{
    vscode.commands.executeCommand("workbench.action.reloadWindow");
}
```

michelleangela (dismissed):

sean-mcmanus (changes requested):
I'm seeing this:
image

It looks like you need to use the backquote in TypeScript to resolve the variables -- I added a comment at the spot.

Also, I'm seeing that when "compilerPath": "C:\\msys64\\mingw64\\bin\\g++.exe" is set in my c_cpp_properties.json -- I thought we're supposed to trust it if it's set in the compilerPath?

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,88 @@ export class DefaultClient implements Client {
...
+        this.updateClientConfigurations();
+    }
+
+    async promptSelectCompiler(command: boolean): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
+        const settings: OtherSettings = new OtherSettings();
+        if (compilerDefaults.compilerPath !== "") {
+            if (!command) {
+                const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", "The compiler ${compilerDefaults.compilerPath} was found on this computer. Do you want to configure it for IntelliSense?"), confirmCompiler, selectCompiler);
```
sean-mcmanus: You need to use the backtick string to expand the variables.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
bobbrow: In this case since the string is localized, you need to use `{0}` in the string and pass in `compilerDefaults.compilerPath` as the 3rd argument.  Localizable strings must be constant and cannot use backtick substitution.

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
browntarik: We can take a look at this in a future PR fix

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure, that sounds good, but I still see `$C:\msys64\mingw64\bin\gcc.exe`.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,86 @@ export class DefaultClient implements Client {
...
+        this.updateClientConfigurations();
+    }
+
+    async promptSelectCompiler(command: boolean): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
+        const settings: OtherSettings = new OtherSettings();
+        if (compilerDefaults.compilerPath !== "") {
+            if (!command) {
+                const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", "The compiler ${0} was found on this computer. Do you want to configure it for IntelliSense?", compilerDefaults.compilerPath), confirmCompiler, selectCompiler);
```
bobbrow: It should be `{0}`, not `${0}`.  (no dollar sign)

sean-mcmanus: Can "on the computer" be removed? That text seems unnecessary?

sean-mcmanus: "configure it for IntelliSense" is not accurate wording, because it's saying do you want to configure gcc for IntelliSense, but gcc is not in need of configuring, it's IntelliSense that needs to be configured, i.e. it should be something like "configure IntelliSense for it." or "configure IntelliSense for this compiler.", etc.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,86 @@ export class DefaultClient implements Client {
...
+                if (value === confirmCompiler) {
+                    compilerPaths = await util.addTrustedCompiler(compilerPaths, compilerDefaults.compilerPath);
+                    settings.defaultCompiler = compilerDefaults.compilerPath;
+                    compilerDefaults = await this.requestCompiler(compilerPaths);
+                    this.updateClientConfigurations();
+                } else if (value === selectCompiler) {
+                    this.handleCompilerQuickPick();
+                } else {
+                    const setCompiler: string = localize("setCompiler.string", "Set Compiler");
+                    const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have a compiler configured. Unless you set your own configurations, IntelliSense may not be functional."), selectCompiler);
```
sean-mcmanus, line 981: Same here, instead of "You do not have a compiler configured." it should say something like "You do not have IntelliSense configured for a compiler."

sean-mcmanus, line 981: I'm a little confused by what "set your own configurations" is supposed to mean. I don't know what it would be changed to though.

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,86 @@ export class DefaultClient implements Client {
...
+                    if (value === setCompiler) {
+                        this.handleCompilerQuickPick();
+                    }
+                }
+            } else {
+                this.handleCompilerQuickPick();
+            }
+        } else {
+            vscode.window.showInformationMessage(localize("noCompiler.message", "No compiler detected, please install a compiler."));
```
sean-mcmanus: The message says "please install a compiler", but might they already have a compiler but it's just not in the location that we look at by default? Should the string be clarified?

Extension/src/LanguageServer/client.ts
browntarik: Let's do a separate review to clarify text after this PR goes into effect

Extension/src/LanguageServer/client.ts
```diff
@@ -868,6 +877,86 @@ export class DefaultClient implements Client {
...
+        }
+        compilerDefaults = await this.requestCompiler(compilerPaths);
+        this.updateClientConfigurations();
+    }
+
+    async promptSelectCompiler(command: boolean): Promise<void> {
+        const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
+        const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
+        const settings: OtherSettings = new OtherSettings();
+        if (compilerDefaults.compilerPath !== "") {
```
sean-mcmanus, line 969: This "if/else" should be removed, i.e. it shouldn't ever show the "No compiler detected, please install a compiler." Having no compiler detected just means we didn't detect a compiler in the hardcoded spots we decide to look, so the list of specific compilers will be empty , but the other options in the quickpick should still be valid/shown.

sean-mcmanus (changes requested):

Colengms (dismissed):

sean-mcmanus (changes requested):
I'm seeing no prompt appear when no compiler is detected.

Extension/src/LanguageServer/client.ts
sean-mcmanus: It looks like may be we want the "else" still but it should show a prompt that shows the selectCompiler.

sean-mcmanus: Oh, I see you added code for that below, but it's at the wrong spot, since it's hitting the 1st else, i.e. compilerDefaults.compilerPath is ''. It is '' because that is what cpptools is setting it too...not sure if that should be changed to undefined somehow.

Extension/src/LanguageServer/client.ts
```diff
@@ -983,6 +1069,22 @@ export class DefaultClient implements Client {
                     } else {
                         this.configuration.CompilerDefaults = compilerDefaults;
```
sean-mcmanus: I'm seeing compilerDefaults be undefined here with a multi-root workspace, which leads to an undefined dereference in the assignment. The code that assigns to that got moved below.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Also, if the `compilerDefaults.compilerPath` check is bypassed, it ends up showing `The compiler was found on this computer` so something may need to handle that.

Extension/src/LanguageServer/client.ts
```diff
@@ -864,6 +873,90 @@ export class DefaultClient implements Client {
...
+                const defaultClient: DefaultClient = <DefaultClient>client;
+                defaultClient.configuration.CompilerDefaults = compilerDefaults;
+                defaultClient.configuration.handleConfigurationChange();
+            }
+        });
+    }
+
+    public async handleCompilerQuickPick(): Promise<void> {
+        const settings: OtherSettings = new OtherSettings();
+        const paths: string[] = compilerDefaults.knownCompilers.map(function (a: configs.KnownCompiler): string { return a.path; });
```
sean-mcmanus: In the "no compiler found" scenario, it's possible this code could be executed (if other workarounds are done), in which case this is undefined (the value returned from cpptools). It seems like function should be made to handle that so it can still show the 3 other options.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -864,6 +873,90 @@ export class DefaultClient implements Client {
...
+        paths.push(localize("selectAnotherCompiler.string", "Select another compiler on my machine"));
+        paths.push(localize("installCompiler.string", "Help me install a compiler"));
+        paths.push(localize("noConfig.string", "Do not configure a compiler (not recommended)"));
+        const index: number = await ui.showSelectCompiler(paths);
```
sean-mcmanus: This index is -1 if the Escape key is used and it doesn't look like that case is handled -- I see index being used to index into an array below.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -864,6 +873,90 @@ export class DefaultClient implements Client {
...
+        if (index === paths.length - 1) {
+            paths[index] = "";
```
sean-mcmanus: I'm seeing this option (setting compilerPath to "") not have any effect. Are you able to repro that?

sean-mcmanus (changes requested):

caesarpreston (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -864,6 +876,122 @@ export class DefaultClient implements Client {
...
+                this.updateClientConfigurations();
+                return;
+            }
+        } else {
+            if (index !== -1) {
+                util.addTrustedCompiler(compilerPaths, paths[index]);
+            }
+        }
+        // If a compiler is selected, update the default.compilerPath user setting.
+        if (index < paths.length - 3) {
```
sean-mcmanus, line 958: You should check index !== -1 here too.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Actually, can you just check for index === -1 earlier on and exit from the function?

Extension/src/LanguageServer/client.ts
```diff
@@ -883,6 +886,32 @@ export class DefaultClient implements Client {
...
 
+    public async showSelectCompiler(paths: string[]): Promise<number> {
+        const options: vscode.QuickPickOptions = {};
+        options.placeHolder = localize("select.compile.commands", "Select a compiler to configure for IntelliSense");
+
+        const items: IndexableQuickPickItem[] = [];
+        for (let i: number = 0; i < paths.length; i++) {
+            let option: string | undefined;
+            let isCompiler: boolean = false;
+            if (paths[i].indexOf("\\") > 0) {
```
sean-mcmanus: As mentioned in chat, I don't think this backslash path will get resolved on Linux/Mac...not sure if that should block the PR or not.

Extension/src/LanguageServer/client.ts
```diff
@@ -2301,6 +2437,13 @@ export class DefaultClient implements Client {
         return this.requestWhenReady(() => this.languageClient.sendRequest(SwitchHeaderSourceRequest, params));
     }
 
+    public requestCompiler(compilerPath: string[]): Thenable<configs.CompilerDefaults> {
```
sean-mcmanus, line 2447: Seems like compilerPath should be named compilerPaths.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -864,6 +876,122 @@ export class DefaultClient implements Client {
...
+                case 'darwin':
+                    vscode.commands.executeCommand('vscode.open', "https://go.microsoft.com/fwlink/?linkid=2217706");
+                    return;
+                default: // Linux
+                    vscode.commands.executeCommand('vscode.open', "https://go.microsoft.com/fwlink/?linkid=2217615");
+                    return;
+            }
+        } else if (index === paths.length - 3) {
+            const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
+            if (result !== undefined && result.length > 0) {
```
sean-mcmanus, line 947: This else is not handled correctly. It ends up sending an empty/undefined compilerPaths.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -980,8 +1101,22 @@ export class DefaultClient implements Client {
                         }
                         // Listen for messages from the language server.
                         this.registerNotifications();
-                    } else {
+                    }
+                    // update all client configurations
+                    this.configuration.setupConfigurations();
+                    initializedClientCount++;
+                    // count number of clients, once all clients are configured, check for trusted compiler to display notification to user and add a short delay to account for config provider logic to finish
+                    if ((vscode.workspace.workspaceFolders === undefined) || (initializedClientCount >= vscode.workspace.workspaceFolders.length)) {
```
sean-mcmanus, line 1112: This code gets called again if the language server crashes. It seems like we wouldn't want this to be called again in that case, right?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: I'm seeing some infinite client creation occurring after a crash -- it looks like some existing code is creating a client in a deactivated state and it's invalid to make a cpptools request for that client since the cpptools process crashed, so it throws an exception. This only repros with a multi-root workspace.

michelleangela (dismissed):

sean-mcmanus (changes requested):
image

I'm seeing this error when there's no compiler found. Should this be fixed before the checkin?

michelleangela (dismissed):

michelleangela:
> image
> 
> I'm seeing this error when there's no compiler found. Should this be fixed before the checkin?

This can be fixed after check in.

sean-mcmanus (approved):

michelleangela (approved):

PR #10171, opened 2022-11-19
Change engine logging and add natvis diagnostics

WardenGnaw:
This PR adds logging categories to engineLogging and adds in the option for natvisDiagnostics with different categories.

See https://github.com/microsoft/MIEngine/pull/1322

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -286,7 +286,7 @@
         "name": "iar",
         "source": "iar",
         "owner": "cpptools",
-        "fileLocation" : "absolute",
+        "fileLocation": "absolute",
         "pattern": {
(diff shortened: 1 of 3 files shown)


sean-mcmanus (approved):

Colengms (approved):

sean-mcmanus:
@WardenGnaw Which release did you want this to get into? team/vcls/vscode is only targetting 1.14.0 unless we do a separate cherry pick into the insiders branch.

sean-mcmanus:
...well, it sounds like you want us to cherry pick this into 1.13.5.

WardenGnaw:
The underlying engine already has these changes for v1.13.5, these are just changes so users can see the new options if they decide to use the string category version. 

If its easier to ship in v1.14.0, we can check these changes for that version.

Issue #10164, opened 2022-11-17
Creation declaration/definition that targets a new file (or an empty one) has 2 extra newlines if clang-format isn't used


sean-mcmanus:
In test.h, do Create Declaration/Definition on
```cpp
void func();
```
with an empty test.cpp.

Bug: The result has 2 extra newlines:

```cpp
#include "test.h"



void func()
{
}
```

It's possible the root cause issue might result in other bugs, not sure though.

sean-mcmanus:
This got fixed in 1.13.5 (pre-release, release candidate): https://github.com/microsoft/vscode-cpptools/releases/tag/v1.13.5, but only if clang-format is used....I've updated the title to reflect this.

bobbrow:
We’re closing this as not planned.
This issue has been inactive for a long time, and we will be unable to investigate and address it in the foreseeable future.

PR #10177, opened 2022-11-21
Fix selection and formatting for CDD (and extra newlines with clang-format)

### LINKED ISSUE #10160: Create declaration/definition does not select the correct range of new inserted text
Sample code:

header file
```cpp

// other.h
class other {
public:
void abc();
void xyz();
void myFunction();
};

```

source file
```cpp
// other.cpp
#include "other.h"

void other::abc() {

}

void other::xyz() {

}
```

**Repro steps:**
1. Use sample files with sample code.
2. Invoke create declaration / definition command on function `myFunction()`.

**Result:** definition of `myFunction()` is created in file other.cpp. Extension navigates to file other.cpp but does not select and put cursor on the newly created definition of `myFunction()`.

**Expected result:** when navigating to the newly created definition or declaration, extension should select the newly inserted text.

### LINKED ISSUE #10161: Create declaration/definition feature should format newly inserted definition or declaration
1. Create declaration of a function.
2. Invoke create declaration / definition command on function declaration.

**Result:** the created definition is not format.

**Expected result:** the created definition should be formatted according to the format style specified in settings.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/10161, https://github.com/microsoft/vscode-cpptools/issues/10160.

Also https://github.com/microsoft/vscode-cpptools/issues/10164 for clang-format due to a by-product of its extra newline removal).

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -3013,8 +3013,43 @@ export class DefaultClient implements Client {
                 };
                 if (modifiedDocument && lastEdit) {
                     await vscode.workspace.applyEdit(workspaceEdit);
-                    const selectionRange: vscode.Range = lastEdit.range; // TODO: range should be the new range after text edit was applied.
+
+                    // Compute numNewlines for formatRange.
+                    let numNewlines: number = 0;
+                    for (let index: number = 0; index < lastEdit.newText.length; ++index) {
+                        if (lastEdit.newText.charAt(index) === "\n") {
+                            ++numNewlines;
+                        }
+                    }
+
+                    // Move the cursor to the new code, accounting for \n or \n\n at the start.
+                    let selectionRange: vscode.Range = lastEdit.range;
+                    if (lastEdit.newText.startsWith("\r\n\r\n") || lastEdit.newText.startsWith("\n\n")) {
+                        selectionRange = new vscode.Range(new vscode.Position(selectionRange.start.line + 2, 0), new vscode.Position(selectionRange.start.line + 2, 0));
+                        numNewlines -= 2;
+                    } else if (lastEdit.newText.startsWith("\r\n") || lastEdit.newText.startsWith("\n")) {
+                        selectionRange = new vscode.Range(new vscode.Position(selectionRange.start.line + 1, 0), new vscode.Position(selectionRange.start.line + 1, 0));
+                        numNewlines -= 1;
+                    } else {
+                        selectionRange = new vscode.Range(new vscode.Position(selectionRange.start.line, 0), new vscode.Position(selectionRange.start.line, 0));
+                    }
                     await vscode.window.showTextDocument(modifiedDocument, { selection: selectionRange });
+
+                    // Run formatRange.
+                    const formatEdits: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
+                    const formatRange: vscode.Range = new vscode.Range(selectionRange.start, new vscode.Position(selectionRange.start.line + numNewlines, 0));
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -2964,6 +2964,16 @@ export class DefaultClient implements Client {
         this.handleRemoveCodeAnalysisProblems(false, identifiersAndUris);
     }
 
+    private static countNewlines(text: string): number {
+        let numNewlines: number = 0;
+        for (let index: number = 0; index < text.length; ++index) {
```
bobbrow: Or you can do it without a loop:
`return (text.match(/\n/g) || []).length;`

Extension/src/LanguageServer/client.ts
```diff
@@ -2964,6 +2964,16 @@ export class DefaultClient implements Client {
         this.handleRemoveCodeAnalysisProblems(false, identifiersAndUris);
     }
 
+    private static countNewlines(text: string): number {
```
bobbrow: Not sure that this needs to be a static function in a class.  It could just be a regular function outside a class (e.g. a "util" function).

Extension/src/LanguageServer/client.ts
```diff
@@ -3013,8 +3027,34 @@ export class DefaultClient implements Client {
                 };
                 if (modifiedDocument && lastEdit) {
                     await vscode.workspace.applyEdit(workspaceEdit);
-                    const selectionRange: vscode.Range = lastEdit.range; // TODO: range should be the new range after text edit was applied.
+                    let numNewlines: number = DefaultClient.countNewlines(lastEdit.newText);
+
+                    // Move the cursor to the new code, accounting for \n or \n\n at the start.
+                    let selectionRange: vscode.Range = new vscode.Range(new vscode.Position(lastEdit.range.start.line + numNewlinesFromPreviousEdits, 0), new vscode.Position(lastEdit.range.start.line + numNewlinesFromPreviousEdits, 0));
```
bobbrow: Slightly more readable might be:
```ts
let startLine = lastEdit.range.start.line;
if (lastEdit.newText.startsWith("\r\n\r\n") || lastEdit.newText.startsWith("\n\n")) {
    startLine += 2;
    numNewlines -= 2;
} else if (lastEdit.newText.startsWith("\r\n") || lastEdit.newText.startsWith("\n")) {
    startLine += 1;
    numNewlines -= 1;
}
const selectionRange: vscode.Range = new vscode.Range(new vscode.Position(startLine + numNewlinesFromPreviousEdits, 0), new vscode.Position(startLine + numNewlinesFromPreviousEdits, 0));
```

bobbrow (dismissed):
I'm fine with this.  Only cosmetic feedback.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure, that seems good to me.

Extension/src/LanguageServer/client.ts
sean-mcmanus: I'll try that

sean-mcmanus:
I'm still investigating a potential bug...

sean-mcmanus:
@bobbrow There is a regression with selection formatting with 1.13.4 with clang-format 15, which can cause the formatting to fail with CDD. I'm looking into whether it's a clang-format bug or if it's some other change.

bobbrow (dismissed):

sean-mcmanus:
@bobbrow I pushed a minor fix of an issue I hit when using vcFormat.

Supsource (approved):

bobbrow (approved):

Colengms (approved):

michelleangela (approved):

PR #10202, opened 2022-11-30
Add quotes to CDD strings.

sean-mcmanus:
This is to match what VS uses and for consistency with the one other string that uses the quotes.


sean-mcmanus:
I assume this can wait for 1.14.0 unless someone feels we should add it to 1.13.6.

michelleangela (approved):

Issue #10035, opened 2022-10-19
Document formatting causes sometimes messy and invalid code


Delgan:
### Environment

- OS and version: Ubuntu 20.04
- VS Code: 1.71.1
- C/C++ extension: 1.12.4
- OS and version of remote machine (if applicable):  Ubuntu 20.04 (dev container)
- GDB / LLDB version: 10.2
- Clang-format version: 14.0

### Bug Summary and Steps to Reproduce

**Bug Summary:**
From time to time, when saving and thus formatting a C++ file, the file ends up in an invalid "state". Code blocks are mixed together for an unknown reason. Different lines are intertwined with each others. I don't know how to reproduce this mistake systematically. It happens relatively rarely (once a week or so).

[Screencast from 05-10-2022 17_04_58.webm](https://user-images.githubusercontent.com/4193924/196697686-e029ff5d-5550-4e9e-8a8e-d6e302eaa7bd.webm)

**Steps to reproduce:**
1. Enable "Format on save" config
2. Modify the file, write lines, undo, redo, etc.
3. Try to save (`ctrl+S`)
4. Observe that file is messed up
5. Undo (`ctrl+Z`)
6. Save again
7. Observe that file is still messed up

### Debugger Configurations

```shell
N/A
```

### Debugger Logs

```shell
N/A
```

### Other Extensions

Possibly related extensions:
* akiramiyakoda.cppincludeguard
* psioniq.psi-header
* cschlosser.doxdocgen
* twxs.cmake
* ms-vscode.cmake-tools

### Additional Information

This problem happened to myself and a colleague of mine on two different computers, so it's not unique to my setup. However, we're sharing the same codebase and the same devcontainer, so it might be related to something in our dev environment.

If I remember correctly, calling the "Format document" command without saving also generates the error. The bug appeared about 2 months ago I would say.

It's not related to the content of the file. If I close VS Code and re-open it, then it works.

It is not easy to reproduce so unfortunately I do not have any additional information to provide. I'll try to update the ticket with logs when it reproduces.

Delgan:
Ok, "great" news: the bug happened again while I disabled almost all my others extensions (`ms-vscode.cpptools` and `ms-vscode.cmake-tools`).
I'm thus pretty confident the bug comes from `vscode-cpptools`.


https://user-images.githubusercontent.com/4193924/196921777-da6f63f5-feb6-467f-84a7-ec56a6fc152d.mp4

<details>
[...]

PR #10227, opened 2022-12-06
Fix document corruption due to delayed evaluation in protocolFilter

Colengms:
This should address https://github.com/microsoft/vscode-cpptools/issues/10035 as well as some document corruption issues related to CDD.

The issue arose in protocolFilter.ts due to use of deferred lambdas.  That caused some TypeScript objects to be re-evaluated later, when the message was actually sent, instead of within the protocolFilter handler at the time the message was intended to be processed.  For example, when processing a `didChange` message, the `TextDocument` has a particular document version number, but we would get a more recent version number when `sendMessage` was actually called after deferred through `notifyWhenLanguageClientReady`.  The mismatching version numbers resulted in incorrect management of the contents of the buffer cache in the native process.

It doesn't seem necessary to defer using `notifyWhenLanguageClientReady`, as the protocolFilter would not be processing messages if it were not ready to.

I also removed the call to `processDelayedDidOpen` from `didChange`, as it could attempt to defer as well.  This was a failsafe anyway.  We should not be getting calls to `didChange` without having previously processed a `didOpen` (delayed or otherwise).  If we did, that would seem to indicate a more serious issue elsewhere.

Alternatively, this perhaps could have been addressed using `await` with the protocol handler, as those can now be async.  However, I expect that could be vulnerable to deadlock.  For example: If it were to try to wait for message B before allowing delivery of message A to complete, that would deadlock the queue.  Instead, the logic within protocolFilter should be careful to ensure that `sendMessage` occurs in a synchronous manner before returning.

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -58,6 +58,7 @@
 * Fix clang-tidy and clang-format not working on CentOS7 and other Linux OS's without glibc 2.27 or greater. [#10019](https://github.com/microsoft/vscode-cpptools/issues/10019)
 * Fix various bugs with the `C_Cpp.codeAnalysis.clangTidy.headerFilter` setting. [#10023](https://github.com/microsoft/vscode-cpptools/issues/10023)
 * Fix Doxygen comment generation when there's a selection. [#10028](https://github.com/microsoft/vscode-cpptools/issues/10028)
+* Fix issue that could cause document corruption. [#10035](https://github.com/microsoft/vscode-cpptools/issues/10035)
 * Fixed crash on Linux/Mac when a full command line is specified in `compilerPath` containing invalid arguments. [PR #10070](https://github.com/microsoft/vscode-cpptools/pull/10070)
 * Fix random "Failed to spawn IntelliSense process: 65520" on Mac. [#10091](https://github.com/microsoft/vscode-cpptools/issues/10091)
 * Fix "Don't hardcode path to kill in UnixUtilities". [#10124](https://github.com/microsoft/vscode-cpptools/issues/10124)
--- a/Extension/src/LanguageServer/protocolFilter.ts
+++ b/Extension/src/LanguageServer/protocolFilter.ts
@@ -9,7 +9,7 @@ import { Middleware } from 'vscode-languageclient';
 import { Client } from './client';
 import * as vscode from 'vscode';
 import { CppSettings } from './settings';
-import { clients, onDidChangeActiveTextEditor, processDelayedDidOpen } from './extension';
+import { clients, onDidChangeActiveTextEditor } from './extension';
 
 export function createProtocolFilter(): Middleware {
     // Disabling lint for invoke handlers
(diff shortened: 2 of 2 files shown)


sean-mcmanus (approved):

michelleangela (approved):

bobbrow:
`notifyWhenLanguageClientReady` has another purpose which is to serialize the messages so that the order is preserved.  If the problem is that the document changes between calls, then I would think saving the old values would be the safer option.

bobbrow:
In other words, I'm concerned that removing the calls to `notifyWhenLanguageClientReady` will cause different ordering problems with the messages.

bobbrow:
I'm trying to reconcile this with the code that was retained.  All of the other messages still use the `notify` queue, so it seems that the actual issue is that we need to clone the object, not that we need to remove the `notify` call.
```ts
    const defaultHandler: (data: any, callback: (data: any) => Promise<void>) => Promise<void> = async (data, callback: (data: any) => void) => { clients.ActiveClient.notifyWhenLanguageClientReady(() => callback(data)); };
    // const invoke1 = async (a: any, next: (a: any) => any) => { await clients.ActiveClient.awaitUntilLanguageClientReady(); return next(a); };
    const invoke2 = async (a: any, b: any, next: (a: any, b: any) => any) => { await clients.ActiveClient.awaitUntilLanguageClientReady(); return next(a, b); };
    const invoke3 = async (a: any, b: any, c: any, next: (a: any, b: any, c: any) => any) => { await clients.ActiveClient.awaitUntilLanguageClientReady(); return next(a, b, c); };
    const invoke4 = async (a: any, b: any, c: any, d: any, next: (a: any, b: any, c: any, d: any) => any) => { await clients.ActiveClient.awaitUntilLanguageClientReady(); return next(a, b, c, d); };
```

Colengms:
I actually overlooked that potential deferral due to `awaitUntilLanguageClientReady`.  I don't believe they are needed or desirable.  protocolFilter should already be serialized.  We know the language service is ready to process messages, as the protocolFilter is invoked as part of that pipeline.   Because those calls are 'awaited', they would not cause messages to be processed out of order.  I'm not sure whether they are vulnerable to the 'delayed evaluation' issue.  Perhaps they are either always completing immediately or represent additional potential repros of the late evaluation bug.  (We _definitely_ saw that the root cause of the issue was that the version number was too recent in some `didChange` messages, and that issue was addressed by longer allowing `sendMessage` to be deferred (without awaiting it).

I don't believe anything we are doing within the protocol filter justifies delaying allowing those messages to continue to be processed immediately.  Can you identify any specific requirement that protocolFilter to delay processing a message until something in particular in the task queue has completed?

bobbrow:
It has nothing to do with the language client being ready, but everything to do with the call to `queueTask` within the `notifyWhenLanguageClientReady`, `awaitUntilLanguageClientReady`, etc calls.  All messages sent to the language server are serialized through this queue to retain their order.  My fear is that removing the call opens the door to processing messages out of the intended order.

It didn't sound like you need to do a deep copy of all message params.  Don't we just need to save the document version to check whether the message should be discarded or not?

PR #10296, opened 2022-12-17
Address issues with request cancelation

Colengms:
Address some issues with unhandled cancelations potentially resulting in exceptions.

Removed some unnecessary uses of `requestWhenReady`, which I'd like to deprecate (but there are a few uses I can't remove yet).

Some of our requests were returning arrays.  I've changed those to return an object containing the array.  This is in order to support generic LSP request cancelation, which involves returning an object containing only an error code (`ServerCancelled` or `RequestCancelled`) and error message.  That's useful to allow cancellation to be handled without knowledge of the type it returns, such as if cancellation occured before the message could be otherwise processed.  In those cases, I'm using the fact that the array is undefined (instead of empty) to detect if the request was cancelled.

This depends on native-side changes that are also in code review.


michelleangela (dismissed):

sean-mcmanus:
@Colengms The diff is fixed if you use the GitHub option to stop showing whitespace.

sean-mcmanus (dismissed):

Colengms:
> > I confirmed that the additional uses of `CancellationError` appear to be handled by VS Code and do not result is exceptions being logged.
> 
> Is this a contract of the API?

They originally added CancellationError due to our users complaining about an error in their log when we would trigger a fresh request by cancelling the previous request (in the way they advised).  That added CancellationError to address that.

https://github.com/microsoft/vscode-cpptools/issues/5487

That issue was not scoped to a specific provider.  That was [added to the API docs](https://code.visualstudio.com/api/references/vscode-api#CancellationError), and there was no clear indication which operations it applied to, seeming to imply that it applies to any operations to which they pass us a CancellationToken.  But, when it didn't work for folding ranges, I opened: https://github.com/microsoft/vscode/issues/125799   Though, using it for FoldingRanges no longer triggers the error that issue was opened about.

Searching their codebase, they now appear to be using it for inlayHints, and lots of 'common' places.  I tested each scenario and only used CancellationToken when it did not result in any error being logged.  I could switch it out for everything other than semanticTokens and inlayHints, if you like.  But the short answer to your question is that they don't appear to have documented which operations support it.

bobbrow:
Based on that documentation you linked, it sounds like we should be using `CancellationError` everywhere that tokens can be cancelled and that we should open bugs for any places where VS Code doesn't handle it properly.

Colengms:
> Based on that documentation you linked, it sounds like we should be using `CancellationError` everywhere that tokens can be cancelled and that we should open bugs for any places where VS Code doesn't handle it properly.

I opened: https://github.com/microsoft/vscode/issues/169698

sean-mcmanus (changes requested):
Why is yarn.lock changed but not package.json? i.e. I think you should update package.json, right?

michelleangela (dismissed):

sean-mcmanus (approved):

sean-mcmanus (approved):

sean-mcmanus:
@Colengms Is this ready to check in now?

PR #10306, opened 2022-12-22
Await in protocol filter

### LINKED ISSUE #1146: LSP `Middleware` may late-evaluate parameters in async handlers, causing incorrect results
VS Code version: Code 1.74.0
OS Version: All
vscode-languageclient: 8.0.1

In the C/C++ extension, we maintain copies of open file buffers in our native LSP process, for various reasons.  We keep those buffers up to date by processing `didOpen`, `didChange`, etc..  Also, in TypeScript, we use a `Middleware` instance to filter these messages (and ensure they are processed serially with regards to other messages we have queued).

The issue is most obvious with the `didChange` message.  The `Middleware` would appear to support async handlers.  The return type of the `didChange` handler is a `Promise<void>`.  In our `Middleware`, we may `await` another function before calling `sendMessage`.  We pass `sendMessage` the original parameters that were passed to us.  We are seeing that when our `didChange` handler cannot complete immediately, the call to `sendMessage` will effectively re-evaluate the originally provided `TextDocumentChangeEvent`, such that it now contains a more recent document version than was associated with the specified `contentChanges`.

```
        didChange: async (textDocumentChangeEvent, sendMessage) => {
            await somethingThatCantCompleteImmediately(); // If this doesn't complete immediately, and an edit races with us,
            await sendMessage(textDocumentChangeEvent);   // this will send an incorrect document version.
        },
```

This mismatch between the change/edit data and the document version results in corruption of our internal buffer, as we are using the version value to ensure we do not apply stale edits, for example.

If the `Middleware` allows async handlers, VS Code should ensure that the parameters passed to them will not change out from under them before they complete.

Deep-copying the parameters might be a work-around, but some are complex, and we don't necessarily know what `sendMessage` does with the parameters we pass to it.  Another work-around might be for us to remove our calls to `sendMessage` and create our own custom messages, passing them only the parameters we care about, pinned to their original values.

I can create a simplified repo if needed.  I figured I would get this issue opened first.  I could potentially provide a fork of cpptools, with some minor modifications to restore the above pattern, if the above info does not make the issue sufficiently obvious.

Colengms:
Based on discussion related to: https://github.com/microsoft/vscode-cpptools/pull/10257

This is a more complete fix for: https://github.com/microsoft/vscode-cpptools/issues/10035 
Previously worked around by: https://github.com/microsoft/vscode-cpptools/pull/10227

This fix is now possible due to an underlying issue in VS Code (actually vscode-languageclient) getting fixed: https://github.com/microsoft/vscode-languageserver-node/issues/1146

Basically, all filters in `protocolFilter.ts` are now async, and use `await` to ensure filtered LSP messages are synchronized with other language client operations.


Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -72,47 +69,51 @@ export function createProtocolFilter(): Middleware {
             }
         },
         didChange: async (textDocumentChangeEvent, sendMessage) => {
+            await clients.ActiveClient.awaitUntilLanguageClientReady();
             const me: Client = clients.getClientFor(textDocumentChangeEvent.document.uri);
-            me.onDidChangeTextDocument(textDocumentChangeEvent);
-            sendMessage(textDocumentChangeEvent);
+            if (me.TrackedDocuments.has(textDocumentChangeEvent.document)) {
```
bobbrow: Shouldn't every file be tracked already?  It seems dangerous to ever skip processing a didChange. Should this be an assert instead?

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -72,47 +69,51 @@ export function createProtocolFilter(): Middleware {
...
+                me.onDidChangeTextDocument(textDocumentChangeEvent);
+                await sendMessage(textDocumentChangeEvent);
+            }
         },
-        willSave: defaultHandler,
+        willSave: invoke1,
         willSaveWaitUntil: async (event, sendMessage) => {
```
bobbrow, line 70: If this handler doesn't use `await` it should not be marked `async`

Colengms: I'm not sure that's necessarily the right rule to apply.  The use of `async` here causes the return type of the lambda to automatically be treated as a `Thenable<TextEdit[]>` without having to explicitly specify that as a return type.  Normally, I would simply use `await` on the call there to `sendMessage`, but that results in a lint warning that the use of `await` is redundant when used in `return await ...`.  If it weren't for the lint error, I would agree.  But, since that occurs, this appears to be the cleanest approach to me.

bobbrow: fair enough.

Extension/src/LanguageServer/protocolFilter.ts
Colengms: We may defer 'didOpen'.  We may not consider the file opened until a `onDidChangeVisibleTextEditors` fires (or, apparently, when we receive a document symbol request for it).  Likely due to asyncs, we may receive a didChange before we finish processing a delayed didOpen.  That's perfectly fine, as we deliver the current state of the file to the native side when we do, and track changes from that point on.  This check avoids sending unnecessary didChange messages to the native side before we've gotten around to considering the file open.  Previously, these would not get discarded until a similar check occurred on the native side, after receiving them.  There is a comment to this effect in the native code, in `textDocument_didChange`, where it exits early when `get_document` does not find the document.

Colengms: That said, I removed it, just in case...

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -69,47 +61,49 @@ export function createProtocolFilter(): Middleware {
...
+            // await clients.ActiveClient.awaitUntilLanguageClientReady();
+            // Don't use awaitUntilLanguageClientReady.
+            // Otherwise, the message can be delayed too long.
             const me: Client = clients.getClientFor(event.document.uri);
             if (me.TrackedDocuments.has(event.document)) {
-                // Don't use me.requestWhenReady or notifyWhenLanguageClientReady;
-                // otherwise, the message can be delayed too long.
                 return sendMessage(event);
```
sean-mcmanus, line 76: Should this be `await sendMesssage(event)`. There's a known bug in which VS Code is not waiting for willSaveWaitUntil (https://github.com/microsoft/vscode/issues/164841).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/protocolFilter.ts
Colengms: `await` is implied here, as the return value of the async lambda is a promise.  It would generate a lint warning/error if `await` is added, as it's considered superfluous.  (I had an `await` here and had to remove it.)

Based on the comment in the code, I left this message unchanged - it doesn't await the language client (and therefore anything else queued waiting for it), so this will still be delivered out of order (quickly).

sean-mcmanus (approved):

michelleangela (approved):

PR #10419, opened 2023-01-25
add second prompt for when a compiler needs config

browntarik:


Extension/src/LanguageServer/client.ts
```diff
@@ -962,35 +975,35 @@ export class DefaultClient implements Client {
...
                     settings.defaultCompiler = compilerDefaults.compilerPath;
                     compilerDefaults = await this.requestCompiler(compilerPaths);
                     this.updateClientConfigurations();
                 } else if (value === selectCompiler) {
-                    this.handleCompilerQuickPick();
+                    this.handleCompilerQuickPick(true);
                 } else {
                     const setCompiler: string = localize("setCompiler.string", "Set Compiler");
-                    const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have a compiler configured. Unless you set your own configurations, IntelliSense may not be functional."), selectCompiler);
+                    const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have a compiler configured. Unless you set your own configurations, IntelliSense may not be functional."), setCompiler);
```
sean-mcmanus: Why did this change from selectCompiler to setCompiler?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I'd prefer they'd all say "Select Compiler", unless we think of a reason to say something different.

Extension/src/LanguageServer/client.ts
```diff
@@ -926,10 +926,23 @@ export class DefaultClient implements Client {
         paths.push(localize("noConfig.string", "Do not configure a compiler (not recommended)"));
         const index: number = await this.showSelectCompiler(paths);
         if (index === -1) {
+            settings.defaultCompiler = "";
+            if (showSecondPrompt) {
+                const setCompiler: string = localize("setCompiler.string", "Set Compiler");
+                const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have a compiler configured. Unless you set your own configurations, IntelliSense may not be functional."), setCompiler);
```
michelleangela: There's a repeat use of this string. For whichever string that gets used for all these 3 cases, could you try saving the string's value into a global variable and reference that variable instead of having several copies of the string?

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: @browntarik It seems like the entire showInformationMessage itself could be made into a function for the cases when the messages are the same.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -915,21 +915,34 @@ export class DefaultClient implements Client {
         return (selection) ? selection.index : -1;
     }
 
-    public async handleCompilerQuickPick(): Promise<void> {
+    public async showPrompt(promptMessage: string): Promise<void> {
+        const value: string | undefined = await vscode.window.showInformationMessage(localize("setCompiler.message", "You do not have IntelliSense configured. Unless you set your own configurations, IntelliSense may not be functional."), promptMessage);
+        if (value === promptMessage) {
```
bobbrow: Is there ever a case when this is true?  It seems odd that we would localize the message twice (once here and once somewhere else and pass it in).

bobbrow: All the cases I see only pass in "Select Compiler" which won't match `value` (and `value` is never displayed in a prompt anywhere as far as I can tell)

browntarik: Perhaps I could be using the API wrong but from what I understood you want the value returned from showInformationMessage to be equal to the localized string that is passed as the second argument. In which case "value == promptMessage" will be true if the user selects the "Select Compiler" option

bobbrow: I think I was confused because of the name of the variables.  I expected `promptMessage` to be the localized thing for `setCompiler.message`.  Maybe rename `promptMessage` to `buttonMessage`.

sean-mcmanus (changes requested):
I get 2 duplicate Set Compiler ("You do not have IntelliSense configured." prompts if no compiler is found and the 1st prompt results in no compiler selection (previously it shows an incorrect message). For the 0 compiler case, showing the 1st prompt doesn't seem valuable, right?

Extension/src/LanguageServer/client.ts
```diff
@@ -962,35 +975,27 @@ export class DefaultClient implements Client {
...
 
-    async promptSelectCompiler(command: boolean): Promise<void> {
+    async promptSelectCompiler(isCommand: boolean): Promise<void> {
         const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
         const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
         const settings: OtherSettings = new OtherSettings();
         if (compilerDefaults.compilerPath !== "") {
-            if (!command && (compilerDefaults.compilerPath !== undefined)) {
+            if (!isCommand && (compilerDefaults.compilerPath !== undefined)) {
                 const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", "The compiler {0} was found on this computer. Do you want to configure it for IntelliSense?", compilerDefaults.compilerPath), confirmCompiler, selectCompiler);
```
sean-mcmanus: Can you remove "on this computer" -- it's extra text the user shouldn't need to read.

sean-mcmanus: Also, all references to text like "configure it for IntelliSense?" should be changed to something like "configure IntelliSense for the compiler?"

sean-mcmanus (changes requested):
Also, after the 2nd prompt (with 1+ possibly compilers found) but none selected as a trusted default, it will now infinitely prompt with the 2nd prompt when Escape is used -- my understanding was that it wasn't supposed to appear  >2 times (i.e. the exact same prompt shouldn't appear more than once).
Repro: Click "Select Compiler", hit Escape, repeat...

Also, using the last (compilerPath = "") option keeps showing the prompt repeatedly (as mentioned in my other comment).

sean-mcmanus (changes requested):
If the user sets `"C_Cpp.default.compilerPath": ""` the prompts still appear -- I don't think it's supposed to, i.e. there are user scenarios where users need to set that and they're configuring system defines/includes manually.

Extension/src/LanguageServer/client.ts
bobbrow: "Do you want to configure IntelliSense with this compiler?"

sean-mcmanus: Yeah, that sounds good to me.

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: @tarik-brown What was the resolution for this?

browntarik: check latest changes

sean-mcmanus (changes requested):
If no compiler is found, it still gives an identical duplicate prompt if Select Compiler is used, then escape is used in the quick pick.

sean-mcmanus:
> If the user sets `"C_Cpp.default.compilerPath": ""` the prompts still appear -- I don't think it's supposed to, i.e. there are user scenarios where users need to set that and they're configuring system defines/includes manually.

I don't see this as fixed yet.

Extension/src/LanguageServer/client.ts
```diff
@@ -1133,6 +1141,7 @@ export class DefaultClient implements Client {
                                 defaultClient.configuration.handleConfigurationChange();
                             }
                         });
+                        compilerDefaults.trustedCompilerFound = false;
```
sean-mcmanus: This should be removed -- I'm guessing it was added for testing.

sean-mcmanus (changes requested):

michelleangela:
Please add in PR description of how the fix was tested. Include steps (aka repro steps) for others to help verify.

sean-mcmanus:
> > If the user sets `"C_Cpp.default.compilerPath": ""` the prompts still appear -- I don't think it's supposed to, i.e. there are user scenarios where users need to set that and they're configuring system defines/includes manually.
> 
> I don't see this as fixed yet.

I fixed it.

sean-mcmanus:
> Please add in PR description of how the fix was tested. Include steps (aka repro steps) for others to help verify.

The testing I did is explained in the email I sent; however, I plan to do more testing tomorrow, since I was continually finding/fixing bugs today so my latest fixes might have broken something that I didn't re-test.

sean-mcmanus:
> If no compiler is found, it still gives an identical duplicate prompt if Select Compiler is used, then escape is used in the quick pick.

I fixed this.

sean-mcmanus (approved):
I'm approving, but it could use more review/testing.

bobbrow (approved):
Please commit the change and let's build the VSIX

michelleangela (approved):

PR #10442, opened 2023-01-31
Fix issues with didOpen synchronization

Colengms:
Addresses issues with certain messages (semantic token, inlay hints, and code folding) being delivered before we've processed `didOpen`.

Adds an implicit didOpen for messages that require didOpen to have already occurred.

Updates delayed didOpen processing to use async/await consistently, so processing of didOpen could be more reliably synchronized with.

Always passes `processDelayedDidOpen` through `requestWhenReady` to avoid the potential race conditions that could occur if multiple callers (i.e. providers) tried to call into it at essentially the same time.

Uses `requestWhenReady` in protocolFilter to ensure all async handling in each message handler is included in what is being sequenced.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -344,7 +344,7 @@ export async function processDelayedDidOpen(document: vscode.TextDocument): Prom
                 }
                 await client.provideCustomConfiguration(document.uri, undefined);
                 client.onDidOpenTextDocument(document);
-                await client.takeOwnership(document);
+                await client.sendDidOpen(document);
                 return true;
             }
         }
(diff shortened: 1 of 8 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -324,37 +324,27 @@ function onDidChangeTextEditorSelection(event: vscode.TextEditorSelectionChangeE
     clients.ActiveClient.selectionChanged(makeCpptoolsRange(event.selections[0]));
 }
 
-export function processDelayedDidOpen(document: vscode.TextDocument): boolean {
+export async function processDelayedDidOpen(document: vscode.TextDocument): Promise<boolean> {
     const client: Client = clients.getClientFor(document.uri);
     if (client) {
         // Log warm start.
+        clients.timeTelemetryCollector.setDidOpenTime(document.uri);
```
bobbrow: This is called on line 335. Why is it called twice now?

sean-mcmanus (approved):

michelleangela (approved):

Issue #7471, opened 2021-05-06
-std=c2x support for C++-style attributes in C-code.


Iipal:
Bug type: Language Service

- OS and Version: Ubuntu 20.04
- VS Code Version: 1.56
- C/C++ Extension Version: 1.3.1

**Expected behavior**

No errors?
inside union `allocator_t`->`data` appears as an unknown data type (2nd screenshot)

**Code sample and logs**
- Code sample && Configurations in `c_cpp_properties.json`: on the screenshot below
- Logs from running `C/C++: Log Diagnostics` from the VS Code command palette: shows nothing

![image](https://user-images.githubusercontent.com/31858171/117289895-d43b6100-ae75-11eb-948c-e5d34ee35cff.png)
![image](https://user-images.githubusercontent.com/31858171/117290536-9854cb80-ae76-11eb-9957-0502ae86211b.png)
> Theme: Darcula with JetBrains Mono font.

Iipal:
even `__auto_type` behavior is strange for `linux-clang-x64` mode even if `__is_identifier` says it's valid;

if I change the mode to `linux-gcc-x64` as well as a compiler to `gcc` - then everything works fine. But why in clang-mode it's said that it's a valid identifier but then I see errors?
![image](https://user-images.githubusercontent.com/31858171/117298386-fd60ef00-ae7f-11eb-9ffb-d9b7bc4d8133.png)

sean-mcmanus:
Can you provide a code sample that compiles with C but gives an error with our extension? I'm seeing gcc and clang both fail to compile 
```
typedef union {
    int i;
} allocator_t [[gnu::transparent_union]];

int main()
{
    return 0;
}
```

error: expected expression
} allocator_t [[gnu::transparent_union]];

sean-mcmanus:
Oh -- I got it to compile with -std=c2x -- we don't support that std version yet. We can use this issue to track adding that.

sean-mcmanus:
I've filed a bug against our shared VS parser at https://developercommunity.visualstudio.com/t/C-standard-2x-isnt-supported-by-Intelli/1416530 . After they enable that we can add support on our side.

TDHolmes:
I hope this lands soon. My project has started adopting these features and causes syntax errors throughout the project (specifically sized enums)

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.14.4

PR #10456, opened 2023-02-02
Accepting c2x compilers

Bernardin-MS:
Issue: #7471

There is a corresponding PR on the native side


Extension/c_cpp_properties.schema.json
```diff
@@ -36,10 +36,14 @@
                             "c99",
                             "c11",
                             "c17",
+                            "c20",
```
bobbrow: I'm not sure if c20 is going to be a thing.  There was ambiguity between c17 and c18 at one point as well and we settled on just keeping one of them.  I "think" we just need c23 in the schema.

michelleangela:
The html file for the IntelliSense Configuration UI (run command `C/C++: Edit Configurations (UI)` to see the UI)  will need to updating too.

https://github.com/microsoft/vscode-cpptools/blob/main/Extension/ui/settings.html

image

Extension/c_cpp_properties.schema.json
Colengms: There is no c20 or gnu20 defined for gcc https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html#C-Dialect-Options  and presumably clang as well, as it copies it's std's from gcc https://clang.llvm.org/docs/UsersManual.html  ("The supported modes for C are c89, gnu89, c94, c99, gnu99, c11, gnu11, c17, gnu17, c2x, gnu2x, and various aliases for those modes.")

michelleangela: Where did `c20` come from? @Bernardin-MS Is there a source that refers to `c20`?

Bernardin-MS: @michelleangela it was a placeholder based off a reference PR

@Colengms so c2x is the only argument we track, right?

Colengms: @Bernardin-MS I'm not sure I understand your question.  "c23" and "gnu23" are valid.  But, "c20" and "gnu20" aren't valid, so could you remove those?  :)

Extension/ui/settings.html
```diff
@@ -581,10 +581,12 @@
         <div class="section-text" data-loc-id="c.standard.description">The version of the C language standard to use for IntelliSense. Note: GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense will emulate the equivalent C standard version.</div>
         <div>
             <select name="inputValue" id="cStandard" class="select-default">
+                <option value="c2x">c2x</option>
```
michelleangela: Should this be `c23`? Is there a reason this is `c2x` and not consistent with the other settings? Note: this file is the UI for the c_cpp_properties.json. The values in both the UI and .json for IntelliSense configuration should be the same.

Same comment for `gnu2x`

michelleangela (changes requested):

Extension/ui/settings.html
Bernardin-MS: thx, typo

michelleangela (approved):

sean-mcmanus (approved):

Colengms (approved):

sean-mcmanus (approved):

sean-mcmanus (changes requested):
This shouldn't get checked in until the C++ side is ready.

bobbrow:
We're holding this one for 1.15

Dkhack (approved):

sean-mcmanus (approved):

Issue #10467, opened 2023-02-03
Custom configurations should support both URI and file paths in `SourceFileConfigurationItem.item` for backwards compatibility


Colengms:
An issue has arisen with the WAVE extension, which we've tracked down to receiving a file path in the `uri` field of `SourceFileConfigurationItem`.

Bob recalls this being supported.  I've not yet been able to find a code change that would have altered this behavior, and a version of cpptools that introduced an issue hasn't been identified.  But it should be straight forward to ensure file paths are supported in this field, which would proactively address the issues WAVE extension users are experiencing.

PR #10470, opened 2023-02-03
Support use of file paths in uri field of custom configuration

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10467

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2763,7 +2763,7 @@ export class DefaultClient implements Client {
                     }
                 }
                 sanitized.push({
-                    uri: uri,
+                    uri,
                     configuration: itemConfig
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -2756,7 +2763,7 @@ export class DefaultClient implements Client {
                     }
                 }
                 sanitized.push({
-                    uri: item.uri.toString(),
+                    uri: uri,
```
bobbrow: If the variable has the same name as the field, you can just put `uri` here instead of `uri: uri`

bobbrow (dismissed):

sean-mcmanus (dismissed):

sean-mcmanus (approved):

PR #10472, opened 2023-02-04
Fix localized strings

michelleangela:
- Fix hint text.
- Fix create declaration/definition failed string format.
- Lower case "comment" in code action command "Generate Doxygen comment" as the other commands are sentence case.

Example of updated string.
image

### DIFF
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -404,7 +404,7 @@
     "refactor_create_declaration_definition_failed":
     {
         "text": "Create Declaration / Definition failed: %s",
-        "hint": "The operation 'Create Declaration / Definition' on a function was not successful. %s is the info why it failed."
+        "hint": "The operation 'Create Declaration / Definition' on a function was not successful. %s is the error info that has a period at the end of the string."
     },
     "refactor_create_default_delete":
     {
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

Extension/src/nativeStrings.json
```diff
@@ -403,8 +403,8 @@
     },
     "refactor_create_declaration_definition_failed":
     {
-        "text": "Create Declaration / Definition failed:",
-        "hint": "The operation 'Create Declaration / Definition' on a function was not successfully."
+        "text": "Create Declaration / Definition failed: %s",
+        "hint": "The operation 'Create Declaration / Definition' on a function was not successful. %s is the info why it failed."
```
sean-mcmanus: Should you mention that %s ends with a period in case a translator incorrectly adds another one?

Another alternative would be to revert that change and it keep it ending with ":". I didn't realize the %s ended in a period previously.

sean-mcmanus (dismissed):

Extension/src/nativeStrings.json
michelleangela: By having the %s it shows additional info about the failure is expected by the caller of the string "Create Declaration / Definition failed:".

sean-mcmanus (approved):

Issue #10406, opened 2023-01-20
[Fortran] debugger only shows the first 1000 elements of arrays


rawezh1:
### Environment

- OS and version: Windows 10 Version 10.0.19045
- VS Code: 1.74.3
- C/C++ extension: 1.13.9
- GDB / LLDB version: mingw-w64 12.1

### Bug Summary and Steps to Reproduce

Hello,
I am trying to debug a Fortran program using cppdbg but vscode only shows 1000 elements. I am able to see the other elements by manually writing expressions in the watch window for each section but the size of the arrays are dynamic, so this is not a viable solution. 
It also seems like that this was an old problem that was fixed in a previous version (1.13.6) but I do not know why it happens in my case. 

### Debugger Configurations

```shell
launch.json:
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Fdebugger Launch",
            "type": "cppdbg",
            "request": "launch",
            "program": "${workspaceFolder}/myapp",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${workspaceFolder}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "preLaunchTask": "build",
            "miDebuggerPath": "C:\\msys64\\mingw64\\bin\\gdb.exe",
            "setupCommands": [
                {
                    "text": "-enable-pretty-printing",
                    "description": "Enable GDB pretty printing",
                    "ignoreFailures": true
                },
                {
                    "description": "Set Disassembly Flavor to Intel",
                    "text": "-gdb-set disassembly-flavor intel",
                    "ignoreFailures": true
                }
            ]
        },
    ]
}
tasks.json:
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "label": "build",
            "type": "shell",
            "command": "make -f Makefile",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "presentation": {
                "reveal": "silent",
                "panel": "shared"
            },
            "problemMatcher": []
        }
    ]
}
Makefile:
########################################################################
####################### Makefile Template ##############################
########################################################################

# Compiler settings - Can be customized.
CC = gfortran
[...]

PR #10476, opened 2023-02-06
Add Ada to the list of supported languages (#10475)

### LINKED ISSUE #10475: Add Ada to supported languages for debugging
### Feature Request

Hello everyone!

I suggest to add Ada to the list of supported languages for debugging. Ada is debuggable through GDB, and modifying the extension's `package.json` file to add Ada makes everything work out of the box.

This would be very useful for us since we are developing our own VS Code extension for Ada (https://github.com/AdaCore/ada_language_server): the idea would be to bind `vscode-cpptools` as an `extensionPack` to allow users to easily debug their Ada code, with a very minimal JSON configuration.

AnthonyLeonardoGracio:
To allow debugging Ada projects with GDB. Fixes https://github.com/microsoft/vscode-cpptools/issues/10475

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3146,6 +3146,7 @@
         "type": "cppdbg",
         "label": "C++ (GDB/LLDB)",
         "languages": [
+          "ada",
           "c",
           "cpp",
           "cuda-cpp",
@@ -4893,6 +4894,7 @@
         "label": "C++ (Windows)",
         "when": "workspacePlatform == windows",
         "languages": [
+          "ada",
           "c",
           "cpp",
           "cuda-cpp",
@@ -5289,6 +5291,9 @@
       }
     ],
     "breakpoints": [
+      {
+        "language": "ada"
+      },
       {
         "language": "c"
       },


Extension/package.json
```diff
@@ -3146,6 +3146,7 @@
         "type": "cppdbg",
         "label": "C++ (GDB/LLDB)",
         "languages": [
+          "ada",
```
sean-mcmanus, line 3149: Can you move this below "rust"? Unless the order is supposed to be alphabetical?

sean-mcmanus (changes requested):

WardenGnaw (approved):

sean-mcmanus:
@WardenGnaw Should we add Fortran too? I noticed https://github.com/microsoft/vscode-cpptools/issues/10406 .

Extension/package.json
AnthonyLeonardoGracio: I think it's alphabetical yes. It probably does not change the extension's behavior in any case.

WardenGnaw:
> @WardenGnaw Should we add Fortran too? I noticed #10406 .

I would only add it if we have a request. GDB supports the following languages C, C ++, D, Go, Objective-C,  Fortran, Pascal, Rust, Assembly, Modula-2, and Ada.

I'm hesitant on adding everything unless there is a request for it.

sean-mcmanus (approved):

Issue #10466, opened 2023-02-03
Wrong tooltip in launch.json for natvisDiagnostics


xisui-MSFT:
### Environment

N/A
- OS and version:
- VS Code:
- C/C++ extension:
- OS and version of remote machine (if applicable):
- GDB / LLDB version:

### Bug Summary and Steps to Reproduce

https://github.com/microsoft/vscode-cpptools/blob/2d3cac0dea96a8b973cbea95cc0a1e302c28ad31/Extension/package.json#L3512, should be c_cpp.debuggers.logging.natvisDiagnostics.description

### Debugger Configurations

```shell
N/A
```

### Debugger Logs

```shell
N/A
```

bobbrow:
@browntarik can you fix up this typo real quick?

bobbrow:
@browntarik 
Bug is on line 135 of `Extension/tools/OptionsSchema.json`.  Fix it there and then run `yarn generateOptionsSchema` in the terminal.  (I think that's how it works, right @WardenGnaw?)

WardenGnaw:
> @browntarik Bug is on line 135 of `Extension/tools/OptionsSchema.json`. Fix it there and then run `yarn generateOptionsSchema` in the terminal. (I think that's how it works, right @WardenGnaw?)

Yep.

PR #10483, opened 2023-02-07
change description for natvisDiagnostics

browntarik:
Addresses #10466

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3515,7 +3515,7 @@
                       },
                       {
                         "type": "boolean",
-                        "description": "%c_cpp.debuggers.logging.engineLogging.description%",
+                        "description": "%c_cpp.debuggers.logging.natvisDiagnostics.description%",
                         "default": false
                       }
                     ]
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -4303,7 +4303,7 @@
                       },
                       {
                         "type": "boolean",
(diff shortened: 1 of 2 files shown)


Extension/package.json
```diff
@@ -4303,7 +4303,7 @@
                       },
                       {
                         "type": "boolean",
-                        "description": "%c_cpp.debuggers.logging.engineLogging.description%",
+                        "description": "%c_cpp.debuggers.logging.natvisDiagnostics.description%",
```
bobbrow, line 4306: Did you run `yarn generateOptionsSchema`?  I believe there are two entries in package.json that should be updated.

browntarik: I only saw one from the issue

browntarik: image

sean-mcmanus (approved):

Extension/package.json
sean-mcmanus: Is this resolved?

WardenGnaw: We only have this supported for MIEngine `cppdbg`. Not `cppvsdbg`.

WardenGnaw (approved):

Extension/package.json
WardenGnaw: This can be resolved. The Options schema fixed the boolean description and there are two instances of the boolean being updated. 

Line 3518 and this.

bobbrow (approved):

PR #10485, opened 2023-02-07
Remove the full path for the cl.exe compilerPath generated for c_cpp_properties.json

browntarik:


Extension/src/LanguageServer/utils.ts
```diff
@@ -30,6 +31,17 @@ export function rangeEquals(range1: vscode.Range | Range, range2: vscode.Range |
     range1.end.line === range2.end.line && range1.end.character === range2.end.character;
 }
 
+export function getFileFromPath(filePath: string): string | undefined {
```
sean-mcmanus: It seems like this should return "string" and never undefined, i.e. return filePath on error or if it doesn't have path segments?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/utils.ts
```diff
@@ -30,6 +31,17 @@ export function rangeEquals(range1: vscode.Range | Range, range2: vscode.Range |
...
+    let compiler: string | undefined;
```
sean-mcmanus: Seems like compiler should be renamed to "filename" or something more generic.

sean-mcmanus: Actually, it seems like the variable should be removed.

Extension/src/LanguageServer/utils.ts
```diff
@@ -30,6 +31,17 @@ export function rangeEquals(range1: vscode.Range | Range, range2: vscode.Range |
...
+    const slash: string = (os.platform() === 'win32') ? "\\" : "/";
+
+    if (filePath.includes(slash)) {
+        if (filePath.split(slash).pop() !== undefined) {
+            compiler = filePath.split(slash).pop();
```
sean-mcmanus: This can just be returned instead of assigned to compiler.

Extension/src/LanguageServer/client.ts
```diff
@@ -906,21 +906,13 @@ export class DefaultClient implements Client {
 
         const items: IndexableQuickPickItem[] = [];
         for (let i: number = 0; i < paths.length; i++) {
-            let option: string | undefined;
-            let isCompiler: boolean = false;
-            const slash: string = (os.platform() === 'win32') ? "\\" : "/";
+            const compiler: string | undefined = getFileFromPath(paths[i]);
+            const isCompiler: boolean = compiler !== undefined;
```
sean-mcmanus: You could change this test to compiler !== paths[i].

Extension/src/LanguageServer/client.ts
```diff
@@ -906,21 +906,13 @@ export class DefaultClient implements Client {
...
-            if (paths[i].includes(slash)) {
-                if (paths[i].split(slash).pop() !== undefined) {
-                    option = paths[i].split(slash).pop();
-                    isCompiler = true;
-                }
-            }
-
-            if (option !== undefined && isCompiler) {
-                const path: string | undefined = paths[i].replace(option, "");
+            if (compiler !== undefined && isCompiler) {
```
sean-mcmanus: This test can be simplified to "if(isCompiler)".

Extension/src/LanguageServer/utils.ts
browntarik: Could you give a code snippet?

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -906,21 +906,13 @@ export class DefaultClient implements Client {
 
         const items: IndexableQuickPickItem[] = [];
         for (let i: number = 0; i < paths.length; i++) {
-            let option: string | undefined;
-            let isCompiler: boolean = false;
-            const slash: string = (os.platform() === 'win32') ? "\\" : "/";
+            const compiler: string = getFileFromPath(paths[i]);
```
michelleangela: Instead of `getFileFromPath()`, can `path.basename(....).toLowerCase()` be used? See examples of existing code where `path.basename` used. Also lower case the result when comparing in case the input is uppercase.

michelleangela (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -396,7 +396,15 @@ export class CppProperties {
             (isUnset(settings.defaultCompileCommands) || settings.defaultCompileCommands === "") && !configuration.compileCommands) {
             // compile_commands.json already specifies a compiler. compilerPath overrides the compile_commands.json compiler so
             // don't set a default when compileCommands is in use.
-            configuration.compilerPath = this.defaultCompilerPath;
+
+            // if the compiler is a cl.exe compiler, replace the full path with the "cl.exe" string.
+            const compiler: string = path.basename(this.defaultCompilerPath).toLowerCase();
+
+            configuration.compilerPath = "cl.exe";
+            if (compiler === "cl.exe") {
```
bobbrow, line 403: Are lines 403 and 404 out of order?

Extension/src/LanguageServer/client.ts
```diff
@@ -906,21 +906,13 @@ export class DefaultClient implements Client {
 
         const items: IndexableQuickPickItem[] = [];
         for (let i: number = 0; i < paths.length; i++) {
-            let option: string | undefined;
-            let isCompiler: boolean = false;
-            const slash: string = (os.platform() === 'win32') ? "\\" : "/";
+            const compiler: string = path.basename(paths[i]).toLowerCase();
+            const isCompiler: boolean = compiler !== paths[i];
```
michelleangela: Does this assume the value of `paths[i]` is lower case? 

Sorry if I misled, I mean to use lower case when comparing with the hardcode literal "cl.exe".

Extension/src/LanguageServer/configurations.ts
```diff
@@ -396,7 +396,15 @@ export class CppProperties {
             (isUnset(settings.defaultCompileCommands) || settings.defaultCompileCommands === "") && !configuration.compileCommands) {
             // compile_commands.json already specifies a compiler. compilerPath overrides the compile_commands.json compiler so
             // don't set a default when compileCommands is in use.
-            configuration.compilerPath = this.defaultCompilerPath;
+
+            // if the compiler is a cl.exe compiler, replace the full path with the "cl.exe" string.
+            const compiler: string = path.basename(this.defaultCompilerPath).toLowerCase();
+
+            configuration.compilerPath = "cl.exe";
```
michelleangela: Did `configuration.compilerPath = "cl.exe";` get moved out of the conditional block?

Extension/src/LanguageServer/client.ts
michelleangela: How come `compiler !== paths[i]` means it is a compiler?

browntarik: If it is not equal, then it is only a compiler name. For example, "cl.exe" was successfully extracted from a path. If it is equal, then it will be equal to the full path option (i.e one of the last three options like "Help me install a compiler"). Thus it will be categorized as false or "not a compiler"

michelleangela: Oh I see, the name "compiler" made me think it has to be validated it's a real compiler instead of is compiler name.

browntarik: should I change it to isValidCompilerName?

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
bobbrow: Maybe call it `compilerName` for clarity

bobbrow: (not essential though)

michelleangela: If you'd like to make the variable reflect what it's checking. Descriptive variables are good for readability.

michelleangela: Maybe change `compiler` to `compilerName` too.

michelleangela (approved):

sean-mcmanus (approved):

sean-mcmanus (approved):

Colengms (approved):

Issue #10076, opened 2022-10-28
New hover setting requires window reload and does not trigger reload window prompt


Colengms:
The new setting to enable/disable hover is passed in the `ServerCapabilities` object in response to LSP initialization. So, if the setting was disabled when the window was loaded, enabling it would require the window to be reloaded to properly initialize LSP to send hover requests.

The autocomplete setting is similar.  Though, changes to the autocomplete setting do trigger a reload window prompt.

VS Code will merge the results from multiple hover providers, and does something similar (? based on score) with completion.  So, I don't believe it's necessary to avoid registering as a hover or completion provider.  We could avoid requiring the user to reload the window when either of these settings are changed, by always registering as a provider for them.

PR #10498, opened 2023-02-09
Add reload prompt when changing hover setting

browntarik:
Addresses #10076


michelleangela (approved):

sean-mcmanus (approved):

sean-mcmanus:
But Colen was suggesting "We could avoid requiring the user to reload the window when either of these settings are changed, by always registering as a provider for them." -- I suppose that could still be done later on though.

PR #10502, opened 2023-02-09
add basic compiler selection telemetry

browntarik:


Extension/src/LanguageServer/client.ts
```diff
@@ -971,6 +971,8 @@ export class DefaultClient implements Client {
         paths.push(localize("installCompiler.string", "Help me install a compiler"));
         paths.push(localize("noConfig.string", "Do not configure a compiler (not recommended)"));
         const index: number = await this.showSelectDefaultCompiler(paths);
+        const action: string = index === -1 ? 'escaped' : index === paths.length - 1 ? 'disable' : index === paths.length - 2 ? 'help' : index === paths.length - 3 ? 'browse' : 'select compiler';
```
michelleangela: For readability, can this be replaced with switch/cases or if/else? Nested trinary are hard to follow.

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -971,6 +971,8 @@ export class DefaultClient implements Client {
...
+        telemetry.logLanguageServerEvent('compilerSelection', { action });
```
sean-mcmanus, line 992: What about escape after browse? Do we want telemetry on that?

bobbrow: I think we can address that in a follow up PR.  There are a few telemetry requests in the spec that weren't implemented.  I'd like to get some basic stuff for 1.14.2 without missing our deadline

Extension/src/LanguageServer/client.ts
bobbrow: Let's put a note to clean it up for 1.14.3.  I mostly just don't want to jeapordize today's pre-release.

bobbrow (dismissed):

sean-mcmanus (dismissed):

michelleangela (dismissed):

michelleangela (approved):

sean-mcmanus (approved):

PR #10513, opened 2023-02-13
Add telemetry for compiler notification and # in default compiler selection

browntarik:
Telemetry Use Cases: 

- A user’s selection in the initial notification (Configure compiler, select compiler button, escaping) 
Event Name: compilerNotification

- A user’s selection in the quickpick (Select a compiler, to get help, they do not need a compiler)  
Event Name: compilerSelection

- How many compilers were found on the user's machine when we show the notification/quickpick  
Event Name: compilerCount


Extension/src/LanguageServer/client.ts
```diff
@@ -950,11 +950,13 @@ export class DefaultClient implements Client {
         const settings: OtherSettings = new OtherSettings();
         const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
         const paths: string[] = [];
+        let compilerCount: number = 0;
         if (compilerDefaults.knownCompilers !== undefined) {
             const tempPaths: string[] = compilerDefaults.knownCompilers.map(function (a: configs.KnownCompiler): string { return a.path; });
             let clFound: boolean = false;
             // Remove all but the first cl path.
             for (const path of tempPaths) {
```
Colengms: Rather than increment per path, you could use `tempPaths.length`.

Extension/src/LanguageServer/client.ts
```diff
@@ -967,6 +969,8 @@ export class DefaultClient implements Client {
                 }
             }
         }
+        const compilerCountString: string = compilerCount.toString();
+        telemetry.logLanguageServerEvent('compilerCount', { compilerCountString });
```
bobbrow: Can you add this to the telemetry properties for the 'compilerSelection' event instead of having a separate event for it?

Extension/src/LanguageServer/client.ts
```diff
@@ -1046,8 +1051,12 @@ export class DefaultClient implements Client {
                     settings.defaultCompiler = compilerDefaults.compilerPath;
                     compilerDefaults = await this.requestCompiler(compilerPaths);
                     DefaultClient.updateClientConfigurations();
+                    action = "configure compiler";
+                    telemetry.logLanguageServerEvent('compilerNotification', { action });
```
Colengms: Can you pass the string instead of creating a new variable for it?

Extension/src/LanguageServer/client.ts
```diff
@@ -967,6 +969,8 @@ export class DefaultClient implements Client {
                 }
             }
         }
+        const compilerCountString: string = compilerCount.toString();
```
sean-mcmanus: Shouldn't this be something like `paths.length()`?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1046,8 +1051,12 @@ export class DefaultClient implements Client {
                     settings.defaultCompiler = compilerDefaults.compilerPath;
                     compilerDefaults = await this.requestCompiler(compilerPaths);
                     DefaultClient.updateClientConfigurations();
+                    action = "configure compiler";
```
bobbrow: How about "confirm compiler" to match the button?

Extension/src/LanguageServer/client.ts
```diff
@@ -1046,8 +1051,12 @@ export class DefaultClient implements Client {
...
+                    telemetry.logLanguageServerEvent('compilerNotification', { action });
                 } else if (value === selectCompiler) {
                     this.handleCompilerQuickPick(true);
+                    action = "show quickpick";
+                    telemetry.logLanguageServerEvent('compilerNotification', { action });
                 } else {
                     this.showPrompt(selectCompiler, true);
```
bobbrow, line 1048: I think we'd want to log this one too.  "dismissed" would be an appropriate name.

Extension/src/LanguageServer/client.ts
```diff
@@ -1046,8 +1051,12 @@ export class DefaultClient implements Client {
                     settings.defaultCompiler = compilerDefaults.compilerPath;
                     compilerDefaults = await this.requestCompiler(compilerPaths);
                     DefaultClient.updateClientConfigurations();
+                    action = "configure compiler";
+                    telemetry.logLanguageServerEvent('compilerNotification', { action });
                 } else if (value === selectCompiler) {
                     this.handleCompilerQuickPick(true);
+                    action = "show quickpick";
+                    telemetry.logLanguageServerEvent('compilerNotification', { action });
```
bobbrow: Since I recommend we log for all three cases, move the call to `logLanguageServerEvent` outside the if/else block after the action is set.

bobbrow (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1037,6 +1041,7 @@ export class DefaultClient implements Client {
         }
         const selectCompiler: string = localize("selectCompiler.string", "Select Compiler");
         const confirmCompiler: string = localize("confirmCompiler.string", "Yes");
+        let action: string;
```
sean-mcmanus, line 1033: Can you remove this variable and just pass the string directly?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: I think it would be paths.length and not tempPaths, since paths has the values shown to users.

Extension/src/LanguageServer/client.ts
```diff
@@ -1026,6 +1026,8 @@ export class DefaultClient implements Client {
                 return;
             }
             settings.defaultCompiler = result[0].fsPath;
+            action = "compiler browsed";
+            telemetry.logLanguageServerEvent('compilerBrowsed', { action });
```
bobbrow: I'd like to keep this in the same 'compilerSelection' event if possible.  Maybe we wrap the whole function in a `try/finally` so that we can log it whenever this function exits since there are so many early returns.

Extension/src/LanguageServer/client.ts
browntarik: It would be tempPaths because we care about the total number of compilers even if extraneous cl compilers are condensed for UX purposes

bobbrow: I think we want to know how many appear in the list.  Let's not count cl.exe a million times.

Extension/src/LanguageServer/client.ts
browntarik: Can you clarify? You can't pass the string as an argument; it has to be a {"key": value} pair.

Extension/src/LanguageServer/client.ts
browntarik: What is the syntax for that?

bobbrow: The syntax is delete this line and the line above and add it to the "measures" (3rd optional argument which accepts numbers instead of strings for the values).

Something like this (I didn't compile)
```ts
telemetry.logLanguageServerEvent('compilerSelection', { action }, { compilerCount });
```

Extension/src/LanguageServer/client.ts
```diff
@@ -967,6 +968,8 @@ export class DefaultClient implements Client {
                 }
             }
         }
+        compilerCount = paths.length;
+        telemetry.logLanguageServerEvent('compilerSelection', { action :"compilerCount"}, { compilerCount })
```
sean-mcmanus: Can you use `{ compilerCount: paths.length }` without the compilerCount local variable?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Also, can you run the formatter on this? I think `action :` should be "`action: ` -- running for the formatter may also fix the linter issues.

bobbrow: These lines are being deleted. It will merge with the other call to log this event

bobbrow: But `{ compilerCount: paths.length }` will work without needing a `compilerCount` variable.

Extension/src/LanguageServer/client.ts
```diff
@@ -971,63 +971,56 @@ export class DefaultClient implements Client {
...
-        }
-        if (index === paths.length - 3) {
-            const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
-            if (result === undefined || result.length === 0) {
-                return;
+            if (index === paths.length - 3) {
+                action = "browse";
+                const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
+                if (result === undefined || result.length === 0) {
+                    return;
```
sean-mcmanus, line 1009: It seems like it would be better if it set something like `action = "browse dismissed"` instead of `"action = "browse"` in the escape case.

sean-mcmanus (changes requested):

michelleangela:
Where is the event for "compilerBrowsed"?

bobbrow:
> Where is the event for "compilerBrowsed" as described in PR description?

It's gone. I asked him to merge it with 'compilerSelection'

bobbrow (dismissed):

Colengms (dismissed):

michelleangela (dismissed):

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: I meant
```TypeScript
            if (index === paths.length - 3) {
                const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
                if (result === undefined || result.length === 0) {
                    action = "browse dismissed";
                    return;
                }
```
i.e. there's no need to set action to "browse dismissed" if it's not dismissed.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
browntarik: True but it doesn't matter regardless since the function will return before reaching the other string

sean-mcmanus (approved):

michelleangela (approved):

Issue #10532, opened 2023-02-16
Automatically load counterpart files in a separate pane, like Xcode Assistant Editor


csnover:
### Feature Request

This may be the same request as #10346; the description in that ticket was not totally clear. Hopefully this one is clear enough to be actionable. Please feel free to close/merge it into that ticket if they are the same thing.

Xcode has a feature called Assistant Editor which creates a split view where the pane on the right automatically loads the counterpart file for whatever is loaded in the main editor on the left. So, for example, if you load a .c file in the main editor, Xcode will automatically load the corresponding .h file in the assistant editor (or vice-versa). This request is to add this feature to vscode-cpptools.

For implementation, the built-in markdown-language-features extension seems like a good place to start since its “Open Preview” commands work exactly the way this feature should be implemented (in particular, “Open Preview to the Side”), except that here the ‘preview’ would be another editor for the counterpart file. (For clarity: both Xcode Assistant Editor and the VSCode Markdown Preview use one tab and change the content automatically based on to what other tab is currently active, rather than opening new tabs for each source file.)

If this sounds like something that would be accepted as a new feature, I may try implementing it myself, depending on free time. If this already exists, my apologies and let me know where it is.

Thanks,

sean-mcmanus:
Yeah, this sounds like a duplicate of https://github.com/microsoft/vscode-cpptools/issues/10346, but I agree with you that issue is hard to understand, so I'll close the other issue as a duplicate of this one.

cjwijtmans:
How i envisioned it was not a pane but being able to set regex rules on a editor group. Which would make it a vscode feature not a cpptools feature.

sean-mcmanus:
@cjwijtmans Sure, can you file a feature request at https://github.com/microsoft/vscode/issues ?

csnover:
Editor group by regular expression has an upstream ticket already and was closed due to lack of community upvotes: https://github.com/microsoft/vscode/issues/136162

cjwijtmans:
> Editor group by regular expression has an upstream ticket already and was closed due to lack of community upvotes: [microsoft/vscode#136162](https://github.com/microsoft/vscode/issues/136162)

I do not see anything about regex there. But sad to see it rejected. I guess we will need cpptools panel then.

PR #10559, opened 2023-02-21
Add Assistant Editor

csnover:
This PR closes issue #10532.

This patch set includes a prerequisite patch to deduplicate the code that detects the document language since the Assistant Editor also needs to do this and the checks were just copied and pasted all over the place.

The main thing this patch set introduces is a MVP version of Assistant Editor. In this implementation, the Assistant Editor can be toggled in the command menu. It opens beside the active tab group and will track the active tab from that group until it is closed.

Multiple Assistant Editors can be active at the same time for different editor groups, so e.g. you can have two editor groups each with their own Assistant Editor.

The Assistant Editor persists its state to the workspace state so when the workspace is reloaded, the Assistant Editor continues to operate normally.

There are some caveats to the current implementation that are largely the result of what seem to be very many missing APIs in VSCode:

* The Assistant Editor tab has no indicator that it is special in some way because there is no API to apply labels to tabs
* The Assistant Editor tab will be deleted and recreated whenever it needs to switch to a new document, because there is no API to replace content of a tab, and no API to embed a standard TextEditor inside a custom Editor that would eliminate the need for any sort of tab management
* The Assistant Editor tab could get lost if someone does a bunch of stuff to change their layout, because the Tab API invalidates the objects it exposes when tab groups change, and also has no API to determine whether a tab was invalidated despite not being closed (see https://github.com/microsoft/vscode-discussions/discussions/480 where I tried asking what the right way to do this is supposed to be), and I do not trust the hacks that currently exist in this code recover the tab. Hopefully this is a bug upstream and the extra tracking code can go away someday
* The Assistant Editor tab will always jump to the end of the tab group it exists in whenever it switches to a new document, because there is no API to specify where a tab should be created within a tab group, and no API to move a tab other than some commands that only operates on the active editor and there is—you probably are noticing a theme by now—no API that I could find to switch focus to a particular tab

There is also a caveat caused by the closed-source (as far as I can tell) binary blob that this extension uses. I had to write redundant (and presumably less robust) code in the Assistant Editor that tries to find the counterpart file for the currently open file, instead of being able to reuse the existing API call that the Switch Header/Source command uses.

I was actually unable to get cpptools to respond to requests at all from the extension debugger when I followed the [build instructions](https://github.com/microsoft/vscode-cpptools/blob/main/Documentation/Building%20the%20Extension.md) and I don’t know what was going on there; the client appeared to be running, but the `Promise` for the `requestSwitchHeaderSource` call would never settle. (Even in a fresh checkout of the `release` branch, the already existing Switch Header/Source command did not work.) I didn’t bother to try troubleshooting this since without access to the cpptools binary source it felt like a waste of time, since I can’t change the stuff in the binary anyway, and it calls this the “didSwitchHeaderSource” command and not the “queryCounterpartFileName” or whatever command, so it would be wrong.

Looking forward to your feedback. Apologies in advance if any comments seem frustrated and please feel free to tone police them if it feels necessary; this is the first time I have worked on a VSCode extension, and I did not expect such a bad developer experience given how well VSCode itself works as an IDE.

Thanks,

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -2925,6 +2925,11 @@
         "category": "C/C++"
       },
       {
+        "command": "C_Cpp.ToggleAssistantEditor",
+        "title": "%c_cpp.command.toggleAssistantEditor.title%",
+        "category": "C/C++"
+      },
+      {
         "command": "C_Cpp.EnableErrorSquiggles",
         "title": "%c_cpp.command.enableErrorSquiggles.title%",
         "category": "C/C++"
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -13,6 +13,7 @@
     "c_cpp.command.configurationEditUI.title": "Edit Configurations (UI)",
     "c_cpp.command.selectDefaultCompiler.title": "Select Default Compiler",
     "c_cpp.command.switchHeaderSource.title": "Switch Header/Source",
+    "c_cpp.command.toggleAssistantEditor.title": "Toggle Assistant Editor",
     "c_cpp.command.enableErrorSquiggles.title": "Enable Error Squiggles",
     "c_cpp.command.disableErrorSquiggles.title": "Disable Error Squiggles",
     "c_cpp.command.toggleIncludeFallback.title": "Toggle IntelliSense Engine Fallback on Include Errors",
--- a/Extension/src/LanguageServer/assistantEditor.ts
+++ b/Extension/src/LanguageServer/assistantEditor.ts
@@ -1,0 +1,463 @@
+// SPDX-License-Identifier: MIT
+
+import * as pathUtil from 'path';
+import * as vscode from 'vscode';
+import { PersistentWorkspaceState } from './persistentState';
+import * as nls from 'vscode-nls';
+import { isCppOrCFile, isCppToolsFile, isHeaderFile } from '../common';
+
+/**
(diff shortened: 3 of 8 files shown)


csnover:
The CI is failing due to a linting issue. I opened #10573 about it and will update this patch set based on what is decided there to get the checks to pass. Thanks!

bobbrow:
Thank you for submitting this PR.  I'm sorry we haven't made much progress on it yet since we've had a couple of other projects in flight, but we have seen it and will find a release to prioritize this for when we've sorted out some of the open issues.

csnover:
No worries, thanks for the update. Running with this change locally there is still an outstanding issue where the assistant tab gets lost sometimes, but since I didn’t receive any feedback for a while I moved my time to something else. Let me know when you are about ready so I can revisit this problem. (Unfortunately, no one upstream ever responded to [my question](https://github.com/microsoft/vscode-discussions/discussions/480) so I am doing the best I can with what I have.)

sean-mcmanus:
I thought @csnover was going to update the linting rules in this branch and not us.

gdamiaN538 (approved):
https://github.com/microsoft/vscode-cpptools.wiki.git

sean-mcmanus:
We're closing this PR since it seems to be stale. Let us know if you think it should be re-opened.

PR #10567, opened 2023-02-22
Remove overriding of insertMode.

### LINKED ISSUE #10553: Autocomplete deletes code on the right side of the line
### Environment

- OS and Version:  macOS 13.2.1
- VS Code Version: 1.75.1
- C/C++ Extension Version: 1.14.3

Bug Summary:

When I autocomplete a param it deletes the content on the right side of the line

https://user-images.githubusercontent.com/3449006/219973582-d5ee7546-dce3-4429-ac87-4ceb96b4ca03.mp4

Steps to reproduce:
1. Copy this code
```c++
struct Something {
    int something;
};
struct Foo {
    Something first, second;
};

Foo foo;
foo.first.something;
```
2. delete `first.` and place the cursor just after `foo`
3. `command + space` or `ctrl + space (in windows)` to toggle the autocomplete menu
4. select option (see video above)

Expected behavior:

I will like the behaviour I usually get when working with typescript, where when I call the autocomplete it respects what is written on the right of the line (see video below)

https://user-images.githubusercontent.com/3449006/219973550-777a4f5c-91d1-47be-b786-3a6ea7693599.mp4

### Configuration and Logs

```shell
{
  "C_Cpp.default.cppStandard": "c++20"
}
```

### Other Extensions

_No response_

### Additional context

_No response_

sean-mcmanus:
Fix https://github.com/microsoft/vscode-cpptools/issues/10553 .

Revert change in https://github.com/microsoft/vscode-cpptools/pull/4892.

I'm unable to find a reason to use the "replace" instead of "insert" as the default. Both TypeScript and VS have the "insert" behavior and I haven't found a bad behavior with "insert".

If we get complaints about the change with 1.15.0, we could possibly revert it for 1.15.1.


bobbrow (changes requested):
We should not make this change. We intentionally did this to match C++ in Visual Studio. Please do not check this in.

sean-mcmanus:
> We should not make this change. We intentionally did this to match C++ in Visual Studio. Please do not check this in.

Then this changes matches the current VS -- do you get different behavior?

bobbrow:
> > We should not make this change. We intentionally did this to match C++ in Visual Studio. Please do not check this in.
> 
> This changes matches the current VS -- do you get different behavior?

Please share your test cases and the VS/VSC results.  I do see one discrepancy where if you CTL+Space after the `.` or `->`, VS will do insert instead of replace, but that seems like a bug to me.  In my testing, invoking auto-complete anywhere else will replace.

sean-mcmanus:
> Please share your test cases and the VS/VSC results. I do see one discrepancy where if you CTL+Space and the cursor is immediately to the right of the `.` or `->`, VS will do insert instead of replace, but that seems like a bug to me. In my testing, invoking auto-complete anywhere else within an existing identifier will replace.

The test case is the user's at https://github.com/microsoft/vscode-cpptools/issues/10553 . Yeah, the cursor is to the right of the `.`. The result is the same after this change and with VS. The result is the same when it's to the right of `.` as well, i.e. it will do a replace with VS and VS Code (even with this setting set to "insert").

bobbrow:
We've had this behavior for the last 6 years, before the setting existed.  I can point you to the PR if you want.  I remember discussing the defaults with the VS Code team years ago when they introduced this setting.  Some languages were doing replace and others were doing insert.  They went with insert because that's what TypeScript was doing and it's their primary language.  They told us we could override the default in our package.json and that's what we did.  We won't be changing our behavior to match TypeScript's at this point.  It will impact too many people who are used to the current behavior.

If we are not doing the right thing for "insert" mode, then we can fix the auto-complete code to honor that setting.  I don't think we currently read it.  But I would expect VS Code to have enough information to ignore the Range in our TextEdit and do the right thing for the "insert" case.  (In other words, I would make it a VS Code bug)

sean-mcmanus:
I don't understand. The behavior seems right to me with "insert" mode, so I don't understand why you're talking about in regards to "insert" mode. The incorrect behavior seems to be with "replace" mode.

bobbrow:
> I don't understand. The behavior seems right to me with "insert" mode, so I don't understand why you're talking about in regards to "insert" mode. The incorrect behavior seems to be with "replace" mode.

The behavior that I called a potential VS Bug?  I believe our behavior is "correct."  It doesn't make sense to not replace when the cursor is to the right of the `.` or `->`.  I didn't have time to install VS 2017 and double check what the behavior was when we implemented this, but I would still call this a VS bug.

sean-mcmanus:
> The behavior that I called a potential VS Bug? I believe our behavior is "correct." It doesn't make sense to not replace when the cursor is to the right of the . or ->. I didn't have time to install VS 2017 and double check what the behavior was when we implemented this, but I would still call this a VS bug.

Yes, with "insert" mode, VS Code does a replace when the cursor is to the right of the . or ->, which matches VS 2022. Using "replace" mode causes it to not match VS.

Also, FYI, I accidentally hit the wrong button and editted your comment instead of quoting it (I believe I restored it though).

sean-mcmanus:
@bobbrow Our current hypothesis is that VS Code fixed the "insert" behavior such that it now behaves so that it matches what VS C++ uses so we no longer want the "replace" behavior -- can you try it out?

sean-mcmanus:
@bobbrow Can you revisit this for 1.15.0? Using "insert" will still do a replace when VS does, see https://github.com/microsoft/vscode-cpptools/issues/10613 .

sean-mcmanus:
Anyone wants "insertMode" to be changed back to the default (i.e this PR), you can create an issue on that.

Issue #10531, opened 2023-02-15
Including standard library headers doesn't work since 1.14.3


K1T3K1:
### Environment

- OS and Version: macOS Ventura 13.1 M1 Pro Arm64 Apple Silicon
- VS Code Version: 1.75.1
- C/C++ Extension Version: 1.14.3 
- If using SSH remote, specify OS of remote machine: 

### Bug Summary and Steps to Reproduce

Bug Summary: can't `#include <header>` from standard library on 1.14.3. Reverting to 1.13.9 fixes the problem

Expected behavior:

### Configuration and Logs

```shell
#include errors detected. Please update your includePath. Squiggles are disabled for this translation unit (/Users/kpz/repositories/cpp_projects/std_lib_impl/function.h).C/C++(1696)
cannot open source file "memory"C/C++(1696)

settings.json:
    "C_Cpp.intelliSenseEngine":"default",
    "C_Cpp.clang_format_fallbackStyle": "{BasedOnStyle: Google, ColumnLimit: 0}",
    "C_Cpp.clang_format_style": "file",
    "C_Cpp.formatting": "default",
    "C_Cpp.vcFormat.newLine.beforeElse": false,
    "C_Cpp.vcFormat.space.beforeInheritanceColon": false,
    "C_Cpp.codeAnalysis.clangTidy.enabled": false,
    "C_Cpp.codeAnalysis.clangTidy.checks.enabled": [
        "clang-analyzer-*"
    ],
    "C_Cpp.codeAnalysis.clangTidy.fallbackConfig": "",
    "C_Cpp.codeAnalysis.runAutomatically": true,
    "C_Cpp.codeAnalysis.clangTidy.useBuildPath": true,
    //"C_Cpp.codeAnalysis.clangTidy.args": ["--extra-arg=--target=gcc-arm-none-eabi"],
    "C_Cpp.default.includePath": [
        "${workspaceFolder}/**",
        //"/Users/kpz/esp/esp-idf/components/**"
```

bobbrow:
Do you use `c_cpp_properties.json` to configure IntelliSense?  If not, you'll need to run the "Select a Compiler" command to tell us which compiler you want us to emulate.  We don't automatically pick one for you anymore unless it's in your $PATH and there should have been a prompt to select one after you upgraded the extension.  This command effectively sets the `C_Cpp.default.compilerPath` setting for you.

K1T3K1:
[...]

PR #10574, opened 2023-02-23
 Add an additional message to the #include error and also add a code action (fixit) for the squiggle to run the "Select Default Compiler" command.

browntarik:
Addresses https://github.com/microsoft/vscode-cpptools/issues/10531


Extension/package.json
```diff
@@ -2917,8 +2917,8 @@
         "category": "C/C++"
       },
       {
-        "command": "C_Cpp.selectDefaultCompiler",
-        "title": "%c_cpp.command.selectDefaultCompiler.title%",
+        "command": "C_Cpp.SelectDefaultCompiler",
+        "title": "%c_cpp.command.SelectDefaultCompiler.title%",
```
michelleangela: `SelectDefaultCompiler` for title should be lower case.

michelleangela (changes requested):

Extension/src/nativeStrings.json
```diff
@@ -437,5 +440,8 @@
         "text": "The operation is not supported for '%s'.",
         "hint": "%s is the function that is not supported for an operation that involves automatically generating code."
     },
-    "unknown_error": "Unknown error."
+    "unknown_error": "Unknown error.",
+    "run_select_compiler": {
+        "text": "Please run the 'Select Default Compiler' command to locate your system headers"
```
sean-mcmanus: This string needs to be updated still.

sean-mcmanus (changes requested):

Extension/bin/messages/de/messages.json
```diff
@@ -1695,8 +1695,8 @@
   null,
   "Für die Vektoroperation sind identische Elementtypen erforderlich.",
   "Die Vektoroperation gilt nicht für Vektoren mit nicht integralem Typ.",
-  "Die Datei \"%s\" kann nicht geöffnet werden: %sq2.",
-  "Die -Datei \"%s\" kann nicht geöffnet werden: %sq2: %s3.",
+  "Die Datei %s kann nicht geöffnet werden: %sq2.",
```
sean-mcmanus, line 1698: I believe this should be `Quelldatei` -- I think we may need code changes to properly translate this. I've filed an internal issue and email about it.

sean-mcmanus (changes requested):

Extension/bin/messages/de/messages.json
sean-mcmanus: ...but this may not block the check in if we can't get it figured out soon enough, i.e. it could be fixed in a follow up PR

browntarik: We can take a look at this in another PR

sean-mcmanus (dismissed):

Extension/package.nls.json
```diff
@@ -11,7 +11,7 @@
     "c_cpp.command.configurationProviderSelect.title": "Change Configuration Provider...",
     "c_cpp.command.configurationEditJSON.title": "Edit Configurations (JSON)",
     "c_cpp.command.configurationEditUI.title": "Edit Configurations (UI)",
-    "c_cpp.command.selectDefaultCompiler.title": "Select Default Compiler",
+    "c_cpp.command.SelectDefaultCompiler.title": "Select Default Compiler",
```
michelleangela: `c_cpp.command.SelectDefaultCompiler.title` needs to match the ID from package.json. So the "SelectDefaultCompiler" part needs to be lower case too.

michelleangela (changes requested):

michelleangela (approved):

sean-mcmanus (approved):

PR #10629, opened 2023-03-03
Fix an uncaught exception

Colengms:
Addresses an issue reported internally by the WAVE team, in which the call to `callTaskWithTimeout` is resulting in an uncaught exception when it times out.


bobbrow (approved):

bobbrow:
Interesting... It looks like the "unable to provide" message box logic is in the `catch` which hints that this line may have been inside the `try` at one point.

sean-mcmanus (approved):

Issue #10527, opened 2023-02-14
Language Status follow up for compilerPath


bobbrow:
When no compiler is configured, the link for IntelliSense Status in the Language Status flyout should run the "select a compiler" command.  Alex's suggested text for the link is "Improve IntelliSense."

sean-mcmanus:
I would prefer something like "Configure IntelliSense" over "Improve IntelliSense". Was there a reason to use "improve" instead?

sean-mcmanus:
I don't understand. If no compiler is configured then we've selected on own. To "configure" implies "an explicit action to configure something correctly". The verb "improve" seems too generic and seems to imply that the link will enable IntelliSense to be better, faster, more accurate, etc., when it's really just "configuring" to work correctly with a particular compiler.

bobbrow:
> If no compiler is configured then we've selected one on own.

I believe this is for the case where we would have shown the prompt to confirm our compiler selection.  If they dismiss that, we don't have a compiler at all.  We configure without it.  I'm not particular about the text here.  Please sync with @AlexandraKemperMS and come to a consensus if you would like to have different text.

bobbrow:
This issue has been replaced by #10685

PR #10641, opened 2023-03-07
Add Select a Compiler to flyout

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10527


Extension/src/LanguageServer/ui_new.ts
```diff
@@ -280,10 +280,10 @@ export class NewUI implements UI {
         const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
 
         // TODO: Integrate with Tarik's feature to determine if compiler/bare-intellisense is configured
```
bobbrow, line 282: Can we remove this TODO comment now?

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -280,10 +280,10 @@ export class NewUI implements UI {
...
-        if (settings.intelliSenseEngine === "disabled") {
+        if (settings.intelliSenseEngine === "default" && !compilerDefaults.trustedCompilerFound) {
```
sean-mcmanus, line 283: I think we also need to check like `(compilerPaths.length !== 1 || compilerPaths[0] !== "")` i.e. in case compilerPath is set to `""` then we don't want to do this.

sean-mcmanus (changes requested):

PR #10669, opened 2023-03-14
Use `vscode.env.language` to setting localization language ID

Colengms:
This uses vscode.env.language, based on guidance in: https://github.com/microsoft/vscode/issues/176995 can be addressed.

Note that we've been using the same logic used by `vscode-nls`.  There is still a potential issue here if VS Code provides us with a language that the user does not have the language pack installed for.  In that case, we will display text in that langauge, but VS Code will default to the OS language(?).


sean-mcmanus (dismissed):

Colengms:
Sean pointed out that we appear to already have been handling the value correctly.  https://github.com/microsoft/vscode/issues/176995 would help, as it would make it clearly a VS Code issue when this value is determined incorrectly.

sean-mcmanus:
@Colengms So did you want to check this in for 1.15.0 to see if it helps or not?

sean-mcmanus (approved):

Colengms:
The underlying issue that led to this change should be addressed by:  https://github.com/microsoft/vscode/pull/177426

Since we've been doing this the same way as `vscode-nls`, and any changes that would break `vscode-nls` would break all extensions that uses `vscode-nls` until they upgrade to a new version, I think our current implementation is still acceptable.

PR #10670, opened 2023-03-14
Don't show the 'select a compiler' code action if the default compiler is already set (and other quick fixes)

sean-mcmanus:
1. Don't show the 'select a compiler' code action if the default compiler is already set.
2. Fix 'Select Default Compiler' should be made to update the workspace (or workspace folder) setting instead of appearing to not work: https://github.com/microsoft/vscode-cpptools/issues/10674
3. Fix Replace "includePath" with "compilerPath" in the 'Edit "includePath" setting' code action with for missing system includes: https://github.com/microsoft/vscode-cpptools/issues/10675
4. Fix clang-tidy 'clang-analyzer-' documentation links not working.: https://github.com/microsoft/vscode-cpptools/issues/10678


Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -168,9 +168,13 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
                 resultCodeActions.push(...disableCodeActions);
                 resultCodeActions.push(...docCodeActions);
                 return;
-            }
-            if (command.command === 'C_Cpp.CreateDeclarationOrDefinition' && (command.arguments ?? []).length === 0) {
+            } else if (command.command === 'C_Cpp.CreateDeclarationOrDefinition' && (command.arguments ?? []).length === 0) {
                 command.arguments = ['codeAction']; // We report the sender of the command
+            } else if (command.command === "C_Cpp.SelectDefaultCompiler") {
+                const defaultCompilerPath: string | undefined = new CppSettings(this.client.RootUri).defaultCompilerPath;
```
bobbrow: I think it would be better to hide the code action if `compilerPath` is set in the active configuration (c_cpp_properties.json) instead.  The user could have selected an unsupported compiler previously, so `C_Cpp.default.compilerPath` could be set, but it's "unsuitable."

sean-mcmanus: sure, I'll also looking into https://github.com/microsoft/vscode-cpptools/issues/10674

sean-mcmanus:
> I'm testing, in particular, I'm looking into why I'm not seeing the "Add to include path" code action for non-system includes.

Oh, it was because I had the configurationProvider set.

Colengms (approved):

Issue #10685, opened 2023-03-15
Add a notification to the status bar when selecting a compiler will improve IntelliSense


bobbrow:
### Feature Request

There are many ways to configure IntelliSense, and to be the most effective we are looking to show this to a specific subgroup of users who will get the most help out of being redirected to the compiler quickpick.  These users are: 
-	Users who have not configured IntelliSense using a compiler, custom configuration provider, or  compile_commands.json 
-	Users who have configured a custom configuration provider or compile_commands, however they are falling back on the base configuration instead 
For these groups, we will add the item (design below) to the status bar. If we determine that users do not meet the requirements above anymore, the status bar icon will be removed. 

 ![image](https://user-images.githubusercontent.com/12818240/225454099-e6077f74-e177-49f8-8336-55ff6e0bd3d1.png)

Clicking on this will pull up the quickpick to select a compiler (the same effect as the Select Default Compiler command in the command pallete). Once a user has configured a compiler for IntelliSense or made a selection in the quickpick (such as selecting a compiler or choosing “”), this item will disappear from the status bar.

bobbrow:
This issue replaces #10527

PR #10690, opened 2023-03-16
Add "Configure Intellisense" Status Bar Item

browntarik:
Work item reference (GitHub issue link, etc.): https://github.com/microsoft/vscode-cpptools/issues/10685

Problem:
• Users are having trouble configuring a compiler for IntelliSense even with our notification on startup
Fix:
• Add a persistent status bar item that will let the user know they do not have a compiler selected. This status bar item will call the  Select Default Compiler command

How tested:
• Manual Testing

Implementation Summary:
• Added compilerStatusBarItem to UI along with corresponding member variables which are set in client.ts


Extension/src/LanguageServer/ui_new.ts
```diff
@@ -112,6 +113,17 @@ export class NewUI implements UI {
         };
         this.ShowReferencesIcon = false;
 
+        this.compilerStatusItem = vscode.window.createStatusBarItem(`c.cpp.references.statusbar`, vscode.StatusBarAlignment.Right, 901);
```
sean-mcmanus: references->compiler

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -112,6 +113,17 @@ export class NewUI implements UI {
...
+        this.compilerStatusItem.name = localize("c.cpp.references.statusbar", "Intellisense Status");
```
sean-mcmanus: references->compiler

sean-mcmanus: Why is this "IntelliSense Status" -- seems like it should just be "Configure IntelliSense"

Extension/src/LanguageServer/ui.ts
```diff
@@ -104,6 +106,17 @@ export class OldUI implements UI {
         };
         this.ShowReferencesIcon = false;
 
+        this.compilerStatusItem = vscode.window.createStatusBarItem(`c.cpp.references.statusbar`, vscode.StatusBarAlignment.Right, 901);
```
sean-mcmanus: references->compiler

Extension/src/LanguageServer/ui.ts
```diff
@@ -104,6 +106,17 @@ export class OldUI implements UI {
...
+        this.compilerStatusItem.name = localize("c.cpp.compilerStatus.statusbar", "Compiler Status");
```
sean-mcmanus: Would "Configure IntelliSense" be more understandable to users since it would match the user text?

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
It still pops up the "You do not have IntelliSense configured" 2nd prompt if it gets configured from the Status Bar item (or some other means) -- is that fixable? I was thinking maybe it could check if the setting had gotten set already. Or maybe we could file another issue to track that.

sean-mcmanus (changes requested):
The status UI doesn't disappear after I set the C_Cpp.default.compilerPath in the settings.json

sean-mcmanus (changes requested):
Can we use a timer to delay showing the Status Bar item till the prompt has appeared for 15 seconds (when the auto-hide will occur) so they don't both appear at the same time?

image

You can use the setTimeout function (there are examples in the code).

Extension/src/LanguageServer/ui.ts
```diff
@@ -25,6 +25,7 @@ export interface UI {
     activeDocumentChanged(): void;
     bind(client: Client): void;
     showConfigurations(configurationNames: string[]): Promise<number>;
+    ShowCompilerStatusIcon(show: boolean): void;
```
bobbrow: it should be a lower-case 's' for "showCompilerStatusIcon"

sean-mcmanus (dismissed):
I think we need to fix the bug https://github.com/microsoft/vscode-cpptools/issues/10691 before adding this status bar UI.

sean-mcmanus (commented):
@bobbrow @browntarik Should the status bar item open the original info prompt instead of the Select Compiler dropdown if a potential compiler is found? That would make it easier to click the Yes button.

AlexandraKemperMS:
> > @bobbrow @browntarik Should the status bar item open the original info prompt instead of the Select Compiler dropdown if a potential compiler is found? That would make it easier to click the Yes button.
> 
> This is an interesting idea. @AlexandraKemperMS what do you think?

I like this idea, but I think for a lot of users the quickpick will be much more noticeable - around half of users dismiss the notification, and that's the ones who respond at all/see it in the first place. Also, the compiler we would normally recommend in the notification would be at the top of the list in the quickpick, so also the first thing to select.

sean-mcmanus:
> I like this idea, but I think for a lot of users the quickpick will be much more noticeable - around half of users dismiss the notification, and that's the ones who respond at all/see it in the first place. Also, the compiler we would normally recommend in the notification would be at the top of the list in the quickpick, so also the first thing to select.

Ok, that makes sense to me.

sean-mcmanus:
> I think we need to fix the bug #10691 before adding this status bar UI.

Bob says this can be fixed afterwards.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -174,6 +174,7 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
             } else if (command.command === 'C_Cpp.CreateDeclarationOrDefinition' && (command.arguments ?? []).length === 0) {
                 command.arguments = ['codeAction']; // We report the sender of the command
             } else if (command.command === "C_Cpp.SelectDefaultCompiler") {
+                command.arguments = ['statusBar'];
```
bobbrow: this should be 'codeAction'

sean-mcmanus: Isn't this the code action and not the status bar?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -296,6 +309,21 @@ export class OldUI implements UI {
...
 
+    private compilerTimout?: NodeJS.Timeout;
+    public async showCompilerStatusIcon(show: boolean): Promise<void> {
+        if (this.compilerTimout) {
+            clearTimeout(this.compilerTimout);
+        }
+        const action: string = "";
+        if (show) {
+            this.compilerTimout = setTimeout(() => { this.compilerStatusItem.show(); }, 15000);
+            telemetry.logLanguageServerEvent('compilerStatusBar', { 'compiler status bar shown' : action });
```
sean-mcmanus: This is missing from the newUI case?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1057,7 +1061,7 @@ export class DefaultClient implements Client {
                     this.showPrompt(selectCompiler, true);
                     action = "dismissed";
                 }
-                telemetry.logLanguageServerEvent('compilerNotification', { action });
+                telemetry.logLanguageServerEvent('compilerNotification', { action }, { sender });
```
bobbrow: I believe the args should be `{ action, sender: getSenderType(sender) }`

(`getSenderType` is defined in extension.ts, but the function should move to common.ts and be marked with `export` so you can use it here)

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Also, shouldn't the telemetry be when it's actually shown?

Extension/src/LanguageServer/client.ts
```diff
@@ -1186,6 +1190,7 @@ export class DefaultClient implements Client {
                         compilerDefaults = await this.requestCompiler(compilerPaths);
                         DefaultClient.updateClientConfigurations();
                         if (!compilerDefaults.trustedCompilerFound && !displayedSelectCompiler && (compilerPaths.length !== 1 || compilerPaths[0] !== "")) {
+                            await ui.showCompilerStatusIcon(true);
                             // if there is no compilerPath in c_cpp_properties.json, prompt user to configure a compiler
                             this.promptSelectCompiler(false);
```
bobbrow: `promptSelectCompiler` should be called with a second argument for "sender".  In this case the name could be 'initialization' or something like that?

Extension/src/LanguageServer/extension.ts
```diff
@@ -541,8 +541,9 @@ function onResetDatabase(): void {
     clients.ActiveClient.resetDatabase();
 }
 
-function selectDefaultCompiler(): void {
-    clients.ActiveClient.promptSelectCompiler(true);
+function selectDefaultCompiler(sender?: any): void {
+    logForUIExperiment("CompilerSelect", sender);
```
bobbrow: I believe this feature is part of a different experiment.  This function does the logging for the language status experiment, so you'll need to set up a new one (look at the existing one for information).

Extension/src/LanguageServer/ui.ts
```diff
@@ -296,6 +309,21 @@ export class OldUI implements UI {
...
     }
 
+    private compilerTimout?: NodeJS.Timeout;
+    public async showCompilerStatusIcon(show: boolean): Promise<void> {
+        if (this.compilerTimout) {
+            clearTimeout(this.compilerTimout);
+        }
+        const action: string = "";
+        if (show) {
+            this.compilerTimout = setTimeout(() => { this.compilerStatusItem.show(); }, 15000);
```
bobbrow: I think there was a condition for doing a timeout vs. just showing it.  Check with @sean-mcmanus on that.

browntarik: @sean-mcmanus what do you think?

Extension/src/LanguageServer/ui.ts
browntarik: I am not sure what you mean by "when it's actually shown". Do you mean put it after the "show" function call?

bobbrow: It means put it inside the lambda for the `setTimeout`.

Also delete the `{ 'compiler status bar shown' : action }` part. `action` should be on the left of the colon, but we don't need this anyway.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: what?

sean-mcmanus: I think you mentioned that you wanted to force the status bar to be shown before the timeout if the user explicitly closes the prompts. I personally didn't think it mattered too much though. Up to you guys.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: yeah

Extension/src/LanguageServer/extension.ts
```diff
@@ -542,8 +540,8 @@ function onResetDatabase(): void {
 }
 
 function selectDefaultCompiler(sender?: any): void {
-    logForUIExperiment("CompilerSelect", sender);
-    clients.ActiveClient.promptSelectCompiler(true,sender);
+    logForCompilerExperiment("CompilerSelect", sender);
```
bobbrow: The sender is already logged in `promptSelectCompiler`. Do we need to log another event here?

What we're actually missing is the code that queries the experiment to determine whether the new status bar button should appear or not.

bobbrow: Fixed this offline with @browntarik.  Will be in the next push

Extension/src/LanguageServer/ui.ts
```diff
@@ -104,6 +106,17 @@ export class OldUI implements UI {
         };
         this.ShowReferencesIcon = false;
 
+        this.compilerStatusItem = vscode.window.createStatusBarItem(`c.cpp.compilerStatus.statusbar`, vscode.StatusBarAlignment.Right, 901);
+        this.compilerStatusItem.name = localize("c.cpp.compilerStatus.statusbar", "Configure IntelliSense");
+        this.compilerStatusItem.text = `$(warning) Configure Intellisense`;
```
bobbrow: This needs to be localized. Do something like:
```ts
  `$(warning) ${this.compilerStatusItem.name}`
```

Extension/src/LanguageServer/ui.ts
```diff
@@ -296,6 +309,25 @@ export class OldUI implements UI {
...
+            clearTimeout(this.compilerTimout);
+        }
+        if (show) {
+            this.compilerTimout = setTimeout(() => {
+                this.compilerStatusItem.show();
+                telemetry.logLanguageServerEvent('compilerStatusBar');
+            }, 15000);
+        } else {
+            this.compilerStatusItem.hide();
+            telemetry.logLanguageServerEvent('compilerStatusBar');
```
bobbrow: Don't log telemetry for the `hide`

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -112,6 +114,17 @@ export class NewUI implements UI {
         };
         this.ShowReferencesIcon = false;
 
+        this.compilerStatusItem = vscode.window.createStatusBarItem(`c.cpp.compilerStatus.statusbar`, vscode.StatusBarAlignment.Right, 901);
+        this.compilerStatusItem.name = localize("c.cpp.compilerStatus.statusbar", "Configure IntelliSense");
+        this.compilerStatusItem.text = `$(warning) Configure Intellisense`;
```
bobbrow: This needs to be localized like in the other UI file.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -403,6 +416,25 @@ export class NewUI implements UI {
...
+            clearTimeout(this.compilerTimout);
+        }
+        if (show) {
+            this.compilerTimout = setTimeout(() => {
+                this.compilerStatusItem.show();
+                telemetry.logLanguageServerEvent('compilerStatusBar');
+            }, 15000);
+        } else {
+            this.compilerStatusItem.hide();
+            telemetry.logLanguageServerEvent('compilerStatusBar');
```
bobbrow: don't log telemtry for `hide`

Extension/src/telemetry.ts
```diff
@@ -139,6 +139,15 @@ export function logLanguageServerEvent(eventName: string, properties?: { [key: s
     sendTelemetry();
 }
 
+export async function showStatusBarIntellisenseIndicator(): Promise<boolean> {
```
sean-mcmanus: This should be IntelliSense (capital S). I'll submit a PR to fix the pre-existing issues.

Extension/src/telemetry.ts
```diff
@@ -139,6 +139,15 @@ export function logLanguageServerEvent(eventName: string, properties?: { [key: s
...
+    if (new CppSettings().experimentalFeatures) {
+        return true;
+    }
+    const experimentationService: IExperimentationService | undefined = await getExperimentationService();
+    const useNewUI: boolean | undefined = experimentationService?.getTreatmentVariable<boolean>("vscode", "showStatusBarIntellisenseIndicator");
```
sean-mcmanus: same here

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -403,6 +416,24 @@ export class NewUI implements UI {
...
+    public async showCompilerStatusIcon(show: boolean): Promise<void> {
+        if (!telemetry.showStatusBarIntellisenseIndicator()) {
+            return;
+        }
+        if (this.compilerTimout) {
+            clearTimeout(this.compilerTimout);
+        }
+        if (show) {
+            this.compilerTimout = setTimeout(() => {
+                this.compilerStatusItem.show();
```
sean-mcmanus, line 430: Doesn't it need to re-check if the condition still exists to show it, since the user could have already set it in before the 15 seconds?

sean-mcmanus (changes requested):

Extension/src/telemetry.ts
bobbrow: This needs to match whatever Alex configured for the treatment variable. It's not publicly facing, so it doesn't matter in the end.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -403,6 +416,26 @@ export class NewUI implements UI {
...
+    public async showCompilerStatusIcon(show: boolean): Promise<void> {
+        if (!telemetry.showStatusBarIntelliSenseIndicator()) {
+            return;
+        }
+        if (this.compilerTimout) {
+            clearTimeout(this.compilerTimout);
+        }
+        if (show) {
+            this.compilerTimout = setTimeout(() => {
+                if (show) {
```
sean-mcmanus: This bool is passed by value, so it's value won't change -- it needs to be re-evaluated, i.e. via either copying the root checks or passing a lambda function.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui_new.ts
browntarik: Can you give a code snippet?

Extension/src/LanguageServer/ui.ts
```diff
@@ -294,6 +307,26 @@ export class OldUI implements UI {
         }
     }
 
+    private compilerTimout?: NodeJS.Timeout;
+    public async showCompilerStatusIcon(show: boolean): Promise<void> {
+        if (!telemetry.showStatusBarIntelliSenseIndicator()) {
+            return;
```
michelleangela, line 313: Does `this.compilerTimout` need to be reset for this case? Or will it be undefined already when `!telemetry.showStatusBarIntelliSenseIndicator()`?

Extension/src/LanguageServer/ui.ts
```diff
@@ -314,19 +314,16 @@ export class OldUI implements UI {
...
-                isShowTrue = showVariable(isShowTrue);
-                if (isShowTrue) {
-                    this.compilerStatusItem.show();
-                }
+                this.compilerStatusItem.show();
                 telemetry.logLanguageServerEvent('compilerStatusBar');
             }, 15000);
         } else {
             this.compilerStatusItem.hide();
+            this.compilerTimout = undefined;
```
bobbrow: This should go in the lambda called by `setTimeout` (e.g. line 322), not here.

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -403,6 +416,26 @@ export class NewUI implements UI {
...
+            this.compilerTimout = undefined;
+        }
+        if (show) {
+            this.compilerTimout = setTimeout(() => {
+                this.compilerStatusItem.show();
+                telemetry.logLanguageServerEvent('compilerStatusBar');
+            }, 15000);
+        } else {
+            this.compilerStatusItem.hide();
+            this.compilerTimout = undefined;
```
bobbrow: This should go in the lambda called by `setTimeout` (e.g. line 431), not here.

Extension/src/LanguageServer/ui.ts
bobbrow: No, we should never enter the rest of the function (the result of this function should never change during the session), so the timeout should never be set.

browntarik: If it reaches this case then I believe it will be undefined already

bobbrow (approved):

sean-mcmanus (approved):

Issue #16, opened 2016-04-17
Call hierarchy


csholmq:
Perhaps the inverse to #15; to be able to view an entire call hierarchy of a function. 

Seen it implemented in Eclipse.

DanielGibson:
Having this feature would be awesome!
In Eclipse it also works for class fields, enum constants and global variables (showing functions they're used in, and expanding the function call hierarchy from there). To be better than Eclipse, making it work for macros as well would be extra cool :)

It's really invaluable to understand and navigate non-small code bases.

sean-mcmanus:
Visual Studio has this, but VS Code doesn't currently have any interface to enable a call hierarchy (other than possibly some unusual interface involving repeated dropdown popups). Maybe someone could file a feature request with VS Code to enable this, if it's not already been requested? This would require the IntelliSense engine from our end. It's not on our radar to implement this yet though (Find All References would be higher priority).

hoffmael:
Adding my vote to this one. This would be my number 1 most important missing feature and use it daily in eclipse to navigate and understand the structure of a large code base. I can live without find all references feature and use global text search with whole word option instead.

sean-mcmanus:
See https://github.com/Microsoft/vscode/issues/16110 .

gulshan:
A new tree view api has been added to vscode. So, this should be doable now.

sean-mcmanus:
Looks like https://code.visualstudio.com/docs/extensionAPI/vscode-api#TreeDataProvider .

hoffmael:
Can I vote twice for this one? The well implemented call hierarchy is the main reason eclipse is still my go-to IDE as its an invaluable tool to navigate 1M+ LOC code bases.

sandipanbiswas:
Awesome feature to have! Right there on the top of bucket list..

jguiban:
+1 for having this useful feature integrated soon

KarolisMil:
I would also appreciate this feature. Highly used feature... I finally could switch from eclipse to Visual code.

sean-mcmanus:
@KarolisMil Yes, this is our second most up-voted feature request (behind find all references). Survey feedback indicates that too many users are having trouble getting IntelliSense setup so we're working on improving that first.

aisnote:
I need the Call hierarchy too!

iradiohead:
anynews

sean-mcmanus:
[...]

PR #10697, opened 2023-03-17
Call hierarchy feature

michelleangela:
Issue: https://github.com/microsoft/vscode-cpptools/issues/16

Add `CallHierarchyProvider` class to implement VS Code's `vscode.CallHierarchyProvider` API.

Note: this changeset also disables indentation check because it is not reliable (https://github.com/typescript-eslint/typescript-eslint/issues/1824)

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -1,0 +1,69 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import * as vscode from 'vscode';
+import { DefaultClient } from '../client';
+import { processDelayedDidOpen } from '../extension';
+
+export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
+    private client: DefaultClient;
+
+    constructor(client: DefaultClient) {
+        this.client = client;
+    }
+
+    public async prepareCallHierarchy(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken):
+    Promise<vscode.CallHierarchyItem | vscode.CallHierarchyItem[] | undefined> {
+        const range: vscode.Range | undefined = document.getWordRangeAtPosition(position);
+        if (range === undefined) {
+            return undefined;
+        }
+
+        const symbol: string = document.getText(range);
+        await this.client.requestWhenReady(() => processDelayedDidOpen(document));
+
+        // Get call items denoted by given document and position
+        // items = await this.client.languageClient.sendRequest(PrepareCallHierarchyRequest, params, token);
+
+        // create a vscode.CallHierarchyItem for each returned call item.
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -0,0 +1,199 @@
...
+    position: Position;
+}
+
+interface CallHierarchyItemResult {
+    item?: CallHierarchyItem;
+
+    /**
+     * If a request is cancelled, `succeeded` will be undefined.
+     */
+    succeeded: boolean;
```
sean-mcmanus, line 56: The comment suggests it should be `succeeded?: boolean`?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: This is not defined as optional on the native side. But for the case when the request was cancelled. So it is not intended as `succeeded?`

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -0,0 +1,199 @@
...
+
+interface CallHierarchyCallsItemResult {
+    calls: CallHierarchyCallsItem[];
+}
+
+const CallHierarchyItemRequest: RequestType<CallHierarchyParams, CallHierarchyItemResult, void> =
+    new RequestType<CallHierarchyParams, CallHierarchyItemResult, void>('cpptools/prepareCallHierarchy');
+
+const CallHierarchyCallsToRequest: RequestType<CallHierarchyParams, CallHierarchyCallsItemResult, void> =
+new RequestType<CallHierarchyParams, CallHierarchyCallsItemResult, void>('cpptools/callHierarchyCallsTo');
```
sean-mcmanus: Can you run Format Document? There are 4 cases where lines are missing indentation.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
sean-mcmanus: Actually, it looks like lint complains about the other 3 cases, so this is the only case.

sean-mcmanus: From looking at https://github.com/typescript-eslint/typescript-eslint/issues/1824 it sounds like we may want to disable the linter indent rule (which would enable Format Document results to be kept). That can be done later in another PR though.

michelleangela: The other cases results in a yarn lint error. So Only the first case can be formatted. I'll update the yarn lint.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: Comment has been updated to clarify condition.

sean-mcmanus (approved):

Colengms (approved):

PR #10717, opened 2023-03-21
Refactor experiment variable

bobbrow:
Move logic for experiments into the telemetry module closer to the experimentation service. This is to set the pattern for future experiments so that they are all defined in the same place.

### DIFF
--- a/Extension/src/telemetry.ts
+++ b/Extension/src/telemetry.ts
@@ -78,10 +78,12 @@ export function getExperimentationService(): Promise<IExperimentationService> | 
 }
 
 export async function showLanguageStatusExperiment(): Promise<boolean> {
+    if (new CppSettings().experimentalFeatures) {
+        return true;
+    }
     const experimentationService: IExperimentationService | undefined = await getExperimentationService();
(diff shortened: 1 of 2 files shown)


Extension/src/telemetry.ts
```diff
@@ -76,6 +77,13 @@ export function getExperimentationService(): Promise<IExperimentationService> |
     return initializationPromise;
 }
 
+export async function showLanguageStatusExperiment(): Promise<boolean> {
+    const experimentationService: IExperimentationService | undefined = await getExperimentationService();
```
sean-mcmanus, line 84: Can you check the setting first? i.e.
```ts
    if (new CppSettings().experimentalFeatures) {
        return true;
    }
    const experimentationService: IExperimentationService | undefined = await getExperimentationService();
    const useNewUI: boolean | undefined = experimentationService?.getTreatmentVariable<boolean>("vscode", "ShowLangStatBar");
    return useNewUI ?? false;
```

sean-mcmanus (changes requested):

Extension/src/telemetry.ts
bobbrow: sure

michelleangela (approved):

sean-mcmanus (approved):

Issue #10742, opened 2023-03-26
cppbuild tasks do not set cwd to the workspace folder


akbyrd:
### Environment

- OS and Version: 22H2 19045
- VS Code Version: 1.76.2
- C/C++ Extension Version: 1.14.5

### Bug Summary and Steps to Reproduce

Bug Summary: shell tasks set the current working directory to the workspace folder. cppbuild tasks do not. This means that converting a working shell task to a cppbuild task does not work.

The working directory is set to the installation directory of vscode, which doesn't seem particularly useful.

Steps to reproduce:
* Execute this task:
    ```
    {
        "label": "build",
        "group": "build",
        "type": "shell",
        "command": "dir",
        "args": [],
    }
    ```
* Notice cwd is the workspace folder
* Execute this task:
    ```
    {
        "label": "build",
        "group": "build",
        "type": "cppbuild",
        "command": "dir",
        "args": [],
    }
    ```
* Notice cwd is the vscode installation folder

Expected behavior:
* shell and cppbuild tasks have the same working directory
* cppbuild tasks use the workspace folder as the working directory

### Configuration and Logs

```shell
N/A
```

sean-mcmanus:
You can workaround this via adding
```json
            "options": {
                "cwd": "${workspaceFolder}"
            },
```
to the task in tasks.json

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.15.2 .

PR #10747, opened 2023-03-27
Set a default cwd for cppbuild tasks.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cpptools/issues/10742 .

There's a TODO about how to make it better for multi-root, but this seems good enough for now.


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -393,6 +393,16 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
         }
         if (this.options.cwd) {
             this.options.cwd = util.resolveVariables(this.options.cwd);
+        } else {
+            const editor: TextEditor | undefined = window.activeTextEditor;
+            let folder: WorkspaceFolder | undefined = editor ? workspace.getWorkspaceFolder(editor.document.uri) : undefined;
+            if (!folder && workspace.workspaceFolders) {
+                // TODO: Use the workspace folder for the tasks.json?
```
michelleangela, line 400: Is there a way to get the folder from tasks.json if tasks.json exists?

michelleangela (approved):

Colengms (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
sean-mcmanus: I couldn't find a way -- that's why I just added  TODO.

Issue #10743, opened 2023-03-26
Dozen instances of cpptools-srv freeze PC after "Replace in files" operation


1vanK:
### Environment

OS and Version: Linux Mint 21 Cinnamon 5.4.12

VS Code Version:
```
Version: 1.76.2
Commit: ee2b180d582a7f601fa6ecfdad8d9fd269ab1884
Date: 2023-03-14T17:57:21.103Z
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Linux x64 5.15.0-60-generic
Sandboxed: Yes
```
C/C++ Extension Version: 
```
Name: C/C++
Id: ms-vscode.cpptools
Description: C/C++ IntelliSense, debugging, and code browsing.
Version: 1.14.5
Publisher: Microsoft
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools

Name: C/C++ Extension Pack
Id: ms-vscode.cpptools-extension-pack
Description: Popular extensions for C++ development in Visual Studio Code.
Version: 1.3.0
Publisher: Microsoft
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools-extension-pack
```

### Bug Summary and Steps to Reproduce

Steps to reproduce:
1) Download repo:
```
git clone https://github.com/dviglo/dviglo repo
git -C repo reset --hard dff0ffaca4c0d5dfb84a37f5fb778b5f1bad109c
```
2) Goto repo folder
3) `code .`
4) `Edit` -> `Replace in Files`
5)
![screen](https://user-images.githubusercontent.com/13021826/227774557-2d6f98e6-21da-46e2-a4ec-d92a96473dd7.png)
6) Click `Replace All`

Bug Summary:

After the successful rename, a large number of `cpptools-srv` instances are launched. Each of them starts an active access to the hard disk (HDD, not SSD). And the OS stops responding. I have to restart the computer with a button on the case.

### Configuration and Logs

```shell
`c_cpp_properties.json`: I have no this file

Log Diagnostics:

-------- Diagnostics - 26.03.2023, 15:24:57
Version: 1.14.5
Current Configuration:
{
    "name": "Linux",
    "includePath": [
        "/home/user/gamedev/dviglo/repo/**"
    ],
    "defines": [],
    "cStandard": "c17",
    "cppStandard": "c++14",
[...]

PR #10748, opened 2023-03-28
Prevent launching lots of IntelliSense processes from find/replace.

sean-mcmanus:
Fix bug https://github.com/microsoft/vscode-cpptools/issues/10743 .

I've also filed a bug on VS Code itself at https://github.com/microsoft/vscode/issues/178461


sean-mcmanus:
@bobbrow Since this bug is capable of freezing a user's entire OS due to hogging all the CPU and memory, I think we should ship this in a 1.14.6 patch.

bobbrow (dismissed):

sean-mcmanus:
@bobbrow But what release do you think this should target? 1.14.6 and/or 1.15.2?

bobbrow:
Yeah, we can do a patch for 1.14.

michelleangela (dismissed):

Colengms (approved):

Issue #10691, opened 2023-03-16
"Do you want to configure IntelliSense with this compiler?" prompt appears even though the compiler is sent by the configurationProvider


sean-mcmanus:
![image](https://user-images.githubusercontent.com/19859882/225740540-666ea9ea-9ad0-41d3-b9e1-1c7579dad74b.png)

The compilerPath sent by the configurationProvider is already being used to configure IntelliSense so we shouldn't be showing the prompt (or the new status UI warning).

PR #10753, opened 2023-03-28
Remove compiler status when configProver/compile_commands is updated

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10691, https://github.com/microsoft/vscode-cpptools/issues/10754


sean-mcmanus:
Does this fix the compile_commands.json case too?

browntarik:
> Does this fix the compile_commands.json case too?

No, I am doing that in a separate PR

sean-mcmanus:
I don't think the timer is a good approach -- we don't know how long it will take for the configuration provider or compile commands compilerPath will be available. In particular, I'm seeing it sometimes take > 10 seconds.

Can we just check if the configuration provider or compile commands is set at all and then disable the compiler selection prompt?

sean-mcmanus (dismissed):
See my other comment

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -1551,6 +1553,15 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
+                if (changedSettings["legacyCompilerArgsBehavior"]) {
```
bobbrow: This looks like a second copy of lines 1553-1555?

Extension/src/LanguageServer/client.ts
```diff
@@ -1551,6 +1553,15 @@ export class DefaultClient implements Client {
...
+                    this.configuration.handleConfigurationChange();
+                }
+                if (changedSettings["default.compilerPath"]) {
+                    ui.showCompilerStatusIcon(false);
+                }
+                if (changedSettings["default.compileCommands"]) {
```
bobbrow: Do we need to add a case for `default.configurationProvider`?  Since the body of the `if` is the same for each changed case, you can probably `||` all the conditions together into a single `if`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1206,14 +1206,16 @@ export class DefaultClient implements Client {
...
-                        compilerDefaults = await this.requestCompiler(compilerPaths);
-                        DefaultClient.updateClientConfigurations();
-                        if (!compilerDefaults.trustedCompilerFound && !displayedSelectCompiler && (compilerPaths.length !== 1 || compilerPaths[0] !== "")) {
-                            await ui.showCompilerStatusIcon(true);
-                            // if there is no compilerPath in c_cpp_properties.json, prompt user to configure a compiler
-                            this.promptSelectCompiler(false, 'initialization');
-                            displayedSelectCompiler = true;
-                        }
+                        const timer: NodeJS.Timeout = global.setTimeout(async () => {
+                            clearTimeout(timer);
```
bobbrow: I'm not sure what this does.  If you're exercising this code, the timer has already fired and doesn't need to be cleared, right?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -348,6 +350,9 @@ export class CppProperties {
         if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady) {
             const configuration: Configuration | undefined = this.CurrentConfiguration;
             if (configuration) {
+                if (configuration.compilerPath !== undefined || configuration.compileCommands !== undefined) {
```
bobbrow: Also check configurationProvider?

bobbrow (changes requested):

browntarik:
> I don't think the timer is a good approach -- we don't know how long it will take for the configuration provider or compile commands compilerPath will be available. In particular, I'm seeing it sometimes take > 10 seconds.
> 
> Can we just check if the configuration provider or compile commands is set at all and then disable the compiler selection prompt?

@bobbrow @AlexandraKemperMS @Colengc What do we think?

bobbrow:
> > I don't think the timer is a good approach -- we don't know how long it will take for the configuration provider or compile commands compilerPath will be available. In particular, I'm seeing it sometimes take > 10 seconds.
> > Can we just check if the configuration provider or compile commands is set at all and then disable the compiler selection prompt?
> 
> @bobbrow @AlexandraKemperMS @Colengc What do we think?

I agree with @sean-mcmanus

AlexandraKemperMS:
> > I don't think the timer is a good approach -- we don't know how long it will take for the configuration provider or compile commands compilerPath will be available. In particular, I'm seeing it sometimes take > 10 seconds.
> > Can we just check if the configuration provider or compile commands is set at all and then disable the compiler selection prompt?
> 
> @bobbrow @AlexandraKemperMS @Colengc What do we think?

I also agree with bob - I think this would have been a nice to have, but the feature works fine if we just disable the indicator if a config provider/compile commands is set

Extension/src/LanguageServer/client.ts
```diff
@@ -1551,6 +1551,12 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
+                if (changedSettings["default.compilerPath"] || changedSettings["default.configurationProvider"]) {
+                    ui.showCompilerStatusIcon(false);
+                }
+                if (changedSettings["default.compileCommands"]) {
```
sean-mcmanus: Can you add this to the previous block?

Extension/src/LanguageServer/client.ts
michelleangela: `changedSettings["default.compileCommands"]` can also be part of the conditional `changedSettings["default.compilerPath"] || changedSettings["default.configurationProvider"]` because both conditional statements result in doing the same logic when both are true.

Extension/src/LanguageServer/client.ts
```diff
@@ -1551,6 +1551,12 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
+                if (changedSettings["default.compilerPath"] || changedSettings["default.configurationProvider"]) {
+                    ui.showCompilerStatusIcon(false);
```
sean-mcmanus: Can you add a check for the condition, i.e. `changedSettings["default.compilerPath"] !== undefined || changedSettings["default.compileCommands"] !== undefined || changedSettings["default.configurationProvider"] !== undefined`, since the changed setting might be to change it a setting to undefined?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

sean-mcmanus:
@bobbrow Can you re-review this?

Colengms (approved):

michelleangela (approved):

bobbrow (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -348,6 +350,9 @@ export class CppProperties {
         if (this.configurationIncomplete && this.defaultIncludes && this.defaultFrameworks && this.vcpkgPathReady) {
             const configuration: Configuration | undefined = this.CurrentConfiguration;
             if (configuration) {
+                if (configuration.compilerPath !== undefined || configuration.compileCommands !== undefined || configuration.configurationProvider !== undefined) {
+                    ui.showCompilerStatusIcon(false);
```
sean-mcmanus, line 354: @tarik-brown FYI, ui is undefined here.

sean-mcmanus: I'll submit a PR with a fix.

PR #10765, opened 2023-03-31
remote attach: use absolute /bin/sh path on Linux

bluca:
When using machinectl as a transport to connect the debugger to a nspawn container, we need to pass full paths when running commands, as that's what machinectl expects. Use /bin/sh when running on Linux. This allows connecting VSCode to nspawn containers.

https://www.freedesktop.org/software/systemd/man/machinectl.html https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html

### DIFF
--- a/Extension/src/Debugger/attachToProcess.ts
+++ b/Extension/src/Debugger/attachToProcess.ts
@@ -112,6 +112,7 @@ export class RemoteAttachPicker {
         let parameterBegin: string = `$(`;
         let parameterEnd: string = `)`;
         let escapedQuote: string = `\\\"`;
+        let shPrefix: string = ``;
 
         const settings: CppSettings = new CppSettings();
         if (settings.useBacktickCommandSubstitution) {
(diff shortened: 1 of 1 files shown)


Extension/src/Debugger/attachToProcess.ts
```diff
@@ -122,12 +123,14 @@ export class RemoteAttachPicker {
 
         // Must use single quotes around the whole command and double quotes for the argument to `sh -c` because Linux evaluates $() inside of double quotes.
         // Having double quotes for the outerQuote will have $(uname) replaced before it is sent to the remote machine.
+        // Also use a full path on Linux, so that we can use transports that need a full patch such as 'machinectl' to connect to nspawn containers.
```
WardenGnaw: Is this Linux or all of Unix? This codepath will also affect macOS.

bluca: I don't run OSX, but /bin/sh is a thing on all unix derivatives

sean-mcmanus: "patch" should be "path" right?

sean-mcmanus (changes requested):

Extension/src/Debugger/attachToProcess.ts
bluca: indeed, fixed

bluca: I've changed to match exactly on "linux", just in case, as machinectl and nspawn run on linux only anyway

WardenGnaw (approved):

sean-mcmanus (approved):

bluca:
@WardenGnaw @sean-mcmanus thanks for the reviews and approvals - who can I ask about clicking the merge button?

PR #10769, opened 2023-04-04
Revert to StatusBarItem for select configuration

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10755


sean-mcmanus (dismissed):

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -438,14 +436,27 @@ export class NewUI implements UI {
...
                 isCppPropertiesJson = activeEditor.document.fileName.endsWith("c_cpp_properties.json");
                 if (isCppPropertiesJson) {
                     vscode.languages.setTextDocumentLanguage(activeEditor.document, "jsonc");
                 }
             }
+
+            // It's sometimes desirable to see the config and icons when making changes to files with C/C++-related content.
+            // TODO: Check some "AlwaysShow" setting here.
+            this.ShowConfiguration = isCpp || isCppPropertiesJson ||
+                activeEditor.document.uri.scheme === "output" ||
```
Colengms, line 455: Should this check if the output channel is one of ours?

Colengms: Do we need to display something when these are open?  The reasoning isn't obvious to me.

Colengms (changes requested):

Extension/src/LanguageServer/ui_new.ts
browntarik: I am not sure; I am unfamiliar with this code. @sean-mcmanus @bobbrow What do you guys think?

sean-mcmanus: This is "pre-existing" code. I don' t know if there's a way to detect if the selected document output channel is one of ours or not. It seems low priority and shouldn't block this PR though. We could file an issue to track that if you want.

sean-mcmanus: We also have an active issue to "always" show the configuration with some setting regardless of the active file.

bobbrow: > I am not sure; I am unfamiliar with this code. @sean-mcmanus @bobbrow What do you guys think?

If this looks the same as what's currently in `ui.ts`, then I'm fine with what's here.  The goal of this change is to revert the configuration selection button back to its old place.

sean-mcmanus: yeah

Colengms (approved):

sean-mcmanus (approved):

PR #10771, opened 2023-04-05
Switch to Select IntelliSense Configuration

### LINKED ISSUE #10756: All in one configuration selection experience
This issue covers the implementation of the new all-in-one experience for onboarding configurations.

This feature will be enabled as part of the new status bar button A|B experiment.  Use code in `telemetry.ts` to determine whether the experiment is active.

* Disable notification messages related to configuration when the experiment is active
  * Allowing a configuration provider
  * Use a compile_commands.json that we discovered
  * "Configure your IntelliSense settings to help find missing headers"
  * We found [compiler], do you want to use it to configure IntelliSense
* We still keep logic that auto-selects stuff without a prompt regardless of the experiment state
  * auto-select configuration provider
  * auto-select trusted compilers
* If we detect that a user is not configured, we show the status bar button for "Configure IntelliSense" when the experiment is active (this should already be implemented). Clicking on this button will show a new QuickPick with all the configuration options (configuration provider/compile commands/compilerPath).  This new quick pick is shown instead of the "Select a Compiler" quick pick which is the current implementation.
  * see Mock up below
  * A user is “configured” if:
    * If the user has set configurationProvider, compileCommands, or compilerPath in c_cpp_properties.json  
      * Even if the configProvider or compileCommands doesn’t cover the active source file, for now, I think we want to consider those folks as “configured” because the new quick pick will suggest that they set one of these fields and that would be confusing.  We can revisit this decision in a “v2” if we change our minds, or the experiment data guides us this way
      * Setting compilerPath to “” should also be consider “configured”
    * If the user has set C_Cpp.default.configurationProvider, C_Cpp.default.compileCommands, or C_Cpp.default.compilerPath in the Workspace/Folder settings (User settings too? I believe this means that they used “select a compiler” already).
    * If we auto-enabled a configuration provider.
    * If we auto-selected a compiler for them and there is no better configuration option available (e.g. configurationProvider or compileCommands)

Mock up
image

Sample code
```ts
class Item implements vscode.QuickPickItem {
    constructor(public label: string, public description?: string, public detail?: string, public kind?: vscode.QuickPickItemKind) {
    }
}

const qp = vscode.window.createQuickPick<Item>();
qp.items = [
    new Item('extensions', undefined, undefined, vscode.QuickPickItemKind.Separator),
    new Item('Use information provided by CMake Tools'),
    new Item('Use information provided by Makefile Tools'),
    new Item('compile_commands.json', undefined, undefined, vscode.QuickPickItemKind.Separator),
    new Item('Use information from ${workspaceFolder}/build/compile_commands.json'),
    new Item('compilers', undefined, undefined, vscode.QuickPickItemKind.Separator),
    new Item('Use system headers from Visual Studio 2022'),
    new Item('Use system headers from g++', 'c:/msys2/mingw64/bin/g++.exe'),
    new Item('Select another compiler on my machine'),
    new Item('Help me install a compiler'),
    new Item('Don\'t configure IntelliSense for this workspace (not recommended)')
];
qp.title = 'Configure IntelliSense';
qp.placeholder = 'How would you like to configure IntelliSense for this workspace?';
qp.show();
qp.onDidAccept(() => {
    // `qp.selectedItems` should have the selected choice
    qp.hide();
    qp.dispose();
});
// there may be other events to subscribe to

### PR DESCRIPTION
**Pull request by sean-mcmanus**
Fixes https://github.com/microsoft/vscode-cpptools/issues/10756 .

Changes "Select Default Compiler" to "Select IntelliSense Configuration", which adds the configuration provider and compile commands options to the quick pick if they're detected.

### CODE REVIEW

File: Extension/src/LanguageServer/client.ts [hunk 1]
```diff
@@ -937,7 +942,14 @@ export class DefaultClient implements Client {
             if (isCompiler) {
                 const path: string | undefined = paths[i].replace(compilerName, "");
                 const description: string = localize("found.string", "Found at {0}", path);
-                items.push({ label: compilerName, description: description, index: i });
+                const label: string = localize("use.compiler", "Use the configuration from querying {0}", compilerName);
+                items.push({ label: label, description: description, index: i });
+            } else if (paths[i] === "configuration providers") {
+                items.push({ label: localize("configuration.providers", "configuration providers"), index: i, kind: vscode.QuickPickItemKind.Separator });
+            } else if (paths[i] === "compile commands") {
+                items.push({ label: localize("compile.commands", "compile commands"), index: i, kind: vscode.QuickPickItemKind.Separator });
```
bobbrow: I was thinking we keep this as `compile_commands.json` (unlocalized).  I'm not sure how "compile commands" translates and whether foreign language users would make that association since the file name is untranslated English.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -173,14 +173,14 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
                 return;
             } else if (command.command === 'C_Cpp.CreateDeclarationOrDefinition' && (command.arguments ?? []).length === 0) {
                 command.arguments = ['codeAction']; // We report the sender of the command
-            } else if (command.command === "C_Cpp.SelectDefaultCompiler") {
+            } else if (command.command === "C_Cpp.SelectIntelliSenseConfiguration") {
```
bobbrow, line 184: I think we still need to keep the old command around for this code action on the std headers.

sean-mcmanus: That code action now invokes the new quick pick (with potentially config providers and compile commands).

Extension/src/LanguageServer/client.ts
sean-mcmanus: sure

Extension/src/LanguageServer/Providers/codeActionProvider.ts
bobbrow: But is that what we want?  We don't enable the code action for cases where configProvider and compile commands are in use.

sean-mcmanus: Yeah, we want that -- if the config provider or compile commands are not set and system headers are not found, we want to have the config providers and compile commands in the quick pick if they're available and not just the compilers.

bobbrow: We need to keep the "select a compiler" quick pick (for now) for the walkthrough experiment.

sean-mcmanus: @bobbrow Ok

Extension/i18n/chs/package.i18n.json
```diff
@@ -17,6 +17,7 @@
 	"c_cpp.command.configurationEditJSON.title": "编辑配置(JSON)",
```
Colengms: Please don't modify  i18n.json files directory.  These are output files form the location process and will get stomped on.  It's not necessary to update them when adding strings.  If a string if missing, it will fall back to English automatically.  (Even if you're trying to preserve a localized string when just renaming the ID, it will still get stomped on with an English version).

Extension/src/LanguageServer/extension.ts
```diff
@@ -536,6 +537,10 @@ function selectDefaultCompiler(sender?: any): void {
     clients.ActiveClient.promptSelectCompiler(true, sender);
 }
 
+function selectIntelliSenseConfiguration(sender?: any): void {
+    clients.ActiveClient.promptSelectIntelliSenseConfiguration(true, sender);
```
Colengms: Shouldn't all(?) of these commands be waiting until the language service is ready before proceeding?  In the case of this function, we might not have received the compiler defaults from the native side until then.

Colengms (changes requested):
In ui_new.ts, there is a comment:

// TODO: Integrate with Tarik's feature to determine if compiler/bare-IntelliSense is configured

The code there is checking the wrong thing (incorrectly checking if IntelliSense is disabled via a setting to determine whether to display a Select Default Compiler).  And I assume it's incorrectly referring to the CheckForCompiler command instead of the SelectDefaultCompiler command.  Could you fix that up approproriate? (Or we can create a new bug for that).

(Walkthrough was the only place that used CheckForCompiler, so I'll remove it in that PR).

Extension/i18n/chs/package.i18n.json
sean-mcmanus: Okay, yeah, I wasn't sure about that.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Not sure, I was just following the existing code for promptSelectCompiler.

Colengms: The `selectDefaultCompiler` command callback is definitely wrong.  I had to add `notifyWhenLanguageClientReady` there in the `update-walkthrough` branch, otherwise use of that command from the walkthrough would easily fail whenever clicked when the extension was not yet fully started up.

sean-mcmanus: ok

sean-mcmanus:
FYI, I'm not done self-reviewing or self-testing, but I could use help.

Extension/src/LanguageServer/client.ts
```diff
@@ -69,6 +69,14 @@ let pendingTask: util.BlockingTask<any> | undefined;
 let compilerDefaults: configs.CompilerDefaults;
 let diagnosticsCollectionIntelliSense: vscode.DiagnosticCollection;
 let diagnosticsCollectionRefactor: vscode.DiagnosticCollection;
+
+interface ConfigStateReceived {
+    compilers: boolean;
+    compileCommands: boolean;
+    configProviders: CustomConfigurationProvider1[] | undefined;
```
bobbrow: style: prefer 
`configProviders?: CustomConfigurationProvider1[]`
over
`configProviders: CustomConfigurationProvider1[] | undefined`

Extension/src/LanguageServer/client.ts
```diff
@@ -792,6 +801,7 @@ export interface Client {
     addFileAssociations(fileAssociations: string, languageId: string): void;
     sendDidChangeSettings(): void;
     isInitialized(): boolean;
+    ShowConfigureIntelliSenseStatus(): boolean;
```
bobbrow: `showConfigureIntelliSenseStatus` (little 's')

Extension/src/LanguageServer/client.ts
```diff
@@ -830,6 +840,9 @@ export class DefaultClient implements Client {
     public lastCustomBrowseConfigurationProviderVersion: PersistentFolderState<Version> | undefined;
     private registeredProviders: PersistentFolderState<string[]> | undefined;
 
+    private configStateReceived: ConfigStateReceived = { compilers: false, compileCommands: false, configProviders: [], timeout: false };
```
bobbrow: The interface allows for undefined `configProviders`. Would that be preferred over empty array?

Extension/src/LanguageServer/client.ts
```diff
@@ -952,15 +977,31 @@ export class DefaultClient implements Client {
...
             secondPromptCounter++;
             if (value === buttonMessage) {
-                this.handleCompilerQuickPick(showSecondPrompt, sender);
+                this.handleIntelliSenseConfigurationQuickPick(showSecondPrompt, sender);
             }
         }
     }
 
-    public async handleCompilerQuickPick(showSecondPrompt: boolean, sender?: any): Promise<void> {
+    public async handleIntelliSenseConfigurationQuickPick(showSecondPrompt: boolean, sender?: any, compilersOnly?: boolean | undefined): Promise<void> {
```
bobbrow: `| undefined` is not needed when the argument is `compilersOnly?`

Extension/src/LanguageServer/client.ts
```diff
@@ -978,26 +1019,29 @@ export class DefaultClient implements Client {
...
-        paths.push(localize("selectAnotherCompiler.string", "Select another compiler on my machine"));
+        paths.push(localize("selectAnotherCompiler.string", "Select another compiler on my machine..."));
         paths.push(localize("installCompiler.string", "Help me install a compiler"));
-        paths.push(localize("noConfig.string", "Do not configure a compiler (not recommended)"));
-        const index: number = await this.showSelectDefaultCompiler(paths);
+        paths.push(localize("noConfig.string", "Do not configure with a compiler (not recommended)"));
+        const index: number = await this.showSelectIntelliSenseConfiguration(paths, compilersOnly);
         let action: string = "";
+        let configurationSelected: boolean = false;
+        const statusBarIndicatorEnabled: boolean = await telemetry.showStatusBarIntelliSenseIndicator();
```
bobbrow: For clarity, I would think "button" was a better description of what we're using over "indicator"

Extension/src/LanguageServer/ui.ts
```diff
@@ -307,24 +308,17 @@ export class OldUI implements UI {
         }
     }
 
-    private compilerTimout?: NodeJS.Timeout;
-    public async showCompilerStatusIcon(show: boolean): Promise<void> {
-        if (!telemetry.showStatusBarIntelliSenseIndicator()) {
+    public async showConfigureIntelliSenseStatusIcon(show: boolean, client?: Client): Promise<void> {
```
bobbrow: For clarity, I would think "button" was a better description of what we're using over "StatusIcon"

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -414,24 +415,17 @@ export class NewUI implements UI {
         }
     }
 
-    private compilerTimout?: NodeJS.Timeout;
-    public async showCompilerStatusIcon(show: boolean): Promise<void> {
-        if (!telemetry.showStatusBarIntelliSenseIndicator()) {
+    public async showConfigureIntelliSenseStatusIcon(show: boolean, client?: Client): Promise<void> {
```
bobbrow: Same here, I would think "button" was a better description of what we're using over "StatusIcon"

Extension/src/LanguageServer/client.ts
sean-mcmanus: It was capitalized to avoid a naming conflict with the lowercased variable name. I saw other methods start with a capital up above. I could change it though still.

sean-mcmanus: Yeah, if you have a suggestion for what to rename showConfigureIntelliSenseStatus to.

Extension/src/LanguageServer/client.ts
sean-mcmanus: yeah

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure, that's just what it was named previously.

Extension/src/LanguageServer/client.ts
bobbrow: You can leave it for now then.  Or call it `showConfigureIntelliSenseButton`.

Extension/src/LanguageServer/client.ts
sean-mcmanus: changed

Extension/src/LanguageServer/client.ts
sean-mcmanus: changed

Extension/src/LanguageServer/client.ts
bobbrow: > Sure, that's just what it was named previously.

Yeah, I know. It doesn't mean we have to live with it forever if there's a better option.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure, ending with Button is fine with me, although it makes it longer. But the issue with Capitalize and lowercase still exists unless we can think of some better naming convention.

Extension/src/LanguageServer/client.ts
sean-mcmanus: changed

Extension/src/LanguageServer/client.ts
bobbrow: FWIW, replace "Status" with "Button" is the same number of characters.

sean-mcmanus: Oh -- I added Button onto Status -- that's not what you wanted, i.e. StatusButton?

Extension/src/LanguageServer/client.ts
```diff
@@ -1017,26 +1061,54 @@ export class DefaultClient implements Client {
...
+                    settings.defaultCompilerPath = util.isCl(paths[index]) ? "cl.exe" : paths[index];
+                }
             }
 
+            ui.showConfigureIntelliSenseStatusButton(false, this);
             util.addTrustedCompiler(compilerPaths, settings.defaultCompilerPath);
             compilerDefaults = await this.requestCompiler(compilerPaths);
             DefaultClient.updateClientConfigurations();
         } finally {
             telemetry.logLanguageServerEvent('compilerSelection', { action, sender: util.getSenderType(sender)}, { compilerCount: paths.length });
```
bobbrow: We need to add a property to differentiate when this is "select a compiler" vs "select IntelliSense Configuration".  Or log a different event.  Ask @AlexandraKemperMS what works better for the experiment.

We need to fix "compilerCount" to be correct.  It's currently reporting the number of items in the quick pick.

Extension/src/LanguageServer/client.ts
bobbrow: I don't really care that much.  I thought it would be easier to just call it the ConfigureIntelliSenseButton instead of adding extra adjectives to it.  It's not really a "status".  It goes on the "status bar", but I didn't want all of that in the name either.

sean-mcmanus: Sure, I went with the shorter Button.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Sure, I'll look into that.

sean-mcmanus: Is what I did okay?

bobbrow: yes, it looks good.

bobbrow (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -926,19 +940,30 @@ export class DefaultClient implements Client {
...
-            const isCompiler: boolean = compilerName !== paths[i];
+            const isCompiler: boolean = isCompilerSection && compilerName !== paths[i];
 
             if (isCompiler) {
                 const path: string | undefined = paths[i].replace(compilerName, "");
                 const description: string = localize("found.string", "Found at {0}", path);
-                items.push({ label: compilerName, description: description, index: i });
+                const label: string = localize("use.compiler", "Use {0}", compilerName);
+                items.push({ label: label, description: description, index: i });
+            } else if (paths[i] === "configuration providers") {
```
bobbrow: suggestion: you might want to add const variables for these strings since they are used in more than one location.

sean-mcmanus: yeah, sounds good

bobbrow (dismissed):

bobbrow (dismissed):

bobbrow (dismissed):

Colengms (dismissed):

michelleangela (approved):

PR #10772, opened 2023-04-05
Fix auto provider detection

Colengms:
We're seeing a scenario in which one provider will register, get automatically selected because it's the only one (and then provide a browse configuration), but not get properly un-selected when a subsequent provider registers.

Removed a check that was opportunistically using a configuration provider if registered and if we received a browse configuration for it, as this scenario would lead to the 'too early' automatically selected config provider being kept.

We have a pattern in place to clear out any modifications we had made to the configuration (by calling `parsePropertiesFile` again) when we need to reprocess it.  But, that was a NO-OP if there was no `c_cpp_properties.json` file, so our modifications would not get removed in that scenario (making the prior registered configuration provider look like the user had explicitly specified it).  I updated that to instead use the default configuration, and I needed to fix one place where we didn't do that reset if the contents were empty.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -828,11 +828,12 @@ export class CppProperties {
         if (!this.configurationJson) {
             return;
         }
+        this.parsePropertiesFile(); // Clear out any modifications we may have made internally.
         const settings: CppSettings = new CppSettings(this.rootUri);
         const userSettings: CppSettings = new CppSettings();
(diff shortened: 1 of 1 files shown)


sean-mcmanus (changes requested):
This breaks the 1 provider auto-select scenario (the 2 provider scenario is fixed). I see this.configuration.CurrentConfigurationProvider return `''` instead of the 1 provider.

sean-mcmanus (approved):

michelleangela (approved):

PR #10792, opened 2023-04-09
Quick fix option to include header file

### LINKED ISSUE #10791: Quickfix for missing header file
### Feature Request

During development it is cumbersome to add missing header files. There should be a quickfix option on an unknown symbol which adds the corresponding header to the current file.

quickfix_for_symbolspng

The languageClient in src/LanguageServer/client.ts can be used to query the symbol information and determine the header location. I already implemented the feature as a draft ;)

m-zellner:
Closes: https://github.com/microsoft/vscode-cpptools/issues/10791

Implement a quick fix option for a missing symbol


Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -199,6 +199,20 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
+            command: {
+                title: "Add include header for symbol",
+                command: "C_Cpp.AddIncludeHeader",
+                arguments: [params.uri, range, token]
+            },
+            edit: undefined,
+            kind: vscode.CodeActionKind.QuickFix
+        };
+
+        resultCodeActions.push(includeHeaderCodeAction);
```
m-zellner, line 214: I couldnt find a better way to do that. Therefore add it by default.

Extension/src/LanguageServer/client.ts
```diff
@@ -1861,6 +1863,81 @@ export class DefaultClient implements Client {
...
+            vscode.window.showErrorMessage("Could not find symbol name");
+            return;
+        }
+
+        const symbolName: string = document.getText(symbolRange);
+        const params: WorkspaceSymbolParams = {
+            query: symbolName
+        };
+
+        const symbols: LocalizeSymbolInformation[] = await this.languageClient.sendRequest(GetSymbolInfoRequest, params, token);
```
m-zellner, line 1881: For large projects this approach is potentially to slow.

Extension/src/LanguageServer/client.ts
```diff
@@ -2581,7 +2658,7 @@ export class DefaultClient implements Client {
                 return false;
             });
         },
-        () => ask.Value = false);
+            () => ask.Value = false);
```
m-zellner: Woops something went wrong

Extension/src/LanguageServer/extension.ts
```diff
@@ -429,11 +431,12 @@ export function registerCommands(enabled: boolean): void {
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.GenerateDoxygenComment', enabled ? onGenerateDoxygenComment : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.CreateDeclarationOrDefinition', enabled ? onCreateDeclarationOrDefinition : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.RescanCompilers', enabled ? onRescanCompilers : onDisabledCommand));
+    commandDisposables.push(vscode.commands.registerCommand('C_Cpp.AddIncludeHeader', enabled ? addIncludeHeader : onDisabledCommand));
 }
 
 function logForUIExperiment(command: string, sender?: any): void {
     const settings: CppSettings = new CppSettings();
-    const properties: {[key: string]: string} = {
+    const properties: { [key: string]: string } = {
```
m-zellner: Mhm I guess im using the wrong formatter. I will change that

Extension/src/LanguageServer/client.ts
```diff
@@ -1861,6 +1863,81 @@ export class DefaultClient implements Client {
         await this.notifyWhenLanguageClientReady(() => this.languageClient.sendNotification(RescanFolderNotification));
     }
 
+    public async addIncludeHeader(filePath: string, range: vscode.Range, token: vscode.CancellationToken): Promise<void> {
```
m-zellner, line 1866: client.ts is already huge. I will extract the function if the feature is approved by a codeowner

M1Z3:
> @microsoft-github-policy-service agree

Extension/src/LanguageServer/Providers/codeActionProvider.ts
Colengms: Hi @M1Z3 .  Thanks for contributing this.  We'd like to do this.  As you aluded to here, we may be able to determine whether the code action is needed based on IntelliSense results, to avoid showing it when not needed.  So, we may do additional work (possibly in the native, closed-source part of the extension) and may not necessarily incorporate exactly what you've provided.

m-zellner: I totally agree. So how we are going to merge this? Do you want to change the IntelliSense implementation before merging this PR?

M1Z3:
Not needed anymore

Issue #10795, opened 2023-04-10
user settings C_Cpp > Default: Compiler Path behavior from setting empty string vs undefined


snehara99:
Hi, so I noticed a possible bug or unexpected behavior regarding the compiler path settings while trying to repro a user issue on vscode-cmake-tools ([3096](https://github.com/microsoft/vscode-cmake-tools/issues/3096)):

If I remove the default `compilerPath` from my user `settings.json` completely (set undefined), `compilerPathIsExplicit` depends only on the `compilerPath` setting in `c_cpp_properties.json`. However, if the default `compilerPath` in user settings is set to an empty string rather than being undefined, `compilerPathIsExplicit` is always `true`. 

I think this difference in behavior between the default compiler path in user settings being an empty string vs being undefined can be confusing, but I also understand this may be expected behavior. I just wanted to point it out.

However, I think a larger issue is when changing the default user settings via the UI rather than the `settings.json` file, an empty string versus undefined is indistinguishable. Erasing the default compiler path results in the default compiler path to be set to an empty string in the user `settings.json` file rather than undefined. I think this could cause unanticipated problems for users using the UI rather than the `settings.json` file.

I'm not sure if this is a bug, but I think it would help if some warning was given in the UI at least. Let me know if you'd like any clarification, thanks!

browntarik:
This is by design but displaying a notification in the language status UI may be helpful.

browntarik:
Since we need the VS Code UI to distinguish between an undefined or blank string for the optimal fix, we decided to keep this as known issue. This is a regression from an older fix we had for this issue so we may take a look at this as well.

Colengms:
https://github.com/microsoft/vscode-cpptools/pull/10797 somewhat addresses the issue in VS Code's settings UI, but does not address the same issue in `C/C++: Edit Configuration (UI)`.

sean-mcmanus:
The settings UI issue is fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.15.2 . I'm not sure if we need another issue to track the Edit Configuration (UI) case or not (feel free to open a new issue if you want).

PR #10797, opened 2023-04-10
Fix regression that resulted in `compilerPath` shown in settings UI

Colengms:
Fixed a regression introduced in https://github.com/microsoft/vscode-cpptools/pull/8129 , which caused the `default.compilerPath` field to become editable in VS Code settings UI.

This is related to: https://github.com/microsoft/vscode-cpptools/issues/10795

In `C_Cpp.default.compilerPath`, we consider the value of a blank string `""` as being distinct from null/undefined.  However, string settings displayed in VS Code settings UI do not make that distinction.  In lieu of removing this distinction from our code (so as not to break existing users), we prevented this field from being editable in the settings UI by using a 'type' that is not strictly just 'string'.

```
"C_Cpp.default.compilerPath": {
  "type": [
    "string",
    "null"
  ],
  "default": null,
  "markdownDescription": "%c_cpp.configuration.default.compilerPath.markdownDescription%",
  "scope": "machine-overridable"
},
```

This results in a link in VS Code's settings UI, referring the user to the JSON file to properly edit that field.

image

Alternatively, we could either:
a) Stop using `""` to indicate not to fall back to a default detected compiler, perhaps in favor of a separate setting. (i.e. `"fallbackToDetectedCompiler": false`.
b) Leave the existing imperfect behavior in place, as this regression had existed for 1.5 years before it was brought to our attention.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -614,7 +614,10 @@
             "scope": "machine-overridable"
           },
           "C_Cpp.default.compilerPath": {
-            "type": "string",
+            "type": [
+              "string",
+              "null"
+            ],
             "default": null,
             "markdownDescription": "%c_cpp.configuration.default.compilerPath.markdownDescription%",
             "scope": "machine-overridable"


sean-mcmanus (approved):

browntarik (approved):

browntarik (commented):
Will this break our users? Do we want to make this change now?

michelleangela (approved):

Issue #10810, opened 2023-04-12
"Configure IntelliSense" status bar button may not be visible if the status bar is full


sean-mcmanus:
We could potentially move it to the right more and/or collapse the text into an icon after 15 seconds.

PR #10829, opened 2023-04-15
Fix configure IntelliSense UI bugs with non-C/C++ files (and other small fixes)

### LINKED ISSUE #10827: "Configure IntelliSense" warning shows up in non-c++ folders if C++ extension gets activated
**Summary:**
The warning "Configure IntelliSense" shows up for folders that have no C++ files if the C++ extension gets activated.

**Repro steps:**
1. Ensure there is no default C++ IntelliSense configuration set under the user scope setting.
2. Open a folder that have no C++ files, like a Typescript project.
3. Open Walkthrough page for C++ extension.
4. Close walkthrough.
5. Observe "Configure IntelliSense" is on the status bar.

**Expected behavior:**
 "Configure IntelliSense" warning should only show up for workspaces with C++ files even if C++ extension gets activated.

sean-mcmanus:
- Fix https://github.com/microsoft/vscode-cpptools/issues/10827

  - Check for the valid file cases before showing the "Configure IntelliSense" UI -- previously it was only checking after the active file was changed.
  - Only show it for settings files if a C/C++ file or json has been made active.
  - Change the "name" to "C/C++ Configure IntelliSense" (shown in the right-click menu not the visible text/title) -- so users know which extension is adding that if they want to disable/hide it.
- Refactor isCpp checks. 
- Also, the fixes below were just 1 line fixes so I included them in this:
  - Fix a bug where the "Configure IntelliSense" would not disappear until reload window is done in the case where it's only being shown due mimicking the prompt for using compile commands or config provider and the dropdown is exited without a selection.
  - Fix a bug where the persistent state for showing the compile commands of config provider prompt would get cleared after the compiler selection dropdown was shown, even if not invoked from the "Configure IntelliSense" button.
  - Move the status bar location to the left of the configuration item so that it's less likely to disappear due to lack of space: part of https://github.com/microsoft/vscode-cpptools/issues/10810 .

### DIFF
--- a/Extension/src/LanguageServer/ui_new.ts
+++ b/Extension/src/LanguageServer/ui_new.ts
@@ -109,8 +109,9 @@ export class NewUI implements UI {
         this.ShowReferencesIcon = false;
 
         this.configureIntelliSenseStatusItem = vscode.window.createStatusBarItem(`c.cpp.configureIntelliSenseStatus.statusbar`, vscode.StatusBarAlignment.Right, 901);
-        this.configureIntelliSenseStatusItem.name = localize("c.cpp.configureIntelliSenseStatus.statusbar", "Configure IntelliSense");
-        this.configureIntelliSenseStatusItem.text = `$(warning) ${this.configureIntelliSenseStatusItem.name}`;
+        this.configureIntelliSenseStatusItem.name = localize("c.cpp.configureIntelliSenseStatus.cppText", "C/C++ Configure IntelliSense");
+        const configureIntelliSenseText: string = localize("c.cpp.configureIntelliSenseStatus.text", "Configure IntelliSense");
+        this.configureIntelliSenseStatusItem.text = `$(warning) ${configureIntelliSenseText}`;
         this.configureIntelliSenseStatusItem.backgroundColor = new vscode.ThemeColor('statusBarItem.warningBackground');
         this.configureIntelliSenseStatusItem.command = {
             command: "C_Cpp.SelectIntelliSenseConfiguration",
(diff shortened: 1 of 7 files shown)


Extension/src/LanguageServer/ui_new.ts
```diff
@@ -416,14 +417,27 @@ export class NewUI implements UI {
...
         if (!telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient) {
             return;
         }
         if (show) {
             this.showConfigureIntelliSenseButton = true;
-            this.configureIntelliSenseStatusItem.show();
+            const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
             telemetry.logLanguageServerEvent('configureIntelliSenseStatusBar');
+            if (activeEditor) {
+                const isCpp: boolean = activeEditor.document.uri.scheme === "file" && (activeEditor.document.languageId === "c" || activeEditor.document.languageId === "cpp" || activeEditor.document.languageId === "cuda-cpp");
```
michelleangela: Create a util function for `isCpp` and `isCppPropertiesJson` since the check is used more than once.

sean-mcmanus: sure

bobbrow: I feel like this was added in another PR, but we didn't take it yet. https://github.com/microsoft/vscode-cpptools/pull/10559/files#diff-1c5f3d6b951c5bd69dc09ef8a4bfa18807bb18d572579c2675f82a8a853284f7

michelleangela (approved):

Colengms (approved):

Issue #10238, opened 2022-12-07
Create declaration/definition - copy to clipboard


michelleangela:
Feature enhancement:

Add the ability to copy to clipboard the definition or declaration to create from a "create declaration or definition" operation.

michelleangela:
This is resolved in https://github.com/microsoft/vscode-cpptools/releases/tag/1.16.0.

PR #10835, opened 2023-04-18
Add copy to clipboard support for CDD

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10238


Extension/package.nls.json
```diff
@@ -32,6 +32,7 @@
     "c_cpp.command.GoToNextDirectiveInGroup.title": "Go to next preprocessor directive in conditional group",
     "c_cpp.command.GoToPrevDirectiveInGroup.title": "Go to previous preprocessor directive in conditional group",
     "c_cpp.command.CreateDeclarationOrDefinition.title": "Create Declaration / Definition",
+    "c_cpp.command.CopyDeclarationOrDefinition.title": "Copy declaration to clipboard",
```
sean-mcmanus: Is this string used? Why does it only mention "declaration" and not "definition"?

sean-mcmanus (changes requested):

Extension/package.nls.json
browntarik: This is currently not user facing in any way but you are correct, the string needs to be changed. If a command is not user facing, should we remove it?

sean-mcmanus: The string is visible in the Extension "Contributes" section where it lists all the commands. If the command isn't needed in the package.json for users to use in the command palette, then you could follow an example like with the C_Cpp.FixAllTypeCodeAnalysisProblems command which isn't in the package.json.

Extension/src/common.ts
```diff
@@ -910,6 +910,17 @@ export async function promptReloadWindow(message: string): Promise<void> {
     }
 }
 
+export async function promptCDDFailed(copy?: boolean): Promise<void> {
+    const copy_message: string = "Failed to copy declaration or definition to clipboard.";
+    const create_message: string = "Failed to create declaration or definition.";
```
michelleangela: "Failed to create declaration or definition." error is determined and handled by the native side logic based on the create declaration definition's API result. And the native side sends a request to typescript to show the error message in a prompt if create declaration definition API result is not successful.

Extension/src/common.ts
```diff
@@ -910,6 +910,17 @@ export async function promptReloadWindow(message: string): Promise<void> {
     }
 }
 
+export async function promptCDDFailed(copy?: boolean): Promise<void> {
+    const copy_message: string = "Failed to copy declaration or definition to clipboard.";
```
michelleangela: This will need to be localized.

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -3513,12 +3514,23 @@ export class DefaultClient implements Client {
                 }
                 lastEdit = new vscode.TextEdit(range, edit.newText);
                 const position: vscode.Position = new vscode.Position(edit.range.start.line - editPositionAdjustment, edit.range.start.character);
+                // If the command was invoked via the copy command, copy the declaration or definition to clipboard.
+                if (copy) {
+                    vscode.env.clipboard.writeText(edit.newText).then(() => {
+                        // Writing was successful
+                    }, () => {
+                        // Writing was unsuccessful
+                        util.promptCDDFailed(true);
```
michelleangela: I think this is the only case where create declaration or definition failed for an explicit request to copy needs to be handled.

Though the case for when create declaration and definition fails and falls back to copy will also need to be handled. Where it handles this will depend on how native side's CDD API result is and how the API's failures are handled.

Extension/src/common.ts
```diff
@@ -912,6 +912,26 @@ export async function promptReloadWindow(message: string): Promise<void> {
     }
 }
 
+export async function promptCDDFailed(copy?: boolean): Promise<void> {
+    const copyMessage: string = localize("cdd.copyMessage", "Failed to copy declaration or definition to clipboard.");
+    const createMessage: string = localize("cdd.createMessage", "Failed to create declaration or definition.");
+
+    if (copy) {
```
Colengms: Could you instead use the bool to decide what value to pass to a single call to `showInformationMessage`.  Using the "?:" operators, you could reduce this to from 5 lines to 1 line.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -327,10 +327,13 @@ interface PublishRefactorDiagnosticsParams {
...
     uri: string;
     range: Range;
+    copyToClipboard: boolean;
 }
 
 export interface CreateDeclarationOrDefinitionResult {
-    changes: { [key: string]: any[] };
+    edit: any;
+    clipboardText: string;
+    // add result for clipboard_text
```
michelleangela: Remove comment "// add result for clipboard_text"?

Extension/src/common.ts
```diff
@@ -912,6 +912,26 @@ export async function promptReloadWindow(message: string): Promise<void> {
     }
 }
 
+export async function promptCDDFailed(copy?: boolean): Promise<void> {
+    const copyMessage: string = localize("cdd.copyMessage", "Failed to copy declaration or definition to clipboard.");
+    const createMessage: string = localize("cdd.createMessage", "Failed to create declaration or definition.");
```
michelleangela: For what scenario is the error `createMessage` used for? The error messages for the operation "create declaration/definition" is set in the native side. There are also strings already defined for this too.

Extension/src/common.ts
```diff
@@ -912,6 +912,26 @@ export async function promptReloadWindow(message: string): Promise<void> {
...
+
+    if (copy) {
+        await vscode.window.showInformationMessage(copyMessage);
+    } else {
+        await vscode.window.showInformationMessage(createMessage);
+    }
+}
+
+export async function addTrustedCompiler(compilers: string[], path: string): Promise<string[]> {
```
michelleangela: Is `addTrustedCompiler` part of this change?

You may need to rebase this feature branch.

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -3502,12 +3505,27 @@ export class DefaultClient implements Client {
...
+        }
+
         const result: CreateDeclarationOrDefinitionResult = await this.languageClient.sendRequest(CreateDeclarationOrDefinitionRequest, params);
         // TODO: return specific errors info in result.
-        if (result.changes === undefined) {
+        if (result.edit.changes.length === undefined && result.clipboardText) {
+            if (!params.copyToClipboard) {
+                util.promptCDDFailed();
+            } else {
+                vscode.env.clipboard.writeText(result.clipboardText).then(() => {
```
sean-mcmanus: Can this just be `vscode.env.clipboard.writeText(result.clipboardText);`? It seems unnecessary to handle a failure, since that is not expected and would be some VS Code issue.

Extension/src/LanguageServer/client.ts
```diff
@@ -327,10 +327,13 @@ interface PublishRefactorDiagnosticsParams {
 export interface CreateDeclarationOrDefinitionParams {
     uri: string;
     range: Range;
+    copyToClipboard: boolean;
 }
 
 export interface CreateDeclarationOrDefinitionResult {
-    changes: { [key: string]: any[] };
+    edit: any;
```
sean-mcmanus: Can the "any" type be avoided?

sean-mcmanus: It looks like it should be WorkspaceEdit, right?

sean-mcmanus (changes requested):

Extension/src/common.ts
browntarik: Is it possible for CDD to fail after a successful request to the native side?

browntarik: This would be for the case that there was a successful CDD request to the native side, but something failed to complete on the typescript side.

Extension/src/common.ts
browntarik: Is it okay to leave this in and merge? I think this was due to a previous merge conflict.

Extension/src/LanguageServer/client.ts
browntarik: I think it is necessary to notify the user in this case since this is not a general copy to clipboard scenario but rather it is invoked via our extension.

sean-mcmanus: I don't know what the difference is -- the vscode.env.clipboard.writeText is an API that our extension invokes and we don't check for failure in our previous usage of it. How/why would copy to clipboard fail?

Extension/src/common.ts
```diff
@@ -912,6 +912,26 @@ export async function promptReloadWindow(message: string): Promise<void> {
     }
 }
 
+export async function promptCDDFailed(copy?: boolean): Promise<void> {
+    const copyMessage: string = localize("cdd.copyMessage", "Failed to copy declaration or definition to clipboard.");
```
sean-mcmanus: I think this should be `to the clipboard`.

Extension/src/common.ts
sean-mcmanus: No, you should remove it -- it looks like it was moved to client.ts.

Extension/src/nativeStrings.json
```diff
@@ -443,5 +443,12 @@
     "unknown_error": "Unknown error.",
     "run_select_intellisense_configuration": {
         "text": "Please run the 'Select IntelliSense Configuration...' command to locate your system headers."
+    },"offer_copy_declaration": {
```
sean-mcmanus: This should start on a new line.

sean-mcmanus: Also, could you format this document? The formatting seems inconsistent.

Extension/src/nativeStrings.json
```diff
@@ -443,5 +443,12 @@
...
+        "text": "Copy declaration of '{0}' to clipboard",
```
sean-mcmanus: As mentioned elsewhere, I think the `to clipboard` should be `to the clipboard`.

Extension/src/LanguageServer/client.ts
browntarik: Technically not because it is nested in a json object, I think I made the json struct incorrectly on the native side but haven't fixed it yet

Extension/src/LanguageServer/client.ts
browntarik: @bobbrow @Colengms @michelleangela @AlexandraKemperMS what do we think about this?

bobbrow: I would avoid using `.then` in a method already marked `async`.  You should use:
```ts
  try {
    await vscode.env.clipoard.writeText(result.clipboardText);
  } catch {
    // handle failure
  }

Comment (browntarik) [reply Extension/src/LanguageServer/client.ts]: But do you support the concept of catching the error in the first place? @bobbrow

Comment (bobbrow) [reply Extension/src/LanguageServer/client.ts]: I'm not opposed. If it never fails no one will ever see the response to the failure.

Comment (sean-mcmanus) [reply Extension/src/LanguageServer/client.ts]: I'm opposed to error handling unless that is what is expected of the callers of the API. Is it? Have you seen any examples of other extensions doing that? If so, then it seems like our other call should add that for consistency with using the API.

File: Extension/src/LanguageServer/client.ts (hunk 10 above)
Comment (sean-mcmanus) [reply Extension/src/LanguageServer/client.ts]: Okay, it seems like I might not be understanding something. Maybe I'll figure it out later.

File: Extension/src/LanguageServer/client.ts (hunk 9 above)
Comment (bobbrow) [reply Extension/src/LanguageServer/client.ts]: It's difficult to tell from the documentation because it doesn't specify, but the contract for `Thenable` is that it is allowed to reject/throw.  It's probably unlikely that it will fail, but we don't have any global try/catch wrappers to see which vscode API's are throwing that we may not be handling.

File: Extension/src/LanguageServer/client.ts [hunk 14]
```diff
@@ -3504,12 +3504,28 @@ export class DefaultClient implements Client {
                     character: range.end.character,
                     line: range.end.line
                 }
-            }
+            },
+            copyToClipboard: false
```
sean-mcmanus: This false can be replaced by `copy` and the `if` below removed.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -3504,12 +3504,28 @@ export class DefaultClient implements Client {
...
         };
 
+        if (copy) {
+            params.copyToClipboard = true;
+        }
+
         const result: CreateDeclarationOrDefinitionResult = await this.languageClient.sendRequest(CreateDeclarationOrDefinitionRequest, params);
         // TODO: return specific errors info in result.
-        if (result.changes === undefined) {
```
Colengms, line 3521: When a 'request' is cancelled using a standard LSP cancelation, the resulting object will contain an error code, and will not be the expected object (so will not have the expected field(s)).  This check if an expected field was undefined handled that scenario. 

For example, in `codeActionProvider.ts`, line 66, you'll see that it's actually triggering a cancellation error exception (valid from providers) when this happens.

In this case, I'd recommend you retain a check such as:

```
if (result.edit === undefined)
   return;
```

To handle the possibility that the LSP request may have been canceled, and the expected object not actually returned.

Colengms (changes requested):

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: @browntarik Why is this resolved? I don't see the resolution? Do you have unpushed changes? Usually issues are only resolved after the resolution has been pushed.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, it still seems like this should be WorkspaceEdit, such as like the other examples in CodeAnalysisDiagnosticRelatedInformation. I'm not seeing what the difference is or why WorkspaceEdit wouldn't work.

Extension/src/LanguageServer/client.ts
michelleangela: If you decide to make `copy` as an optional parameter, you can use the nullish coalescing `??` operator to evaluate the value of `copy` to determine what value to assign to `copyToClipboard`.

**Example:**
`copyToClipboard: copy ?? false;`

**Explanation:**
When `copy` is **not** undefined, it will use the value on the left hand side of `??` to assign to `copyToClipboard`. If `copy` evaluates as undefined, it will use `false` the value on the right hand side of `??` as the value to assign to `copyToClipboard`.

Extension/src/LanguageServer/client.ts
```diff
@@ -3504,12 +3507,32 @@ export class DefaultClient implements Client {
...
+
+        if (result.edit === undefined) {
+            return;
+        }
         // TODO: return specific errors info in result.
-        if (result.changes === undefined) {
+        if (result.edit.changes.length === undefined && result.clipboardText) {
+
+            if (!params.copyToClipboard) {
+                util.promptCDDFailed();
```
michelleangela: Is this still necessary if the language server is already sending a message that CD failed when the command is not "copy"?

Extension/src/LanguageServer/client.ts
```diff
@@ -3504,12 +3507,32 @@ export class DefaultClient implements Client {
...
+        if (copy) {
+            params.copyToClipboard = true;
+        }
+
         const result: CreateDeclarationOrDefinitionResult = await this.languageClient.sendRequest(CreateDeclarationOrDefinitionRequest, params);
+
+        if (result.edit === undefined) {
+            return;
+        }
         // TODO: return specific errors info in result.
```
michelleangela: the comment "// TODO: return specific errors info in result." can be removed since the errors are sent separately from the language server.

Extension/src/LanguageServer/client.ts
```diff
@@ -3504,12 +3507,32 @@ export class DefaultClient implements Client {
...
-        if (result.changes === undefined) {
+        if (result.edit.changes.length === undefined && result.clipboardText) {
+
+            if (!params.copyToClipboard) {
+                util.promptCDDFailed();
+            } else {
+                try {
+                    await vscode.env.clipboard.writeText(result.clipboardText);
```
michelleangela: Does this notify the user that the text was copied to clipboard to indicate that something happened?

bobbrow: No, we do not need to notify the user that it's copied.  VS doesn't and it's just one more thing in the way for the user.  They ran the command and since it should be fast they should be able to trust that it just worked.  We can notify them if it fails though.

Extension/src/LanguageServer/client.ts
browntarik: Yes, because copyToClipboard gives no clear visual indication to the user if it failed or not. This is different from create failing because the success case is the modification of a file.

michelleangela: The conditional `if (result.edit.changes.length === undefined && result.clipboardText)` is true either case "create..." or "copy....". If this logic should only be executed for "copy..." case then there should be a check in the conditional for that operation only.

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
michelleangela: To clarify, what are these errors `util.promptCDDFailed()` and `util.promptCDDFailed(true)` meant for?

browntarik: Initially it was to notify the user visually if either CDD failed or CopytoClipboard failed. I think inititially we agreed to do both in our first spec meeting but now it makes sense not to show this notification for if CDD fails. I still think this notification is appropriate for if copyToClipboard fails since the user will have no other visual cue (overall this scenario will be extremely rare anyways).

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,40 @@ export class DefaultClient implements Client {
...
+        if (result.edit === undefined) {
+            await vscode.window.showInformationMessage(result.errorText); // Copy/Create Declaration/Definition was completely unsuccessful due to api failure.
+            return;
+        }
+
+        // Handle CDD error messaging
+        if (result.errorText) {
+            try {
+                await vscode.env.clipboard.writeText(result.clipboardText);
+                await vscode.window.showInformationMessage(result.errorText + " Declaration/definition was copied to clipboard."); // CDD failed but fallback to clipboard was succesful
```
sean-mcmanus: succesful->successful

sean-mcmanus: As mentioned elsewhere, it seems like "to clipboard" is unnecessary.

sean-mcmanus: I think this should be changed to an error message with a button added that only does the copy if the user clicks the button.

Extension/src/nativeStrings.json
```diff
@@ -393,55 +393,60 @@
...
+        "text": "Copy declaration of '{0}' to the clipboard",
+        "hint": "{0} is the name of a C/C++ function."
+    },
+    "offer_copy_definition": {
+        "text": "Copy definition of '{0}' to the clipboard",
+        "hint": "{0} is the name of a C/C++ function."
+    },
+    "refactor_copy_declaration_definition_failed": {
+        "text": "Copying Declaration / Definition to clipboard failed: %s",
+        "hint": "The operation 'Copy Declaration / Definition' on a function was not successful. %s is the error info that has a period at the end of the string."
```
sean-mcmanus, line 449: This 1st sentence of the hint seems unnecessary.

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
sean-mcmanus: ...oh I see the same thing in other hints above. I'd prefer if those were removed too.

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,40 @@ export class DefaultClient implements Client {
...
+            return;
+        }
+
+        // Handle copy to clipboard.
+        if (params.copyToClipboard) {
+            if (result.clipboardText) {
+                try {
+                    await vscode.env.clipboard.writeText(result.clipboardText);
+                } catch {
+                    // handle failure
```
sean-mcmanus: I don't see anyone else on GitHub that has a try catch specifically around `await vscode.env.clipboard.writeText`, so I'd prefer if this were removed, since it's unnecessary.

Extension/src/LanguageServer/client.ts
Colengms: Unresolving this comment, as I don't understand why `any` is needed here.

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,40 @@ export class DefaultClient implements Client {
...
+
+        // Handle CDD error messaging
+        if (result.errorText) {
+            try {
+                await vscode.env.clipboard.writeText(result.clipboardText);
+                await vscode.window.showInformationMessage(result.errorText + " Declaration/definition was copied to clipboard."); // CDD failed but fallback to clipboard was succesful
+
+            } catch {
+                // handle failure
+                await vscode.window.showInformationMessage(result.errorText); // CDD failed but fallback was unsuccessful.
```
Colengms: Never await a `showInformationMessage` (or other vscode functions that display UI).  This will cause the code here to pause until the user dismisses the UI (which could be never).  You can omit the await here.

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,40 @@ export class DefaultClient implements Client {
...
+        }
+
+        // Handle copy to clipboard.
+        if (params.copyToClipboard) {
+            if (result.clipboardText) {
+                try {
+                    await vscode.env.clipboard.writeText(result.clipboardText);
+                } catch {
+                    // handle failure
+                    await vscode.window.showInformationMessage(localize("cdd.copyMessage", "Copying Declaration/Definition to clipboard failed.")); // Copy to clipboard system call failed.
```
Colengms: Please don't await vscode notifications.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: To clarify, it's not always bad to await these, such as when you need to know which of multiple options the user clicked on, and provided that occurs within an async that is not being awaited on at some higher level.  But, awaiting here will stall the async execution of this particular async context, and that's probably not what you want.

Extension/src/nativeStrings.json
browntarik: Going to keep this for consistency.

Extension/src/LanguageServer/client.ts
browntarik: I would prefer to keep it

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,39 @@ export class DefaultClient implements Client {
...
+            },
+            copyToClipboard: copy ?? false
         };
 
         const result: CreateDeclarationOrDefinitionResult = await this.languageClient.sendRequest(CreateDeclarationOrDefinitionRequest, params);
-        // TODO: return specific errors info in result.
-        if (result.changes === undefined) {
+        // Create/Copy returned no result.
+        if (result.edit === undefined) {
+            await vscode.window.showInformationMessage(result.errorText); // Copy/Create Declaration/Definition was completely unsuccessful due to api failure.
```
Colengms: Another `await` of `showInformationMessage`, which probably isn't desirable.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: So, we should only use await when the code that follows needs the result (or needs to happen afterwards), right? So the other awaits below should be removed too?

Extension/src/nativeStrings.json
sean-mcmanus: Sure, but couldn't both be removed for consistency too? Or do you mean consistency with something else?

Extension/src/LanguageServer/client.ts
Colengms: Or await may be needed to ensure the ordering of calls (A completes before B is called).  I'm not familiar with the other calls here being awaited.  I just noticed these UI calls specifically as they won't complete until/unless the user actually actions on them, so can cause indefinite stalls.

Extension/src/LanguageServer/client.ts
```diff
@@ -3506,12 +3510,31 @@ export class DefaultClient implements Client {
...
+        if (result.edit === undefined) {
+            vscode.window.showInformationMessage(result.errorText); // Copy/Create Declaration/Definition was completely unsuccessful due to api failure.
+            return;
+        }
+
+        // Handle CDD error messaging
+        if (result.errorText) {
+            let copiedToClipboard: boolean = false;
+            if (result.clipboardText && !params.copyToClipboard) {
+                await vscode.env.clipboard.writeText(result.clipboardText);
```
sean-mcmanus: I think this await should be removed or it should be changed to "void" to signal that it's not awaiting. Same as the other one below.

sean-mcmanus (dismissed):

sean-mcmanus (dismissed):

michelleangela:
The string `refactor_function_copied_to_clipboard` needs to be removed as it's no longer used on the language sever side.

Colengms (dismissed):

Colengms (approved):

michelleangela (approved):

PR #10901, opened 2023-04-28
Fix some spelling errors and formatting.

sean-mcmanus:


michelleangela (dismissed):

Extension/src/LanguageServer/ui_new.ts
```diff
@@ -606,9 +606,9 @@ export class NewUI implements UI {
         options.placeHolder = this.selectACommandString;
 
         const items: IndexableQuickPickItem[] = [];
-        items.push({ label: localize({ key: "active.analysis", comment: [this.codeAnalysisTranslationHint]}, "Run Code Analysis on Active File", this.codeAnalysisProgram), description: "", index: 0 });
```
Colengms, line 609: These 'comment' strings do not work as expected.  All strings passed to localize for `key`, `comment`, and the (primary) string need to be literals, due to how `vscode-nls-dev` processes source files.  I've confirmed that these comments do not appear (as <note> nodes) in the exported XLF file for these entries.

Do you want to fix that in this PR?

Colengms (changes requested):

Extension/src/LanguageServer/ui_new.ts
sean-mcmanus: sure

sean-mcmanus: done

Colengms (approved):

PR #10929, opened 2023-05-08
Add server-originated notification to cancel reference

Colengms:
Exposes the `cancelReferences` function in `client.ts` via a server-originated notification, allowing the server side to trigger cancelation of FAR-related features without getting out of sync with the progress UI on the TypeScript side.


Extension/src/LanguageServer/client.ts
```diff
@@ -2324,6 +2325,7 @@ export class DefaultClient implements Client {
         this.languageClient.onNotification(ReportCodeAnalysisProcessedNotification, (e) => this.updateCodeAnalysisProcessed(e));
         this.languageClient.onNotification(ReportCodeAnalysisTotalNotification, (e) => this.updateCodeAnalysisTotal(e));
         this.languageClient.onNotification(DoxygenCommentGeneratedNotification, (e) => void this.insertDoxygenComment(e));
+        this.languageClient.onNotification(CanceledReferencesNotification, this.cancelReferences);
```
sean-mcmanus, line 2328: Can you check if `workspaceReferences.referencesCanceled = true;` is set? That might need to be set, not sure.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: The call to `cancelReferences` will set `workspaceReferences.referencesCanceled = true` (if not already cancelling).  I tested the scenario, and the cancelation appears to work as expected.

sean-mcmanus (approved):

michelleangela (approved):

PR #10936, opened 2023-05-10
Call Hierarchy Callers Of - add progress and cancellation

michelleangela:
This changeset is for the call hierarchy callers of operation.

- Add progress notification window to show the progress of a "callers of" search. UI is similar to the reference search progress notification window. Re-uses implementation of reference search notification progress but without the "preview" feature.
- Add the ability to cancel a "callers of" operation from the progress notification window. When it is canceled, no results are returned.

image

Note: this has a corresponding language server changeset.

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -112,7 +112,7 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
 
     public async provideCallHierarchyIncomingCalls(item: vscode.CallHierarchyItem, token: vscode.CancellationToken):
         Promise<vscode.CallHierarchyIncomingCall[] | undefined> {
-        return new Promise<vscode.CallHierarchyIncomingCall[] | undefined>((resolve, reject) => {
+        const getIncomingCalls: any = new Promise<vscode.CallHierarchyIncomingCall[] | undefined>((resolve, reject) => {
             if (item === undefined) {
                 resolve(undefined);
                 return;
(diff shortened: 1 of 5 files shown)


sean-mcmanus (dismissed):

Colengms (dismissed):

sean-mcmanus (dismissed):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -114,23 +112,65 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
...
-        }
-
-        await this.client.awaitUntilLanguageClientReady();
-        const params: CallHierarchyParams = {
-            textDocument: { uri: item.uri.toString() },
-            position: Position.create(item.range.start.line, item.range.start.character)
-        };
-        const response: CallHierarchyCallsItemResult = await this.client.languageClient.sendRequest(CallHierarchyCallsToRequest, params, token);
-        if (token.isCancellationRequested || response.calls === undefined) {
-            throw new vscode.CancellationError();
```
Colengms, line 128: I noticed that new code no longer does `throw new vscode.CancellationError();`.  This is the appropriate way to notify VS Code that a provider has internally cancelled the requested operation (so it doesn't think it completed successfully with empty results, or failed and needs to be retried).  I'm not certain the nuances of FAR and CallHierarchy, but just wanted to call attention to that, in case throwing CancellationError on cancel might result in different UI behavior.

michelleangela: This is what find all references does. I first tried doing `throw new vscode.CancellationError();` but it would not complete the whole operation correctly.

michelleangela: I think it is because the request to the language server changed from a request to a notification with a callback. But even if it's all handled in a promise that doesn't resolve until the callback happens, it still doesn't work well with `throw new vscode.CancellationError();`.

michelleangela: I thought of another way to handle the `new vscode.CancellationError();`. I'll try it out and see what happens.

michelleangela: Added back `new vscode.CancellationError();` by handle the resolved or reject outcome of the `const getIncomingCalls: any = new Promise<vscode.CallHierarchyIncomingCall[] | undefined>`.

Colengms (approved):

PR #10941, opened 2023-05-11
Remove internal properties from writing.

sean-mcmanus:
But is there a way to repro a bug with 1.15.4? I haven't been able to yet.


sean-mcmanus:
...can't be reproed.

Issue #10943, opened 2023-05-12
"Select IntelliSense Configuration" and Run&Debug button do not offer user-specified compilerPath


Colengms:
Repro steps:
* Install a non-standard C++ compiler into an arbitrary location.  My repro was using a compiler toolchain downloaded here: https://github.com/mstorsjo/llvm-mingw/releases
* Specify a path to that compiler in the `compilerPath` field of `c_cpp_properties.json`. (IntelliSense properly queries the compiler and uses its settings for IntelliSense).  In my case, this was; `"compilerPath": "Z:\\llvm-mingw-20220906-msvcrt-x86_64\\bin\\clang++"`
* Use either the "Select IntelliSense Configuration" command or click on the "Debug C++ file" (run and debug) button. 
![image](https://github.com/microsoft/vscode-cpptools/assets/49173979/bd69cadb-99b6-4b89-965a-3d43e1d0fc97)

Note that the compiler specified in `compilerPath` is not offered as an option on either of those lists.

Also, there are two similarly named commands: "C/C++: Select IntelliSense Configuration..." and "C/C++: Select a Configuration...".  Perhaps these names could be revisited to make them less ambiguous.

Colengms:
The Run & Debug scenario is more obviously wrong, as the user should be able to build using the same `compilerPath` they've configured IntelliSense with.

sean-mcmanus:
@Colengms What did you have in mind for the command names? That's a completely separate issue so I'm not sure why it's added to this issue. "C/C++: Select IntelliSense Configuration..." is short for "C/C++: Select IntelliSense Configuration Source..." but I was worried that would be too long.

michelleangela:
This is resolved in https://github.com/microsoft/vscode-cpptools/releases/tag/1.16.0.

PR #10961, opened 2023-05-17
Fix addTrustedCompiler.

sean-mcmanus:
TypeScript changes related to https://github.com/microsoft/vscode-cpptools/issues/10943 .

- The previous code wasn't sending the added trusted compilers to the native side so the known compilers wasn't getting updated.
- The previous code was sending trusted compilers as an array when that isn't needed.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -69,14 +69,6 @@ export function hasTrustedCompilerPaths(): boolean {
     return trustedCompilerPaths.length !== 0;
 }
 
-export function addTrustedCompiler(path: string): void {
-    // Detect duplicate paths or invalid paths.
-    if (trustedCompilerPaths.includes(path) || path === null || path === undefined) {
-        return;
-    }
-    trustedCompilerPaths.push(path);
-}
-
 // Data shared by all clients.
 let languageClient: LanguageClient;
 let firstClientStarted: Promise<void>;
@@ -213,7 +205,7 @@ interface ReportStatusNotificationBody extends WorkspaceFolderParams {
 }
 
 interface QueryDefaultCompilerParams {
-    trustedCompilerPaths: string[];
+    newTrustedCompilerPath: string;
 }
 
 interface CppPropertiesParams extends WorkspaceFolderParams {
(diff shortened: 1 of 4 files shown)


fearthecowboy (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -3745,6 +3735,17 @@ export class DefaultClient implements Client {
...
         this.model.referencesCommandMode.Value = mode;
     }
+
+    public addTrustedCompiler(path: string): Promise<void> {
+        // Detect duplicate paths or invalid paths.
+        if (trustedCompilerPaths.includes(path) || path === null || path === undefined) {
+            return Promise.resolve();
+        }
+        trustedCompilerPaths.push(path);
+        return this.requestCompiler(path).then((result) => {
```
Colengms: Should this use `await` instead of `then` ?  It seems like this function could be simplified slightly if async.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: No -- I didn't want to force all the callers to be async.

sean-mcmanus: Oh, I was mistaken.

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -3747,6 +3737,15 @@ export class DefaultClient implements Client {
     public setReferencesCommandMode(mode: refs.ReferencesCommandMode): void {
         this.model.referencesCommandMode.Value = mode;
     }
+
+    public async addTrustedCompiler(path: string): Promise<void> {
+        // Detect duplicate paths or invalid paths.
+        if (trustedCompilerPaths.includes(path) || path === null || path === undefined) {
+            return Promise.resolve();
```
Colengms: As an `async` function, you could just return instead of resolving a promise.

sean-mcmanus: Ah, okay thanks.

sean-mcmanus: Fixed

Colengms (dismissed):

michelleangela (dismissed):

michelleangela (approved):

Issue #10933, opened 2023-05-10
Remove old status bar code and keep the language status bar


bobbrow:
The Language Status Bar experiment is complete and we're ready to remove the old status bar code and clean up the telemetry.  We want to remove 'experiment' from the event names and remove the 'newUI' and 'uiOverride' properties since we don't need it anymore.

PR #11030, opened 2023-05-30
Removed LanguageStatus experiment code and merged ui.ts and ui_new.ts

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10933


Extension/src/LanguageServer/ui.ts
```diff
@@ -56,32 +57,52 @@ interface ConfigurationStatus {
...
+enum LanguageStatusPriority {
+    First = 0,
+    High = 1,
+    Mid = 2,
+    Low = 3
+}
+
 const commandArguments: string[] = ['oldUI']; // We report the sender of the command
 
 export class OldUI implements UI {
```
sean-mcmanus: It seems like OldUI should be moved into UI or at least renamed?

Extension/src/LanguageServer/ui.ts
```diff
@@ -56,32 +57,52 @@ interface ConfigurationStatus {
...
 }
 
+enum LanguageStatusPriority {
+    First = 0,
+    High = 1,
+    Mid = 2,
+    Low = 3
+}
+
 const commandArguments: string[] = ['oldUI']; // We report the sender of the command
```
sean-mcmanus: It seems like this variable and its usage should be removed.

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
Shouldn't ui_new.ts be removed? The diff might look easier to understand if ui.ts were removed and ui_new.ts renamed to ui.ts, not sure.

Extension/src/LanguageServer/ui.ts
```diff
@@ -56,32 +56,52 @@ interface ConfigurationStatus {
...
+    First = 0,
+    High = 1,
+    Mid = 2,
+    Low = 3
+}
+
+const commandArguments: string[] = []; // We report the sender of the command
 
-export class OldUI implements UI {
+export class CurrentUI implements UI {
```
Colengms: "CurrentUI"?  Unless we think we'll need to implement another replacement soon or additional alternative classes for this UI (at which point, the interface would likely need to be updated/different anyway), the otherwise unused interface could be removed, and any unnecessary plumbing removed to reduce code that needs to be maintained.

Also (existing issue), "UI" (or OldUI, NewUI, CurrentUI) is woefully vague.  It tells someone unfamiliar with the code very little about what it is.  If/when any other bit of UI were needed for other purposes, it could be confusing.  How about naming it "LanguageStatusUI", or something similarly specific?

Colengms (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -119,90 +139,227 @@ export class OldUI implements UI {
...
-        const showIcon: boolean = val || this.isRunningCodeAnalysis;
-        const twoStatus: boolean = val && this.isRunningCodeAnalysis;
-        this.ShowFlameIcon = showIcon;
-        this.intelliSenseStatusBarItem.text = showIcon ? "$(flame)" : "";
-        this.intelliSenseStatusBarItem.tooltip = (val ? this.updatingIntelliSenseTooltip : "")
-            + (twoStatus ? " | " : "")
-            + (this.isRunningCodeAnalysis ? this.runningCodeAnalysisTooltip : "");
+        const settings: CppSettings = new CppSettings((vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) ? vscode.workspace.workspaceFolders[0]?.uri : undefined);
+
+        // TODO: Integrate with Tarik's feature to determine if compiler/bare-IntelliSense is configured
```
michelleangela: Is this comment still pending?

Extension/src/LanguageServer/ui.ts
```diff
@@ -56,32 +56,52 @@ interface ConfigurationStatus {
...
+    private readonly updatingIntelliSenseText: string = localize("updating.intellisense.text", "IntelliSense: Updating");
+    private readonly idleIntelliSenseText: string = localize("idle.intellisense.text", "IntelliSense: Ready");
+    private readonly missingIntelliSenseText: string = localize("absent.intellisense.text", "IntelliSense: Not configured");
+    private readonly codeAnalysisRunningText: string = localize("running.analysis.text", "Code Analysis: Running");
+    private readonly codeAnalysisPausedText: string = localize("paused.analysis.text", "Code Analysis: Paused");
+    private readonly codeAnalysisModePrefix: string = localize("mode.analysis.prefix", "Code Analysis Mode: ");
+    private codeAnalysProgress: string = "";
+    // Prevent icons from appearing too often and for too short of a time.
+    private readonly iconDelayTime: number = 1000;
+    get isNewUI(): boolean { return true; };
```
michelleangela: Is `get isNewUI(): boolean { return true; };` still needed? If there's no other UI to compare, then this query is not necessary.

michelleangela (changes requested):

Extension/src/LanguageServer/ui.ts
michelleangela: If it's still pending, could you update the comment with the issue number or name of feature?

sean-mcmanus: I think the TODO should be removed.

Extension/src/LanguageServer/ui.ts
browntarik: I changed it to LanguageStatusUI

Extension/src/LanguageServer/client.ts
```diff
@@ -2499,18 +2499,9 @@ export class DefaultClient implements Client {
...
-            }
+            this.model.isInitializingWorkspace.Value = true;
         } else if (message.endsWith("Indexing")) {
-            if (ui.isNewUI) {
-                this.model.isIndexingWorkspace.Value = true;
-                this.model.isInitializingWorkspace.Value = false;
-            } else {
-                this.model.isParsingWorkspace.Value = true;
-            }
+            this.model.isIndexingWorkspace.Value = true;
```
michelleangela, line 2504: The original code had these properties set if the new UI was being used:

```              
this.model.isIndexingWorkspace.Value = true;
this.model.isInitializingWorkspace.Value = false;
```

It looks like `this.model.isInitializingWorkspace.Value = false;` got deleted or was it no longer needed?

browntarik: Thanks for catching that!

sean-mcmanus (approved):

sean-mcmanus (changes requested):
IntelliSense Rescan command appears missing.

sean-mcmanus:
> IntelliSense Rescan command appears missing.

Oh, sorry, it looks like I was using a different machine with wrong binaries.

sean-mcmanus (approved):

michelleangela (approved):

Colengms (approved):

Extension/src/LanguageServer/ui.ts
```diff
@@ -538,23 +660,21 @@ export class OldUI implements UI {
...
-        this.intelliSenseStatusBarItem.dispose();
+        this.browseEngineStatusItem.dispose();
+        this.intelliSenseStatusItem.dispose();
         this.referencesStatusBarItem.dispose();
+        this.codeAnalysisStatusItem.dispose();
     }
 }
 
-export async function getUI(): Promise<UI> {
+export async function getUI(): Promise<LanguageStatusUI> {
```
bobbrow, line 670: This doesn't need to be async anymore

Issue #11032, opened 2023-05-31
In windows, `"console": "integratedTerminal"` option works better?


caiohamamura:
In this line, the `externalTerminal` is hardcoded. But for Windows `cl.exe` as suggested by the [VSCode documentation](https://code.visualstudio.com/docs/cpp/config-msvc), at least, this won't work. We have to manually change everytime to `integratedTerminal` to work, or else the Debugger won't attach.

https://github.com/microsoft/vscode-cpptools/blob/e394edb63096b1b6995c3e1dd8b1ad9d47603690/Extension/src/Debugger/configurationProvider.ts#L437

Can anyone else confirm this, or am I missing something?

If that is true shouldn't the extension check if the user is running `cl.exe` and use `integratedTerminal` instead by default?

michelleangela:
Fixed in pre-release 1.16.1 https://github.com/microsoft/vscode-cpptools/releases/tag/v1.16.1.

michelleangela:
Fix is now available in [release version 1.16.2](https://github.com/microsoft/vscode-cpptools/releases/tag/v1.16.2).

PR #11035, opened 2023-05-31
configurationProvider.ts: use integratedTerminal for cl.exe

caiohamamura:
Solves #11032

This patch will check if the compiler is `cl.exe` and use `integratedTerminal` instead of `externalTerminal`.

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -434,7 +434,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
                 if (newConfig.type === DebuggerType.cppdbg) {
                     newConfig.externalConsole = false;
                 } else {
-                    newConfig.console = "externalTerminal";
+                    newConfig.console = "integratedTerminal";
                 }
                 const isWindows: boolean = platformInfo.platform === 'win32';
                 // Extract the .exe path from the defined task.


sean-mcmanus (approved):

caiohamamura:
Sorry @sean-mcmanus I just requested review from you again by mistake, I'm not changing anything, just rebasing to the most recent master to avoid the need for a commit merge when you actually merge this PR.

sean-mcmanus (dismissed):

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -431,8 +431,11 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
                 newConfig.name = configPrefix + compilerName + " " + this.buildAndDebugActiveFileStr();
                 newConfig.preLaunchTask = task.name;
+                const isCl: boolean = compilerName === "cl.exe";
                 if (newConfig.type === DebuggerType.cppdbg) {
                     newConfig.externalConsole = false;
+                } else if (isCl) {
+                    newConfig.console = "integratedTerminal";
                 } else {
                     newConfig.console = "externalTerminal";
```
WardenGnaw: Is there a case where `newConfig.type` is `cppvsdbg` and not using `cl.exe`?

CppVsDbg is designed to be used only with MSVC.

I believe the change can be simplified to this one line being
```suggestion
                    newConfig.console = "integratedTerminal";
```

sean-mcmanus: I don't know of any case where cppvsdbg would be useable without cl.exe, so that seems good to me.

caiohamamura:
So just this should be enough then? 

7f64104

sean-mcmanus (approved):

WardenGnaw (approved):

michelleangela (approved):

michelleangela:
@caiohamamura 
This has been merged to main branch. Thank you for contributing!

PR #11056, opened 2023-06-07
Fix references cancellation and preview

### LINKED ISSUE #10998: Fix Cancellation for References/Rename/Call Hierarchy
If a "references/rename/call hierarchy incoming calls" request is done and canceled, it may cause an incomplete response to the VS Code  and result in the UI to look like the request is still in progress.

Re-work how cancelation or interruption of these requests are done and handled.

michelleangela:
This changeset simplifies how preview works for "find all references". It also makes cancellation more straightforward and sequential for "find all references", "rename" and "call hierarchy callers of" by changing their messages to the language server to a `RequestType` object. This avoids race conditions in reading cancellation states among the different requests and also avoids the chance of not resolving a request from reading/setting incorrect states due to race conditions.

This change also has a corresponding changeset on the language server code base.

**Changes for Preview and Cancel for Find All References**
- When doing a preview, the partial or preview results is now shown in the "Other References" view. This no longer resolves a current request to show any confirmed reference type from partial results in the VS Code References view.
- When cancelling, any in progress results are now shown in the "Other References" view. This is because cancelling the request will return an empty result object from the request message returned from the language server.
- A find all references request will now only complete when the language server has finished processing the request or when the processing gets cancelled.

**Request Changes**
- The typescript message sent to the language server for getting "find all references", "rename", and "call hierarchy callers of" is now a `RequestType` object. This is to ensure synchronization of a language server response to its typescript sender. This also removes the requirement of having to manually manage cancellation of requests.
- Cancellation is now done by calling cancel on a local cancellation token passed to the request to the language server rather than by sending a separate cancellation notification to the language server.
- A new request sends a cancellation event to cancel a current request by calling `ReferencesManager.cancelCurrentReferenceRequest`.

**Other Changes**
- Refactored some objects and variables to the appropriate classes/files that use them exclusively.

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -5,9 +5,9 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import * as Telemetry from '../../telemetry';
-import { DefaultClient, CancelReferencesNotification, workspaceReferences } from '../client';
+import { DefaultClient, workspaceReferences } from '../client';
 import { processDelayedDidOpen } from '../extension';
-import { CallHierarchyResultCallback } from '../references';
+import { CallHierarchyResultCallback, CancellationSender } from '../references';
 import { Position, Range, RequestType, TextDocumentIdentifier } from 'vscode-languageclient';
 import { makeVscodeRange } from '../utils';
 
@@ -77,7 +77,7 @@ export interface CallHierarchyCallsItemResult {
     calls: CallHierarchyCallsItem[];
 }
 
-export enum CallHierarchyRequestStatus {
+enum CallHierarchyRequestStatus {
     Unknown,
     Succeeded,
     Canceled,
@@ -95,6 +95,7 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
     // Indicates whether a request is from an entry root node (e.g. top function in the call tree).
     private isEntryRootNodeTelemetry: boolean = false;
     private client: DefaultClient;
+
     constructor(client: DefaultClient) {
         this.client = client;
     }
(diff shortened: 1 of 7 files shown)


Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -125,73 +126,60 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
...
-            const CallHierarchyCallsToEvent: string = "CallHierarchyCallsTo";
-            if (item === undefined) {
-                this.logTelemetry(CallHierarchyCallsToEvent, CallHierarchyRequestStatus.Failed);
-                resolve(undefined);
+            // Listen to cancellation from an incoming new request, user or language server.
+            workspaceReferences.onCancellationRequested(sender => { requestCanceled = sender; });
+
+            // Define the callback that will process results.
+            const resultCallback: CallHierarchyResultCallback = (result: CallHierarchyCallsItemResult | null, progressBarDuration?: number) => {
+                if (requestCanceled !== CancellationSender.None || result === null || result?.calls === undefined) {
```
sean-mcmanus: This `result == null` check should not be required because `result?.call === undefined` will be true in that case.

michelleangela: Fix pending. Fixed for line 160 too.

Extension/src/LanguageServer/Providers/renameProvider.ts
```diff
@@ -3,110 +3,100 @@
...
+                // Send the request to language server
+                const params: RenameReferencesParams = {
+                    newName: newName,
+                    position: Position.create(position.line, position.character),
+                    textDocument: { uri: document.uri.toString() }
+                };
+                workspaceReferences.startRename(params);
             } else {
-                callback();
+                // Only complete the request at this point if the request to language server
```
sean-mcmanus: Can you change all (existing) uses of `to language server` to `to the language server` and `when language` to `when the language`?

Extension/src/LanguageServer/client.ts
```diff
@@ -3655,56 +3616,20 @@ export class DefaultClient implements Client {
...
+            workspaceReferences.UpdateProgressUICounter(this.model.referencesCommandMode.Value);
 
-    public clearPendingReferencesCancellations(): void {
-        if (DefaultClient.referencesPendingCancellations.length > 0) {
-            while (DefaultClient.referencesPendingCancellations.length > 1) {
-                const pendingCancel: ReferencesCancellationState = DefaultClient.referencesPendingCancellations[0];
-                DefaultClient.referencesPendingCancellations.pop();
-                pendingCancel.reject();
+            // If the search is find all references, preview partial results.
+            // This will cause language server to send partial results to display
```
sean-mcmanus: `cause language server`->`cause the language server`

Extension/src/LanguageServer/Providers/renameProvider.ts
michelleangela: This comment was removed, as logic changed.

Extension/src/LanguageServer/client.ts
michelleangela: fixed

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -124,76 +130,54 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
...
+        if (item === undefined) {
+            this.logTelemetry(CallHierarchyCallsToEvent, CallHierarchyRequestStatus.Failed);
+            return undefined;
+        }
+
+        // Listen to a cancellation for this request. When this request is cancelled,
+        // use a local cancellation source to implicitly cancel a token.
+        let requestCanceled: CancellationSender | undefined;
+        const cancelSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
+        this.disposables.push(token.onCancellationRequested(() => {
```
Colengms: If a new instances of these are created every time the operation is called, a lot of disposables will accumulate and a ton of them may need to be disposed of when the provider is disposed.  Instead, could you set aside these two disposables and dispose them on subsequent iterations?  That would avoid having an indeterminate number of them lingering.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -124,76 +130,54 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
...
+            requestCanceled = CancellationSender.ProviderToken;
+            cancelSource.cancel();
+        }));
+        this.disposables.push(workspaceReferences.onCancellationRequested(sender => {
```
Colengms: Rather than register a callback on this event every time a new request occurs (which will accumulate a ton of callbacks on that event, as these aren't currently being removed/disposed), maybe register just once for each provider, in the providers' constructors?  Then, if that used `this.cancelSource.cancel()`, it would always apply to the current in-progress operation.

Colengms (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: At the beginning of each new request, the event which invokes `workspaceReferences.onCancellationRequested` is called, won't the new request cancel itself if `this.cancelSource.cancel()` is used?

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: `private dispose()`  is called after each request. But this may dispose newly created listeners. I'll change it to locally track the listeners.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
Colengms: I'm not sure I follow.  `workspaceReferences.cancelCurrentReferenceRequest(CancellationSender.NewRequest);` would get called (and call `this.cancelSource.,cancel()`) before you would replace `this.cancelSource` with a new one.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
Colengms: Wouldn't calling dispose() there be a bug, as a subsequent request may already have started?  You could be disposing of the new ones, while they're still needed.

Colengms (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: Yes, that's why I created local disposables instead of adding new ones in a class variable.

Colengms (dismissed):

Colengms (dismissed):

sean-mcmanus (changes requested):
I see Find All References cancels Call Hierarchy, but Call Hierarchy isn't canceling Find All References -- can you add cancel handling to prepareCallHierarchy be make that cancel Find All References (and Call Hierarchy)?

michelleangela:
> I see Find All References cancels Call Hierarchy, but Call Hierarchy isn't canceling Find All References -- can you add cancel handling to prepareCallHierarchy be make that cancel Find All References (and Call Hierarchy)?

Added.

Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
```diff
@@ -3,98 +3,63 @@
...
+
+        // Reset anything that can be cleared before procossing the result.
+        // Note: ReferencesManager.resetReferences is called in ReferencesManager.showResultsInPanelView
+        workspaceReferences.resetProgressBar();
+        cancellationTokenListener.dispose();
+        requestCanceledListener.dispose();
+
+        // Process the result.
+        if (cancelSource.token.isCancellationRequested || response.referenceInfos === null || response.isCanceled) {
+            throw new vscode.CancellationError();
```
sean-mcmanus: This appears to cause VS Code to show `Exception has occurred: TypeError: Cannot destructure property 'range' of 'e.location' as it is undefined.`. Can you change it to `return undefined;`? That seemed to fix it for me. I'm not sure where it's documented about the usage of vscode.CancellationError or if we should report a VS Code bug on it or not.

The error is from code at
https://github.com/microsoft/vscode/blob/69246279e5b60be0f3cfd77da6f64a211a55e64a/extensions/references-view/src/references/model.ts#L288

i.e. it seems like they could add a check to prevent that, but maybe they'd say we shouldn't throw the CancellationError.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
michelleangela: That's why the previous code returned undefined.

Colengms: https://github.com/microsoft/vscode/issues/169698

> The gist is that all language operations are cancellable and that it is OK to throw the cancellation error*. In fact we do that automatically by racing against the provided token (see [these lines](https://insiders.vscode.dev/github/microsoft/vscode/blob/1d73bc45c2060131e1877a539deb68ed1f873e2e/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L1799-L1801))

> We'll switch to throwing CancellationError from all of our providers on cancel. If we run into any problems with that, we'll open specific issues. Closing this one.

sean-mcmanus: This is a regression compared to 1.15.4 -- I think we should return undefined for all the cases that VS Code shows an error popup

michelleangela: Added comment to clarify usage of undefined.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -88,33 +88,52 @@ export enum CallHierarchyRequestStatus {
...
         };
-        const response: CallHierarchyItemResult = await this.client.languageClient.sendRequest(CallHierarchyItemRequest, params, token);
-        if (token.isCancellationRequested || response.succeeded === undefined) {
+        const response: CallHierarchyItemResult = await this.client.languageClient.sendRequest(CallHierarchyItemRequest, params, cancelSource.token);
+
+        cancellationTokenListener.dispose();
+        requestCanceledListener.dispose();
+
+        if (cancelSource.token.isCancellationRequested || response.succeeded === undefined) {
             throw new vscode.CancellationError();
```
sean-mcmanus: This throw causes VS Code to show 
image
so I think it should be changed to return undefined, although a bug should be logged for them to fix that -- I haven't located the spot in their code which is triggering the error message.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: Fixed.

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -125,76 +148,56 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
...
+
+        // Send the request to the language server.
+        let result: vscode.CallHierarchyIncomingCall[] | undefined;
+        const params: CallHierarchyParams = {
+            textDocument: { uri: item.uri.toString() },
+            position: Position.create(item.range.start.line, item.range.start.character)
+        };
+        const response: CallHierarchyCallsItemResult = await this.client.languageClient.sendRequest(CallHierarchyCallsToRequest, params, cancelSource.token);
+
+        // Reset anything that can be cleared before procossing the result.
```
sean-mcmanus: procossing ->processing

Extension/src/LanguageServer/Providers/renameProvider.ts
```diff
@@ -3,111 +3,77 @@
...
+        workspaceReferences.startRename();
+        const workspaceEditResult: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
+        const params: ReferencesParams = {
+            newName: newName,
+            position: Position.create(position.line, position.character),
+            textDocument: { uri: document.uri.toString() }
+        };
+        const response: ReferencesResult = await this.client.languageClient.sendRequest(RenameRequest, params, cancelSource.token);
+
+        // Reset anything that can be cleared before procossing the result.
```
sean-mcmanus: procossing ->processing

Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
```diff
@@ -3,98 +3,67 @@
...
+        // Send the request to the language server.
+        const locationsResult: vscode.Location[] = [];
+        const params: ReferencesParams = {
+            newName: "",
+            position: Position.create(position.line, position.character),
+            textDocument: { uri: document.uri.toString() }
+        };
+        const response: ReferencesResult = await this.client.languageClient.sendRequest(FindAllReferencesRequest, params, cancelSource.token);
+
+        // Reset anything that can be cleared before procossing the result.
```
sean-mcmanus: procossing ->processing

Extension/src/LanguageServer/references.ts
```diff
@@ -75,6 +80,23 @@ export enum ReferencesCommandMode {
...
+    /* Cancellation was from a new request */
+    NewRequest,
+
+    /* Cancellation was from the provider cancellation token */
+    ProviderToken,
+
+    /* Cancellation was from the language server */
+    LanguageServer,
+
+    /* Cancelaltion was from the user selecting the cancel button in the reference search progress bar */
```
sean-mcmanus: Cancelaltion ->Cancellation

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #10388, opened 2023-01-16
Add wildcard support for includeDir

yne:
For example, using the following tree:

- rootUri/
  - sources/
    - main/
      - main.c
    - vendor/
      - khash.h
  - headers/
    - C001
      - a.h
    - C002
      - b.h

|The following directory | will give
|-------------------|---------------
| `sources/**` | plain old recursive path like before
| `headers/C*/`| dynamic path
| `missing/`   | reported as "not found"
| `missing/*`  | expanded as empty result (silent fail)

Fixes #723

PR #11085, opened 2023-06-13
Modernize packages, code patterns

fearthecowboy:
This PR modernizes some of the cpptools project.

During my effort to add some significant new code to the project, I ran into quite a few things that prevented that, so I've stopped and did a refactoring pass first so that we can modernize some of the development.
 
## Updates:
  - updates the TS compiler to 5.1.3
  - updates Mocha and ESLint to modern versions
  - removes deprecated TSLint components, using ESLint components only
  - newer webpack to work with newer TS version
  - newer gulp-typescript to work with TS version
  - updated types for node to 16.20
  - reset yarn.lock and regenerated (updates to latest versions of packages)
   
## Fixes:
  - remove lodash (only used for deep clone, switching to JSON `stringify`/`parse` )
  - switch to ES2022 target (The VSCode target of 1.69 or above supports everything in ES2022)
    - eliminates extra shims in TS => JS compilation.
    - (ie, uses actual async instead of downlevel shims - removed 39 copys of redundant `__awaiter` declarations )
    - allows use of newer members in ES standard library. (with updated node typings)
  - even after adding in new code, still ~15000 lines/500k less in the output main.js
  - refactored out   `requestWhenReady`,  `notifyWhenLanguageClientReady`, `awaitUntilLanguageClientReady`, and `queueTask` in favor of a `.ready` promise. 
  - handful of related cleanups
  - ensured there are no dangling promises.
  
## Added some code before compiler detection work comes in
There are a few reusable bits of code in `src/Automation` that simplified this PR, so I'm adding them in before my next PR with the compiler detection work 

This includes
  - `ManualPromise` - a `Promise` object that can be resolved/rejected external to the constructor (ie, doesn't need `resolve` and `reject` parameters, offers members instead)
  - `Signal`/`ManualSignal` - a special kind of `Promise` that can be reset (either automatically in `Signal` or manually in `ManualSignal`) to the unresolved state so that the next time it is awaited it can block again. 
  - `sleep()` - an async `Promise`-based replacement for using setTimeout
  - a static typeguard class `is` that can be used as a consistent pattern for value identification and type narrowing.
  
## Async refactoring
With the newer ESLint, we can turn on checking of dangling promises, which precipitated a whole slew of fixes to clean them up 

Which brings to point, some better practices for safer Async usage:
 - all async calls must be marked `await` or `void` or `return`ed
 - `Promise`s that have no consumer should be appended with `.catch(logAndReturn.undefined)` to ensure that an unhandled promise doesn't happen (and at least the error should show up in the log)
 - when possible, `return` the final `Promise` instead of using `await` - allows for aggressive tail-call optimization in V8
 - enable more functions to be `async` instead of returning void - this opens up some possibilities to full call tracking later, and gives the consumer the choice whether to `await` or not. (This is OK for vscode event handlers too)
 - begin moving to using `Promise` everywhere possible instead of `Thenable` - there is no benefit to `Thenable`, use of `Promise` properly will make async work simpler 

## Still might add in this PR
- clean up `clients.ActiveClient` vs `getActiveClient()`  -- pick one and stick with it. 
- clean up `ui` vs `getUI()` - consistency - same thing

## In a followup PR:
 - investigate and fix initialization and remove `BlockingTask` concept and move to a resettable `Signal` -- should make blocking at the high-order tasks much cleaner


Extension/.eslintrc.js
```diff
@@ -109,48 +113,6 @@ module.exports = {
...
-                        "check-branch",
-                        "check-operator",
-                        "check-separator",
-                        "check-preblock",
-                        "check-type"
-                    ]
-                }
-            }
-        ]
+
```
fearthecowboy: I will look into ensuring that old-style TSLint rules are represented by their modern equivalents

Extension/gulpfile.js
```diff
@@ -6,7 +6,7 @@
 'use strict';
 
 const gulp = require('gulp');
-const eslint = require('gulp-eslint');
+const eslint = require('gulp-eslint-new');
```
fearthecowboy, line 9: `gulp-eslint` is unmaintained, and `gulp-eslint-new` is used as a replacement.

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -30,7 +30,7 @@ export class CppdbgDebugAdapterDescriptorFactory extends AbstractDebugAdapterDes
         super(context);
     }
 
-    async createDebugAdapterDescriptor(session: vscode.DebugSession, executable?: vscode.DebugAdapterExecutable): Promise<vscode.DebugAdapterDescriptor> {
+    async createDebugAdapterDescriptor(_session: vscode.DebugSession, _executable?: vscode.DebugAdapterExecutable): Promise<vscode.DebugAdapterDescriptor> {
```
fearthecowboy, line 33: unused parameters can be prefixed with underscore to indicate that it's unused, but not deleted.

Extension/.eslintrc.js
```diff
@@ -57,6 +58,10 @@ module.exports = {
         "@typescript-eslint/triple-slash-reference": "error",
         "@typescript-eslint/type-annotation-spacing": "error",
         "@typescript-eslint/unified-signatures": "error",
+        "@typescript-eslint/no-floating-promises": "error",
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" } ],
```
sean-mcmanus: Can you run the formatter? I'm seeing tab/space differences.

fearthecowboy: 👍 

Yeah, I'm seeing them too, my default settings are slightly different, and the `.vscode/settings.json`  in the repo doesn't have  indent sizes specified, so I was going to bring that up in the meeting, and then set it accordingly, and then make sure everything is formatted correctly before coming out of draft mode. 

I'm worried about it completely freaking out and making it look like massive changes due to indenting.

sean-mcmanus: It looks like we use 4 spaces for indentation, but I don't know if/how that is configured. Maybe you can set the editor.indentSize to for in .vscode/settings.json?

fearthecowboy: That's what I'd like to do - there are some files that are using tabs and some using a different indent, so before I was going to force it all one way, I wanted to talk about it in the meeting. 

Ideally, it'd be perfectly consistent, and this issue doesn't ever come back.

Extension/src/main.ts
```diff
@@ -114,7 +115,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
             const config: string = path.join(vscode.workspace.workspaceFolders[i].uri.fsPath, ".vscode/c_cpp_properties.json");
             if (await util.checkFileExists(config)) {
                 const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(config);
-                vscode.languages.setTextDocumentLanguage(doc, "jsonc");
+                await vscode.languages.setTextDocumentLanguage(doc, "jsonc");
```
sean-mcmanus: Why is this await and not void? i.e. seems like it should be void?

sean-mcmanus (changes requested):

sean-mcmanus:
In your description `all async calls must be marked await or void or returned` is `returned` supposed to be `return`ed? (or is there a `returned` syntax I'm not aware of?

fearthecowboy:
> a `return`ed syntax I'm not aware of?

If at the end of an async function you use `await ...` and then `return; ` or `const foo = await ...` and `return foo` you should just `return foo;`  -- returning a promise in an `async` function is exactly the same as `await` then `return` -- awaiting before adds no value if the output of the function is the thing you're awaiting.

Extension/src/main.ts
fearthecowboy: Personal Intuition -- there may be a few things that I marked `await` based on what I saw.  We'll examine some of these tomorrow and I'll adjust based on intent.

Extension/package.json
```diff
@@ -6020,57 +6021,54 @@
     "@types/glob": "^7.1.3",
     "@types/minimatch": "^3.0.5",
     "@types/mkdirp": "^0.5.2",
-    "@types/mocha": "^8.2.2",
-    "@types/node": "^14.14.0",
+    "@types/mocha": "^10.0.0",
+    "@types/node": "18.11.9",
```
sean-mcmanus: I noticed you removed the `^` from some of the dependencies. Was there a reason for that? Can those be added back?

Extension/src/Automation/Async/manual-promise.ts
```diff
@@ -0,0 +1,102 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
```
sean-mcmanus: Can these be put in Extension/src/Async? i.e. can the Automation folder be removed?

fearthecowboy: My next PR adds ALOT more files into the `Automation` folder that are not in the same directory

sean-mcmanus: Sure, but does the Async folder have to be a subfolder of Automation? Is it Automation related?

fearthecowboy: It's part of a much larger 'automation' suite.  (see https://github.com/fearthecowboy/vscode-cpptools/tree/compiler-detection/Extension/src/Automation ) 

I can promote them all higher, but that makes a lot more folders at the upper level, which seemed less in line with the code that's already there.

sean-mcmanus: No, I'm fine with keeping it in that subfolder, but why is the folder called "Automation"? Could it be renamed to something else that would explain its purpose better?

Extension/src/LanguageServer/client.ts
```diff
@@ -1184,32 +1182,29 @@ export class DefaultClient implements Client {
...
+                    await this.showPrompt(selectCompiler, true, sender);
                     action = "dismissed";
                 }
                 telemetry.logLanguageServerEvent('compilerNotification', { action });
             } else if (!isCommand && (compilerDefaults.compilerPath === undefined)) {
-                this.showPrompt(selectCompiler, false, sender);
+                await this.showPrompt(selectCompiler, false, sender);
             } else {
-                this.handleIntelliSenseConfigurationQuickPick(isCommand, sender);
+                await this.handleIntelliSenseConfigurationQuickPick(isCommand, sender);
```
sean-mcmanus: I think these last 2 `await` can be `return`? Similar code is above in promptSelectCompiler and that uses `return`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1184,32 +1182,29 @@ export class DefaultClient implements Client {
...
                     action = "confirm compiler";
-                    ui.ShowConfigureIntelliSenseButton(false, this);
+                    await ui.ShowConfigureIntelliSenseButton(false, this);
                 } else if (value === selectCompiler) {
-                    this.handleIntelliSenseConfigurationQuickPick(true, sender);
+                    await this.handleIntelliSenseConfigurationQuickPick(true, sender);
                     action = "show quickpick";
                 } else {
-                    this.showPrompt(selectCompiler, true, sender);
+                    await this.showPrompt(selectCompiler, true, sender);
```
sean-mcmanus: I think these `await` should be `void`. Same as in the promptSelectCompiler function above.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1040,61 +1041,58 @@ export class DefaultClient implements Client {
...
+                        void vscode.commands.executeCommand('vscode.open', "https://go.microsoft.com/fwlink/?linkid=2217615");
                         return;
                 }
             }
             if (index === paths.length - 3) {
                 const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
                 if (result === undefined || result.length === 0) {
                     if (showSecondPrompt && !!compilerDefaults && !compilerDefaults.trustedCompilerFound && !fromStatusBarButton) {
-                        this.showPrompt(selectIntelliSenseConfig, true, sender);
+                        await this.showPrompt(selectIntelliSenseConfig, true, sender);
```
sean-mcmanus: I think this `await` should be `void`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1119,7 +1117,7 @@ export class DefaultClient implements Client {
                     ask.Value = false;
                 }
                 if (!configurationSelected) {
-                    this.handleConfigStatusOrPrompt();
+                    await this.handleConfigStatusOrPrompt();
```
sean-mcmanus, line 1124: I think this `await` should be `return`.

Extension/src/LanguageServer/client.ts
```diff
@@ -1150,19 +1148,19 @@ export class DefaultClient implements Client {
                     await this.addTrustedCompiler(settings.defaultCompilerPath);
                     DefaultClient.updateClientConfigurations();
                     action = "confirm compiler";
-                    ui.ShowConfigureIntelliSenseButton(false, this);
+                    void ui.ShowConfigureIntelliSenseButton(false, this);
                 } else if (value === selectCompiler) {
-                    this.handleIntelliSenseConfigurationQuickPick(true, sender, true);
+                    await this.handleIntelliSenseConfigurationQuickPick(true, sender, true);
```
sean-mcmanus: `await`->`void`

Extension/src/LanguageServer/client.ts
```diff
@@ -1184,32 +1182,29 @@ export class DefaultClient implements Client {
...
             } else {
-                this.handleIntelliSenseConfigurationQuickPick(isCommand, sender);
+                await this.handleIntelliSenseConfigurationQuickPick(isCommand, sender);
             }
         }
     }
 
     /**
-     * All public methods on this class must be guarded by the "pendingTask" promise. Requests and notifications received before the task is
+     * All public methods on this class must be guarded by the ".ready" promise. Requests and notifications received before the task is
```
sean-mcmanus: Is this supposed to be `.ready`? Why not just `ready`? Or `this.ready`?

Extension/src/LanguageServer/client.ts
```diff
@@ -1756,153 +1753,147 @@ export class DefaultClient implements Client {
...
+            return; // There is no c_cpp_properties.json to edit because there is no folder open.
+        }
+        this.configuration.handleConfigurationChange();
+        if (this.configStateReceived.configProviders === undefined) {
+            this.configStateReceived.configProviders = [];
+        }
+        this.configStateReceived.configProviders.push(provider);
+        const selectedProvider: string | undefined = this.configuration.CurrentConfigurationProvider;
+        if (!selectedProvider || this.showConfigureIntelliSenseButton) {
+            await this.handleConfigStatusOrPrompt("configProviders");
```
sean-mcmanus: `await`->`void`

sean-mcmanus (changes requested):

Extension/package.json
fearthecowboy: fixed.

Extension/src/LanguageServer/client.ts
fearthecowboy: Yes! good catch

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -453,60 +450,55 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
                         } else {
-                            // Try the usr path for non-windows platforms.
-                            const usrDebuggerPath: string = path.join("/usr", "bin", debuggerName);
-                            if (!isWindows && await util.checkFileExists(usrDebuggerPath)) {
-                                newConfig.miDebuggerPath = usrDebuggerPath;
-                            } else {
-                                logger.getOutputChannelLogger().appendLine(localize('debugger.path.not.exists', "Unable to find the {0} debugger. The debug configuration for {1} is ignored.", `\"${debuggerName}\"`, compilerName));
-                                return resolve(undefined);
-                            }
+                            logger.getOutputChannelLogger().appendLine(localize('debugger.path.not.exists', "Unable to find the {0} debugger. The debug configuration for {1} is ignored.", `\"${debuggerName}\"`, compilerName));
```
sean-mcmanus, line 496: Is this supposed to have `void` in front? Was that supposed to generate a warning?

Extension/src/LanguageServer/Providers/onTypeFormattingEditProvider.ts
```diff
@@ -14,11 +14,12 @@ export class OnTypeFormattingEditProvider implements vscode.OnTypeFormattingEdit
     }
 
     public async provideOnTypeFormattingEdits(document: vscode.TextDocument, position: vscode.Position, ch: string, options: vscode.FormattingOptions, token: vscode.CancellationToken): Promise<vscode.TextEdit[]> {
+
```
sean-mcmanus: Why is there a newline added here? i.e. what newline rule is being followed? Our existing code base usually doesn't have a blank newlines at the start of functions, but it might be okay if there was a reason for it.

sean-mcmanus:
FYI, I was planning to finish code reviewing this on Friday.

Extension/src/LanguageServer/Providers/onTypeFormattingEditProvider.ts
fearthecowboy: Wasnt put there by the formatter, I cleaned it up.

Extension/src/Debugger/configurationProvider.ts
fearthecowboy: it's not `async`...

Extension/src/LanguageServer/client.ts
fearthecowboy: In this case, it can't -- its in a `finally` block, and `return`ing from the finally is an unsafe operation.

Extension/.vscode/tasks.json
```diff
@@ -166,6 +166,34 @@
...
                 }
             ]
+        },
+        {
+            "label": "Compile:Watch (unit-tests)",
+            "type": "shell",
+            "options": {
+                "cwd": "${workspaceFolder}"
+            },
+            "command": "node_modules/.bin/tsc --build ${workspaceFolder}/test.tsconfig.json --watch",
```
bobbrow: The other tasks use yarn plus script from the package.json. Should we follow that pattern here as well?

Extension/src/Automation/Async/manual-promise.ts
```diff
@@ -0,0 +1,102 @@
...
+     */
+    finally(onfinally?: (() => void) | null | undefined): Promise<T> {
+        return this.promise.finally(onfinally);
+    }
+
+    /**
+     * A method to manually resolve the Promise.
+     */
+    public resolve: (value?: T | PromiseLike<T> | undefined) => void = (v) => {
+        void v; /* */
```
bobbrow: Since `resolve` and `reject` are set in the constructor, should we change these to `readonly` and initialize to `undefined` here?  I was initially confused and wondering how the promise actually got resolved since the constructor code is below this.

Extension/src/Automation/Async/manual-signal.ts
```diff
@@ -0,0 +1,94 @@
...
+     */
+    finally(onfinally?: (() => void) | null | undefined): Promise<T> {
+        return this.promise.finally(onfinally);
+    }
+
+    /**
+     * A method to manually resolve the Promise.
+     *
+     * This also resets this instance to a new promise interally, so that any new awaiters will not be instantly resolved.
+     * @param value
```
bobbrow: Does it need to call `this.reset()` to actually do the reset?  (same for `reject`)

Extension/src/Automation/Async/manual-signal.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { ManualPromise } from './manual-promise';
+import { Resetable } from './resolvable';
+
+/**
+ * A signal is an externally fulfillable promise ( @see ManualPromise )
+ * that once awaited, will automatically reset to the unwaited state
```
bobbrow: I think this is a copy/paste error from `signal.ts`

Extension/src/Automation/Async/manual-signal.ts
bobbrow: Since this is the manual signal, I think the comment might just be a copy/paste error.

Extension/src/Automation/Async/manual-signal.ts
fearthecowboy: Fixing...

Extension/src/Automation/Async/manual-promise.ts
fearthecowboy: The initialization is *inside* the call to the promise constructor, so isn't techically in scope of the constructor itself, so `readonly` can't be used.

Extension/src/Automation/Async/manual-signal.ts
fearthecowboy: Yeah, incorrect comment. Fixing...

Extension/.vscode/tasks.json
fearthecowboy: Sure!

Extension/src/Automation/Async/manual-promise.ts
fearthecowboy: renaming to `Utility`

Extension/src/main.ts
fearthecowboy: moved to void

fearthecowboy:
> But when is this planned to go out of draft?

I've moved it out of draft now.

Extension/src/Automation/Async/signal.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { ManualPromise } from './manual-promise';
+import { Resolveable } from './resolvable';
+
+/**
+ * A signal is an externally fulfillable promise ( @see ManualPromise )
+ * that once awaited, will automatically reset to the unwaited state
```
bobbrow: >awaited [](http://example.com/codeflow?start=13&length=7)

semantics: I think it would be more correct to say that once it is fulfilled it will be reset automatically.  The await itself does not reset this promise (i.e. it could be awaited in more than one place before the promise is fulfilled), rather the resolve/reject resets the promise.

fearthecowboy: Good Point

Extension/src/Automation/System/guards.ts
```diff
@@ -0,0 +1,67 @@
...
+    static iterable<T = unknown>(instance: any): instance is Iterable<T> {
+        return typeof instance !== 'string' && !!instance[Symbol.iterator];
+    }
+
+    static asyncIterable(instance: any): instance is AsyncIterable<unknown> {
+        return !!instance[Symbol.asyncIterator];
+    }
+
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    static Constructor(instance: any): instance is Constructor {
```
bobbrow: nit: lower case 'c' for method name (or is it impossible because it's a keyword)?

bobbrow:
>     return input === undefined || isArrayOfString(input);

move to `guard.ts` or update implementations to use the ones in `guard.ts` until these can be removed.

---
Refers to: Extension/src/common.ts:335 in f3ac7cb. [](commit_id = f3ac7cb9fcee974c2890a52fd9821e5e096e1d19, deletion_comment = False)

Extension/src/cppTools1.ts
```diff
@@ -64,6 +64,7 @@ export class CppTools1 implements CppToolsTestApi, CppToolsTestExtension {
     }
 }
 
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
bobbrow: If we don't need `NullCppTools` anymore, just delete it.  It used to be returned during activation if cpptools was unable to run on that platform, but I see that code is gone now.

bobbrow: The platform compatibility check is done in `checkVsixCompatibility`, but it appears that NullCppTools didn't come along for the ride.

Extension/src/main.ts
```diff
@@ -291,13 +292,12 @@ async function checkVsixCompatibility(): Promise<void> {
...
             }
             if (promise) {
-                promise.then(async (value) => {
-                    if (value === moreInfoButton) {
-                        await vscode.commands.executeCommand("markdown.showPreview", vscode.Uri.file(getLocalizedHtmlPath("Reinstalling the Extension.md")));
-                    } else if (value === ignoreButton) {
-                        ignoreMismatchedCompatibleVsix.Value = true;
-                    }
-                });
+                const value = await promise;
```
bobbrow: I don't think we should await this promise because it's tracking a popup message.  `.then` is the right pattern for that.  We could remove `async` from the handler though and just not `await` the `executeCommand` because there's no need to do so.

Extension/src/telemetry.ts
```diff
@@ -137,13 +126,7 @@ export function logLanguageServerEvent(eventName: string, properties?: { [key: s
...
 
     if (initializationPromise) {
-        try {
-            // Use 'then' instead of 'await' because telemetry should be "fire and forget".
-            initializationPromise.then(sendTelemetry);
-            return;
-        } catch (e) {
-            // Continue even if we were not able to initialize the experimentation platform.
-        }
+        void initializationPromise.then(sendTelemetry).catch(logAndReturn.undefined);
```
bobbrow: this line should `return` and have the same chain as the debugger telemetry function.

Extension/src/Automation/System/guards.ts
fearthecowboy: Indeed, `constructor` is one of the keywords that can't be a member name. (

Extension/src/cppTools1.ts
fearthecowboy: removed.

fearthecowboy:
> move to `guards.ts` or update implementation

I'll add that to the list of things to clean up in round two -- I've got a list going in #11119

Extension/src/SSH/sshCommandRunner.ts
```diff
@@ -260,229 +256,232 @@ export function runInteractiveSshTerminalCommand(args: ITerminalCommandArgs): Pr
...
-
-            if (windowListener) {
-                windowListener.dispose();
-                windowListener = undefined;
-            }
-
-            if (tokenListener) {
-                tokenListener.dispose();
-            }
+    // return new Promise(async (resolve, reject) => {
```
sean-mcmanus: Is this commented line supposed to be here?

Extension/ui/settings.ts
```diff
@@ -352,7 +357,7 @@ class SettingsApp {
                 // Get HTML element containing the string, as we can't localize strings in HTML js
                 const noCompilerSpan: HTMLSpanElement = <HTMLSpanElement>document.getElementById(elementId.noCompilerPathsDetected);
                 const option: HTMLOptionElement = document.createElement("option");
-                option.text = noCompilerSpan.textContent;
+                option.text = noCompilerSpan.textContent ||"";
```
sean-mcmanus: No space after ||? Is this file formatted? Should this be `??` or was that going to be handled in the follow up PR?

Extension/src/Utility/System/guards.ts
```diff
@@ -0,0 +1,67 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
```
sean-mcmanus, line 5: Why is this file called "guard" -- is there a better name that could be used? Should it be called "is.ts"?

Extension/src/Utility/System/guards.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Socket } from 'node:net';
+import { isPromise } from 'node:util/types';
+import { AsyncConstructor, Constructor, Primitive } from './types';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export class is {
```
sean-mcmanus, line 11: Is this going to cause any usability issue in regards to the TypeScript "is" keyword or is the overloaded usage desired? Would it be better as something different like isType?

sean-mcmanus (dismissed):

Extension/src/Utility/System/guards.ts
bobbrow: These kinds of functions are referred to as type guards, so the name actually works.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

Extension/src/Utility/System/guards.ts
fearthecowboy: `is` is not a reserved word.

sean-mcmanus: Yeah, I didn't say it was a reserved word, I just mentioned that it's a keyword.

Extension/src/Utility/System/guards.ts
sean-mcmanus: But is "guard" too general? Could the file be typeGuard.ts so that's more descriptive?

sean-mcmanus:
@fearthecowboy Can the user's PR at https://github.com/microsoft/vscode-cpptools/pull/10388 be reviewed and merged before this PR gets checked in so they don't have to make more changes?

Extension/src/Utility/System/guards.ts
fearthecowboy: `is` is not a keyword in JavaScript/TypeScript at all.

Extension/src/SSH/sshCommandRunner.ts
fearthecowboy: ah, nuts. no, I'll clean that up

Extension/ui/settings.ts
fearthecowboy: I found an eslint rule to enforce the spacing on this. Fixed

Extension/src/Utility/System/guards.ts
fearthecowboy: I'm happy with guard.

fearthecowboy:
> @fearthecowboy Can the user's PR at #10388 be reviewed and merged before this PR gets checked in so they don't have to make more changes?

Yes, that can be merged first.

Extension/src/Utility/System/guards.ts
sean-mcmanus: I'm referring to https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates -- the TypeScript lexical tokenizer reports it as a keyword in regards to lexical colorization. I'm not sure what it's called technically.

sean-mcmanus (dismissed):

sean-mcmanus (dismissed):

michelleangela (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -1182,35 +1183,32 @@ export class DefaultClient implements Client {
                 const value: string | undefined = await vscode.window.showInformationMessage(localize("selectCompiler.message", "The compiler {0} was found. Do you want to configure IntelliSense with this compiler?", compilerDefaults.compilerPath), confirmCompiler, selectCompiler);
                 if (value === confirmCompiler) {
                     settings.defaultCompilerPath = compilerDefaults.compilerPath;
-                    await this.addTrustedCompiler(settings.defaultCompilerPath);
```
Colengms, line 1185: This looks like it changes behavior.  `addTrustedCompiler` will asynchronously set `compilerDefaults`, which is referenced by `updateClientConfigurations()`.  If not awaited here, the updated value may not be used.

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -4,21 +4,21 @@
...
 import { Middleware } from 'vscode-languageclient';
+import * as util from '../common';
 import { Client } from './client';
-import * as vscode from 'vscode';
 import { clients, onDidChangeActiveTextEditor, processDelayedDidOpen } from './extension';
-import * as util from '../common';
 
 export function createProtocolFilter(): Middleware {
     // Disabling lint for invoke handlers
-    const invoke1 = (a: any, next: (a: any) => any): any => clients.ActiveClient.requestWhenReady(() => next(a));
```
Colengms, line 15: Although an existing bug, the intention was for these invocations to never run in parallel.  Filtered message B should never be delivered to the server before filtered message A (due to asyncs).  We actually had some pretty terrible issues around this, which may apparently only have been addressed due to favorable timing.  Could you update these with the right new pattern to use to queue up `next` as a blocking task, such that it must be complete before `ready` is resolved for all other callers?  The intention was to do the same for all uses of `requestWhenReady` that are passed `processDelayedDidOpen`, ensuring none of these messages will get delivered before a preceding `processDelayedDidOpen` had fully completed.  ...  We may want to severe the concept of "(language client is) ready" from this message-ordering concern, as calls that care if the language client is 'ready' don't necessarily also need to participate in message ordering.

Extension/src/main.ts
```diff
@@ -99,11 +100,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
                     // If switching from disabled to enabled, we can continue activation.
                     currentIntelliSenseEngineValue = settings.intelliSenseEngine;
                     languageServiceDisabled = false;
-                    LanguageServer.activate();
+                    return LanguageServer.activate();
                 } else {
                     // We can't deactivate or change engines on the fly, so prompt for window reload.
                     reloadMessageShown = true;
-                    util.promptForReloadWindowDueToSettingsChange();
+                    return util.promptForReloadWindowDueToSettingsChange();
```
Colengms: Looking through the code to VS Code, it looks like it expects this activate to be an async.  (src\vs\workbench\api\common\extHostExtensionActivator.ts, in VS Code repo)  Wouldn't adding this return here, which awaits `showInformationMessage` under the hood, potentially cause us to never complete activation?

This is one reason why I'd prefer we do not ever propagate the return/await Promises that will only resolve when a user responds to UI.  Even for potential perf measurements - I don't think we will want to care how long it took the user to take an action, if ever.  It would be safer not to incur the risk of someone accidentally adding an await that may never resolve.

Colengms (changes requested):

Extension/src/main.ts
Colengms: It looks like I overlooked that this was happening in a lambda.  My concern about returning/awaiting promises from UI calls that may never be actioned on, is still valid.  :)

Extension/src/LanguageServer/client.ts
fearthecowboy: Weird. No idea why that one got changed. Fixing.

Extension/src/LanguageServer/protocolFilter.ts
fearthecowboy: I'd rather add that to the follow-up task #11119 so that I can make sure that when the fix is done, that it solves it correctly, and try and write some tests to make sure that it's correct.

While this PR doesn't fix this issue, it doesn't alter it either.

Extension/src/main.ts
fearthecowboy: The other reason that you return the Promise is to give the caller the _choice_ to know when something is completed, and opt to wait or do something after the call is complete (via `.then(...)` ) , which can be handy even if it's "after the user has done XXX then do YYY"

In this case, there's nothing to do, so resolving.

Colengms: We do not ever need or want to give the caller the choice to wait for a UI operation to complete, in which we do not require any response from the user.  It's unnecessary to support this, and doing so introduces a way in which a simple mistake can lead to a deadlock.

Extension/src/LanguageServer/protocolFilter.ts
Colengms: > We actually had some pretty terrible issues around this, which may apparently only have been addressed due to favorable timing.

My concern is that your change may actually re-introduce the issue.  I'm not entirely convinced that it does not change the behavior, due to the prior implementation having successfully addressed repros of the bug that you're saying technically still exists. We should have Sean test this branch against his repro of the original issues to ensure it doesn't re-introduce the issue, preferably before this is checked in.  (or before shipping, if your follow-up doesn't make it in before that.)  Please make the follow-up a high priority.

Extension/src/main.ts
fearthecowboy: Ok, reverting

Extension/src/LanguageServer/protocolFilter.ts
sean-mcmanus: Yeah, that bug was https://github.com/microsoft/vscode-cpptools/issues/10035 -- I was able to repro it pretty readily via typing functions in a header like `        int fooo01() const;` during the initial loading period and the document corruption could be seen after formatting, but it may be harder to repro if your computer is "too fast". We have to make sure edits don't get sent out of order or the document can get corrupt. We haven't gotten any reports of document corruption since that issue was fixed with 1.13.6.

sean-mcmanus (approved):

Extension/src/LanguageServer/protocolFilter.ts
sean-mcmanus: I don't repro a bug, but it could just be a really low repro probability.

Colengms (approved):

Issue #10934, opened 2023-05-10
Add telemetry for configuration


bobbrow:
When a user updates their configuration, we need to log telemetry for how and where it was configured. For example, if a user sets their default.compilerPath setting to configure IntelliSense we need to log it.

PR #11093, opened 2023-06-14
Add configuration telemetry

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10934

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1059,6 +1059,7 @@ export class DefaultClient implements Client {
                         return;
                 }
             }
+            const showButtonSender: string = "quickPick";
             if (index === paths.length - 3) {
                 const result: vscode.Uri[] | undefined = await vscode.window.showOpenDialog();
                 if (result === undefined || result.length === 0) {
@@ -1080,12 +1081,12 @@ export class DefaultClient implements Client {
                     await this.configuration.updateCustomConfigurationProvider(provider.extensionId);
                     this.onCustomConfigurationProviderRegistered(provider);
                     telemetry.logLanguageServerEvent("customConfigurationProvider", { "providerId": provider.extensionId });
-                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, "quickPick");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, showButtonSender);
                     return;
                 } else if (index < compileCommandsIndex) {
                     action = "select compile commands";
                     this.configuration.setCompileCommands(this.compileCommandsPaths[index - configProvidersIndex - 1]);
-                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, "quickPick");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, showButtonSender);
                     return;
                 } else {
                     action = "select compiler";
@@ -1094,7 +1095,7 @@ export class DefaultClient implements Client {
                 }
             }
 
-            ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "quickPick");
+            ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, showButtonSender);
             await this.addTrustedCompiler(settings.defaultCompilerPath);
             DefaultClient.updateClientConfigurations();
         } finally {
@@ -1686,12 +1687,13 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
+                const showButtonSender: string = "settingsChanged";
                 if (changedSettings["default.configurationProvider"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, showButtonSender);
                 } else if (changedSettings["default.compileCommands"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, showButtonSender);
                 } if (changedSettings["default.compilerPath"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, showButtonSender);
                 }
                 this.configuration.onDidChangeSettings();
                 telemetry.logLanguageServerEvent("CppSettingsChange", changedSettings, undefined);
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1042,7 +1042,7 @@ export class DefaultClient implements Client {
                 if (showSecondPrompt) {
                     this.showPrompt(selectIntelliSenseConfig, true, sender);
                 }
-                ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.NotConfigured, "disablePrompt");
+                ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.NotConfigured, "disablePrompt");
                 return;
             }
             if (index === paths.length - 2) {
@@ -1080,12 +1080,12 @@ export class DefaultClient implements Client {
                     await this.configuration.updateCustomConfigurationProvider(provider.extensionId);
                     this.onCustomConfigurationProviderRegistered(provider);
                     telemetry.logLanguageServerEvent("customConfigurationProvider", { "providerId": provider.extensionId });
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.ConfigProvider, "quickPick");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, "quickPick");
                     return;
                 } else if (index < compileCommandsIndex) {
                     action = "select compile commands";
                     this.configuration.setCompileCommands(this.compileCommandsPaths[index - configProvidersIndex - 1]);
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompileCommands, "quickPick");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, "quickPick");
                     return;
                 } else {
                     action = "select compiler";
@@ -1094,7 +1094,7 @@ export class DefaultClient implements Client {
                 }
             }
 
-            ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompilerPath, "quickPick");
+            ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "quickPick");
             await this.addTrustedCompiler(settings.defaultCompilerPath);
             DefaultClient.updateClientConfigurations();
         } finally {
@@ -1150,7 +1150,7 @@ export class DefaultClient implements Client {
                     await this.addTrustedCompiler(settings.defaultCompilerPath);
                     DefaultClient.updateClientConfigurations();
                     action = "confirm compiler";
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompilerPath, "promptSelectCompiler");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "promptSelectCompiler");
                 } else if (value === selectCompiler) {
                     this.handleIntelliSenseConfigurationQuickPick(true, sender, true);
                     action = "show quickpick";
@@ -1184,7 +1184,7 @@ export class DefaultClient implements Client {
                     await this.addTrustedCompiler(settings.defaultCompilerPath);
                     DefaultClient.updateClientConfigurations();
                     action = "confirm compiler";
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompilerPath, "promptSelectIntelliSense");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "promptSelectIntelliSense");
                 } else if (value === selectCompiler) {
                     this.handleIntelliSenseConfigurationQuickPick(true, sender);
                     action = "show quickpick";
@@ -1687,11 +1687,11 @@ export class DefaultClient implements Client {
                     this.configuration.handleConfigurationChange();
                 }
                 if (changedSettings["default.configurationProvider"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.ConfigProvider, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, "settingsChanged");
                 } else if (changedSettings["default.compileCommands"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompileCommands, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, "settingsChanged");
                 } if (changedSettings["default.compilerPath"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, true, this, ConfigurationType.CompilerPath, "settingsChanged");
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "settingsChanged");
                 }
                 this.configuration.onDidChangeSettings();
                 telemetry.logLanguageServerEvent("CppSettingsChange", changedSettings, undefined);
@@ -2734,13 +2734,13 @@ export class DefaultClient implements Client {
             this.showConfigureIntelliSenseButton = false;
         }
         if (!configProviderNotSet) {
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, true, this, ConfigurationType.ConfigProvider, "handleConfig");
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.ConfigProvider, "handleConfig");
         } else if (!compileCommandsNotSet) {
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, true, this, ConfigurationType.CompileCommands, "handleConfig");
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.CompileCommands, "handleConfig");
         } else if (!compilerPathNotSet) {
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, true, this, ConfigurationType.CompilerPath, "handleConfig");
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.CompilerPath, "handleConfig");
         } else {
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, true, this, ConfigurationType.NotConfigured, "handleConfig");
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.NotConfigured, "handleConfig");
         }
 
         if (showConfigStatus && !displayedSelectCompiler && !statusBarIndicatorEnabled) {
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2738,8 +2738,8 @@ export class DefaultClient implements Client {
                 this.showConfigureIntelliSenseButton = false;
             }
         } else if (showConfigStatus && !displayedSelectCompiler) {
-                this.promptSelectIntelliSenseConfiguration(false, "notification");
-                displayedSelectCompiler = true;
+            this.promptSelectIntelliSenseConfiguration(false, "notification");
+            displayedSelectCompiler = true;
         }
 
         const configProviderType: ConfigurationType = this.configuration.ConfigProviderAutoSelected ? ConfigurationType.AutoConfigProvider : ConfigurationType.ConfigProvider;
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2730,11 +2730,18 @@ export class DefaultClient implements Client {
 
         showConfigStatus = showConfigStatus || (configurationNotSet &&
             !!compilerDefaults && !compilerDefaults.trustedCompilerFound && trustedCompilerPaths && (trustedCompilerPaths.length !== 1 || trustedCompilerPaths[0] !== ""));
-        if (showConfigStatus) {
-            this.showConfigureIntelliSenseButton = true;
-        } else {
-            this.showConfigureIntelliSenseButton = false;
+
+        if (statusBarIndicatorEnabled) {
+            if (showConfigStatus) {
+                this.showConfigureIntelliSenseButton = true;
+            } else {
+                this.showConfigureIntelliSenseButton = false;
+            }
+        } else if (showConfigStatus && !displayedSelectCompiler) {
+                this.promptSelectIntelliSenseConfiguration(false, "notification");
+                displayedSelectCompiler = true;
         }
+
         const configProviderType: ConfigurationType = this.configuration.ConfigProviderAutoSelected ? ConfigurationType.AutoConfigProvider : ConfigurationType.ConfigProvider;
         const compilerType: ConfigurationType = this.configuration.CurrentConfiguration?.compilerPathIsExplicit ? ConfigurationType.CompilerPath : ConfigurationType.AutoCompilerPath;
         const configType: ConfigurationType =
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -1726,8 +1726,12 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
-                if (changedSettings["default.compilerPath"] !== undefined || changedSettings["default.compileCommands"] !== undefined || changedSettings["default.configurationProvider"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, this);
+                if (changedSettings["default.compilerPath"] !== undefined) {
```
Colengms: A lot of users specify multiple of these.  Could you check these in the same order they would be tried when multiple are specified?  That order would be (generally) :
* config provider
* compile_commands.json
* compilerPath
(the same order these are checked, in cpp_properties::provide_args())

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1242,6 +1239,16 @@ export class CppProperties {
         }
 
         this.applyDefaultIncludePathsAndFrameworks();
+        const configuration: Configuration | undefined = this.CurrentConfiguration;
+        if (configuration) {
+            if (configuration.compilerPath !== undefined) {
```
Colengms: Similar to previous comment about check order.  That way, we would more accurately report their 'effective' configuration type.

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,14 +435,15 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
+    public async ShowConfigureIntelliSenseButton(show: boolean, client: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
+        if (configurationType !== undefined && sender !== undefined && this.showConfigureIntelliSenseButton && !show) { // Only log telemetry when the button is being hidden.
```
bobbrow: I think it would be good to log when the button is shown because we may not see them ever fix it.  Can you add the "show" property to the Properties for this event?

With this in mind, there is another call missed in `client.ts` around line 2775 that should pass in the type and sender (it will actually be a few lines down since lines were added in this PR)

bobbrow (commented):
Did we decide not to log this for ActiveConfigChanged (`ui.ts` line 507)?

Extension/src/LanguageServer/ui.ts
```diff
@@ -118,7 +125,9 @@ export class LanguageStatusUI {
             title: this.configureIntelliSenseStatusItem.name,
             arguments: ['statusBar']
         };
-        this.ShowConfigureIntelliSenseButton(false, this.currentClient);
+        if (this.currentClient !== undefined) {
+            this.ShowConfigureIntelliSenseButton(false, this.currentClient);
```
Colengms: This is confusing to me.  Since this is within the constructor, it shouldn't need to 'remove' the indicator.  Wouldn't it be sufficient to set `this.currentClient.setShowConfigureIntelliSenseButton(false)` here instead of incurring the rest of the logic in `ShowConfigureIntelliSenseButton()`?

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,14 +435,15 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
+    public async ShowConfigureIntelliSenseButton(show: boolean, client: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
+        if (configurationType !== undefined && sender !== undefined && this.showConfigureIntelliSenseButton && !show) { // Only log telemetry when the button is being hidden.
+            telemetry.logLanguageServerEvent('showConfigureIntelliSenseButton', { configurationType, sender });
+        }
         if (!await telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient) {
```
Colengms: Should this `telemetry.showStatusBarIntelliSenseButton` check occur first?  Or, if the intention is to log this telemetry even if that particular feature is disabled (experiment is not active for the user), I see at least one place where `ShowConfigureIntelliSenseButton` will not get called if the feature is not enabled. (In `handleConfigStatusOrPrompt`, if `statusBarIndicatorEnabled` is false).  I'm guessing we'll want to make sure all the calls to this function get through in order to provide telemetry, even when that experiment is not enabled?

Colengms (changes requested):

Extension/src/LanguageServer/ui.ts
browntarik: I don't understand, why would we want to make calls to the function if the experiment is not enabled?

Extension/src/LanguageServer/ui.ts
```diff
@@ -507,7 +520,6 @@ export class LanguageStatusUI {
         client.ActiveConfigChanged(value => {
             this.ActiveConfig = value;
             this.currentClient = client;
-            this.ShowConfigureIntelliSenseButton(client.getShowConfigureIntelliSenseButton(), client);
```
sean-mcmanus: I believe this should not be removed -- it's for when switching documents in a multi-root workspace, i.e. when the user switches to a file that belongs to a workspace that isn't configured. I'd recommend testing with a multi-root workspace to make sure that scenario still works correctly.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,12 +435,16 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
+    public async ShowConfigureIntelliSenseButton(show: boolean, logTelemetry: boolean, client?: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
+        if (configurationType !== undefined && sender !== undefined && logTelemetry) {
```
bobbrow: I don't think you need a `logTelemetry` boolean for this function.  Is there a case where you would pass in the configurationType and the sender and not want to log it?  I didn't see one in this PR.

browntarik: There was one but we went back on it but it could be useful to keep the boolean just because there may be cases in the future where we want to eliminate redundant telemetry

bobbrow: Please remove it.  It serves no purpose today and is easy to add later if we ever do actually need it.

Extension/src/LanguageServer/ui.ts
```diff
@@ -118,7 +125,9 @@ export class LanguageStatusUI {
             title: this.configureIntelliSenseStatusItem.name,
             arguments: ['statusBar']
         };
-        this.ShowConfigureIntelliSenseButton(false, this.currentClient);
+        if (this.currentClient !== undefined) {
```
bobbrow: It seems like this does the opposite of what it did before.  It used to pass in the currentClient, now it doesn't ever pass in a client.  Is that intentional?

bobbrow (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1241,6 +1238,17 @@ export class CppProperties {
             this.resetToDefaultSettings(true);  // I don't think there's a case where this will be hit anymore.
         }
 
+        const configuration: Configuration | undefined = this.CurrentConfiguration;
+        if (configuration) {
+            if (configuration.configurationProvider !== undefined) {
+                getUI().then((ui: LanguageStatusUI) => ui.ShowConfigureIntelliSenseButton(false, this.client, ConfigurationType.ConfigProvider, "baseConfiguration"));
```
sean-mcmanus: This appears to be causing the button to disappear after a configuration change or am I mistaken?

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -2724,15 +2728,22 @@ export class DefaultClient implements Client {
 
         showConfigStatus = showConfigStatus || (configurationNotSet &&
             !!compilerDefaults && !compilerDefaults.trustedCompilerFound && trustedCompilerPaths && (trustedCompilerPaths.length !== 1 || trustedCompilerPaths[0] !== ""));
+        if (showConfigStatus) {
+            this.showConfigureIntelliSenseButton = true;
```
sean-mcmanus: This may cause bugs in which the configuration status will appear even if the statusBarIndicatorEnabled is false. You should only set this if statusBarIndicatorEnabled is true, because other code that handles the activeDocumentChanged will check this variable. You can use the original check and just pass in showConfigStatus as the 1st param.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,12 +433,16 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
+    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
+        if (configurationType !== undefined && sender !== undefined) {
+            const showButton: string = show ? 'true' : 'false';
+            telemetry.logLanguageServerEvent('showConfigureIntelliSenseButton', { configurationType, sender, showButton });
```
sean-mcmanus, line 444: I see this telemetry getting sent 8 times on startup for a 2 workspace folder project with a configurationProvider set. I think maybe the check `client !== this.currentClient` that is done below should be added beforehand.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
bobbrow: That would be fine, but we do want to log telemetry before exiting the function, so the check for the experiment to exit early should remain below.

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,12 +433,16 @@ export class LanguageStatusUI {
...
+        }
         if (!await telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient) {
             return;
         }
         this.showConfigureIntelliSenseButton = show;
-        if (client) {
+        if (client !== undefined) {
             client.setShowConfigureIntelliSenseButton(show);
         }
         if (show) {
```
sean-mcmanus, line 454: Do we still need the ` telemetry.logLanguageServerEvent('configureIntelliSenseStatusBar');` telemetry below?

Extension/src/LanguageServer/client.ts
sean-mcmanus: Actually, I don't think this causes a bug, but it'd still be safer (or less potentially confusing) to not change the this.showConfigureIntelliSenseButton state when not in a mode where that is used.

Extension/src/LanguageServer/ui.ts
bobbrow: It's not the subject of this PR. We can discuss offline and add an issue if so.

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,14 +433,21 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
-        if (!await telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient) {
-            return;
-        }
+    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
         this.showConfigureIntelliSenseButton = show;
```
sean-mcmanus: I don't think this is the correct change -- you're causing setShowConfigureIntelliSenseButton to be run when without checking `!await telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient`.

This PR shouldn't cause any functionality change....well, I'm not sure if this change actually affects anything, but it's setting showConfigureIntelliSenseButton when the original intent was for that to not be changed if the checks below failed.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/ui.ts
sean-mcmanus: Yeah, but this PR adds telemetry to the same function that seems like it could be duplicate?

Extension/src/LanguageServer/ui.ts
sean-mcmanus: This issue wasn't resolved by the changes.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: i.e. I think the code that was moved up should be moved back down and the check for `client !== this.currentClient` should be moved up.

Extension/src/LanguageServer/ui.ts
sean-mcmanus: ...oh, looks like the telemetry doesn't track the same thing.

Extension/src/LanguageServer/ui.ts
bobbrow: I fixed it.

Extension/src/LanguageServer/ui.ts
bobbrow: I fixed it.

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1241,6 +1238,17 @@ export class CppProperties {
...
+            } else if (configuration.compileCommands !== undefined) {
+                getUI().then((ui: LanguageStatusUI) => ui.ShowConfigureIntelliSenseButton(false, this.client, ConfigurationType.CompileCommands, "baseConfiguration"));
+            } else if (configuration.compilerPath !== undefined) {
```
sean-mcmanus: This is logging compilerPath == "" as ConfigurationType.CompilerPath instead of ConfigurationType.NotConfigured....whereas the other code paths above log that same condition as NotConfigured.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
bobbrow: You're right.  Setting compilerPath to "" should count as "configured."

Extension/src/LanguageServer/client.ts
```diff
@@ -2724,15 +2728,22 @@ export class DefaultClient implements Client {
...
+        } else {
+            this.showConfigureIntelliSenseButton = false;
+        }
+        if (!configProviderNotSet) {
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.ConfigProvider, "handleConfig");
+        } else if (!compileCommandsNotSet) {
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.CompileCommands, "handleConfig");
+        } else if (!compilerPathNotSet) {
+            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, ConfigurationType.CompilerPath, "handleConfig");
```
sean-mcmanus: This ConfigurationType.CompilerPath also occurring when compilerPath is "" instead of NotConfigured.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Oh, is that configured?, I'm not sure, I just noticed it was inconsistent.

Extension/src/LanguageServer/client.ts
sean-mcmanus: ...or is that intentional? I was seeing the opposite in the configurationChange handler, so it was sending inconsistent results for the same config.

Extension/src/LanguageServer/configurations.ts
bobbrow: I don't think we show the "configure IntelliSense" button if they set compilerPath to `""`, do we?

Extension/src/LanguageServer/ui.ts
```diff
@@ -426,12 +433,20 @@ export class LanguageStatusUI {
     private showConfigureIntelliSenseButton: boolean = false;
     private configureIntelliSenseTimeout?: NodeJS.Timeout;
 
-    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client): Promise<void> {
-        if (!await telemetry.showStatusBarIntelliSenseButton() || client !== this.currentClient) {
+    public async ShowConfigureIntelliSenseButton(show: boolean, client?: Client, configurationType?: ConfigurationType, sender?: string): Promise<void> {
```
michelleangela, line 438: Change the type of `sender` to an enum, similar to `enum ConfigurationType` to avoid typos in re-usage of the same senders.

Extension/src/LanguageServer/client.ts
bobbrow: `compilerPathNotSet` is only testing for `undefined`.  If it's `""`, that should count as configured.

Extension/src/LanguageServer/ui.ts
bobbrow: We don't do this with the other APIs accepting `sender`.  Maybe a const string would be helpful, but I don't think the intent is to keep this telemetry forever.

Extension/src/LanguageServer/client.ts
```diff
@@ -2724,15 +2728,22 @@ export class DefaultClient implements Client {
 
         showConfigStatus = showConfigStatus || (configurationNotSet &&
             !!compilerDefaults && !compilerDefaults.trustedCompilerFound && trustedCompilerPaths && (trustedCompilerPaths.length !== 1 || trustedCompilerPaths[0] !== ""));
+        if (showConfigStatus) {
+            this.showConfigureIntelliSenseButton = true;
+        } else {
+            this.showConfigureIntelliSenseButton = false;
+        }
+        if (!configProviderNotSet) {
```
michelleangela: Can this block of `if/else` be changed to something like:

```ts
let configType: ConfigurationType = none;
if (condition a) {
      configType = ConfigProvider;
} else if (condition b) {
    configType = CompileCommands;
} else if (......) {
    // etc.
}

// Call ShowConfigureIntelliSenseButton after evaluting the value for ConfigurationType
ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this, configType, "handleConfig");
```

bobbrow: sure

Extension/src/LanguageServer/ui.ts
michelleangela: If string is commonly used, then I suggest to create a local `const string` where `ShowConfigureIntelliSenseButton` is called multiple times in a function with the same sender.

bobbrow: done

sean-mcmanus (changes requested):
Is telemetry supposed to be sent when switching configs? I don't see that. That seems inconsistent because it seems to handle a direct settings change. Maybe I'm not understanding something though.

Oh, looks like it sends telemetry on default settings changes only and not the c_cpp_properties.json changes -- I think that's related to an existing bug that I logged.

I don't know if that matters or not. It just seems inconsistent to log C_Cpp.default.* settings changes and not the c_cpp_properties.json changes or config switches that yield different active c_cpp_properties.json configs.

michelleangela (dismissed):

bobbrow:
> Is telemetry supposed to be sent when switching configs? I don't see that. That seems inconsistent because it seems to handle a direct settings change. Maybe I'm not understanding something though.
> 
> Oh, looks like it sends telemetry on default settings changes only and not the c_cpp_properties.json changes -- I think that's related to an existing bug that I logged.
> 
> I don't know if that matters or not.

`handleConfigurationChange` is called in a lot of scenarios, including when the active configuration changes (search `selectionchanged` to start tracing through it).  I believe it is also called when the c_cpp_properties.json changes (`this.configFileWatcher.onDidChange`).

bobbrow:
I can rename "baseConfiguration" to "configurationChanged" if that helps clarify.

sean-mcmanus:
> `handleConfigurationChange` is called in a lot of scenarios, including when the active configuration changes (search `selectionchanged` to start tracing through it). I believe it is also called when the c_cpp_properties.json changes (`this.configFileWatcher.onDidChange`).

Ah, you're right, I was just confused.

sean-mcmanus (dismissed):

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -2724,15 +2730,20 @@ export class DefaultClient implements Client {
 
         showConfigStatus = showConfigStatus || (configurationNotSet &&
             !!compilerDefaults && !compilerDefaults.trustedCompilerFound && trustedCompilerPaths && (trustedCompilerPaths.length !== 1 || trustedCompilerPaths[0] !== ""));
+        if (showConfigStatus) {
+            this.showConfigureIntelliSenseButton = true;
+        } else {
+            this.showConfigureIntelliSenseButton = false;
+        }
+        const configType: ConfigurationType =
```
sean-mcmanus, line 2747: Should AutoConfigProvider be handled here?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
bobbrow: yes

sean-mcmanus (changes requested):
There's an issue with the AutoConfigProvider case not being handled correctly.

Extension/src/LanguageServer/client.ts
bobbrow: reverted this

Extension/src/LanguageServer/client.ts
```diff
@@ -2731,11 +2737,22 @@ export class DefaultClient implements Client {
             } else {
                 this.showConfigureIntelliSenseButton = false;
             }
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this);
         } else if (showConfigStatus && !displayedSelectCompiler) {
-            this.promptSelectIntelliSenseConfiguration(false, "notification");
-            displayedSelectCompiler = true;
+                this.promptSelectIntelliSenseConfiguration(false, "notification");
```
bobbrow: oops

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -2731,11 +2737,22 @@ export class DefaultClient implements Client {
             } else {
                 this.showConfigureIntelliSenseButton = false;
             }
-            ui.ShowConfigureIntelliSenseButton(this.showConfigureIntelliSenseButton, this);
         } else if (showConfigStatus && !displayedSelectCompiler) {
             this.promptSelectIntelliSenseConfiguration(false, "notification");
             displayedSelectCompiler = true;
         }
+
+        const configProviderType: ConfigurationType = this.configuration.ConfigProviderAutoSelected ? ConfigurationType.AutoConfigProvider : ConfigurationType.ConfigProvider;
```
sean-mcmanus, line 2745: I think the telemetry should not be sent for the config provider if the config provider hasn't been registered yet and it's relying on the cache -- i.e. I think a 2nd configProviderNotSet variable should be saved that excludes the cache check and that should be checked before sending config provider telemetry. Otherwise, it sends invalid results early on (i.e. that change after the 1st config provider registers).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1686,8 +1687,13 @@ export class DefaultClient implements Client {
                 if (changedSettings["legacyCompilerArgsBehavior"]) {
                     this.configuration.handleConfigurationChange();
                 }
-                if (changedSettings["default.compilerPath"] !== undefined || changedSettings["default.compileCommands"] !== undefined || changedSettings["default.configurationProvider"] !== undefined) {
-                    ui.ShowConfigureIntelliSenseButton(false, this);
+                const showButtonSender: string = "settingsChanged";
+                if (changedSettings["default.configurationProvider"] !== undefined) {
+                    ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.ConfigProvider, showButtonSender);
```
sean-mcmanus, line 1692: The auto-config type doesn't appear to be checked here, i.e. I see it send ConfgProvider instead of AutoConfigProvider after clearing out the setting.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
bobbrow: Like this?
```
const configProviderSet = settings.defaultConfigurationProvider ||
        this.configuration.CurrentConfiguration?.configurationProvider ||
        this.configuration.CurrentConfiguration?.configurationProviderInCppPropertiesJson;
```

Extension/src/LanguageServer/client.ts
bobbrow: I think one of the primary goals for this telemetry is to identify the users who have figured out how to configure their stuff.  If we fired ConfigProvider in the past for the user because this default.configurationProvider setting had a value, then we will have already categorized this user as someone who knows how to configure themselves and it wouldn't really matter if another ConfigProvider type was sent for them.  So to avoid churning the code too much before we release, I'd like to ignore this one.

sean-mcmanus: sure

Extension/src/LanguageServer/client.ts
sean-mcmanus: I was expecting the telemetry to not be sent if configProviderNotSet is true and configProviderNotSetAndNoCache is false -- otherwise it sends notConfigured.

bobbrow: I can't think of any cases where we wouldn't want to send telemetry. Sending notConfigured in this case sounds correct (unless we auto-selected a compiler).

sean-mcmanus: The potential problem is that when users are auto configured, we'll always send notConfigured (before the 1st config provider has registerd but there's a cache) when in fact they'll use the cache and then it'll be auto configured after the config provider registers. Its seems better to exclude telemetry in this case since it's just noise.

bobbrow: The queries we'll write for this won't be impacted by that noise.  If the config provider never registers, then they are actually not configured.

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: ok

Colengms (approved):

Issue #10500, opened 2023-02-09
Create Declaration / Definition incorrectly adds the definition into a function


sean-mcmanus:
test.h
```cpp
void func();
```

test.cpp
```cpp
int main()
{
    return 0;
}
```

Bug: It creates invalid code:

```cpp
#include "test.h"
int main()
{

    void func()
    {
    }
    return 0;
}
```

michelleangela:
This is the same issue as https://github.com/microsoft/vscode-cpptools/issues/10464. The issue is in typescript, where there are two text edits and it calculates or adjusts the edit positions.

If there are at least 2 spaces after the last function or namespace scope, then it doesn't repro.

Issue #10464, opened 2023-02-03
Create Declaration / Definition incorrectly adds the definition into a namespace it doesn't belong to


sean-mcmanus:
test.cpp
```cpp
namespace ns {

}
```

test.h
```cpp
class foo {
public:
    foo();
};
```

1. Do Create Declaration / Definition on the foo() constructor in test.h.

Bug: It adds the foo() definition to the namespace block that it doesn't belong to (IntelliSense squiggles). The bug doesn't repro in VS -- the definition is correctly placed after the namespace block.

michelleangela:
If there are at least 2 lines after the last function or namespace scope, then it doesn't repro.

So this is the same issue as #10500

PR #11110, opened 2023-06-20
remove extra adjustments to textEdit for CDD

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10500, https://github.com/microsoft/vscode-cpptools/issues/10464

This change removes the adjustments that were previously added to account for two text edits.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -3509,12 +3509,10 @@ export class DefaultClient implements Client {
             for (const edit of result.edit.changes[file]) {
                 const range: vscode.Range = makeVscodeRange(edit.range);
                 // Get new lines from an edit for: #include header file.
-                if (lastEdit && lastEdit.newText.includes("#include")) {
-                    if (lastEdit.range.isEqual(range)) {
-                        // Destination file is empty.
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -3515,18 +3514,10 @@ export class DefaultClient implements Client {
                         // Destination file is empty.
                         // The edit positions for #include header file and definition or declaration are the same.
                         selectionPositionAdjustment = (lastEdit.newText.match(/\n/g) || []).length;
```
michelleangela: The line `if (lastEdit.range.isEqual(range)) {` can be combined with the previous line into one conditional since the else for `if (lastEdit.range.isEqual(range))` is removed.

sean-mcmanus (approved):

browntarik:
For the scenario where a user generates a definition in a source file with no #include with an existing namespace and another existing function in the header file it will misplace the definition due to miscalculation of neighboring functions. 

example.cpp

namespace uniqueName {
int GlobalFunction3()
{
    return 0;
}
}

michelleangela (approved):

Issue #11135, opened 2023-06-27
Add support for `/*` `/**` `/` and `\\**`  syntax in include wildcards


browntarik:
Follow-up for  https://github.com/microsoft/vscode-cpptools/pull/10388

sean-mcmanus:
On Windows, headers/C*/ doesn't work, but headers/C* works.

PR #11136, opened 2023-06-27
resolve paths with `/`, `/*`, `/**` and `\\**`

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11135


Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,8 +737,16 @@ export class CppProperties {
 
     private resolveAndSplit(paths: string[] | undefined, defaultValue: string[] | undefined, env: Environment, glob = false): string[] {
         const result: string[] = [];
+        let recursiveDoubleStar: boolean = false;
+        let recursiveForwardSlash: boolean = false;
+        let recursiveForwardSlashSingleStar: boolean = false;
+        let recursiveDoubleBackwardSlashDoubleStar: boolean = false;
         if (paths) {
             paths = this.resolveDefaults(paths, defaultValue);
+            recursiveDoubleStar = paths[0].endsWith('**') && !paths[0].endsWith('\\**');
```
sean-mcmanus: This needs to be moved below into the `for (let res of result)` loop. Test case is multiple paths in includePath.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
browntarik: The issue with that is that the strings get normalized so I can't check correctly for correct path endings

browntarik: Never mind, I have a fix.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
+                if (recursiveDoubleStar || recursiveForwardSlashSingleStar) {
+                    res = res.slice(0, res.length - 2);
+                } else if (recursiveForwardSlash) {
+                    res = res.slice(0, res.length - 1);
+                } else if (recursiveDoubleBackwardSlashDoubleStar) {
+                    res = res.slice(0, res.length - 4);
+                }
+
+                // fastGlob can't deal with backslash-separated path => remove them
+                const normalized: string = res.replace(/\\/g, '/');
```
Colengms: Likely an existing bug, but backslashes as path separators are specific to windows.  On other platforms, backslashes can appear in files/directory names. Could you conditionalize any backslash-related logic under an `isWindows` check?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
-                res = res.slice(0, res.length - 2);
-            }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
-            // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
-            const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                fastGlob.sync(normalized, { onlyDirectories: true, cwd}) : [res];
-            globResult.push(...matches.map(s => recursive ? s + '**' : s));
+        let recursiveDoubleStar: boolean = false;
```
Colengms: Could you move these variable declarations to where they are initialized in the for loop, to reduce the number of lines?  It doesn't look like these vars are needed outside of the scope of a single for loop iteration.

```
                let recursiveDoubleStar: boolean = res.endsWith('**') && res.endsWith('\\**');
                let recursiveForwardSlash: boolean = res.endsWith('/');
                let recursiveForwardSlashSingleStar: boolean = res.endsWith('/*');
                let recursiveDoubleBackwardSlashDoubleStar: boolean = res.endsWith('\\**');
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
+                }
+
+                // fastGlob can't deal with backslash-separated path => remove them
+                const normalized: string = res.replace(/\\/g, '/');
+                const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+                // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
+                const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
+                    fastGlob.sync(normalized, { onlyDirectories: true, cwd }) : [res];
+
+                if (recursiveForwardSlash) {
```
Colengms: The if condition already checks the value of the variable, the second check isn't needed?

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
-            const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                fastGlob.sync(normalized, { onlyDirectories: true, cwd}) : [res];
-            globResult.push(...matches.map(s => recursive ? s + '**' : s));
+        let recursiveDoubleStar: boolean = false;
+        let recursiveForwardSlash: boolean = false;
+        let recursiveForwardSlashSingleStar: boolean = false;
+        let recursiveDoubleBackwardSlashDoubleStar: boolean = false;
+        if (paths) {
+            for (let res of paths) {
+                recursiveDoubleStar = res.endsWith('**') && res.endsWith('\\**');
```
bobbrow: Is it necessary to check `res.endsWith('**')` or can that be removed?  It seems like a subset of `res.endsWith('\\**')`

bobbrow: Or did you mean for it to be `/**`  (since I don't see that case anywhere)?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
+                recursiveForwardSlash = res.endsWith('/');
+                recursiveForwardSlashSingleStar = res.endsWith('/*');
```
bobbrow: Should there also be a `\\*` case? (backwardSlashSingleStar)

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
+                recursiveDoubleBackwardSlashDoubleStar = res.endsWith('\\**');
```
bobbrow: This seems like a duplicate of `recursiveDoubleStar`

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +739,52 @@ export class CppProperties {
...
+                // fastGlob will expand the ending double wildcard. temporary strip them before expanding
+                if (recursiveDoubleStar || recursiveForwardSlashSingleStar) {
+                    res = res.slice(0, res.length - 2);
+                } else if (recursiveForwardSlash) {
+                    res = res.slice(0, res.length - 1);
+                } else if (recursiveDoubleBackwardSlashDoubleStar) {
+                    res = res.slice(0, res.length - 4);
```
bobbrow: Should it be `length - 3`?  There's only 3 characters in this one.

Extension/src/LanguageServer/configurations.ts
michelleangela: Also, if the variables are only needed in the conditional statements, then directly querying `res.endsWith(...)` in the conditional maybe sufficient?

michelleangela (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +741,47 @@ export class CppProperties {
...
+            for (let res of paths) {
+                let counter: integer = 0;
+                let replacementString: string = '';
+                const lastIndex: integer = res.length - 1;
+                // Detect and glob all wildcard variations by looking at last character in the path first.
+                for (let i: integer = lastIndex; i >= 0; i--) {
+                    if (res[i] === '*') {
+                        counter++;
+                        replacementString = replacementString.replace (/^/, res[i]);
+                    } else if (res[i] === '/' || res[i] === '\\') {
```
bobbrow: I think `isWindows` should be required in order for backslash to count:
```
(res[i] === '/' || (res[i] === '\\' && isWindows))
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +741,47 @@ export class CppProperties {
...
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
-            // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
-            const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                fastGlob.sync(normalized, { onlyDirectories: true, cwd}) : [res];
-            globResult.push(...matches.map(s => recursive ? s + '**' : s));
+        if (paths) {
+            for (let res of paths) {
+                let counter: integer = 0;
```
bobbrow: I'm not familiar with this type. Can we just use `number`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +741,47 @@ export class CppProperties {
...
+                let replacementString: string = '';
+                const lastIndex: integer = res.length - 1;
+                // Detect and glob all wildcard variations by looking at last character in the path first.
+                for (let i: integer = lastIndex; i >= 0; i--) {
+                    if (res[i] === '*') {
+                        counter++;
+                        replacementString = replacementString.replace (/^/, res[i]);
```
bobbrow: Instead of building `replacementString` character by character, do it outside the for loop with a slice.
```ts
const suffix = res.slice(res.length - counter - 1);  // not tested, could be off by 1...
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +741,47 @@ export class CppProperties {
...
+                    } else if (res[i] === '/' || res[i] === '\\') {
+                        replacementString = replacementString.replace (/^/, res[i]);
+                        counter++;
+                        break;
+                    } else {
+                        break;
+                    }
+                }
+                res = res.slice(0, res.length - counter);
```
bobbrow, line 772: should it be `res.length - counter - 1`?

Extension/src/LanguageServer/configurations.ts
browntarik: This doesn't work because this will build the string backwards

Extension/src/LanguageServer/configurations.ts
browntarik: No, this works fine because we already index with an offset

Extension/src/LanguageServer/configurations.ts
bobbrow: what?  It should trim off the last `counter` characters.

Extension/src/LanguageServer/configurations.ts
bobbrow: where is the offset?  `counter` is initialized with `0`.

Extension/src/LanguageServer/configurations.ts
browntarik: Ohh I see

browntarik: Is there a benefit to building the string outside of the for loop? building it outside causes other bugs. I don't see an issue with the way it is now

Extension/src/LanguageServer/configurations.ts
browntarik: the counter already counts the appropriate number of characters, I can demo it to you

bobbrow: Sure. I was thinking the second parameter was an 'index', not a 'length' and an index is one less than the length to that index.

Extension/src/LanguageServer/configurations.ts
bobbrow: It should give you the same result.  Let's look at it together.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +740,47 @@ export class CppProperties {
...
-            paths.forEach(entry => {
-                const entries: string[] = util.resolveVariables(entry, env).split(util.envDelimiter).map(e => this.resolvePath(e, false)).filter(e => e);
-                result.push(...entries);
-            });
-        }
-        if (!glob) {
-            return result;
+            if (paths && !glob) {
+                paths.forEach(entry => {
+                    const entries: string[] = util.resolveVariables(entry, env).split(util.envDelimiter).map(e => this.resolvePath(e, false)).filter(e => e);
```
sean-mcmanus: util.resolveVariables isn't getting called if glob is true

sean-mcmanus (changes requested):

bobbrow (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +740,50 @@ export class CppProperties {
...
-                res = res.slice(0, res.length - 2);
-            }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
-            // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
-            const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                fastGlob.sync(normalized, { onlyDirectories: true, cwd}) : [res];
-            globResult.push(...matches.map(s => recursive ? s + '**' : s));
+        if (paths) {
```
michelleangela: The path checks at line 741 and 753 can be turned into one at the beginning of the function:

example:

```ts
// If path is not defined return result immediately.
const result: string[] = []
if (paths === undefined) {
    return result;
}
// Do work for when path has value.
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -739,29 +740,50 @@ export class CppProperties {
...
         if (paths) {
             paths = this.resolveDefaults(paths, defaultValue);
-            paths.forEach(entry => {
-                const entries: string[] = util.resolveVariables(entry, env).split(util.envDelimiter).map(e => this.resolvePath(e, false)).filter(e => e);
-                result.push(...entries);
-            });
-        }
-        if (!glob) {
-            return result;
+            if (paths && !glob) {
```
michelleangela: What's the check for paths on line 743 for? `this.resolveDefaults(paths, defaultValue)` always returns an array but length of array could be 0. Is it supposed to check if paths has items instead?

Extension/src/LanguageServer/configurations.ts
browntarik: that is what we want because as far as I can tell, resolve variables is just for resolving the path delimiters

Extension/src/LanguageServer/configurations.ts
browntarik: That was to make sure it wasn't undefined but that is fixed through your earlier suggestion.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Doesn't it resolve variables? i.e. I think it does

browntarik: Oh I see I totally forgot about that functionality. Thanks for pointing it out

Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,31 +738,51 @@ export class CppProperties {
...
             paths.forEach(entry => {
                 const entries: string[] = util.resolveVariables(entry, env).split(util.envDelimiter).map(e => this.resolvePath(e, false)).filter(e => e);
                 result.push(...entries);
             });
-        }
-        if (!glob) {
             return result;
         }
 
         const globResult: string[] = [];
```
michelleangela: `const result: string[] = [];` can be used for both cases and `const globResult: string[] = [];` can be removed. Since if `glob` is true, then `result` won't be used for the first case.

michelleangela (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,33 +738,58 @@ export class CppProperties {
...
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
             // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
             const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                fastGlob.sync(normalized, { onlyDirectories: true, cwd}) : [res];
-            globResult.push(...matches.map(s => recursive ? s + '**' : s));
+                fastGlob.sync(normalized, { onlyDirectories: true, cwd }) : [res];
+
+            result.push(...matches.map(s => s + suffix));
+            result.forEach(entry => {
```
sean-mcmanus: This doesn't appear correct. It's returning after 1 iteration of the for loop?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
bobbrow: I think because you're changing the value of `result` inside the callback.  You probably want to use `reduce` instead and push into the accumulation array (even though the name is counter-intuitive).

bobbrow: But don't we need to resolve variables before the glob?

browntarik: We need to resolve the variables before the glob but we cannot normalize the paths before the globbing. I though the resolving the variables and normalizing the paths went hand in hand but it looks like I can possibly separate the two.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,33 +738,57 @@ export class CppProperties {
...
+                let cwd: string = '';
+                if (isWindows) {
+                    normalized = res.replace(/\\/g, '/');
+                    cwd = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+                }
+                // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
+                const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
+                    fastGlob.sync(normalized, { onlyDirectories: true, cwd }) : [res];
+                if (isGlobPattern){
+                    result.push(...matches.map(s => s + suffix));
```
michelleangela: `result.push` is done on lines 748 and 787. Would this create duplicate entries in the returned list?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,33 +738,57 @@ export class CppProperties {
...
+                let normalized: string = '';
+                let cwd: string = '';
+                if (isWindows) {
+                    normalized = res.replace(/\\/g, '/');
+                    cwd = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+                }
+                // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
+                const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
+                    fastGlob.sync(normalized, { onlyDirectories: true, cwd }) : [res];
+                if (isGlobPattern){
```
michelleangela: Instead of `if (isGlobPattern)`, can it check `res.includes('*'))` directly instead of checking a var since `isGlobPattern` is only used here?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -737,33 +738,57 @@ export class CppProperties {
...
-        const globResult: string[] = [];
-        for (let res of result) {
-            // fastGlob will expand the ending double wildcard. temporary strip them before expanding
-            const recursive: boolean = res.endsWith('**');
-            if (recursive) {
-                res = res.slice(0, res.length - 2);
+        paths = this.resolveDefaults(paths, defaultValue);
+        paths.forEach(entry => {
+            const entries: string[] = util.resolveVariables(entry, env).split(util.envDelimiter).map(e => this.resolvePath(e, false)).filter(e => e);
+            if (!entry.includes('*')){
```
michelleangela: Space before `{` in `if (!entry.includes('*')){`. Same on lines 757, 786.

Extension/src/LanguageServer/configurations.ts
michelleangela: Sorry, I didn't see `res` is getting modified after the check `res.includes('*')`.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: This `if (!entry.includes('*')){` seems incorrect (e.g. if glob is false, then the entries with ** are removed). I think that check should be removed and it should iterate through `result` instead of `paths` below.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
browntarik: No because the first result.push will only push non-glob patterns and the second result.push will only push glob patterns

Extension/src/LanguageServer/configurations.ts
browntarik: This is not the case if you are referring to examples like "sources/**".  I would suggest testing it. Please let me know if you find a bug repro and I will fix it.

Extension/src/LanguageServer/configurations.ts
michelleangela: So the first case excludes non-glob patterns from the returned list if `glob` is false? And if `paths` is getting used for glob candidates, will it include resolved variables?

Is the main logic like this:

1. resolve variables for **all** paths in `paths` input. Place paths with resolved variables in some result container.

2. if `glob` is false, return result container immediately.

3. if `glob` is true, process items from `result container`. Then return a new list which now includes resolved globs that were initially included in `result container` but not processed.

browntarik: Just refactored

michelleangela (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+        const resolvedGlob: string[] = [];
+        for (let res of resolvedVariables) {
+            let counter: number = 0;
+            let slashFound: boolean = false;
+            const lastIndex: number = res.length - 1;
+            // Detect and glob all wildcard variations by looking at last character in the path first.
+            for (let i: number = lastIndex; i >= 0; i--) {
+                if (res[i] === '*') {
+                    counter++;
+                } else if (res[i] === '/' || (res[i] === '\\' && isWindows)) {
```
sean-mcmanus: Can you change this to `isWindows && res[i] === '\\'`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+                    counter++;
+                    slashFound = true;
+                    break;
+                } else {
+                    break;
+                }
+            }
+            let suffix: string = '';
+            if (slashFound) {
+                suffix = res.slice(res.length - counter);
```
sean-mcmanus, line 771: Did you test with `${workspaceFolder}/*/subFolder`? I believe this code isn't handling that correctly, because it appears to treat the /folder as a suffix when it needs to be passed to the fastGlob. Otherwise, I think there's a bug because without the suffix fastGlob could return results that don't have a subFolder under it, i.e. the subFolder part is required as part of the matching.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Oh, I'm wrong, that's handled by the else clause.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+        }
+        const resolvedGlob: string[] = [];
+        for (let res of resolvedVariables) {
+            let counter: number = 0;
+            let slashFound: boolean = false;
+            const lastIndex: number = res.length - 1;
+            // Detect and glob all wildcard variations by looking at last character in the path first.
+            for (let i: number = lastIndex; i >= 0; i--) {
+                if (res[i] === '*') {
+                    counter++;
```
Colengms, line 760: Is it valid to have more than 2 "*'?  What should happen if more than 2 are present?

Extension/src/LanguageServer/configurations.ts
browntarik: That scenario works, the subfolder is treated as a suffix but the path is globbed correctly.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            let normalized: string = '';
+            let cwd: string = '';
+            if (isWindows) {
+                normalized = res.replace(/\\/g, '/');
+                cwd = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            }
+            const isGlobPattern: boolean = res.includes('*');
```
Colengms: Won't `res` always have at least one `*` in it, if set to anything (and is `slashFound` is true)?  Could you combine `slashFound` and `isGlobPattern` into the same bool.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+                suffix = res.slice(res.length - counter);
+                res = res.slice(0, res.length - counter);
             }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            let normalized: string = '';
+            let cwd: string = '';
+            if (isWindows) {
+                normalized = res.replace(/\\/g, '/');
```
sean-mcmanus, line 777: normalized needs to be set on non-Windows.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+            let suffix: string = '';
+            if (slashFound) {
+                suffix = res.slice(res.length - counter);
+                res = res.slice(0, res.length - counter);
             }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            let normalized: string = '';
+            let cwd: string = '';
```
Colengms: Isn't this going to leave `cwd` a blank string?  Shouldn't it be set to `this.rootUri?.fsPath || ''`?

Similarly for `normalized`.  Shouldn't that be set to `res` when not on Windows?

Also, shouldn't all references to `res` below be referencing `normalized`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+            }
+            let suffix: string = '';
+            if (slashFound) {
+                suffix = res.slice(res.length - counter);
+                res = res.slice(0, res.length - counter);
             }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            let normalized: string = '';
```
michelleangela: Could you also test on Linux for the variables `normalized` and `cwd` cases? Or make sure all test cases also work on Linux?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+            let cwd: string = '';
+            if (isWindows) {
+                normalized = res.replace(/\\/g, '/');
+                cwd = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            }
+            const isGlobPattern: boolean = res.includes('*');
+            if (isGlobPattern) {
             // fastGlob silently strip non-found paths. limit that behavior to dynamic paths only
```
sean-mcmanus: Can you re-word this existing comment to something like `// fastGlob silently strips non-found paths. Limit that behavior to dynamic paths only.` And I think it needs to be tabbed over.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
     }
 
     private resolveAndSplit(paths: string[] | undefined, defaultValue: string[] | undefined, env: Environment, glob = false): string[] {
```
Colengms: Existing code, but could you add the type to this arg?  I'm not sure why lint isn't reporting an issue that it's missing a type.

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
+                res = res.slice(0, res.length - counter);
             }
-            // fastGlob can't deal with backslash-separated path => remove them
-            const normalized: string = res.replace(/\\/g, '/');
-            const cwd: string = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
+            let normalized: string = '';
+            let cwd: string = '';
+            if (isWindows) {
+                normalized = res.replace(/\\/g, '/');
+                cwd = this.rootUri?.fsPath?.replace(/\\/g, '/') || '';
```
sean-mcmanus: This `||` should be changed to `??` -- Garrett's follow-up PR is adding a linter rule for that (assuming you still need that after addressing the Linux issue).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Seems fine to me?

browntarik: I am assuming that would result in an invalid path, the code would parse it but glob would return an empty result.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -736,34 +737,57 @@ export class CppProperties {
...
-            if (recursive) {
-                res = res.slice(0, res.length - 2);
+            return resolvedVariables;
+        }
+        const resolvedGlob: string[] = [];
+        for (let res of resolvedVariables) {
+            let counter: number = 0;
+            let slashFound: boolean = false;
+            const lastIndex: number = res.length - 1;
+            // Detect and glob all wildcard variations by looking at last character in the path first.
```
sean-mcmanus: The reference to "and glob" is confusing since there's no actual glob. Can you just use "Detect all" or something equivalent?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: But this case is only for *** at the end, which isn't passed to fastGlob, which is fine.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: No, `${workspaceFolder}/` has slashFound true but no *.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I think Garrett's change may have disabled the typing requirement, but I think that was desired, i.e. like auto in C++. But adding the explicit type seems more consistent in this case.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -771,15 +771,15 @@ export class CppProperties {
                 suffix = res.slice(res.length - counter);
                 res = res.slice(0, res.length - counter);
             }
-            let normalized: string = '';
-            let cwd: string = '';
+            let normalized = res;
+            let cwd: string = this.rootUri?.fsPath || '';
```
Colengms: Should this also be changed to use `??`?  Also, it would be simpler to refer to `cwd` again later, since you know it's not `null`/`undefined`.

```
cwd = cwd.replace(/\\/g, '/');
```

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
browntarik: tested on WSL and everything works fine

sean-mcmanus (approved):

Colengms (approved):

PR #11175, opened 2023-07-10
Added support for Toybox to remote process picker

michalmaka:
Hello and thanks for maintaining this repo!

I've come along a problem where my remote target needed to have different arguments for `ps` command in order to retrieve remote processes list. It was using ToyBox's `ps` and I added option for specifying custom command for getting processes list.

Current approach allows either using `ps` with different arguments or even using other command, see examples:

```
      "pipeTransport": {
        ..
        "customRemoteProcessCommand": "ps -A"
      },
```

```
      "pipeTransport": {
        ..
        "customRemoteProcessCommand": "uname && /home/root/get_pid.sh"
      },
```

One needs to have in mind that the custom command needs to return output in the same format as currently used commands do

### DIFF
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -271,6 +271,7 @@
     "c_cpp.debuggers.pipeTransport.pipeArgs.description": "Command line arguments passed to the pipe program to configure the connection.",
     "c_cpp.debuggers.pipeTransport.pipeEnv.description": "Environment variables passed to the pipe program.",
     "c_cpp.debuggers.pipeTransport.quoteArgs.description": "If the pipeProgram's individual arguments contain characters (such as spaces or tabs), should it be quoted? If 'false', the debugger command will no longer be automatically quoted. Default is 'true'.",
+    "c_cpp.debuggers.pipeTransport.customRemoteProcessCommand.description": "If you have *nix system that doesn't support default command for getting processes list, you can override it here. For example, you can use `ps -A` or any other custom command or script that will be properly parsed by this extension.",
     "c_cpp.debuggers.logging.description": "Optional flags to determine what types of messages should be logged to the Debug Console.",
     "c_cpp.debuggers.logging.exceptions.description": "Optional flag to determine whether exception messages should be logged to the Debug Console. Defaults to true.",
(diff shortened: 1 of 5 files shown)


michalmaka:
microsoft-github-policy-service agree company="Software Mansion"

michalmaka:
@microsoft-github-policy-service agree company="Software Mansion"

WardenGnaw:
> One needs to have in mind that the custom command needs to return output in the same format as currently used commands do

This would lead to more issues for users tyring to get their command to match what `PsProcessParser` expects and make it difficult for the extension authors decide to add more columns to the process listing. If more columns were added or reorganized, this would also lead to breaking user's launch.jsons `customRemoteProcessCommand`.

We would reccomend writing a vscode input variable instead, as users can run their own vscode command and parse their own ouput to the value they want for the launch.json. 

https://code.visualstudio.com/docs/editor/variables-reference#_input-variables

E.g. launch.json
```json
{
  "configurations": [
    {
      "type": "cppdbg",
      "request": "attach",
      "name": "Run Program",
      "program": "${workspaceFolder}/a,out",
      "processId": "${input:pickProcess}"
    }
  ],
  "inputs": [
    {
      "id": "pickProcess",
      "type": "command",
      "command": "extension.my.cysto.testPicker"
    }
  ]
}
```

michalmaka:
what about forcing such custom command to return data as CSV? Having this implemented for custom command will avoid this issue

WardenGnaw:
> what about forcing such custom command to return data as CSV? Having this implemented for custom command will avoid this issue

That sounds good, I think there also need to be an associated field indicating the format of the output. 

E.g.
`$PID,$EXE,$CMDLINE`
and use a custom parser find those token and create the correct `Process` object.

Another question is that is there a reason why ToyBox's ps deviates from the standard usage? Is there a good way to detect that the `ps` being used is ToyBoy's version?

michalmaka:
@WardenGnaw I've changed implementation to just support Toybox detection and proper command for getting remote process list.

For some reason Toybox's `ps` works in a bit different way.

WardenGnaw (approved):

sean-mcmanus (approved):

Issue #6020, opened 2020-08-25
Markdown comment support for tooltip display of hover


DeadRabbits307:
**Type: LanguageService**

Like the support for doxygen documentation (#658) the support for Markdown would be nice. Like in Rust:

![2020-08-25-091128_1278x674_scrot](https://user-images.githubusercontent.com/33861896/91145912-78012a00-e6b6-11ea-9e98-3dd8d04e89c6.png)

![2020-08-25-093703_1278x1380_scrot](https://user-images.githubusercontent.com/33861896/91146105-b860a800-e6b6-11ea-9958-2a9c87a8b52b.png)

sean-mcmanus:
VS Code inherently supports markdown in doc comments, but we added code to explicitly disable/escape it for code that has comments that are not intended to be interpreted as markdown. We'd need to add a setting to re-enable the markdown.

Leedehai:
Following @sean-mcmanus's response above.. Where can we re-enable that in code? I was trying to tweak files in ~/.vscode/extensions/ms-vscode.cpptools-1.0.1 so I can render the comment in Markdown without having to wait for a new Cpptools release :) It looks we just need to delete the code that is used to "explicitly disable/escape" Markdown.

sean-mcmanus:
The markdown escaping is done in our closed source cpptools process. I haven't looked into whether there's a way to undo the escaping via the open source TypeScript.

ericdrobinson:
> We'd need to add a setting to re-enable the markdown.

I was surprised that there isn't a way to get markdown formatting already. We use this feature in other languages to great effect. Would very much love to see it enabled in C/C++.

bshoshany:
Hi @sean-mcmanus, I came here to request this exact same feature and found that it was already requested. It's been more than a year since this issue opened, I wonder if any progress has been made into enabling Markdown comments in C/C++? This would be extremely helpful in writing more meaningful and informative code documentation for my C++ libraries. Thanks!

sean-mcmanus:
@bshoshany There hasn't been any progress yet. If you see this issue get assigned to a milestone then you could expect progress.

bshoshany:
[...]

PR #11177, opened 2023-07-10
Add Setting to Enable Markdown Comment Formatting 

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/6020


Extension/package.json
```diff
@@ -457,6 +457,17 @@
             ],
             "scope": "resource"
           },
+          "C_Cpp.markdownCommentFormatting": {
```
sean-mcmanus: Formatting in the settings currently refers to actual formatting of code, so it seems like it be less confusing this were named something like "markdownCommentRendering" or "displayMarkdownComments" or "renderMarkdownComments", "markdownInComments", "preserveMarkdownInComments", etc.

sean-mcmanus (changes requested):

Extension/package.nls.json
```diff
@@ -242,7 +242,8 @@
...
     "c_cpp.configuration.enhancedColorization.markdownDescription": { "message": "If enabled, code is colorized based on IntelliSense. This setting only applies if `#C_Cpp.intelliSenseEngine#` is set to `default`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.configuration.codeFolding.description": "If enabled, code folding ranges are provided by the language server.",
-    "c_cpp.configuration.hover.description": "If disabled, hover details are no longer provided by the language server.",
+    "c_cpp.configuration.markdownCommentFormatting.description": "If enabled, markdown comment formatting is provided by the language server.",
+    "c_cpp.configuration.hover.description": "Select whether markdown features will be available in comment hover formatting. Default indicates only a subset of markdown features are available.",
```
sean-mcmanus: The phrase "comment hover formatting" doesn't really make sense to me, since there's no actual "formatting" being done. It's more like "comment hover rendering". Unless maybe "formatting" has a special meaning in the domain of Markdown that I'm not aware of...even if so, I'm not sure if our C/C++ users will understand what the usage of "formatting" is supposed to mean.

sean-mcmanus: It seems like "Default indicates" should be something else like "By default,". The phrase "Default indicates" sounds like it's referring to a `Default` setting value.

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -457,6 +457,17 @@
             ],
             "scope": "resource"
           },
+          "C_Cpp.commentFormatting": {
```
bobbrow: Without "markdown" in the setting name, I don't think there's enough context for "subsetEnabled" to be meaningful.

I was going to suggest "markdownInComments" but it looks like @sean-mcmanus already suggested that name, so +1 for that suggestion.

sean-mcmanus: Yeah, as stated previously, I don't think formatting should be in the setting name, because then when users search for formatting settings this would appear in the results, and it's not a formatting setting. "markdownInComments" seems fine to me.

Extension/package.nls.json
```diff
@@ -242,7 +242,10 @@
...
     "c_cpp.configuration.enhancedColorization.markdownDescription": { "message": "If enabled, code is colorized based on IntelliSense. This setting only applies if `#C_Cpp.intelliSenseEngine#` is set to `default`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.configuration.codeFolding.description": "If enabled, code folding ranges are provided by the language server.",
-    "c_cpp.configuration.commentFormatting.description": "Select whether markdown features will be available in comment formatting. By default, only a subset of markdown features are available.",
+    "c_cpp.configuration.markdownInComments.description": "Select whether markdown features will be available in the hover window. By default, only a subset of markdown formatting will be applied to comments in the hover window.",
```
bobbrow: Should we prefer the word "tooltip" over "window" for the hover information?

sean-mcmanus: Yes (or another preferred term if you can find one).

Extension/package.nls.json
```diff
@@ -242,6 +242,10 @@
...
     "c_cpp.configuration.enhancedColorization.markdownDescription": { "message": "If enabled, code is colorized based on IntelliSense. This setting only applies if `#C_Cpp.intelliSenseEngine#` is set to `default`.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "c_cpp.configuration.codeFolding.description": "If enabled, code folding ranges are provided by the language server.",
+    "c_cpp.configuration.markdownInComments.description": "Select whether markdown features will be available in the hover tooltip. By default, only a subset of markdown formatting will be applied to comments in the hover tooltip.",
```
sean-mcmanus: "markdown formatting"->"markdown".

sean-mcmanus (changes requested):

Extension/package.nls.json
```diff
@@ -242,6 +242,10 @@
...
+    "c_cpp.configuration.markdownInComments.subsetEnabled.description": "Enable all markdown features in the hover tooltip except those that include the '_' and '*' characters.",
+    "c_cpp.configuration.markdownInComments.enabled.description": "Enable all markdown features in the hover tooltip",
```
sean-mcmanus: Missing period.

Extension/package.nls.json
sean-mcmanus: or "markdown features"?

sean-mcmanus (approved):

Colengms (approved):

michelleangela (approved):

PR #11191, opened 2023-07-14
Remove walkthrough experiment and update the Help links to point to it

bobbrow:
This reverts #10812, fixes a string, and enables the walkthrough for all users.

(note: I'm running into some issues with getting the commands to work as expected, so opening the PR in draft mode for now)


Colengms (dismissed):

bobbrow:
Works around https://github.com/microsoft/vscode/issues/187958 now.

Extension/src/LanguageServer/client.ts
```diff
@@ -1048,17 +1049,39 @@ export class DefaultClient implements Client {
...
-                        return;
+                        step = step + "mac";
+                        break;
                     default: // Linux
-                        void vscode.commands.executeCommand('vscode.open', "https://go.microsoft.com/fwlink/?linkid=2217615");
-                        return;
+                        step = step + "linux";
+                        break;
                 }
+                void vscode.commands.executeCommand(
```
sean-mcmanus, line 1074: It seems like this should be `return` instead of `void`, right?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: Should it?  Seems like we don't need to pause our own execution waiting for VS Code to process this command.

sean-mcmanus: I don't see how `return` pauses our execution. Wouldn't that be `await`?

sean-mcmanus: Oh, you mean if VS Code itself awaited the result? Okay, sure, `void` makes sense then.

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Or should we use `return` in case the caller actually wants to wait for the command to execute?

bobbrow: What I think is confusing is that some paths in this function return the promise from the last function call and others don't. Prior to #11085, none of these awaitable functions were awaited nor were their promises returned.  I reverted a change that was made before #11085.

EDIT: It doesn't look like the pattern established was to return the promises from `vscode.commands.executeCommand` in this function.

Colengms (approved):

Issue #11119, opened 2023-06-22
TS Modernization - Code Cleanup 


fearthecowboy:
##  More code cleanup tasks

In the effort to reduce some complexity and technical debt in the extension codebase - the following are the outstanding things to address after the TS Modernization PR is completed:  #11085 

- clean up `clients.ActiveClient` vs `getActiveClient()`  -- pick one and stick with it. 
- clean up `ui` vs `getUI()` - consistency - same thing
- elminiate use of `null`
- clean up use of os.platform() everywhere with the use of a constants
- unify use of guards and replace util.is* with guards
- replace filesystem wrappers (those that provide promise-based fs functions) with use of built-in node:fs.promises.XXX functions
- examine all use of new Promise(...) to ensure that there isn't a cleaner use available.
- enable `plugin:@typescript-eslint/strict` in eslint and fix all errors (about ~450), including:
    - `A record is preferred over an index signature `                                                             
    - `Do not delete dynamically computed property keys `                                                          
    - `Expected a 'for-of' loop instead of a 'for' loop with this simple iteration `                               
    - `Expected an error object to be thrown `                                                                     
    - `Explicit enum value must only be a literal value (string, number, boolean, etc)`                            
    - `Literals should be exposed using readonly fields `                                                          
    - `Prefer using an optional chain expression instead, as it is more concise and easier to read `                
    - `Prefer using nullish coalescing operator ('??') instead of a logical or ('||'), as it is a safer operator ` 
    - `The generic type arguments should be specified as part of the constructor type arguments `                  
    - `The two values in this comparison do not have a shared enum type `                                          
    - `Unexpected class with only static properties `                                                              
    - `Unnecessary conditional, both sides of the expression are literal values `                                  
    - `Unnecessary conditional, expected left-hand side of '??' operator to be possibly null or undefined `        
    - `Unnecessary conditional, the types have no overlap `                                                        
    - `Unnecessary conditional, value is always falsy `                                                            
    - `Unnecessary conditional, value is always truthy `                                                           
    - `Unnecessary optional chain on a non-nullish value `                                                         
    - `Use 'as XXX' instead of '<XXX>' ` 
    - `Use 'includes()' method instead ` 
    - `Use 'String#startsWith' method instead `                                                                    
    - `Use an 'interface' instead of a 'type'`
    - `Use the 'String#endsWith' method instead `                                                                  
    - `Useless constructor `                                                                                       
    - `void is not valid as a constituent in a union type `                                                        
[...]

PR #11212, opened 2023-07-18
Fix out-of-order message handling

fearthecowboy:
This addresses a bug caused by the modernization PR #11119 where the ordering of message handling wasn't being preserved.

This splits the `.ready` and `queueBlockingTask()` into separate distinct concepts.

The `.ready` now only blocks the initial initialization of the client, and during the execution of the `provideCustomConfiguration` method.

The `queueBlockingTask()` is replaced by a `queue()` method is implemented with a proper queue. The calls to `queue` return a `Promise` for the execution of the provided `task` so that the actual execution of the task can be awaited after it has been run when the queue is processed to that point.

This removes the `BlockingTask` class in lieu of the queue here, which also allows the developer to see how much is in the queue while debugging.  (you can also await the `.dispatching` member to wait for the queue to finish processing -- This may not have practical benefits, but can be handy during debugging)

I also updated the code that runs the integrationTests from the command line, so that it installs `.vscode-test` into the tmp folder instead of the Extension folder as VSCode can get persnickety and lock the folder when the editor is open to the Extension folder. It also adds better isolation from the user's instance of vscode (by regulating the user-dir and extension-dir)


Extension/src/LanguageServer/extension.ts
```diff
@@ -382,8 +382,7 @@ function onDidChangeVisibleTextEditors(editors: readonly vscode.TextEditor[]): v
     editors.forEach(async (editor) => {
         if (util.isCpp(editor.document)) {
             const client: Client = clients.getClientFor(editor.document.uri);
-            await client.ready;
-            await processDelayedDidOpen(editor.document);
+            await client.queue(()=> processDelayedDidOpen(editor.document));
```
sean-mcmanus: There's no space before => but I see a space before the => in other spots.

Extension/package.json
```diff
@@ -1,6094 +1,6095 @@
...
-          "autoDetect",
-          "${cwd}"
+    "main": "./dist/main",
+    "contributes": {
+        "walkthroughs": [
+            {
+                "id": "cppWelcome",
+                "title": "%c_cpp.walkthrough.title%",
+                "description": "%c_cpp.walkthrough.description%",
+                "when": "false",
```
sean-mcmanus: This (the "when") should be removed.

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -1,6094 +1,6095 @@
...
-                  "%c_cpp.debuggers.cppvsdbg.console.newExternalWindow.description%"
-                ],
-                "description": "%c_cpp.debuggers.cppvsdbg.console.description%",
-                "default": "internalConsole"
-              },
-              "sourceFileMap": {
-                "type": "object",
-                "description": "%c_cpp.debuggers.sourceFileMap.description%",
-                "default": {
-                  "<sour
```
bobbrow, line 6052: If we're going to change the indentation size for this file, please also update `tools/GenerateOptionsSchema.ts` so that it writes JSON back with 4 spaces instead of 2.

Extension/src/LanguageServer/client.ts
```diff
@@ -32,6 +32,8 @@ import { LanguageClient, ServerOptions } from 'vscode-languageclient/node';
 import * as nls from 'vscode-nls';
 import { DebugConfigurationProvider } from '../Debugger/configurationProvider';
 import { CustomConfigurationProvider1, getCustomConfigProviders, isSameProviderExtensionId } from '../LanguageServer/customProviders';
+import { ManualPromise } from '../Utility/Async/manual-promise';
```
bobbrow: Oh, did we check in files with hyphens in their names?  Our convention has been "camelCase", not hyphens.

fearthecowboy: I can rename those.

Extension/package.json
fearthecowboy: Aha.. that's why its getting reset. 

willdo

Extension/src/LanguageServer/client.ts
```diff
@@ -843,6 +846,20 @@ export class DefaultClient implements Client {
     private configStateReceived: ConfigStateReceived = { compilers: false, compileCommands: false, configProviders: undefined, timeout: false };
     private showConfigureIntelliSenseButton: boolean = false;
 
+    /** A queue of asynchronous tasks that need to be processed befofe ready is considered active.  */
+    private tasks = new Array<[ManualPromise<unknown>|undefined,() => Promise<unknown>]>();
```
bobbrow: It might be nice to define an interface for this instead of using an array with unnamed fields. `[ManualPromise<unknown>|undefined,() => Promise<unknown>]`

Extension/src/LanguageServer/extension.ts
fearthecowboy: O_o 

I wonder why the formatter fixes this sometimes and not others.  I will see if I can find out why and make it conform

Extension/src/LanguageServer/client.ts
fearthecowboy: That's a [Tuple](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types), it is literally the typescript prescribed way to do this.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: The existing code style has space before => (I think that is preferred).

Extension/src/LanguageServer/client.ts
bobbrow: I would disagree that TypeScript prescribes using tuples in this scenario (I didn't see it in your link).  But this is a personal preference suggestion.  I've had to read through enough unfamiliar code that used tuples to form an opinion against their use almost universally.  The lack of self-documenting fields is the biggest detractor for me as I review a lot of code and don't always have IntelliSense features available when I do that.  You could just as easily use an unnamed interface here and provide names for the fields so that you would know exactly what this data structure holds without having to search how it's constructed or used.

`{taskResult: ManualPromise<unknown> | undefined; task: () => Promise<unknown>}`

Extension/tools/GenerateOptionsSchema.ts
```diff
@@ -64,7 +66,7 @@ function refReplace(definitions: any, ref: any): any {
 
     // Retrieve ObjectName from definitions. (TODO: Does not retrieve inner objects)
     // Need to deep copy, there are no functions in these objects.
-    const reference: any = JSON.parse(JSON.stringify(definitions[referenceName]));
+    const reference: any = JSON.parse(JSON.stringify(definitions[referenceName],null,4));
```
sean-mcmanus: We always (?) use spaces after `,` if there's something after it.

bobbrow: This one doesn't need an update though because the output is immediately parsed again rather than written to disk.  Only the update on line 137 is needed.

Extension/src/LanguageServer/extension.ts
```diff
@@ -368,6 +368,7 @@ export async function processDelayedDidOpen(document: vscode.TextDocument): Prom
                 }
                 await client.provideCustomConfiguration(document.uri, undefined);
                 // client.takeOwnership() will call client.TrackedDocuments.add() again, but that's ok. It's a Set.
+                client.onDidOpenTextDocument(document);
```
fearthecowboy, line 371: Restoring call that got dropped before the original modernization PR. 
Thanks to Michelle and Colen for helping me track that down.

Extension/src/LanguageServer/client.ts
```diff
@@ -843,6 +846,20 @@ export class DefaultClient implements Client {
     private configStateReceived: ConfigStateReceived = { compilers: false, compileCommands: false, configProviders: undefined, timeout: false };
     private showConfigureIntelliSenseButton: boolean = false;
 
+    /** A queue of asynchronous tasks that need to be processed befofe ready is considered active.  */
+    private queue = new Array<[ManualPromise<unknown>,() => Promise<unknown>]|[ManualPromise<unknown>]>();
+
+    /** returns a promise that waits initialization and/or a change to configuration to complete  (i.e. language client is ready-to-use) */
```
sean-mcmanus: 2 spaces

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -1980,123 +1999,128 @@ export class DefaultClient implements Client {
...
-                    if (await vscode.window.showInformationMessage(message, dismiss, disable) === disable) {
-                        settings.toggleSetting("configurationWarnings", "enabled", "disabled");
-                    }
+                if (await vscode.window.showInformationMessage(message, dismiss, disable) === disable) {
+                    settings.toggleSetting("configurationWarnings", "enabled", "disabled");
                 }
             }
-        });
+        } finally {
+            DefaultClient.isStarted.resolve();
```
bobbrow, line 2122: There is an early return before this try/finally on line 2021.  What happens if this `resolve` doesn't happen?

Extension/src/LanguageServer/client.ts
```diff
@@ -2180,24 +2203,82 @@ export class DefaultClient implements Client {
...
+     *
+     * Use `await <client>.ready` when you need to ensure that the client is initialized, and to run in order
+     * Use `enqueue()` when you want to ensure that subsequent calls are blocked until a critical bit of code is run.
+     *
+     * This is lightweight, because if the queue is empty, then the only thing to wait for is the client itself to be initialized
      */
-    get ready(): Promise<void> {
-        return this.isSupported ? Promise.resolve(pendingTask?.getPromise() || undefined) : fail(localize("unsupported.client", "Unsupported client"));
+    get ready():  Promise<unknown> {
+        if(!DefaultClient.dispatching.isCompleted || DefaultClient.queue.length) {
```
bobbrow: I'm surprised vscode isn't auto-formatting these `if(`'s for you.  There should be a space between keywords and parentheses.

Extension/src/LanguageServer/client.ts
```diff
@@ -1980,123 +1999,128 @@ export class DefaultClient implements Client {
             return;
         }
         telemetry.logLanguageServerEvent('provideCustomConfiguration', { providerId });
+        void this.provideCustomConfigurationAsync(docUri, requestFile, replaceExisting, onFinished, provider);
+    }
 
-        return this.queueBlockingTask(async () => {
-            const tokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
-            console.log("provideCustomConfiguration");
+    private async provideCustomConfigurationAsync(docUri: vscode.Uri,requestFile: string|undefined,  replaceExisting: boolean|undefined, onFinished: () => void, provider: CustomConfigurationProvider1) {
```
Colengms: Nit: We usually have spaces around the `|`

Extension/src/LanguageServer/client.ts
```diff
@@ -2180,24 +2203,82 @@ export class DefaultClient implements Client {
...
+     * This is a lighter-weight complement to `enqueue()`
+     *
+     * Use `await <client>.ready` when you need to ensure that the client is initialized, and to run in order
+     * Use `enqueue()` when you want to ensure that subsequent calls are blocked until a critical bit of code is run.
+     *
+     * This is lightweight, because if the queue is empty, then the only thing to wait for is the client itself to be initialized
      */
-    get ready(): Promise<void> {
-        return this.isSupported ? Promise.resolve(pendingTask?.getPromise() || undefined) : fail(localize("unsupported.client", "Unsupported client"));
+    get ready():  Promise<unknown> {
```
Colengms: Does this need to be a `Promise<unknown>` or can it still be a `Promise<void>`?

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: My preference would be to either run the formatter or keeping the formatting manually consistent, i.e. spaces around `|` and `=>`, etc.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I'd prefer `if (`.

Extension/src/LanguageServer/client.ts
```diff
@@ -2180,24 +2203,82 @@ export class DefaultClient implements Client {
...
+        return DefaultClient.isStarted;
     }
 
     /**
-     * Queue a task that blocks all future tasks until it completes. This is currently only intended to be used
-     * during language client startup and for custom configuration providers.
-     * @param task The task that blocks all future tasks
+     * Enqueue a task to ensure that the order is maintained. The tasks are executed sequentially after the client is ready.
+     *
+     * this is a bit more expensive than `.ready` - this ensures the task is absolutely finished executing before allowing
```
sean-mcmanus, line 2232: I'd prefer sentences in comments to start with a capital letter. There were lots of other new comments added in this PR that are sentences that don't start with a capital letter.

Extension/src/LanguageServer/client.ts
```diff
@@ -2180,24 +2203,82 @@ export class DefaultClient implements Client {
...
+    private static async dispatch() {
+        // ensure that this is OK to start working
+        await this.isStarted;
+
+        // reset the promise for the dispatcher
+        DefaultClient.dispatching.reset();
+
+        do {
+            // pick items up off the queue and run then one at a time until the queue is empty
+            const [promise,task] = DefaultClient.queue.shift() || [];
```
sean-mcmanus: Should this be `?? []`?

Extension/src/LanguageServer/client.ts
sean-mcmanus: I also see `while(` below.

Extension/src/LanguageServer/client.ts
fearthecowboy: OOh. good catch.

fearthecowboy: fixed

Extension/src/LanguageServer/client.ts
fearthecowboy: uh, it looks like it can be `<void>` again.

Extension/src/LanguageServer/client.ts
```diff
@@ -1303,91 +1319,8 @@ export class DefaultClient implements Client {
...
-                    this.isSupported = false;   // Running on an OS we don't support yet.
-                    if (!failureMessageShown) {
-                        failureMessageShown = true;
-                        void vscode.window.showErrorMessage(localize("unable.to.start", "Unable to start the C/C++ language server. IntelliSense features will be disabled. Error: {0}", String(err)));
-                    }
-                }
-            });
         } catch (errJS) {
             const err: NodeJS.ErrnoException = errJS as NodeJS.ErrnoException;
             this.isSupported = false;   // Running on an OS we don't support yet.
```
sean-mcmanus, line 1326: There's extra white space before the comment here. I thought we previously had a linter rule for that.

Extension/test/integrationTests/IntelliSenseFeatures/runTest.ts
```diff
@@ -1,35 +1,75 @@
...
+import { existsSync } from 'fs';
+import { mkdir as md, stat } from 'fs/promises';
+import { tmpdir } from 'os';
+import { resolve } from 'path';
+
+// The folder containing the Extension Manifest package.json
+// Passed to `--extensionDevelopmentPath`
+const extensionDevelopmentPath = resolve(__dirname, '../../../../');
+
+const isolated = resolve(tmpdir(), '.vscode-test', createHash('sha256').update(extensionDevelopmentPath).digest('hex').substring(0,6) );
```
sean-mcmanus, line 13: There's an extra space here.

Extension/src/LanguageServer/client.ts
fearthecowboy: I didn't actually change that line, it appears that the codebase has quite a few cases where that is the case.
There's no rule enabled for that at this point. 

image

sean-mcmanus (approved):

Colengms (approved):

michelleangela (approved):

fearthecowboy:
FYI I manually verified that it passes the E2E test in CI via ADO.

Extension/src/LanguageServer/client.ts
```diff
@@ -1980,123 +1999,129 @@ export class DefaultClient implements Client {
...
-                }
-                if (!uris.length) {
-                    return [];
-                }
-                let configs: util.Mutable<SourceFileConfigurationItem>[] = [];
+        const response: QueryTranslationUnitSourceResult = await this.languageClient.sendRequest(QueryTranslationUnitSourceRequest, params);
+        if (!response.candidates || response.candidates.length === 0) {
+        // If we didn't receive any candidates, no configuration is needed.
+            onFinished();
+            DefaultClient.isStarted.resolve();
```
bobbrow, line 2021: Future-proofing: it might be safer to wrap the whole function body in try/finally to ensure the resolve always happens (in case any other early returns appear in here someday.

PR #11237, opened 2023-07-24
Fix config provider ordering.

sean-mcmanus:
Fix config provider ordering, as suggested by @Colengms

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1856,9 +1856,11 @@ export class DefaultClient implements Client {
         if (!currentProvider.isReady) {
             return;
         }
+
         await Promise.all([
             this.clearCustomConfigurations(),
-            this.handleRemoveAllCodeAnalysisProblems(),
+            this.handleRemoveAllCodeAnalysisProblems()]);
+        await Promise.all([
             ...[...this.trackedDocuments].map(document => this.provideCustomConfiguration(document.uri, undefined, true))
         ]);
     }


Extension/src/LanguageServer/client.ts
```diff
@@ -1856,9 +1856,11 @@ export class DefaultClient implements Client {
         if (!currentProvider.isReady) {
             return;
         }
+
         await Promise.all([
```
fearthecowboy, line 1860: Does `clearCustomConfigurations` need to complete before `handleRemoveAllCodeAnalysisProblems` ?

if not, then I'm ok with this.

if it does, just await them individually

fearthecowboy (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -1856,9 +1856,11 @@ export class DefaultClient implements Client {
...
             this.clearCustomConfigurations(),
-            this.handleRemoveAllCodeAnalysisProblems(),
+            this.handleRemoveAllCodeAnalysisProblems()]);
```
sean-mcmanus, line 1862: Is the formatting okay? The formatter doesn't make any changes.

Colengms (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: The order doesn't matter between those two calls.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Formatting seems okay to me...I don't see any existing code using the previous formatting.

PR #11238, opened 2023-07-24
Fix TypeScript defaultFormatter setting.

sean-mcmanus:
Using dbaeumer.vscode-eslint causes VS Code to give an error popup when formatting is done. Using vscode.typescript-language-features still enables eslint to give formatter linter errors...although there's the protential mismatch issue in which the formatter will create code he linter complains about.

@fearthecowboy If you have a better fix I can abandon this PR.

### DIFF
--- a/Extension/.vscode/settings.json
+++ b/Extension/.vscode/settings.json
@@ -37,7 +37,7 @@
     },
     "[typescript]": {
         "editor.tabSize": 4,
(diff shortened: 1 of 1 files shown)


Colengms (approved):

sean-mcmanus:
Looks like the failure is just Format Selection -- so just don't do that and rely on format on save.

Issue #11157, opened 2023-07-04
Create definition via `Quick Fix...` hover - not working


H-G-Hristov:
Type: <b>Bug</b>

# Scenario 1

1. Create a class
2. Click on the `...`
3. Select `QuickFix: Create definition in <filename>.cpp`

- Nothing happens.

# Scenario 2

1. Create a class
2. COMMAND CLICK ON THE METHOD NAME
2. Click on the `...`
3. Select `QuickFix: Create definition in <filename>.cpp`

- Works.

Also if classes contain nested types these types are not prepended with the class name, e.g.

```c++

namespace N {
class C {
public:
  enum class E;
  struct S;

  std::expected<S, E> GetStuff(const S& s);
};
}
```
Produces: `std::expected<S, E> C::GetStuff(const S& s);`, which is not optimal.

This probably was reported before but I could't find it.

Extension version: 1.16.3
VS Code version: Code 1.79.2 (Universal) (695af097c7bd098fbf017ce3ac85e09bbc5dda06, 2023-06-14T08:58:52.392Z)
OS version: Darwin arm64 22.5.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 Pro (8 x 24)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|2, 3, 2|
|Memory (System)|16.00GB (0.78GB free)|
|Process Argv|--crash-reporter-id a7eaff19-94ae-4653-b917-d9aaab19da5d|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492cf:30256860
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
[...]

PR #11239, opened 2023-07-24
'Create Declaration / Definition' command via 'Quick Fix...' hover 

michelleangela:
Issue https://github.com/microsoft/vscode-cpptools/issues/11157

Note: this has a corresponding language server change.

**Problem:**
When 'Create Declaration / Definition' code action is invoked via `Quick Fix...` when hovering over a function that has `...` under it and the current selection or cursor position is in a different location from the function, no declaration or definition gets created. This is because the range which is used for processing the request comes from the current active selection or cursor position. If the range does not match the actual location of the function then the data for the request is invalid.

**Fix:**
During creating of code actions, save the range for the suggested 'Create Declaration / Definition' to use later as input for the range data.

**Test to verify fix:**

1. Select a line/character in the file other than the function to invoke "create definition..." on.
2. Hover over a function to "create definition..." on.
3. Click on "Quick Fix..." from the tool tip that pops up.
4. Select "Create definition..." from the code action (aka Quick Fix...) selections that pop up.
5. Result: definition is created.

**Regression test cases that invoke 'Create Declaration / Definition' from different entry points:**

- via `Quick Fix..` hovering over a function with `...` under it and the cursor or selection is in different location from function's location. (This is the case that was fixed in this changeset.)
- via `Light Bulb` after selecting a function with `...` under it and then clicking on the light bulb.
- via context menu by right-clicking on function and selecting `Create Declaration / Definition`. 
- via command palette by selecting a function and running command `Create Declaration / Definition`.

### DIFF
--- a/Extension/src/LanguageServer/Providers/codeActionProvider.ts
+++ b/Extension/src/LanguageServer/Providers/codeActionProvider.ts
@@ -187,7 +187,7 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
                 resultCodeActions.push(...docCodeActions);
                 return;
             } else if ((command.command === 'C_Cpp.CreateDeclarationOrDefinition' || command.command === 'C_Cpp.CopyDeclarationOrDefinition')
-                && (command.arguments ?? []).length === 0 && command.range != undefined) {
+                && (command.arguments ?? []).length === 0 && command.range !== undefined) {
                 const args: CreateDeclDefnCommandArguments = {
                     sender: 'codeAction',
                     range: command.range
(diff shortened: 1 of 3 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -23,12 +23,18 @@ interface CodeActionCommand {
...
     edit?: TextEdit;
     uri?: string;
+    range?: Range;
 }
 
 interface GetCodeActionsResult {
     commands: CodeActionCommand[];
 }
 
+export interface CreateDeclDefnCommandArguments {
```
Colengms, line 33: Do you need a new object with two fields, and a single instance of that in the list of arguments?  Couldn't you just add both of these as arguments?  That would seem to be a little bit simpler, overall.

Colengms (changes requested):

Extension/src/LanguageServer/Providers/codeActionProvider.ts
michelleangela: Will address during refactor of code actions.

Issue #10191, opened 2022-11-26
Use XDG_CACHE_HOME on Linux systems for the default databaseFilename


pwaller:
### Environment

- OS and Version: NixOS 22.11
- VS Code Version: 1.73.1
- C/C++ Extension Version: v1.11.0
- Other extensions you installed (and if the issue persists after disabling them): N/A
- If using SSH remote, specify OS of remote machine:  N/A
- Numerous workspaces

### Bug Summary and Steps to Reproduce

I hack on a number of large C/C++ projects including the linux kernel and llvm-project.

What I see is that my `$HOME/.config` directory contains many large binaries (tens of gigabytes, individual files >2.5GiB each), where I would rather it did not. These are the `.config/Code/User/workspaceStorage/<randomhex>/ms-vscode.cpptools/.browse.VC.db` files.

I consider `XDG_CONFIG_HOME` is meant to be the place I can store configuration I intend version control and to backup, but I can't easily do this if it contains a lot of large binaries.

I expected these large binaries to live in `XDG_CACHE_HOME`, which is where I go to delete things if I'm running out of disk space, the place where things should live which can be recomputed.

As an additional note, I see multiple `<randomhex>` directories where the `workspace.json` indicates that these several caches of the large caches belong to the same workspace-on-disk, so there is some unexpected duplication going on as well. The `workspace.json` is identical. I presume these are outdated and unused, and I would naturally clear them out by blatting my `$XDG_CACHE_HOME` directory, not something I intend to do with my `$XDG_CONFIG_HOME`.

### Expected behavior

I think that large recomputable binary databases should be stored in `$XDG_CACHE_DIR/vscode-cpptools` alongside the existing ipch cache.

### Code sample and Logs

```shell
N/A, but these repos are examples of large ones:

https://github.com/llvm/llvm-project
https://github.com/torvalds/linux
```

sean-mcmanus:
Yes, the normal behavior, which I repro, is for the database files to be written to the .cache location -- see the previously fixed bug at https://github.com/microsoft/vscode-cpptools/issues/3979 . i.e. something unexpected is happening for you -- do you have any idea why we may not be using your XDG_CACHE_DIR? We could also review that for possible causes.

pwaller:
I have not configured `C_Cpp.default.browse.databaseFilename`.

[...]

PR #11246, opened 2023-07-25
remove all code related to storagePath

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10191

The plumbing for storage path will be removed from the typescript and moved to the native side. This will allow us to not be limited by the storage path generated by VS Code.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -539,7 +539,6 @@ export interface TextDocumentWillSaveParams {
 interface InitializationOptions {
     packageVersion: string;
     extensionPath: string;
-    storagePath: string;
     freeMemory: number;
     vcpkgRoot: string;
     intelliSenseCacheDisabled: boolean;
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -925,7 +923,6 @@ export class DefaultClient implements Client {
     public get AdditionalEnvironment(): { [key: string]: string | string[] } {
         return {
             workspaceFolderBasename: this.Name,
-            workspaceStorage: this.storagePath,
```
sean-mcmanus, line 928: Oh, does this regress user settings with ${workspaceStorage}? Or what uses this?

michelleangela:
Could you add to PR description how removing storagePath addresses https://github.com/microsoft/vscode-cpptools/issues/10191?

Extension/src/LanguageServer/client.ts
browntarik: I think this would regress those settings, @bobbrow was there a certain way we should handle this scenario?

sean-mcmanus: If the TypeScript side needs the workspaceStorage and we want to computer it in cpptools, then this method could add a method to request it from cpptools.

bobbrow: I synced with @browntarik this morning. I don't think we'll be moving forward with this PR.

Issue #10808, opened 2023-04-11
Selecting a compiler from "Select IntelliSense Configuration" doesn't update the compilerPath in c_cpp_properties.json


sean-mcmanus:
It probably should. It currently will just update the C_Cpp.default.compilerPath, but the existence of compilerPath in c_cpp_properties.json causes the C_Cpp.default.compilerPath to not be used.

michelleangela:
Pending to discuss the expected behavior of running "Select IntelliSense Configuration" when `compilerPath` in c_cpp_properties.json exists.

michelleangela:
Discussed with @AlexandraKemperMS .

The command "Select IntelliSense Configuration" should change both `compilerPath` (if it exists) in c_cpp_properties.json and the `C_Cpp.default.compilerPath`.

sean-mcmanus:
@michelleangela That sounds fine to me.

michelleangela:
Fix available in pre-release https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.1.

PR #11255, opened 2023-07-27
"Select IntelliSense Configuration" updates `compilerPath` in  c_cpp_properties.json

michelleangela:
Issue https://github.com/microsoft/vscode-cpptools/issues/10808

This change updates the `compilerPath` in c_cpp_properties.json only if it is already set when a compiler is configured from the command "Select IntelliSense Configuration".

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -624,7 +624,7 @@ export class CppProperties {
         }, () => { }).catch(logAndReturn.undefined);
     }
 
-    public async updateCompilerPath(path: string): Promise<void> {
+    public async updateCompilerPathIfSet(path: string): Promise<void> {
         return this.handleConfigurationEditJSONCommand(() => {
             this.parsePropertiesFile(); // Clear out any modifications we may have made internally.
             const config: Configuration | undefined = this.CurrentConfiguration;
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -624,6 +624,21 @@ export class CppProperties {
         }, () => { }).catch(logAndReturn.undefined);
     }
 
+    public async updateCompilerPath(path: string): Promise<void> {
```
sean-mcmanus: It seems like this should be renamed to updateCompilerPathIfSet to avoid someone accidentally calling it when it's not set and the intention is to set it?

sean-mcmanus (changes requested):
rename suggestion

sean-mcmanus (approved):

Colengms (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -624,6 +624,21 @@ export class CppProperties {
         }, () => { }).catch(logAndReturn.undefined);
     }
 
+    public async updateCompilerPathIfSet(path: string): Promise<void> {
+        return this.handleConfigurationEditJSONCommand(() => {
```
bobbrow, line 628: Doesn't this create a `c_cpp_properties.json` file if one doesn't exist?  We shouldn't be creating a new one, only updating an existing one if it sets the `compilerPath`.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
michelleangela: I forgot about that. I'll fix it.

PR #11258, opened 2023-07-27
Add vscode.d.ts to .eslintignore.

sean-mcmanus:


Extension/.eslintignore
```diff
@@ -2,3 +2,4 @@
 test/**/index.ts
 test/**/runTest.ts
 tools/prepublish.js
+vscode.d.ts
```
bobbrow: What is vscode.d.ts doing in the root folder?  I thought it should only be in the node_modules folder, or does this pattern match any file with this name throughout the tree?  (and why would eslint be running in the node_modules folder anyway?)

sean-mcmanus: `"postinstall": "npx vscode-dts dev && npx vscode-dts main"` puts it in the root.

sean-mcmanus: From https://github.com/microsoft/vscode-dts#readme

sean-mcmanus: Looks like the node_modules location is the old one, i.e. they have code to delete that.
```typescript
  if (fs.existsSync('node_modules/vscode/vscode.d.ts')) {
    fs.unlinkSync('node_modules/vscode/vscode.d.ts')
```

bobbrow: Sounds good.  It sounds like it should be in `.vscodeignore` too then, or is it needed for the vsix?

bobbrow: And use the same pattern as in the `.gitignore` so that both `.d.ts` files are excluded?

sean-mcmanus: Yeah, good catch, I didn't notice that.

fearthecowboy (approved):

Issue #10421, opened 2023-01-25
Extension always shows some C++ file specific commands in the command palette 


mjbvz:
### Environment

- C/C++ Extension Version: v1.14.0 pre-released

### Bug Summary and Steps to Reproduce

1. Install extension
2. Open a non c++ file, such as a `.py` file
3. Open command plate and search `c++`

**bug**
A bunch of commands that are specific to c/c++ files are shown in the command palette. A good example is `Go to next preprocessor directive`

![Screenshot 2023-01-25 at 3 46 57 PM](https://user-images.githubusercontent.com/12821956/214719230-7ec4ddd1-beec-43b8-bd8a-5dabc42b4e74.png)

These commands should  be reviewed and any that only apply in c++ files should be hidden when not in relevant files

### Configuration and Logs

```shell
NA
```

### Additional context

These commands can be hidden by adding an `enablement` clause to the `command` contribution, or using the `menus.commandPalette` contribution

Colengms:
It looks like the following are specific to the active file and should be hidden if the active file is not C/C++/CUDA:

```
SwitchHeaderSource
EnableErrorSquiggles
DisableErrorSquiggles
ToggleDimInactiveRegions
RestartIntelliSenseForFile
GoToNextDirectiveInGroup
GoToPrevDirectiveInGroup
CreateDeclarationOrDefinition
RunCodeAnalysisOnActiveFile
BuildAndDebugFile
BuildAndRunFile
AddDebugConfiguration
GenerateDoxygenComment
```

It looks like the following can be removed entirely:

```
ToggleIncludeFallback
TakeSurvey
```

sean-mcmanus:
I ran across this issue while investigating the pending Extract Function commands and code actions changes in package.json.

michelleangela:
Fix available in pre-release https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.1.

michelleangela:
Fixed in release version https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.2.

PR #11259, opened 2023-07-28
Clean up the command palette.

sean-mcmanus:
Address most of https://github.com/microsoft/vscode-cpptools/issues/10421 . The regex usage is based on what the TypeScript extension does (I ran across this while looking into their Extract Function support, so I decided to clean it up now).

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3221,7 +3221,7 @@
             {
                 "command": "C_Cpp.SwitchHeaderSource",
                 "key": "Alt+O",
-                "when": "editorLangId == 'c' && editorTextFocus || editorLangId == 'cpp' && editorTextFocus || editorLangId == 'cuda-cpp' && editorTextFocus"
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
Do we need to file another issue to address anything else or is this good enough for now?

fearthecowboy (commented):
Regexes are fun!

`/^(cuda-cpp|c|cpp)$/` might be a bit more readable but I'm not sure that it makes a differnce

fearthecowboy (dismissed):

sean-mcmanus:
> Regexes are fun!
> 
> `/^(cuda-cpp|c|cpp)$/` might be a bit more readable but I'm not sure that it makes a differnce

Sure, that seems better, I'll try that.

sean-mcmanus:
I changed it to `/^(c|(cuda-)?cpp)$/` to avoid the extra -cpp.

fearthecowboy (approved):

michelleangela (approved):

aldirenio:
ola  galera  boa  noite 
  eu  quero instalar   o  c++   e   suas  dependencias   mais  ta  osso  alquem  poria  me  ajudar

PR #11261, opened 2023-07-28
Add keyboard support for Inline Macro.

### LINKED ISSUE #11260: Inline Macro requires invoking hover first
It's noticeable if you navigate to the macro using the keyboard instead of the mouse.

Invoking hover via Ctrl+k, Ctrl+i is a workaround.

The fix would involve detecting if the cursor is on a macro, change the code action to send a command to get the expanded macro, and then do the edit. Currently the code action just invokes the edit that was obtained from the hover.

UPDATE: Actually, I thought of a simpler fix by just invoking hover when the keyboard code action is triggered.

sean-mcmanus:
Based on what the TypeScript extensions does, since I'm doing something similar for Extract Function. Fixes https://github.com/microsoft/vscode-cpptools/issues/11260 .

The keyboard shortcut can be change/replaced/etc. (see https://code.visualstudio.com/docs/editor/refactoring ).

image

Also, normalized loc strings to use `.` instead of `_` and fixed spelling errors in string types (9 misspelled words).


michelleangela (dismissed):

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -218,7 +226,47 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
+                const editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+                if (!editor) {
+                    return false;
+                }
+                const result: vscode.Hover[] = <vscode.Hover[]>(await vscode.commands.executeCommand('vscode.executeHoverProvider', document.uri, range.start));
+                if (result.length === 0) {
+                    return false;
+                }
+                const hoverResult: vscode.MarkdownString = <vscode.MarkdownString>result[0].contents[0];
+                if (!hoverResult.value.includes(localize("expands_to", "Expands to:"))) {
```
bobbrow: nit: we usually use `.` instead of `_` for localization keys. (`expands.to`)

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -218,7 +226,47 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
+                    if (command.edit) {
+                        processCommand(command);
+                        return true;
+                    }
+                }
+                return false;
+            };
+            if (!await processInlineMacro()) {
+                const disabledCodeAction: vscode.CodeAction = {
+                    title: localize({ key: "inline_macro", comment: ["'Inline' is a command and not an adjective, i.e. like 'Expand macro'."] }, "Inline macro"),
```
bobbrow: `inline.macro`

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: ok

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: I ran a regex to find other cases of _ , i.e. looks like I had used that for some code analysis localize calls as well, so I'll fix those too.

bobbrow: Changing all the existing ones right now might reset all of the localization for them.  Maybe wait to do that until after 1.17 ships?

sean-mcmanus: @Colengms Do you know when we'll have translations by?

sean-mcmanus: I undid the other loc changes.

michelleangela (dismissed):

Colengms (dismissed):

bobbrow:
@sean-mcmanus is this ready to check in?

bobbrow (approved):

Colengms (approved):

Issue #11247, opened 2023-07-26
Call Hierarchy child items aren't formatted correctly (default arguments are shown)


sean-mcmanus:
See the screenshot. The root definition is formatted correctly but the children are not, which makes the arguments harder to read.

![image](https://github.com/microsoft/vscode-cpptools/assets/19859882/8ac6f7ad-e066-48cd-aa6c-96f9bc6ede02)

Repro code is like

```cpp
#include <unordered_map>

int foo() { return 0; }

bool func(std::unordered_map<int, int> * d = nullptr)
{
    return foo() == 1;
}
```

The bug should be easy to fix by doing the formatting that is done for the root definition on the child items.

michelleangela:
**Before fix:**
![image](https://github.com/microsoft/vscode-cpptools/assets/38734287/beeee721-1830-4efa-aef2-6b0a2bd35bd7)


**After fix:**
![image](https://github.com/microsoft/vscode-cpptools/assets/38734287/c32169c7-153b-4fbd-b45f-149741ef80fa)

michelleangela:
Fix available in pre-release https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.1.

PR #11264, opened 2023-07-28
Use `selectionRange` for call hierarchy request input

michelleangela:
**GitHub issue:**
https://github.com/microsoft/vscode-cpptools/issues/11247

This is a corresponding change to a language server change, which has the actual fix to https://github.com/microsoft/vscode-cpptools/issues/11247.

**Change:**
Use the property `selectionRange` as the position input for a call hierarchy incoming/outgoing request. Previously `selectionRange` and `range` had the same values. A changed in language server now populates `range` with the function definition range if the data is available.

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -30,7 +30,7 @@ interface CallHierarchyItem {
     /**
      * The resource identifier of this item.
      */
-    uri: string;
+    file: string;
 
     /**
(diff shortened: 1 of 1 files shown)


sean-mcmanus (dismissed):
This actually fixes https://github.com/microsoft/vscode-cpptools/issues/11247 ?

sean-mcmanus:
> This actually fixes #11247 ?

Oh, I see, there's a cpptools fix too.

sean-mcmanus (dismissed):

sean-mcmanus (approved):

Issue #11254, opened 2023-07-26
File path info in the Call Hierarchy should show relative paths for files in the current workspace


sean-mcmanus:
Call Hierarchy file info on the right shows the full path instead of the workspace relative path. This makes it harder to see the important part of the path and it could show potentially unwanted path details like usernames/projects in screenshots/videos.

![image](https://github.com/microsoft/vscode-cpptools/assets/19859882/dae1e89c-5fed-45d0-8a1c-964062a0f498)

The path should match or be similar to what VS Code shows in the Open Editors pane (D2 is a folder at the root of the workspace).

![image](https://github.com/microsoft/vscode-cpptools/assets/19859882/02b9aafe-c465-4578-adc9-3b77b2e60f11)

michelleangela:
Fix available in pre-release https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.1.

PR #11265, opened 2023-07-29
Show relative path for call hierarchy item if available

michelleangela:
**GitHub issue:**
https://github.com/microsoft/vscode-cpptools/issues/11254

**Change:**
Show the path relative to workspace folder for file path info of call hierarchy items. Previously it showed full path.

**Before changes:**
image

**After changes:**
image

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -230,7 +230,9 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
 
     private makeVscodeCallHierarchyItem(item: CallHierarchyItem): vscode.CallHierarchyItem {
         const containerDetail: string = (item.detail !== "") ? `${item.detail} - ` : "";
-        const fileDetail: string = `${path.basename(item.uri)} (${path.dirname(item.uri)})`;
+        const dirPath: string = path.relative(this.client.RootPath, path.dirname(item.uri));
+        const fileDetail: string = dirPath.length === 0 ?
+            `${path.basename(item.uri)}` : `${path.basename(item.uri)} (${dirPath})`;
         return new vscode.CallHierarchyItem(
             item.kind, item.name, containerDetail + fileDetail,
             vscode.Uri.file(item.uri),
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
```diff
@@ -230,7 +230,9 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
 
     private makeVscodeCallHierarchyItem(item: CallHierarchyItem): vscode.CallHierarchyItem {
         const containerDetail: string = (item.detail !== "") ? `${item.detail} - ` : "";
-        const fileDetail: string = `${path.basename(item.uri)} (${path.dirname(item.uri)})`;
+        const dirPath: string = path.relative(this.client.RootPath, path.dirname(item.uri));
```
sean-mcmanus: It should only show the relative path if the file is located under the rootPath. I think you can handle this via checking for ".." the dirPath (not sure if there's another recommended way).

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
michelleangela: Yes, if `path.dirname(item.uri)` does not exist in `this.client.RootPath`, then `path.relative` returns the full path. Is this what you meant?

sean-mcmanus: @michelleangela No, I just tested it, e.g. on a call to std::string::length() (also the documentation of path.relative says that it'll return a path with `..`). Or do I need cpptools side changes?

image

michelleangela: There's no language server changes. I tested it with a different external folder and maybe have missed the `...`

michelleangela: I didn't see the `..` because I tested with a folder from a different drive 

image

sean-mcmanus: Ah, that makes sense, since you can't use `..` across to reference a file on another drive.

sean-mcmanus (dismissed):

sean-mcmanus (approved):

Issue #11252, opened 2023-07-26
Remove vcpkg code action from missing include scenarios


sean-mcmanus:
This used to work at some point with an internal build, but I'm not sure if it ever worked with a public release (i.e. the ones I tried all don't work).

The code action implementation is supposed to check if it's a system header and show different code actions.

You can set compilerPath to "" for a repro.

I reviewed the code and it seems to be setup to do the right thing, but it's going wrong at some point either in cpptools or the TypeScript that handles the result.

![image](https://github.com/microsoft/vscode-cpptools/assets/19859882/16668534-bd1b-4741-9fc4-c96c881686e4)

bobbrow:
It would be nice to remove the "vcpkg" code action as part of this if possible.

sean-mcmanus:
@bobbrow You mean remove the vcpkg code action entirely? Even for non-system includes?

bobbrow:
No, just for system includes since it's not going to help for that case.

sean-mcmanus:
@bobbrow Yeah, the code already handles/checks that -- it's just a matter of figuring out why it's not working as expected (like it used to).

browntarik:
The edit compiler path code action appearing is expected behavior. A fix for the vcpkg code action can be found here: https://github.com/microsoft/vscode-cpptools/pull/11280

michelleangela:
Fix available in pre-release https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.1.

PR #11280, opened 2023-08-02
Prevent vcpkg code action from appearing in all unrelated code action menus

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11252

The vcpkg code action keeps appearing because it is not checking if the include code action is in the list of code actions. This PR will prevent vcpkg from being added to the list before this happens

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -229,10 +229,6 @@ export async function activate(): Promise<void> {
     codeActionProvider = vscode.languages.registerCodeActionsProvider(selector, {
         provideCodeActions: async (document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext): Promise<vscode.CodeAction[]> => {
             const ports: string[] = await lookupIncludeInVcpkg(document, range.start.line);
-            const actions: vscode.CodeAction[] = ports.map<vscode.CodeAction>(getVcpkgClipboardInstallAction);
-            if (ports.length <= 0) {
-                return [];
-            }
 
             if (!await clients.ActiveClient.getVcpkgEnabled()) {
                 return [];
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -228,6 +228,12 @@ export async function activate(): Promise<void> {
     ];
     codeActionProvider = vscode.languages.registerCodeActionsProvider(selector, {
         provideCodeActions: async (document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext): Promise<vscode.CodeAction[]> => {
+            const ports: string[] = await lookupIncludeInVcpkg(document, range.start.line);
```
bobbrow: I don't think this should be the first early return check.  I would make this the last one (right before `telemetry.logLanguageServerEvent`)

Extension/src/LanguageServer/extension.ts
```diff
@@ -228,6 +228,12 @@ export async function activate(): Promise<void> {
...
+            const actions: vscode.CodeAction[] = ports.map<vscode.CodeAction>(getVcpkgClipboardInstallAction);
```
bobbrow: This line should be at the end of the function because we don't need to do this work if we're not going to return the `actions`

Extension/src/LanguageServer/extension.ts
```diff
@@ -228,6 +228,8 @@ export async function activate(): Promise<void> {
     ];
     codeActionProvider = vscode.languages.registerCodeActionsProvider(selector, {
         provideCodeActions: async (document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext): Promise<vscode.CodeAction[]> => {
+            const ports: string[] = await lookupIncludeInVcpkg(document, range.start.line);
```
bobbrow: Move this line down by 242 since we don't need to do this work until then.

bobbrow (approved):

sean-mcmanus (approved):

PR #11300, opened 2023-08-07
default to c++14 on mac for build and run active file

bobbrow:
This is a temporary workaround to enable users to compile the code in the getting started guide/walkthrough without having to create and edit tasks.json.

### DIFF
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -189,10 +189,6 @@ export class CppBuildTaskProvider implements TaskProvider {
                     ['-fcolor-diagnostics', '-fansi-escape-codes', '-g', '${file}', '-o', programName] :
                     ['-fdiagnostics-color=always', '-g', '${file}', '-o', programName];
 
-            // TODO: Remove when compiler query work goes in and we can determine the standard version from TypeScript
-            if (isClang && os.platform() === 'darwin') {
-                args.unshift('-std=c++14');
-            }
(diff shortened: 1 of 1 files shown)


Colengms (dismissed):

sean-mcmanus (approved):
Is c++14 sufficient or was it c++17?

sean-mcmanus (changes requested):
Wait a sec -- won't this break C users, i.e. when F5 is done with a .c file?

error: invalid argument '-std=c++14' not allowed with 'C'

i.e. can you add a check for the file extension before adding that -std flag?

sean-mcmanus:
Also, I believe the default would be a GNU version like "gnu14++" instead of "c++14".

bobbrow:
> Is c++14 sufficient or was it c++17?

c++14 was sufficient in my tests.  I was thinking c+11 was also sufficient, I'd need to check again...

> Wait a sec -- won't this break C users, i.e. when F5 is done with a .c file?

Good catch.

> Also, I believe the default would be a GNU version like "gnu++14" instead of "c++14".

I thought that was primarily for g++, not clang.  Is it?

sean-mcmanus:
The default for apple clang is currently gnu++99 according to the docs at https://opensource.apple.com/source/clang/clang-137/src/tools/clang/docs/UsersManual.html .

bobbrow:
Latest change is not tested yet.  Don't check in yet.  I'll validate this tomorrow.

bobbrow:
Tested on mac. This is ready for review.

sean-mcmanus (approved):

Colengms (approved):

PR #11308, opened 2023-08-11
log machine id mappings for VS to VS Code

bobbrow:


Extension/src/id.ts
```diff
@@ -0,0 +1,64 @@
...
+    const output = await execChildProcess('getmac');
+    const regex = /(?:[a-z0-9]{2}[:\-]){5}[a-z0-9]{2}/gmi;
+    return output.match(regex) ?? [];
+}
+
+/**
+ * Code below is adapted from:
+ *  - vscode\src\vs\base\node\id.ts
+ */
+
```
sean-mcmanus, line 55: This extra blank line should be removed, right? Or is there a significance to the blank line? Should the comment of "code below" be just "function below"? or was there more going to be added?

Extension/src/id.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    // VS Code uses os.networkInterfaces() which has different sorting and availability,
+    // but all MAC addresses are returned by getmac.exe. The ID VS Code uses may change
+    // based on changes to the network configuration. Log the extras so we can assess
+    // how frequently this impacts the machine id.
+    for (const macAddress of macAddresses) {
+        const additional = await getMachineId(macAddress);
+        if (additional) {
+            logLanguageServerEvent('machineIdMap', {additional});
```
sean-mcmanus, line 33: It seems like the additional and primary mac addresses are being logged the same -- is that okay or would there be a need to distinguish between the two?

sean-mcmanus: Oh -- are primary/additional properties in the telemetry?

sean-mcmanus: i.e. it's a simplification of { primary: primary }, right?

sean-mcmanus: Yeah, that's what it's doing (I found an existing example `telemetry.logLanguageServerEvent('configurationSelection', { action, `.

sean-mcmanus (dismissed):

Extension/src/id.ts
bobbrow: I had originally added more code thinking I might log the additional hashes for mac/linux too, but ultimately removed that code.  I left the comment in case we did add more there.

sean-mcmanus (approved):

Colengms (approved):

PR #11309, opened 2023-08-11
Rework Database Storage Path Logic

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10191


Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,35 @@ export class DefaultClient implements Client {
...
+            if (fsPath) {
+                workspaceHash = path.basename(path.dirname(fsPath))
+                legacyStoragePath = fsPath;
             }
         }
 
-        if (!storagePath) {
-            storagePath = this.RootPath ? path.join(this.RootPath, "/.vscode") : "";
+        if (!legacyStoragePath) {
+            legacyStoragePath = this.RootPath ? path.join(this.RootPath, "/.vscode") : "";
```
bobbrow: remove the `/` from the second parameter.

Extension/src/common.ts
```diff
@@ -1300,6 +1300,31 @@ export function isValidIdentifier(candidate: string): boolean {
     return true;
 }
 
+export function getBaseStoragePath(){
```
bobbrow: run the formatter and linter on this file

Extension/src/common.ts
```diff
@@ -1300,6 +1300,31 @@ export function isValidIdentifier(candidate: string): boolean {
     return true;
 }
 
+export function getBaseStoragePath() {
```
sean-mcmanus: Should this function be declared with a return type?

Extension/src/common.ts
```diff
@@ -1300,6 +1300,31 @@ export function isValidIdentifier(candidate: string): boolean {
...
+        default: // Linux
+            defaultCachePath = "\\vscode-cpptools\\";
+            pathEnvironmentVariable = process.env["XDG_CACHE_HOME"];
+            if (!pathEnvironmentVariable) {
+                pathEnvironmentVariable = os.homedir();
+            }
+            break;
+    }
+
+    const result: string | undefined = pathEnvironmentVariable?.concat(defaultCachePath);
```
sean-mcmanus: Can this be directly returned instead of saving it to a result variable?

sean-mcmanus: Can you use path.join and remove the beginning \\ from the strings?

Extension/src/LanguageServer/client.ts
```diff
@@ -539,6 +540,8 @@ export interface TextDocumentWillSaveParams {
 interface InitializationOptions {
     packageVersion: string;
     extensionPath: string;
+    baseStoragePath: string | undefined;
```
sean-mcmanus: The other code puts legacyStoragePath before baseStoragePath? Can you do that make it consistent?

sean-mcmanus: Well, the native side puts legacyStoragePath after baseStoragePath so maybe that should be done instead.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
...
         if (workspaceFolder && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
-            storagePath = path.join(storagePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+            legacyStoragePath = path.join(legacyStoragePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+        }
+
+        this.legacyStoragePath = legacyStoragePath;
+        this.baseStoragePath = baseStoragePath;
+
+        if (baseStoragePath && workspaceHash) {
+            storagePath = baseStoragePath + workspaceHash;
```
sean-mcmanus: Can you use path.join instead of `+`?

Extension/src/common.ts
```diff
@@ -1300,6 +1300,31 @@ export function isValidIdentifier(candidate: string): boolean {
...
+export function getBaseStoragePath() {
+    let defaultCachePath: string = "";
+    let pathEnvironmentVariable: string | undefined = "";
+    switch (os.platform()) {
+        case 'win32':
+            defaultCachePath = "\\Microsoft\\vscode-cpptools\\";
+            pathEnvironmentVariable = process.env["LOCALAPPDATA"];
+            break;
+        case 'darwin':
+            defaultCachePath = "\\Library\\Caches\\vscode-cpptools\\";
```
bobbrow: It seems odd to use backslashes for paths on linux and mac.  Can you flip these to forward slashes?

Extension/src/LanguageServer/client.ts
sean-mcmanus: storagePath isn't set in the open file scenario (no workspaceHash)? Is that a bug?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
         this.rootFolder = workspaceFolder;
         this.rootRealPath = this.RootPath ? (fs.existsSync(this.RootPath) ? fs.realpathSync(this.RootPath) : this.RootPath) : "";
 
-        let storagePath: string | undefined;
+        const baseStoragePath: string | undefined = util.getBaseStoragePath();
+        let legacyStoragePath: string | undefined;
+        let storagePath: string | undefined = "";
```
michelleangela: It looks like `storagePath` can just be a string and doesn't need to be undefined too.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
...
+
         if (workspaceFolder && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
-            storagePath = path.join(storagePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+            legacyStoragePath = path.join(legacyStoragePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+        }
+
+        this.legacyStoragePath = legacyStoragePath;
+        this.baseStoragePath = baseStoragePath;
+
+        if (baseStoragePath && workspaceHash) {
```
michelleangela: Is `if (baseStoragePath && workspaceHash)` to check these are undefined or not empty?

sean-mcmanus:
@browntarik Just a reminder to the test the Open File scenario.

Extension/src/LanguageServer/client.ts
bobbrow: VS Code only provides a storage path when there is a folder open.  There is a "globalStorageUri" though which might always exist.

bobbrow: ~~And it looks like `storagePath` is deprecated and we should be using `storageUri` now instead.  Maybe we should fix that now while we're touching the code.~~

EDIT: nevermind. We already do.  It's just named `storagePath` in our classes.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
...
+        if (!legacyStoragePath) {
+            legacyStoragePath = this.RootPath ? path.join(this.RootPath, ".vscode") : "";
         }
+
         if (workspaceFolder && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
-            storagePath = path.join(storagePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+            legacyStoragePath = path.join(legacyStoragePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+        }
+
+        this.legacyStoragePath = legacyStoragePath;
```
sean-mcmanus: It seems like you can remove the legacyStoragePath/baseStoragePath variables if you just reference this.legacyStoragePath/this.baseStoragePath directly.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
         this.rootFolder = workspaceFolder;
         this.rootRealPath = this.RootPath ? (fs.existsSync(this.RootPath) ? fs.realpathSync(this.RootPath) : this.RootPath) : "";
 
-        let storagePath: string | undefined;
+        const baseStoragePath: string | undefined = util.getBaseStoragePath();
+        let legacyStoragePath: string | undefined;
```
sean-mcmanus: This doesn't need to be undefined if it's initialized to "" like the storagePath is (but see my other comment about potentially removing the variable).

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,34 @@ export class DefaultClient implements Client {
...
+        let storagePath: string | undefined = "";
+        let workspaceHash: string = "";
+
         if (util.extensionContext) {
```
michelleangela: Can this be simplified to:

```ts
        if (util.extensionContext?.storageUri?.fsPath) {
            legacyStoragePath = util.extensionContext?.storageUri?.fsPath;
            workspaceHash = path.basename(path.dirname(legacyStoragePath));
        }
```

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: I think it can be further simplified

```TypeScript
this.legacyStoragePath = util.extensionContext?.storageUri?.fsPath ?? "";
if (!this.legacyStoragePath.empty()) {
    workspaceHash = path.basename(path.dirname(this.legacyStoragePath));
}
```

That's with the local variable removed and avoid multiple calls to util.extensionContext.

Extension/src/LanguageServer/client.ts
browntarik: Can you give an example?

Extension/src/LanguageServer/client.ts
browntarik: not empty

Extension/src/LanguageServer/client.ts
browntarik: I don't think this accounts for multiple workspace folders. See:

     if (workspaceFolder && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
            legacyStoragePath = path.join(legacyStoragePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
      }

Extension/src/LanguageServer/client.ts
michelleangela: Should it use `empty()` to check if the string is not empty?

Extension/src/common.ts
```diff
@@ -1300,6 +1300,33 @@ export function isValidIdentifier(candidate: string): boolean {
     return true;
 }
 
+export function getBaseStoragePath(): string | undefined {
+    let defaultCachePath: string = "";
+    let pathEnvironmentVariable: string | undefined = "";
```
michelleangela: `let pathEnvironmentVariable: string | undefined` can be left uninitialized since it can be undefined.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,20 +1299,29 @@ export class DefaultClient implements Client {
...
+            this.legacyStoragePath = this.RootPath ? path.join(this.RootPath, ".vscode") : "";
         }
+
         if (workspaceFolder && vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
-            storagePath = path.join(storagePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
+            this.legacyStoragePath = path.join(this.legacyStoragePath, util.getUniqueWorkspaceStorageName(workspaceFolder));
         }
+
+
+        if ((this.baseStoragePath?.length !== undefined) && (workspaceHash.length > 0)) {
```
michelleangela: Why is this checking for length is undefined `(this.baseStoragePath?.length !== undefined)`? Is length only a number?

michelleangela (changes requested):

Extension/src/LanguageServer/client.ts
michelleangela: Can this `baseStoragePath` be only string too?

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,21 +1299,26 @@ export class DefaultClient implements Client {
...
 
-        let storagePath: string | undefined;
-        if (util.extensionContext) {
-            const path: string | undefined = util.extensionContext.storageUri?.fsPath;
-            if (path) {
-                storagePath = path;
-            }
-        }
+        this.baseStoragePath = util.getBaseStoragePath();
+        this.legacyStoragePath =  "";
```
sean-mcmanus: 2 spaces before "" -- does format document work?

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yes

Extension/src/LanguageServer/client.ts
```diff
@@ -925,7 +930,7 @@ export class DefaultClient implements Client {
     public get AdditionalEnvironment(): { [key: string]: string | string[] } {
         return {
             workspaceFolderBasename: this.Name,
-            workspaceStorage: this.storagePath,
+            workspaceStorage: this.legacyStoragePath,
```
Colengms: This value is used by `resolveVariables` any time the user specifies the `workspaceStorage` variable.  Shouldn't the new storage location be used here, and the legacy storage path never used anymore (except to delete an old database) ?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: No, it's fine -- the workspaceStorage is still the workspaceStorage/legacyStoragePath -- the only thing that is changing is that we're not storing the database at that spot -- other stuff might still be stored there.

sean-mcmanus: ...if we wanted to add a databaseStoragePath or something that would be a new variable.

sean-mcmanus: My assumption is that users are not using this variable to refer to our default database location.

Colengms: If there is a reason to store our database in a different storage location, it's unclear to me why the same reasoning would not apply to anything else that similarly needs storage.

sean-mcmanus: Because it's a large file that is a cache that is potentially deletable (going to XDG_CACHE_HOME on Linux), where-as other workspace storage is supposed to be small and may not be deletable.

Colengms: Should we use a better name for the variable, then, since it's still actively used and not a 'legacy' holdover?

sean-mcmanus: Better name for what variable?

sean-mcmanus: We figured out the renaming...

Extension/src/LanguageServer/client.ts
```diff
@@ -539,6 +540,8 @@ export interface TextDocumentWillSaveParams {
 interface InitializationOptions {
     packageVersion: string;
     extensionPath: string;
+    cacheStoragePath: string | undefined;
```
michelleangela: Change `cacheStoragePath` to string. The object is not declared as optional on the language server changes so it does not need to be undefined here.

Extension/src/LanguageServer/client.ts
```diff
@@ -1617,7 +1627,9 @@ export class DefaultClient implements Client {
         const initializationOptions: InitializationOptions = {
             packageVersion: util.packageJson.version,
             extensionPath: util.extensionPath,
-            storagePath: this.storagePath,
+            storagePath: this.databaseStoragePath,
```
michelleangela: Is `storagePath` supposed to be `databaseStoragePath` so that it corresponds to the object received on the language server side?

browntarik: Thank you for catching this! This was causing a bug I couldn't figure out

Extension/src/LanguageServer/client.ts
```diff
@@ -829,7 +832,9 @@ export class DefaultClient implements Client {
     private rootPathFileWatcher?: vscode.FileSystemWatcher;
     private rootFolder?: vscode.WorkspaceFolder;
     private rootRealPath: string;
-    private storagePath: string;
+    private cacheStoragePath: string;
```
Colengms: This might not be obvious, but we create multiple instances of `client` (when using a workspace with multiple roots/folders).  As such, cacheStoragePath and databaseStoragePath probably shouldn't be client-specific, as we'll only be passing one of each value in the initialization object (sent only for the first client).  workspaceStoragePath should actually continue be specific to each client, since that is also used in `AdditionalEnvironment` (as that's used for variable expansion, which may occur in folder/client specific contexts).

I'd suggest making these two global or static.  Or, better yet, if these aren't needed for anything other than passing their values up in initializationOptions, maybe these don't need to be member variables at all.

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,21 +1299,26 @@ export class DefaultClient implements Client {
...
         this.rootRealPath = this.RootPath ? (fs.existsSync(this.RootPath) ? fs.realpathSync(this.RootPath) : this.RootPath) : "";
 
-        let storagePath: string | undefined;
-        if (util.extensionContext) {
-            const path: string | undefined = util.extensionContext.storageUri?.fsPath;
-            if (path) {
-                storagePath = path;
-            }
-        }
+        this.cacheStoragePath = util.getCacheStoragePath();
```
Colengms: As alluded to in my last comment, can you move the setup of `cacheStoragePath` and `databaseStoragePath` into `createLanguageClient`, immediately before these strings are added to `initializationOptions`, since that is the only place these are used?

Colengms (dismissed):

Extension/.vscode/settings.json
```diff
@@ -37,7 +37,7 @@
     },
     "[typescript]": {
         "editor.tabSize": 4,
-        "editor.defaultFormatter": "dbaeumer.vscode-eslint",
+        "editor.defaultFormatter": "vscode.typescript-language-features",
```
michelleangela: Revert change to settings.json

Extension/src/LanguageServer/client.ts
```diff
@@ -1294,21 +1298,19 @@ export class DefaultClient implements Client {
...
         this.rootRealPath = this.RootPath ? (fs.existsSync(this.RootPath) ? fs.realpathSync(this.RootPath) : this.RootPath) : "";
 
-        let storagePath: string | undefined;
-        if (util.extensionContext) {
-            const path: string | undefined = util.extensionContext.storageUri?.fsPath;
-            if (path) {
-                storagePath = path;
-            }
-        }
+        this.workspaceStoragePath = "";
```
michelleangela: Since `this.workspaceStoragePath` is already getting assigned on line 1303 with either a value or empty value, then this first assignment can be removed.

Extension/src/LanguageServer/client.ts
```diff
@@ -1617,10 +1619,19 @@ export class DefaultClient implements Client {
             currentCaseSensitiveFileSupport.Value = workspaceSettings.caseSensitiveFileSupport;
         }
 
+        const cacheStoragePath: string = util.getCacheStoragePath();
+        let databaseStoragePath: string = "";
```
michelleangela: `databaseStoragePath` can be const via:

```ts
const databaseStoragePath: string = (cacheStoragePath.length > 0) && (workspaceHash.length > 0) ? 
    path.join(cacheStoragePath, workspaceHash) : "";
```

michelleangela (dismissed):

sean-mcmanus (approved):

michelleangela (approved):

Issue #11335, opened 2023-08-17
Crash recovery not working properly


bobbrow:
I start to have the same issue since the extension auto-updated to 1.17.3 today. I'm running AlmaLinux 8.8 with kernel 4.18.0-477.15.1.el8_8.x86_64. The VSCode version is 1.81

Not sure if this is any useful, but the output message is:

```
[Error - 2:14:08 PM] The language server crashed. Restarting...
[Error - 2:14:08 PM] Sending document notification textDocument/didOpen failed
Error: Attempting to use languageClient before initialized
	at DefaultClient.get languageClient [as languageClient] (/u/vinson3z/.vscode-server/extensions/ms-vscode.cpptools-1.17.3-linux-x64/dist/main.js:47283:19)
	at DefaultClient.sendDidOpen (/u/vinson3z/.vscode-server/extensions/ms-vscode.cpptools-1.17.3-linux-x64/dist/main.js:48450:20)
	at DefaultClient.takeOwnership (/u/vinson3z/.vscode-server/extensions/ms-vscode.cpptools-1.17.3-linux-x64/dist/main.js:48439:20)
	at processDelayedDidOpen (/u/vinson3z/.vscode-server/extensions/ms-vscode.cpptools-1.17.3-linux-x64/dist/main.js:53325:30)
	at runMicrotasks (<anonymous>)
	at processTicksAndRejections (node:internal/process/task_queues:96:5)
	at async Function.dispatch (/u/vinson3z/.vscode-server/extensions/ms-vscode.cpptools-1.17.3-linux-x64/dist/main.js:48476:44)
```

Is there any known workaround for this problem?

Fixed the problem by downgrading to 1.16.3. Downgrading from 1.17.3 to 1.17.2 didn't work.

PS: This happens only when I try to use VSCode with a very large project. For a medium-sized project, it works fine.

_Originally posted by @vineetsoni in https://github.com/microsoft/vscode-cpptools/issues/10651#issuecomment-1682190996_

bobbrow:
`DefaultClient.takeOwnership` used to call `this.sendDidOpen` inside a `this.requestWhenReady` callback.  That was changed to `await this.ready;` in this PR: https://github.com/microsoft/vscode-cpptools/pull/11085, but that `await this.ready;` was removed in this PR: https://github.com/microsoft/vscode-cpptools/pull/11212.

@fearthecowboy, do you remember why that line was removed?

bobbrow:
[...]

PR #11337, opened 2023-08-17
Enqueue the call to sendDidOpen

fearthecowboy:
Fixing #11335


Extension/src/LanguageServer/client.ts
```diff
@@ -2264,13 +2269,13 @@ export class DefaultClient implements Client {
...
      */
     private static async dispatch() {
-        // ensure that this is OK to start working
-        await this.isStarted;
-
         // reset the promise for the dispatcher
         DefaultClient.dispatching.reset();
 
         do {
+            // ensure that this is OK to start working
```
fearthecowboy, line 2276: It turns out that the dispatching can have stuff going when something crashes, so we move the check to inside the loop

bobbrow (approved):

sean-mcmanus (approved):

PR #11356, opened 2023-08-23
Update localized intellisense crash string

browntarik:


bobbrow (dismissed):

Extension/src/nativeStrings.json
```diff
@@ -448,4 +448,4 @@
         "text": "Copying Declaration / Definition to clipboard failed: %s",
         "hint": "The operation 'Copy Declaration / Definition' on a function was not successful. %s is the error info that has a period at the end of the string."
     }
-}
\ No newline at end of file
+}
```
sean-mcmanus, line 455: I think the json formatter/linter is supposed add a newline at the end, but it was removed?

Extension/src/nativeStrings.json
```diff
@@ -248,7 +248,7 @@
     "memory_limit_shutting_down_intellisense": "Shutting down IntelliSense server: {0}. Memory usage is {1} MB and has exceeded the {2} MB limit.",
     "failed_to_query_for_standard_version": "Failed to query compiler at path \"{0}\" for default standard versions.  Compiler querying is disabled for this compiler.",
     "unrecognized_language_standard_version": "Compiler query returned an unrecognized language standard version.  The latest supported version will be used instead.",
-    "intellisense_process_crash_detected": "IntelliSense process crash detected.",
+    "intellisense_process_crash_detected": "IntelliSense process crash detected. {0}",
```
sean-mcmanus: I'm a little worried about this. It'll have something like "handle_completion" at the end, without any connection to the previous sentence. Can you just have 2 strings, the 2nd saying something like "IntelliSense process crashed while processing '{0}'"?

bobbrow: I think we should leave it as-is. Maybe put it in parentheses if you're really worried about it since it's not something our users should really care about.

sean-mcmanus: Yeah, but if it says "IntelliSense processed crash" then customers are going to care about the crash. I'd prefer that we not have to give repeated explanations in GitHub issues about what that additional "handle_completion" text is. And we want to make sure they report that additional text. I don't see any advantage is being cryptic about it.

bobbrow: Then I vote to leave it out.  They're not going to know what `handle_completion` means either way.

sean-mcmanus: @bobbrow We will when they give us the logging -- e.g. "IntelliSense crashed while processing 'extract_function'" after the user selects some code. Otherwise, the logging will have a lot of messages and it will be hard to tell which IntelliSense operation is the cause. That info is important in getting a crash repro, since it will require the user provide us with the selection they made and not just the code. For other IntelliSense crashes, the selection doesn't matter (as far as I've ever seen).

Also, that is the "top" of the IntelliSense crash call stack, so we could potentially correlate that with other known crashes that have that at the top of the call stack, which otherwise we wouldn't know which crash call stack it could be.

sean-mcmanus: I talked to Tarik -- I'd be fine with the `{0}` being replaced by `.` or `: <function_name>`.

browntarik: Sean and I discussed a compromise.

sean-mcmanus (approved):

Extension/src/nativeStrings.json
```diff
@@ -248,7 +248,10 @@
     "memory_limit_shutting_down_intellisense": "Shutting down IntelliSense server: {0}. Memory usage is {1} MB and has exceeded the {2} MB limit.",
     "failed_to_query_for_standard_version": "Failed to query compiler at path \"{0}\" for default standard versions.  Compiler querying is disabled for this compiler.",
     "unrecognized_language_standard_version": "Compiler query returned an unrecognized language standard version.  The latest supported version will be used instead.",
-    "intellisense_process_crash_detected": "IntelliSense process crash detected.",
+    "intellisense_process_crash_detected": {
+        "text": "IntelliSense process crash detected{0}",
+        "hint": "This is an error message to be reported when the IntelliSense process crashes. The string will either end with a period `detected.` or with a colon detailing the feature that caused the crash `dectected: <feature name>."
```
sean-mcmanus: I don't think this 1st sentence is necessary for proper translation, but I don't think it would harm anything.

sean-mcmanus (changes requested):
Unfortunately, the `.` and `:` can't be in the `{0}` because I just realized that they're translated differently, i.e. they need to be part of the loc string -- can you just create 2 strings, one that ends with `: {0}` and one that ends with `.` (the original)?

Extension/src/nativeStrings.json
```diff
@@ -249,9 +249,13 @@
     "failed_to_query_for_standard_version": "Failed to query compiler at path \"{0}\" for default standard versions.  Compiler querying is disabled for this compiler.",
     "unrecognized_language_standard_version": "Compiler query returned an unrecognized language standard version.  The latest supported version will be used instead.",
     "intellisense_process_crash_detected": {
-        "text": "IntelliSense process crash detected{0}",
+        "text": "IntelliSense process crash detected.",
         "hint": "This is an error message to be reported when the IntelliSense process crashes. The string will either end with a period `detected.` or with a colon detailing the feature that caused the crash `dectected: <feature name>."
```
sean-mcmanus: The 2nd sentence of the hint is no longer valid.

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
```diff
@@ -248,7 +248,14 @@
...
     "failed_to_query_for_standard_version": "Failed to query compiler at path \"{0}\" for default standard versions.  Compiler querying is disabled for this compiler.",
     "unrecognized_language_standard_version": "Compiler query returned an unrecognized language standard version.  The latest supported version will be used instead.",
-    "intellisense_process_crash_detected": "IntelliSense process crash detected.",
+    "intellisense_process_crash_detected": {
+        "text": "IntelliSense process crash detected.",
+        "hint": "This is an error message to be reported when the IntelliSense process crashes. The string will either end with a period `detected.` or with a colon detailing the feature that caused the crash `dectected: <feature name>."
+    },
+    "intellisense_feature_crash_detected": {
+        "text": "IntelliSense process crash detected: {0}",
+        "hint": "This is an error message to be reported when the IntelliSense process crashes due to an IntelliSense feature.`{0}` will be substitued with `<feature name>`."
```
sean-mcmanus: Can you remove hint info that isn't necessary to translate the string? I don't think the 1st sentence is necessary to translate the string.

Extension/src/nativeStrings.json
michelleangela: Is "or with a colon detailing the feature that caused the crash `dectected: <feature name>." referring to the other string intellisense_feature_crash_detected? If so, this hint should be removed form intellisense_process_crash_detected.

michelleangela (dismissed):

Extension/src/nativeStrings.json
```diff
@@ -249,6 +249,10 @@
     "failed_to_query_for_standard_version": "Failed to query compiler at path \"{0}\" for default standard versions.  Compiler querying is disabled for this compiler.",
     "unrecognized_language_standard_version": "Compiler query returned an unrecognized language standard version.  The latest supported version will be used instead.",
     "intellisense_process_crash_detected": "IntelliSense process crash detected.",
+    "intellisense_feature_crash_detected": {
+        "text": "IntelliSense process crash detected: {0}",
+        "hint": "`{0}` will be substitued with `<feature name>`."
```
sean-mcmanus: substitued->substituted

sean-mcmanus (changes requested):

sean-mcmanus (approved):

michelleangela (approved):

PR #11394, opened 2023-08-31
Make `yarn` tasks imply `yarn install` where appropriate

fearthecowboy:
After chatting with @Colengc and @sean-mcmanus about implied `yarn install` I've refactored the scripts a bit more in `package.json` and now `compile` and `watch` and `webpack` and others will call `yarn install` (which in itself is very fast) -- and will verify the prep steps and run those only if necessary. (which are slower, and less like to be needed to run often)


sean-mcmanus (dismissed):

Extension/package.json
```diff
@@ -6121,24 +6121,25 @@
...
-        "lint": "eslint -c .eslintrc.js --report-unused-disable-directives src test ui .scripts",
-        "compile": "(yarn verify prep || yarn prep) && tsc --build tsconfig.json",
-        "watch": "(yarn verify prep || yarn prep )&& tsc --build tsconfig.json --watch",
-        "rebuild": "yarn clean && yarn prep && yarn compile",
+        "prep": "yarn prep:dts && yarn copy-walkthrough-media && yarn generate-native-strings && yarn translations-generate",
+        "lint": "yarn install && eslint -c .eslintrc.js --report-unused-disable-directives src test ui .scripts",
+        "compile": "yarn install && (yarn verify prep --quiet || yarn prep) && yarn build",
+        "watch": "yarn install && (yarn verify prep --quiet || yarn prep) && tsc --build tsconfig.json --watch",
+        "rebuild": "yarn install && yarn clean && yarn prep && yarn build",
         "vscode:prepublish": "yarn clean && yarn webpack",
```
Colengms, line 6132: Do you want to move the `translations-generate` step to `vscode:prepublish` only?  Or may a separate version of `compile`/`webpack` that someone could run only if they really need localized strings generated (and which `vscode:prepublish` would leverage, so happens for the VSIX).  It looks like you're now checking to ensure this only happens once now?  FYI, locally, it may be needed only zero times.  :)

fearthecowboy: It's only doing it once and it's pretty good at making sure that it's not being done needlessly.
I didn't want to explode to more tasks, or complicate things further at this point. 

I suspect we may have more opportunities to refine this as I go look at the other pipelines anyway.

Extension/.scripts/verify.ts
```diff
@@ -3,13 +3,50 @@
...
+    if (failing) {
+        process.exit(1);
+    }
+}
+
+export async function compiled() {
+    let failing = false;
+    failing = (await checkCompiled() && (quiet || error(`Compiled files are not present. Run ${green('yarn compile')} to fix it.`))) || failing;
+
+    if (failing){
```
michelleangela: Add space between `)` and `{`. Is linter not working?

fearthecowboy: There was no rule for that. Added one.

Colengms (approved):

michelleangela (approved):

Issue #11271, opened 2023-07-31
Add setting to control whether or not to add `#include` for refactoring actions (i.e. create declaration/definition)


oldbitsnbytes:
### Environment

- OS and Version: latest/any
- VS Code Version: latest/any
- C/C++ Extension Version:latest/any 
- If using SSH remote, specify OS of remote machine: not applicable.

### Bug Summary and Steps to Reproduce

Bug Summary: 
You shall/must NOT assume:
1-  include is "local include" 
2-  might never been already chain-included.

Steps to reproduce:
1. any c++ project
2  new class/struct method member.
3. ctrl+. "Create definition of $funcname in ... $file.cc"
4. .cc: 

#include "$file.h" -> You cannot assume that you systematically have to add this!

Expected behavior:
Please, ** Let the programmer decides where and how to include the file. **

### Configuration and Logs

```shell
not applicable.
```

### Other Extensions

not applicable.

### Additional context

not applicable.

bobbrow:
We include the file if it's not present as a nod to this guideline: [Avoid dependencies on implicitly included names](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names)

We also offer an option to copy the definition/declaration to the clipboard so that you can insert it any way you want.  This does not add any `#include` lines.  Does this address your need?
![image](https://github.com/microsoft/vscode-cpptools/assets/12818240/95590846-8787-4c84-b8d4-8efebf93ef69)

oldlonecoder:
My apologies for me been rude. I feel that this issue might be closed and put it under the carpet. [To be taken with a grain of salt]; 

Changing the title to "Don't add #include to source file" is a brilliant one for describing the idea. 

What else can I say?

michelleangela:
Draft PR for this issue.
Pull request 496507: cpptools - do not add #include if header or source file exists - Repos (visualstudio.com)<https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/VS/pullrequest/496507>

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.18.0 (via the C_Cpp.refactoring.includeHeader setting).

PR #11449, opened 2023-09-20
Add Include Header setting for refactoring actions

michelleangela:
Issue: https://github.com/microsoft/vscode-cpptools/issues/11271

To customize whether or not to add include header during refactoring code actions.


Extension/package.json
```diff
@@ -3050,6 +3050,22 @@
                         "default": false,
                         "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.markdownDescription%",
                         "scope": "application"
+                    },
+                    "C_Cpp.includeHeader": {
```
bobbrow: Right now this is in the "Miscellaneous" sub-heading.  I think it should go in the `"title": "%c_cpp.subheaders.intelliSense.title%"` object unless we plan to add a refactoring group.

Should this be renamed to `C_Cpp.refactor.includeHeader` to scope it for refactorings?

michelleangela: Scoped to refactor in case other refactoring settings are added in the future. Also, to clarify this only applies to refactoring operations.

bobbrow (dismissed):

Extension/package.json
```diff
@@ -561,6 +561,22 @@
                         "description": "%c_cpp.configuration.inactiveRegionBackgroundColor.description%",
                         "scope": "resource"
                     },
+                    "C_Cpp.refactor.includeHeader": {
```
sean-mcmanus: I think "refactor" should be "refactoring" to match the other VS Code references to the term "refactoring", e.g. files.refactoring.autoSave

sean-mcmanus (changes requested):

sean-mcmanus (approved):

Issue #1162, opened 2017-10-24
Extract to function (free function or member function)


JamesRamm:
Thanks for the great work on this; I am successfully using vscode-cpptools to work on a >2m line Qt project with no hitches so far :+1: :smile: 

Refactoring is a huge task on such a project - support for extracting a method or variable would be great!

I.e select a section of code -> right click -> extract method (keybinding would be great aswell)

lambage:
*bump*

hoffmael:
Advanced refactoring is highly desired for my workflow. This would bring C++ up to speed with other VS code languages that support the refactoring features listed here: https://code.visualstudio.com/docs/editor/refactoring

Doppp:
Bump again.

Audhumbla5:
bump - being able to refactor, extract methods and create definition from declaration is incredibly useful for efficiency.

rockboynton:
bump

fv2018:
bump

GustavoOS:
Now that Visual Studio has Refactoring abilities for other languages, any chances to bring it to C++?

ferdinandhelmer:
extract method, create declaration from implementation and change signature would be a *giant* plus. Great work on the C++ extension in general though, I am really liking it! Keep up the good spirit!

nadya-p:
Moving a function from .h to .cpp (or in the opposite direction) is another example of useful C++ specific refactoring. Some nice C++ refactorings implemented in QtCreator are given [here](https://doc.qt.io/qtcreator/creator-editor-refactoring.html).

n9wxu:
bump

Lennon925:
@sean-mcmanus Any plan about this recently?

sean-mcmanus:
@Lennon925 It's not on our plans yet -- it's our 10th most upvoted feature and we're currently working on some of the 9th and below most upvoted features for our 1.0 release (after 0.27.0).

Lennon925:
> @Lennon925 It's not on our plans yet -- it's our 10th most upvoted feature and we're currently working on some of the 9th and below most upvoted features for our 1.0 release (after 0.27.0).

Thanks and Expect it very much~

TherionIM:
bump

stefanstrat:
bump

FrightRisk:
Same here. Would live to highlight a section of code and have it extracted and turned it into a function.

AnthonyParkssonos:
[...]

PR #11467, opened 2023-09-26
Extract to function

sean-mcmanus:
Add "Extract to function" (https://github.com/microsoft/vscode-cpptools/issues/1162) -- and "Extract to free function" (temporarily disabled), "Extract to member function", and "Expand selection (to enable 'Extract to function')".

Currently requires `C_Cpp.experimentFeatures` set to `true` until the major outstanding issues can be fixed, some subset of the query: https://github.com/microsoft/vscode-cpptools/issues?q=is%3Aopen+is%3Aissue+label%3A%22Feature%3A+Extract+to+function%22 .


Extension/package.json
```diff
@@ -6197,6 +6224,18 @@
                     "title": "%c_cpp.codeActions.refactor.inline.macro.title%",
                     "description": "%c_cpp.codeActions.refactor.inline.macro.description%"
                 }
+            },
+            {
+                "languages": [
+                    "c",
+                    "cpp",
+                    "cude-cpp"
```
bobbrow: typo: `cuda-cpp`

sean-mcmanus: @bobbrow Good catch, but FYI, I have not self-code reviewed this PR yet, and it has not been cleaned up at all. So I wouldn't recommend reviewing it until it's ready. That's why it's in draft still and no code reviewers were added.

bobbrow: Don't worry, I just took a quick look because I'm motivated to get this feature in. I was looking at the outline more than anything and happened to see this.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -214,6 +218,26 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
                 } else {
                     return;
                 }
+            } else if (command.command === "C_Cpp.ExtractToFunction" ||
+                command.command === "C_Cpp.ExtractToMemberFunction") {
+                if (!hasExperimentalFeatures) {
+                    return;
+                }
+                if (command.arguments && command.arguments.length === 1) {
+                    disabledReason = command.arguments[0];
```
bobbrow: Can we add a `disabledReason` on the JSON object instead of hacking `arguments` to mean something other than the arguments that we pass to the code action?

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -223,7 +247,8 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
                     arguments: command.arguments
                 },
                 edit: wsEdit,
-                kind: codeActionKind
+                kind: codeActionKind,
+                disabled: disabledReason ? { reason: disabledReason } : undefined
```
bobbrow: This ternary operator shouldn't be needed.  `disabledReason` is `string | undefined` and you should be able to assign it directly.

sean-mcmanus: I thought it was needed to create an object with a `reason: string` property and not a `reason: string | undefined` property, but I can double check.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: sure, I'll check

Extension/src/LanguageServer/Providers/codeActionProvider.ts
bobbrow: You're right. I misread the code.

Extension/src/LanguageServer/Providers/renameProvider.ts
```diff
@@ -24,6 +24,16 @@ export class RenameProvider implements vscode.RenameProvider {
     }
 
     public async provideRenameEdits(document: vscode.TextDocument, position: vscode.Position, newName: string, _token: vscode.CancellationToken): Promise<vscode.WorkspaceEdit | undefined> {
+        // Bypass the normal rename processing during Extract to function,
```
Colengms, line 27: I've expressed my concerns about leveraging "Rename" for this feature at all.  It doesn't seem to be the right approach, regardless of whether TypeScript uses it (which seems to have been a convenience).  Have you followed up with the VS Code team with an issue to address this with UI that doesn't need to be awkwardly routed through Rename?

Colengms (changes requested):

Extension/src/LanguageServer/Providers/renameProvider.ts
sean-mcmanus: I haven't filed a VS Code issue because I think they're unlikely to change the TypeScript refactor experience -- you're free to file an issue if you want. I think the plan is just to change this to prompt for the name before the extracted code is generated, but in a follow up change for 1.18.1

sean-mcmanus: See the TODO in handleExtractToFunction:
```ts
        // TODO: Show a quick pick to get the name before generating the code.
        // That would allow the formatting to be done without waiting for the rename.
        // Also, it's less error prone and eliminates a class of bugs in which the
        // rename position can be incorrect.
```

Colengms (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -743,6 +748,27 @@ async function onCreateDeclarationOrDefinition(args?: any): Promise<void> {
     return getActiveClient().handleCreateDeclarationOrDefinition(false, args?.range);
 }
 
+async function onExtractToFunction(extractAsGlobal: boolean, extractAsMemberFunction: boolean, sender?: string): Promise<void> {
```
michelleangela: extractAsGlobal, extractAsMemberFunction, ExtractToFunction can be enums instead of checking a bool.

sean-mcmanus: Using 2 bools seems simpler. What enum did you have in mind?

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: done

Extension/src/LanguageServer/extension.ts
michelleangela: The enum names would simply reflect the type of extraction being done. Having an enum would avoid errors of passing in incorrect bool values. So instead of a combination of true or false, you can pass in an enum extract type and would be readable what the caller is intending to do based on the enum value passed in.

michelleangela (dismissed):

Extension/src/LanguageServer/extension.ts
michelleangela: Just a coding suggestion though. It's not necessary.

michelleangela (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -194,6 +194,21 @@ interface WorkspaceFolderParams {
     workspaceFolderUri?: string;
 }
 
+interface SelectionParams {
+    uri: string;
+    range: Range;
+}
+
+export interface VsCodeUriAndRange {
```
Colengms, line 202: Minor nit: Could you move this to the file that uses it?  (It looks like only one file?)   Here, it's in the middle of serializable classes, and the fields suggest this isn't serializable.  Also, client.ts is getting pretty bloated.

Colengms (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: It's only used in client.ts

sean-mcmanus: I think it was intended to possibly be used in other places, but I might get rid of it in my subsequent PR since it involves the the rename data which I'm probably going to get rid of so I can get rid of this type then.

PR #11468, opened 2023-09-26
Force module to node16 for ts-node

fearthecowboy:
This is needed if the system typescript is 5.2.2+

### DIFF
--- a/Extension/.scripts/tsconfig.json
+++ b/Extension/.scripts/tsconfig.json
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "target": "ES2022",
-        "module": "CommonJS",
+        "module": "node16",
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

Extension/.scripts/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "target": "ES2022",
-        "module": "CommonJS",
+        "module": "node16",
```
bobbrow, line 4: When I auto-complete this, it gives me `"Node16"` with a capital N

bobbrow: Oh, but the docs use the lowercase version.  Nevermind.

Issue #11447, opened 2023-09-19
`${env:MY_VAR}` is not handled correctly on macOS and Linux if it returns undefined value


ryuujo1573:
### Environment

- OS and Version: macOS 13
- VS Code Version: any
- C/C++ Extension Version: latest

### Bug Summary and Steps to Reproduce

Bug Summary:

When specifying the `includePath` property of `c_cpp_properties.json` configurations,
I was told that either `${env` or `SOME_VAR}/lib/include/**` is not a path.
<img width="841" alt="Screenshot 2023-09-19 at 12 00 08" src="https://github.com/microsoft/vscode-cpptools/assets/16525512/2ab30a81-2986-42bf-8d05-0f20685d5ea0">

> The hidden path is cwd

Steps to reproduce:
1. add a item of `${env:SOME_VAR}/lib/include/**` to `includePath`.
2. boom

Expected behavior:
[This should work](https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables)

### Configuration and Logs

```shell
{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${workspaceFolder}/**",
                "${env:EMSCRIPTEN_ROOT}/libexec/system/include/**"
            ],
            "defines": [
                "WASM"
            ],
            "macFrameworkPath": [
                "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks"
            ],
            "compilerPath": "/usr/bin/clang",
            "cStandard": "c17",
            "cppStandard": "c++17",
            "intelliSenseMode": "macos-clang-arm64"
        }
    ],
    "version": 4
}
```

```[tasklist]
### Tasks
```

michelleangela:
@ryuujo1573 
To use env variables in the [c_cpp_properties.json file](https://code.visualstudio.com/docs/cpp/c-cpp-properties-schema-reference), they need to be defined as follows:

![image](https://github.com/microsoft/vscode-cpptools/assets/38734287/4e70ca9f-ff55-4b17-9056-02c296429941)

Could you try defining `EMSCRIPTEN_ROOT` in c_cpp_properties.json under the config ` "env": {  }`?

michelleangela:
Update: the parser for resolving variables on macOS and Linux is broken. It works for Windows.

ryuujo1573:
Sorry for late reply 😢 

- [x] `${env:VAR}` parsing is broken on posix os :)

[...]

PR #11480, opened 2023-09-28
Handle undefined environment variable

michelleangela:
Issue https://github.com/microsoft/vscode-cpptools/issues/11447.

This fixes the processing of an environment variable in `CppProperties.resolveAndSplit` and `CppProperties.resolvePath`.

**Problem:**
In `util.resolvedVariable`, `process.env` returns undefined if an environment variable is not found or available, and `util.resolvedVariable` will return the match `${env:VAR}` if not resolved.  The `${env:VAR}` will then get split and appended to other paths incorrectly and processed as if it's the variable was resolved and would lead to incorrect values for other variables or paths.

**Fix:**
If an environment variable is not found or resolved, do not try to further change or process it.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -708,6 +708,10 @@ export class CppProperties {
         this.onSelectionChanged();
     }
 
+    private variableHasEnv(value: string): boolean {
+        return value.includes("env:");
+    }
+
     private resolveDefaults(entries: string[], defaultValue?: string[]): string[] {
         let result: string[] = [];
         entries.forEach(entry => {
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -762,8 +766,14 @@ export class CppProperties {
         }
         paths = this.resolveDefaults(paths, defaultValue);
         paths.forEach(entry => {
-            const entries: string[] = util.resolveVariables(entry, env).split(path.delimiter).map(e => glob ? this.resolvePath(e, false) : e).filter(e => e);
-            resolvedVariables.push(...entries);
+            const resolvedVariable: string = util.resolveVariables(entry, env);
```
bobbrow, line 765: I was thinking we could just resolve undefined environment variables to the empty string in `util.resolveVariables`.  Then you wouldn't have to do any checks about leftover `${env:}` stuff.

bobbrow: I believe most terminals behave that way when resolving env variables that don't exist (resolve to empty string)

michelleangela: I purposely returned the unresolved variable as is so that it will show up as unresolved in the logs.

michelleangela: I'll make the `util.resolveVariables` return empty if an env is unresolved and have the caller save the original variable string.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/configurations.ts
michelleangela: After investigating the other callers and scenarios that use the utility `resolveVariables`, I determined that returning an empty string for an unresolved `env` variable would break other case or scenarios where an empty string is valid such as `""` value for compilerPath or where `""` is returned for `${default}`.

Rather than changing the expected behavior of this utility function and changing the callers and potentially breaking them, it is less risky to have the caller check the returned value for further post processing. I removed the helper function to check if a returned value has "env" or not, as that currently is only needed in two places.

If needed, I propose a different PR to change the design/usage of all utilities that resolve variables to ensure a consistent  /robust behavior for all callers.

sean-mcmanus (approved):

PR #11488, opened 2023-09-29
Add Missing Include Code Action

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10791


Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -97,13 +97,17 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
-            if (command.edit) {
+            if (command.edit && command.command !== "C_Cpp.AddMissingInclude" ) {
                 // Inline macro feature.
                 codeActionKind = CodeActionProvider.inlineMacroKind;
                 wsEdit = new vscode.WorkspaceEdit();
                 wsEdit.replace(document.uri, makeVscodeRange(command.edit.range), command.edit.newText);
                 hasInlineMacro = true;
-            } else if (command.command === "C_Cpp.RemoveAllCodeAnalysisProblems" && command.uri !== undefined) {
+            } else if ( command.edit && command.command === "C_Cpp.AddMissingInclude") {
+                wsEdit = new vscode.WorkspaceEdit();
```
Colengms, line 100: I assume this is necessary because the VS Code API requires a vscode object instance instead of what the LSP message is returning (which used to be acceptable via the LSP protocol).  If so, this is not specific to your code-action.  Any code-action that returns an edit from the native side would require the same conversion.  Could you make this code common for all (not otherwise handled) code-actions, so we can avoid adding unnecessary special cases?

Extension/src/LanguageServer/extension.ts
```diff
@@ -531,6 +532,14 @@ async function onRescanCompilers(sender?: any): Promise<void> {
     return clients.ActiveClient.rescanCompilers(sender);
 }
 
+async function onAddMissingInclude(args?: any): Promise<void> {
+    const sender: any | undefined = util.isString(args?.sender) ? args.sender : args;
+    const properties: Record<string, string> = {
+        sender: util.getSenderType(sender)
+    };
+    telemetry.logLanguageServerEvent('CopyDeclDefn', properties);
```
Colengms: Should `CopyDeclDefn` be changed to something else?

Colengms (dismissed):

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -92,13 +92,18 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
             let title: string = getLocalizedString(command.localizeStringParams);
             let wsEdit: vscode.WorkspaceEdit | undefined;
             let codeActionKind: vscode.CodeActionKind = vscode.CodeActionKind.QuickFix;
-            if (command.edit) {
+            if (command.edit && command.command !== "C_Cpp.AddMissingInclude") {
```
sean-mcmanus: This should be
```
if (command.edit) {
    wsEdit = new vscode.WorkspaceEdit();
    wsEdit.replace(document.uri, makeVscodeRange(command.edit.range), command.edit.newText);
    if (command.command === "edit") {
        // Inline macro feature.
        codeActionKind = CodeActionProvider.inlineMacroKind;
        hasInlineMacro = true;
    }
} else if ...
```

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: I think your comment would be addressed by the change I suggested?

Extension/src/LanguageServer/Providers/codeActionProvider.ts
sean-mcmanus: Also my PR at https://github.com/microsoft/vscode-cpptools/pull/11467 changes `edit` to `workspaceEdits` so I think you'll need to make some minor changes after that gets in.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Yes

sean-mcmanus:
```
    const sender: any | undefined = util.isString(args?.sender) ? args.sender : args;
    const properties: Record<string, string> = {
        sender: util.getSenderType(sender)
    };
```
should be removed too

Extension/src/nativeStrings.json
```diff
@@ -3,6 +3,7 @@
     "include_not_found_in_browse_path": "Include file not found in browse.path.",
     "update_browse_path": "Edit \"browse.path\" setting",
     "add_to_include_path": "Add to \"includePath\": {0}",
+    "add_missing_include_path": "Add {0}",
```
sean-mcmanus: I think this should be
```json
    "add_missing_include_path": {
        "text": "Add '{0}'",
        "hint": "{0} is C++ code to add, such as '#include <string>'"
    },
```
since there could be a potential translation ambiguity with just "Add {0}" (since some languages have verbs that change based on the target of the verb).

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
sean-mcmanus: ...and I'm not sure yet if it should be `{0}` or `'{0}'` or `\"{0}\"`

sean-mcmanus: VS uses `'{0}'` so I'd say go with that.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -97,12 +97,15 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
                 wsEdit.replace(document.uri, makeVscodeRange(command.edit.range), command.edit.newText);
-                hasInlineMacro = true;
-            } else if (command.command === "C_Cpp.RemoveAllCodeAnalysisProblems" && command.uri !== undefined) {
+                if (command.command === "edit") {
+                    // Inline macro feature.
+                    codeActionKind = CodeActionProvider.inlineMacroKind;
+                    hasInlineMacro = true;
+                }
+            }
+            else if (command.command === "C_Cpp.RemoveAllCodeAnalysisProblems" && command.uri !== undefined) {
```
sean-mcmanus: `else if` should go on the previous line, i.e. `} else if (`

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -97,11 +97,16 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
             let wsEdit: vscode.WorkspaceEdit | undefined;
             let codeActionKind: vscode.CodeActionKind = vscode.CodeActionKind.QuickFix;
             if (command.edit) {
-                // Inline macro feature.
-                codeActionKind = CodeActionProvider.inlineMacroKind;
                 wsEdit = new vscode.WorkspaceEdit();
+                if (command.command === 'C_Cpp.AddMissingInclude') {
+                    command.edit.newText = command.edit.newText + "\r\n";
```
sean-mcmanus: I think this should be `\n`. I think only Windows uses \r\n.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -97,11 +98,20 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
             let wsEdit: vscode.WorkspaceEdit | undefined;
             let codeActionKind: vscode.CodeActionKind = vscode.CodeActionKind.QuickFix;
             if (command.edit) {
-                // Inline macro feature.
-                codeActionKind = CodeActionProvider.inlineMacroKind;
                 wsEdit = new vscode.WorkspaceEdit();
+                if (command.command === 'C_Cpp.AddMissingInclude') {
+                    if (os.platform() === 'win32') {
+                        command.edit.newText = command.edit.newText + "\r\n";
```
sean-mcmanus: This conditional isn't necessary. VS Code will convert `\n` to `\r\n` for CLRF files on Windows. We have other code that just adds `\n` such as in `src\LanguageServer\Providers\documentFormattingEditProvider.ts`.

Extension/src/LanguageServer/Providers/codeActionProvider.ts
```diff
@@ -97,11 +98,20 @@ export class CodeActionProvider implements vscode.CodeActionProvider {
...
+                    } else {
+                        command.edit.newText = command.edit.newText + "\n";
```
sean-mcmanus: This should be `+= "\n"`.

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #11493, opened 2023-10-02
Don't send file change notifications for excluded files

bobbrow:
I recently noticed a lot of `cpptools/fileChanged` notifications for files in the `.git` folder.  To reduce the noise, I'm filtering out those files that we're treating as excluded.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1602,6 +1602,7 @@ export class DefaultClient implements Client {
             void this.sendDidChangeSettings();
         }
         const changedSettings: Record<string, string> = this.settingsTracker.getChangedSettings();
+        this.excludePatterns?.invalidate();
 
         await this.ready;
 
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
The processing is already excluded in cpptools -- it's just logging you're seeing. Are you sure you want this change?

sean-mcmanus:
The pattern matching code on the cpptools side doesn't exactly match the one you're using here.

sean-mcmanus:
I think the LSP handler in cpptools could be changed to not log for those messges and logging could be added when the file isn't excluded. That seems like a better fix to me.

bobbrow:
It hadn't occurred to me that the exclusions were happening on the native side because it was logging those messages and that made me think the filtering wasn't happening.  Yeah, perhaps if those messages could be suppressed from that side we wouldn't need this.  We'd have to introduce a new way to filter the logging based on some processing though since right now it just logs messages as events appear.  If you feel strongly that this is the wrong approach I can close the PR.

bobbrow:
Discussed offline. This will be made better in the C++ part of the code.

PR #11503, opened 2023-10-04
Reuse workflow for CI builds

michelleangela:
Note on Xvfb: `GabrielBB/xvfb-action@v1.6` is deprecated, it suggests to use `coactions/setup-xvfb@v1`. Another example to start https://github.com/microsoft/vscode-test/blob/master/sample/azure-pipelines.yml.


.github/workflows/jobs-compile-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Install Dependencies
+        run: yarn install ${{ inputs.yarn-args }}
+        working-directory: Extension
+
+      - name: Compile Sources
+        run: yarn run compile
```
Colengms: @fearthecowboy Are these still the correct command lines to use when building in the lab?

.github/workflows/ci_linux.yml
```diff
@@ -7,44 +7,7 @@ on:
...
-
-      - name: Run unit tests
-        run: yarn test
-        working-directory: Extension
-
-#      # NOTE : We can't run the test that require the native binary files
-#      #        yet -- there will be an update soon that allows the tester to 
-#      #        acquire them on-the-fly
-#      - name: Run simple vscode unit tests
-#        uses: GabrielBB/xvfb-action@v1.6
```
bobbrow, line 41: It's maybe worth keeping a note that running the tests that pop open VS Code need this xvfb action to allow for the UI to appear on mac/linux.  When these tests are re-enabled, you'll need to do it differently for Windows vs. mac/linux.

michelleangela: Added conditionals for these tests that are platform dependent.

sean-mcmanus (dismissed):

Colengms (approved):

PR #11532, opened 2023-10-13
update changelog

bobbrow:


Extension/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 ### New Features
 * Add an 'Extract to Function' (or Member Function) code action after selecting code. [#1162](https://github.com/microsoft/vscode-cpptools/issues/1162)
   * Currently, it's only enabled when `C_Cpp.experimentFeatures` is `true`. Also, 'Extract to Free Function' is disabled.
-* Add the option to install a compiler, accessible through the walkthrough. [PR #11286](https://github.com/microsoft/vscode-cpptools/pull/11286)
+* Add a command to install a compiler, accessible through the walkthrough and Select IntelliSense Configuration experience. [#10525](https://github.com/microsoft/vscode-cpptools/issues/10525)
```
sean-mcmanus: We had "a command" previously, but @AlexandraKemperMS said to use "the option".

`and Select IntelliSense Configuration experience` should be `and the 'Select IntelliSense Configuration' experience`.

What is the point of updating the changelog right now? We normally only do that before a release. Can you update our "internal" changelog instead, and/or update the "live" release page at https://github.com/microsoft/vscode-cpptools/releases/tag/v1.18.0 .

sean-mcmanus (changes requested):

Extension/CHANGELOG.md
bobbrow: The point of updating it is so that it's correct in the next release.

bobbrow: You're right about the option bit though. I forgot that this command is not accessible via the command palette.

Extension/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 ### New Features
 * Add an 'Extract to Function' (or Member Function) code action after selecting code. [#1162](https://github.com/microsoft/vscode-cpptools/issues/1162)
   * Currently, it's only enabled when `C_Cpp.experimentFeatures` is `true`. Also, 'Extract to Free Function' is disabled.
-* Add the option to install a compiler, accessible through the walkthrough. [PR #11286](https://github.com/microsoft/vscode-cpptools/pull/11286)
+* Add an option to install a compiler, accessible through the walkthrough and the Select IntelliSense Configuration experience. [#10525](https://github.com/microsoft/vscode-cpptools/issues/10525)
```
Colengms: We noticed that it isn't accessible in Select IntelliSense Configuration on Windows. Do you want to be more concise about that?  :)

bobbrow: I don't think we want to go into all the specifics in the changelog.  How about I shorten it to "Compiler acquisition improvements" and update the Issue to make sure it covers the full behvior?

sean-mcmanus (approved):
This is fine, but you

Colengms (approved):

sean-mcmanus (approved):

Issue #10613, opened 2023-02-28
Enable an insert mode that never replaces (like TypeScript instead of VS)


sean-mcmanus:
This issue prevails, even with `"editor.suggest.insertMode": "insert", "[cpp]": { "editor.suggest.insertMode": "insert" }`:

![actual](https://user-images.githubusercontent.com/9538837/221896188-806284f2-69f7-4ba7-b872-81c5ac34faa1.gif)

I would expect the editor with the above settings to never ever ever delete characters, e. g. expect the following outcome:

![expected](https://user-images.githubusercontent.com/9538837/221896339-a6c4b8a1-ef0a-4273-99a9-81ab58473bef.png)

_Originally posted by @Longhanks in https://github.com/microsoft/vscode-cpptools/issues/10553#issuecomment-1448365253_

sean-mcmanus:
It looks likes this would be fixed by switching to returning https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertReplaceEdit (instead of the textEdit that is currently returned).

Longhanks:
I think this goes against the spirit of the setting and should be considered a bug, rather than a feature request. Maybe the VS Code devs can chime in to clarify the proper idea of `"insertMode": "insert"`.

sean-mcmanus:
@sean-mcmanus Sure, we can consider it a bug, but I'm not sure when we'll get around to fixing it (although the issue is closed, it could still get fixed if we decide to prioritize it later on).

loikkant:
This is how the issue replicates in the "folder" settings:
![Screencast from 16 08 2023 10 21 12](https://github.com/microsoft/vscode-cpptools/assets/37328774/c948eeef-3959-4651-a659-5cdfde935b0b)

When one selects the non-default mode (insert), it swaps back to the default (replace) as soon as the focused UI element is changed. To me this makes it seem like a bug even further.

I find it extremely annoying that the "insert" mode is not working.

Adnn:
I'm joining from #11353, is there anything we can do to help prioritize this issue?

Adnn:
By the way, as there existed previous comments discussing the intention of the settings and whether it was a bug:

>   // Controls whether words are overwritten when accepting completions. Note that this depends on extensions opting into this feature.
>   //  - insert: Insert suggestion without overwriting text right of the cursor.
>   //  - replace: Insert suggestion and overwrite text right of the cursor.

[...]

PR #11562, opened 2023-10-20
Remove defaults to allow insert/replace edits

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10613


bobbrow:
We are not changing the default mode.

bobbrow (changes requested):
We are not changing the default mode.  It should be "replace."  The fix for this issue in not in TypeScript.

PR #11569, opened 2023-10-23
Notify users if there's a pre-release version available.

### LINKED ISSUE #11550: Restore the prompt to install our pre-release
### Feature Request

Prior to explicit support in VS Code for pre-releases, we used to invite users to try out our "insiders" builds.  When the support was added into VS Code, we removed this prompt and tracked the participation rate.  Over the past year or so we saw the participation rate decline such that we're no longer getting sufficient feature coverage in our pre-releases.

We should add back the prompt to our extension so that we can get more valuable feedback from our pre-releases.

spebl:
Check if there's a pre-release available and that the user isn't already using it, and if so, pop up a prompt to see if they want to install it. A little tough to test this locally, but the parts I can't test I've taken from previous PRs/other MS vscode extensions. Any tips to test the pre-release check (from __metadata) and the install command (took it from a previous related PR) is very welcome, as well as any other comments on the code style/use since it's my first PR in the space.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1165,8 +1165,10 @@ export function UpdateInsidersAccess(): void {
 }
 
 export async function preReleaseCheck(): Promise<void> {
-    // First we need to make sure the user isn't already on a pre-release version.
-    if (util.getCppToolsTargetPopulation() === TargetPopulation.Public) {
+    const displayedPreReleasePrompt: PersistentState<boolean> = new PersistentState<boolean>("CPP.displayedPreReleasePrompt", false);
+
+    // First we need to make sure the user isn't already on a pre-release version and hasn't dismissed this prompt before.
+    if (!displayedPreReleasePrompt.Value && util.getCppToolsTargetPopulation() === TargetPopulation.Public) {
         // Get the info on the latest version from the marketplace to check if there is a pre-release version available.
         const response = await fetch('https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery', {
             method: 'POST',
@@ -1182,6 +1184,7 @@ export async function preReleaseCheck(): Promise<void> {
 
         // If the user isn't on the pre-release version, but one is available, prompt them to install it.
         if (preReleaseAvailable) {
+            displayedPreReleasePrompt.Value = true;
             const message: string = localize("prerelease.message", "There is a new pre-release version of the C/C++ extension, would you like to install it?");
             const yes: string = localize("yes.button", "Yes");
             const no: string = localize("no.button", "No");
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -11,6 +11,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { Range } from 'vscode-languageclient';
 import * as nls from 'vscode-nls';
+import { TargetPopulation } from 'vscode-tas-client';
 import { logAndReturn } from '../Utility/Async/returns';
 import * as util from '../common';
 import { PlatformInformation } from '../platform';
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1162,39 @@ export function UpdateInsidersAccess(): void {
         void vscode.commands.executeCommand("workbench.extensions.installExtension", "ms-vscode.cpptools", { installPreReleaseVersion: true }).then(undefined, logAndReturn.undefined);
     }
 }
+
+export async function PreReleaseCheck(): Promise<void> {
```
Colengms: Could you switch to camel-case for function names (to match the existing style)?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1162,39 @@ export function UpdateInsidersAccess(): void {
...
+    // First we need to make sure the user isn't already on a pre-release version.
+    if (!util.packageJson.__metadata?.preRelease === true) {
+        // Get the info on the latest version from the marketplace to check if there is a pre-release version available.
+        const response = await fetch('https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery', {
```
Colengms, line 1173: Shouldn't we opt people into the prerelease train, even if there isn't one currently available? 
  https://github.com/microsoft/vscode/issues/139528

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1162,39 @@ export function UpdateInsidersAccess(): void {
...
+        const data = await response.json();
+        const preReleaseAvailable = data.results[0].extensions[0].versions[0].properties.some((e: Object) => Object.values(e).includes("Microsoft.VisualStudio.Code.PreRelease"));
+
+        // If the user isn't on the pre-release version, but one is available, prompt them to install it.
+        if (preReleaseAvailable) {
+            const message: string = localize("prerelease.message", "There is a new pre-release version of the C/C++ extension, would you like to install it?");
+            const yes: string = localize("yes.button", "Yes");
+            const no: string = localize("no.button", "No");
+            vscode.window.showInformationMessage(message, yes, no).then((selection) => {
+                switch (selection) {
```
browntarik: I believe that since you're just comparing selection to yes, you can use an if statement. Is there a reason for the default case?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1162,39 @@ export function UpdateInsidersAccess(): void {
...
+    // First we need to make sure the user isn't already on a pre-release version.
+    if (!util.packageJson.__metadata?.preRelease === true) {
+        // Get the info on the latest version from the marketplace to check if there is a pre-release version available.
+        const response = await fetch('https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery', {
+            method: 'POST',
+            headers: {
+                Accept : 'application/json; api-version=3.0-preview',
+                'Content-Type' : 'application/json',
+            }, 
+            body: '{"filters": [{"criteria": [{"filterType": 7, "value": "ms-vscode.cmake-tools"}]}], "flags": 529}'
```
sean-mcmanus: Is this supposed to be ms-vscode.cpptools instead of cmake-tools?

sean-mcmanus (changes requested):

Colengms (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1162,39 @@ export function UpdateInsidersAccess(): void {
         void vscode.commands.executeCommand("workbench.extensions.installExtension", "ms-vscode.cpptools", { installPreReleaseVersion: true }).then(undefined, logAndReturn.undefined);
     }
 }
+
+export async function PreReleaseCheck(): Promise<void> {
+    // First we need to make sure the user isn't already on a pre-release version.
+    if (!util.packageJson.__metadata?.preRelease === true) {
```
bobbrow: We have a function in common.ts called `getCppToolsTargetPopulation()` that we usually use to determine if the user is on the pre-release or not.

Extension/src/LanguageServer/extension.ts
bobbrow: I think it makes more sense to wait until one is available or else it looks like we didn't do anything.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Didn't we have a linter rule on this before?

Extension/src/LanguageServer/extension.ts
spebl: Yep, I copied the example a little too literally.

Extension/src/LanguageServer/extension.ts
spebl: I lean towards what Bob said, that we want to see if there's anything to offer before showing a prompt. But happy to change it if we decide otherwise.

Extension/src/LanguageServer/extension.ts
spebl: Thanks! The terms of "Insiders" and "Pre-Release" are new to me and I was a little confused if they were the same or not, and in which scenario.

bobbrow: Sorry about that.  "Insiders" is the old name, "Pre-release" is the new name, but they represent the same thing.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1163,36 @@ export function UpdateInsidersAccess(): void {
...
+            body: '{"filters": [{"criteria": [{"filterType": 7, "value": "ms-vscode.cpptools"}]}], "flags": 529}'
+        });
+
+        const data = await response.json();
+        const preReleaseAvailable = data.results[0].extensions[0].versions[0].properties.some((e: object) => Object.values(e).includes("Microsoft.VisualStudio.Code.PreRelease"));
+
+        // If the user isn't on the pre-release version, but one is available, prompt them to install it.
+        if (preReleaseAvailable) {
+            displayedPreReleasePrompt.Value = true;
+            const message: string = localize("prerelease.message", "There is a new pre-release version of the C/C++ extension, would you like to install it?");
```
Colengms: This is a "comma splice" (two independent clauses).  Either:
a. Separate into two sentences.
b. Use a semicolon.
c. Use a conjunction. (less preferred).

My opinion would be to use 2 sentences.

Colengms (changes requested):

Extension/src/LanguageServer/extension.ts
spebl: Updated the wording to use two sentences. Also open to completely rewording the prompt, I just slightly modified the old "switch to marketplace for pre-release" prompt to make sense in the general case, but if there's a more specific message we want to get a across I'm open to it.

bobbrow (dismissed):

Colengms (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1162,3 +1163,36 @@ export function UpdateInsidersAccess(): void {
...
+            body: '{"filters": [{"criteria": [{"filterType": 7, "value": "ms-vscode.cpptools"}]}], "flags": 529}'
+        });
+
+        const data = await response.json();
+        const preReleaseAvailable = data.results[0].extensions[0].versions[0].properties.some((e: object) => Object.values(e).includes("Microsoft.VisualStudio.Code.PreRelease"));
+
+        // If the user isn't on the pre-release version, but one is available, prompt them to install it.
+        if (preReleaseAvailable) {
+            displayedPreReleasePrompt.Value = true;
+            const message: string = localize("prerelease.message", "There is a new pre-release version of the C/C++ extension. Would you like to install it?");
```
sean-mcmanus: It seems like "There is a new pre-release version of the C/C++ extension. Would you like to install it?" should be something like "A pre-release version of the C/C++ extension is available. Would you like to switch to using the pre-release version?". Otherwise, it sounds like it's asking if the user wants to just install the one particular pre-release version that is new.

spebl: I like that new wording. I'll swap it out. Thanks!

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I'm fine with shortening the 2nd sentence to using "it" too -- I didn't notice that issue until after.

spebl: done!

sean-mcmanus (approved):

Colengms (approved):

PR #11580, opened 2023-10-24
Update walkthrough text for the customize debugging step.

### LINKED ISSUE #11320: Walkthrough Improvement: Updated wording for "Customize Debugging" step
In the "Customize Debugging" step of the walkthrough, we currently state: 

"You can customize your debug configuration (e.g. to pass arguments to your program at run time) by selecting "Add Debug Configuration" to the right of the play button. The custom debug configuration is saved to your project's launch.json file" 

However, at this point in the walkthrough, a user has only opened a cpp file as a single file. Seeing the gear icon to add a debug configuration is only available if they have opened their code in a folder. We need to instruct users in this step that they need to open their code in a folder. 

Proposed new text for the "Customize debugging" step: 

"To customize your debug configuration, select the Explorer in the activity bar and open a folder that includes your C++ file. Open the C++ file, and select "Add Debug Configuration" to the right of the play button. The new debug configuration is saved to your project's launch.json file."

spebl:
Implementation of the proposed text change mentioned in #11320.


Extension/package.nls.json
```diff
@@ -991,9 +991,9 @@
...
-    "c_cpp.walkthrough.customize.debugging.windows.description": "You can customize your debug configuration (e.g. to pass arguments to your program at run time) by selecting \"Add Debug Configuration\" to the right of the play button. The custom debug configuration is saved in your project's launch.json file. \n[Learn More](https://code.visualstudio.com/docs/cpp/config-clang-mac#_debug-helloworldcpp)",
+    "c_cpp.walkthrough.customize.debugging.mac.description": "To customize your debug configuration, select the Explorer in the activity bar and open a folder that includes your C++ file. Open the C++ file, and select \"Add Debug Configuration\" to the right of the play button. The new debug configuration is saved to your project's launch.json file. \n[Learn More](https://code.visualstudio.com/docs/cpp/config-linux#_debug-helloworldcpp)",
```
sean-mcmanus, line 994: I think `Explorer` should be `Explorer view` since that is what VS Code seems to refer to it as in it's documentation.

AlexandraKemperMS: It double checked the [VS Code docs](https://code.visualstudio.com/docs/getstarted/userinterface#_explorer) - they also refer to it as the Explorer if you are clicking on the icon

sean-mcmanus: Ah, ok

sean-mcmanus (approved):

sean-mcmanus (changes requested):
The previous text didn't mention "C++ file" -- what about C users who are debugging? Should we use C or C++ file?

sean-mcmanus:
> The previous text didn't mention "C++ file" -- what about C users who are debugging? Should we use C or C++ file?

Oh, I see, it's in the context of the C++ focused walkthrough.

sean-mcmanus (approved):

Issue #11619, opened 2023-11-03
'Extract to function' has the wrong code selected when space is selected


sean-mcmanus:
Use code

```cpp
int main()
{
    return 0; // extract to function
}
```

The result has the wrong selection:

<img width="179" alt="image" src="https://github.com/microsoft/vscode-cpptools/assets/19859882/995427d1-656c-4d2c-8f52-3d28305fa50d">

Previously formatting would fail: https://github.com/microsoft/vscode-cpptools/issues/11459 -- I thought I had tested/fixed the selection too.

PR #11621, opened 2023-11-03
Fix 'Extract to function' formatting/selection bugs.

### LINKED ISSUE #11614: 'Extract to function' doesn't correctly format the header declaration
It invokes the formatter on the header, but something is causing the formatting to not work as expected. I don't think this repros with VS (but I haven't checked). This repros with 1.18.0 and 1.18.1.

Repro is like

test.cpp
```cpp
namespace foo {
int main()
{
    return 0; // extract to function on this line
}
```

test.h
```
namespace foo
{
    int otherFunc();
}
```

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/11614 (incorrect header formatting) and https://github.com/microsoft/vscode-cpptools/issues/11619 (incorrect selection) and fixes other unreported issues (e.g. duplicate formatting calls and unnecessary document switching).

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -3497,6 +3497,9 @@ export class DefaultClient implements Client {
         let workspaceEdits: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
         let replaceEditRange: vscode.Range | undefined;
         let hasProcessedReplace: boolean = false;
+        // NOTE: References to source/header are in reference to the more common case when it's
+        // invoked on the source file (alternatively named the first file). When invoked on the header file,
+        // the header file operates as if it were the source file (isSourceFile stays true).
         const sourceFormatUriAndRanges: VsCodeUriAndRange[] = [];
         const headerFormatUriAndRanges: VsCodeUriAndRange[] = [];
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -3497,7 +3497,8 @@ export class DefaultClient implements Client {
         let workspaceEdits: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
         let replaceEditRange: vscode.Range | undefined;
         let hasProcessedReplace: boolean = false;
-        const formatUriAndRanges: VsCodeUriAndRange[] = [];
+        const sourceFormatUriAndRanges: VsCodeUriAndRange[] = [];
```
bobbrow, line 3503: Maybe add a comment here explaining that this is not necessarily a "source" file.  It could be a header if the codeAction originated from there.

bobbrow (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: @bobbrow I added a comment.

bobbrow (approved):

PR #11702, opened 2023-11-17
Fix the pre-release prompt.

sean-mcmanus:


sean-mcmanus:
@bobbrow Should we get this into 1.18.5? Otherwise, users of 1.18.x prior to November 17th won't get the pre-release prompt.

Issue #11422, opened 2023-09-09
Build command ignores puncutations in source path


shashwatrathod:
### Environment

- OS and Version: Windows 11 with WSL 2 running Ubuntu 20.04.4
- VS Code Version: 1.81.1
- C/C++ Extension Version: 1.17.5
- If using SSH remote, specify OS of remote machine: N/A

### Bug Summary and Steps to Reproduce

Bug Summary:
If the source path contains punctuation, the extension's build command ignores the punctuation in that path. Since the entered file path doesn't exist a 'No such file or directory' error is thrown. 

Steps to reproduce:
1. Create a new directory `Sept'23` (Notice the `'`).
2. Create a new `.cpp` file inside this directory (`helloworld.cpp`)
3. Try to trigger the build using the 'Run' button on the top-right corner of VSCode's nav bar.

The build will fail and you should see a 'No such file or directory' error in the terminal.

Expected behavior:
The build command uses the source path `/path/to/Sept23/helloworld.cpp` instead of `/path/to/Sept'23/helloworld.cpp` due to which it fails. The punctuation should have been taken into account.

### Configuration and Logs

```shell
{
  "configurations": [
    {
      "name": "Linux",
      "includePath": ["${workspaceFolder}/**"],
      "defines": [],
      "compilerPath": "/usr/bin/g++",
      "cStandard": "c17",
      "cppStandard": "gnu++14",
      "intelliSenseMode": "linux-gcc-x64"
    }
  ],
  "version": 4
}
```

### Additional context

Here is part of the terminal output:
```
Starting build...
/usr/bin/g++ -fdiagnostics-color=always -g /mnt/d/Shashwat/Fall'23/main.cpp -o /mnt/d/Shashwat/Fall'23/main
g++: error: /mnt/d/Shashwat/Fall'23/main.cpp -o /mnt/d/Shashwat/Fall'23/main: No such file or directory
g++: fatal error: no input files
compilation terminated.

Build finished with error(s).

 *  The terminal process failed to launch (exit code: -1). 
 *  Terminal will be reused by tasks, press any key to close it. 
 ```

Colengms:
Hi @shashwatrathod .  Thanks for reporting this.  It looks like this path is not properly escaped to account for shell command-line special characters before being used in a shell command line.

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.19.2

Colengms:
[...]

PR #11734, opened 2023-11-22
Fix shell quoting for command line arguments

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11422

### DIFF
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -172,7 +172,12 @@ export class CppBuildTaskProvider implements TaskProvider {
         const isClang: boolean = !isCl && compilerPathBase.toLowerCase().includes("clang");
         // Double-quote the command if needed.
         let resolvedcompilerPath: string = isCl ? compilerPathBase : compilerPath;
-        resolvedcompilerPath = util.quoteArgument(resolvedcompilerPath);
+        if (os.platform() === 'win32') {
+            resolvedcompilerPath = util.quoteArgumentWindows(resolvedcompilerPath);
+        }
+        else {
+            resolvedcompilerPath = util.quoteArgumentUnix(resolvedcompilerPath);
+        }
 
         if (!definition) {
             const isWindows: boolean = os.platform() === 'win32';
@@ -388,7 +393,13 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
         util.createDirIfNotExistsSync(exePath);
 
         this.args.forEach((value, index) => {
-            value = util.quoteArgument(util.resolveVariables(value));
+            if (os.platform() === 'win32') {
+                value = util.quoteArgumentWindows(util.resolveVariables(value));
+            } 
+            else 
+            {
+                value = util.quoteArgumentUnix(util.resolveVariables(value));
+            }
             activeCommand = activeCommand + " " + value;
             this.args[index] = value;
         });
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -170,11 +170,9 @@ export class CppBuildTaskProvider implements TaskProvider {
         const compilerPathBase: string = path.basename(compilerPath);
         const isCl: boolean = compilerPathBase.toLowerCase() === "cl.exe";
         const isClang: boolean = !isCl && compilerPathBase.toLowerCase().includes("clang");
-        // Double-quote the command if it is not already double-quoted.
+        // Double-quote the command if needed.
         let resolvedcompilerPath: string = isCl ? compilerPathBase : compilerPath;
-        if (resolvedcompilerPath && !resolvedcompilerPath.startsWith("\"") && resolvedcompilerPath.includes(" ")) {
-            resolvedcompilerPath = "\"" + resolvedcompilerPath + "\"";
-        }
+        resolvedcompilerPath = util.quoteArgument(resolvedcompilerPath);
 
         if (!definition) {
             const isWindows: boolean = os.platform() === 'win32';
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -435,6 +436,10 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
 
         let child: cp.ChildProcess | undefined;
         try {
+            // Add shell quoting for command line arguments.
```
Colengms: If the path to the compiler contains a special character that would need to be escaped, is that handled correctly?

A few lines up, there is: ```command = `cmd /c chcp 65001>nul && ${command}`;```.  That is Windows-only, but it is adding some shell-dependent syntax to `command`.  That whole command is expected (by `cp.spawn` when `options.shell` is true) to contain shell quoting/escaping, if needed.  Can you confirm if that original `command` (its value before that line) has been quoted/escaped properly?  Whether a command line contains shell quoting/escaping or not, are mutual-exclusive concepts, so it's important that we're always explicit about it.  Someone may have blindly combined that bit of shell command line without checking if the existing command was also quoted/escaped.

Colengms (changes requested):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
browntarik: I've refactored the shell quote logic to reflect a well documented approach: https://learn.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way

This approach works on all platforms (Windows, Mac and Linux) correctly.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -170,10 +170,13 @@ export class CppBuildTaskProvider implements TaskProvider {
         const compilerPathBase: string = path.basename(compilerPath);
         const isCl: boolean = compilerPathBase.toLowerCase() === "cl.exe";
         const isClang: boolean = !isCl && compilerPathBase.toLowerCase().includes("clang");
-        // Double-quote the command if it is not already double-quoted.
+        // Double-quote the command if needed.
         let resolvedcompilerPath: string = isCl ? compilerPathBase : compilerPath;
-        if (resolvedcompilerPath && !resolvedcompilerPath.startsWith("\"") && resolvedcompilerPath.includes(" ")) {
-            resolvedcompilerPath = "\"" + resolvedcompilerPath + "\"";
+        if (os.platform() === 'win32') {
+            resolvedcompilerPath = util.quoteArgumentWindows(resolvedcompilerPath);
```
sean-mcmanus: Would it be better to make the API be util.quoteArgument -- i.e. check the platform in the implementation? It seems like it would to me.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
browntarik: Yep, this has been updated

Extension/src/common.ts
```diff
@@ -1359,27 +1359,64 @@ export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Pr
...
+        }
+    }
+
+    // Close the quoted argument with a single quote and return it
+    quotedArgument += '\'';
+
+    return quotedArgument;
+}
+
+export function quoteArgumentWindows(argument: string): string {
```
Colengms: Could you merge these two into the same function, which internally checks if running on Windows or not?  That way, all callers would not have to do their own checks to determine which one to invoke.

Colengms (changes requested):

Extension/src/common.ts
sean-mcmanus: Yeah, that's what I said yesterday :)

Extension/src/LanguageServer/cppBuildTaskProvider.ts
sean-mcmanus: Where's the update?

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -390,7 +393,13 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
         util.createDirIfNotExistsSync(exePath);
 
         this.args.forEach((value, index) => {
-            value = util.normalizeArg(util.resolveVariables(value));
+            if (os.platform() === 'win32') {
+                value = util.quoteArgumentWindows(util.resolveVariables(value));
+            }
+            else
+            {
```
sean-mcmanus: { should go on the previous line (maybe run the formatter?)

Extension/src/common.ts
```diff
@@ -1359,27 +1359,64 @@ export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Pr
         return promiseBuilder(nextItem);
     }, Promise.resolve());
 }
+export function quoteArgumentUnix(argument: string): string {
```
sean-mcmanus: We generally put a (blank) newline before the start of a function.

Extension/src/common.ts
bobbrow: Or keep the individual ones, but don't export them and have an exported function 'quoteArgument' that forwards to the correct helper function.

Extension/src/common.ts
```diff
@@ -1359,27 +1359,64 @@ export function sequentialResolve<T>(items: T[], promiseBuilder: (item: T) => Pr
...
-    const unescapedSpaces: string | undefined = arg.split('').find((char, index) => index > 0 && char === " " && arg[index - 1] !== "\\");
-    if (!unescapedSpaces && !process.platform.includes("win")) {
-        return arg;
-    } else if (arg.includes(" ")) {
-        arg = arg.replace(/\\\s/g, " ");
-        return "\"" + arg + "\"";
-    } else {
-        return arg;
+    // Initialize a variable for the quoted argument, starting with a single quote
+    let quotedArgument = '\'';
```
sean-mcmanus: All the `\'` (in this function) can be replaced by `'` if you use `""` quotes (and don't use `""` if the string has `"` or you need to use `\"`, such as in the code in the other function below)

sean-mcmanus (approved):

Colengms (approved):

Issue #7759, opened 2021-07-01
Generate IntelliSense results progressively


Colengms:
Currently, the C/C++ Extension will process an entire translation unit before returning results (squiggles) for a file.  Those results are monolithic.  When a new set is generated, they replace the prior set.

My understanding is that VS will generate IntelliSense results progressively.  It will return an initial block of results after having parsed up until the end of the currently visible region of the file, in order to populating the visible area as quickly as possible.  Then it will progressively populate the rest of the file, potentially in multiple sections, allowing squiggles to show up more or less as they are determined.

This issue is tracking doing something similar in the C/C++ Extension, instead of always processing an entire file at a time.

Other features, such as "Creating Definition from Declaration (and vice versa)", are implemented in a similar progressive manner in VS.  We could also implement that feature in the extension by processing the entire file, but it may be worthwhile to establish some patterns for returning progressive results that could be used by both features.

bobbrow:
This potentially addresses #3626

PR #11735, opened 2023-11-22
Enable progressive IntelliSense updates

Colengms:
* Enables progressive IntelliSense updating ( https://github.com/microsoft/vscode-cpptools/issues/7759 ) 
* Updates all visible documents/editors instead of just the active document/editor.
* Made semantic token and inlay hint providers simpler. Their purposes are now only to trigger VS Code requests for these as the native side pushes them down to the TypeScript side.
* Removes delaying of `didOpen`, which previously was used to defer TU creation until a file was actually made visible.  Instead, the native side will only perform a TU update once it receives a separate visibility notification.
* Fixed some issues with messages being processed out of order.


Extension/src/LanguageServer/client.ts
```diff
@@ -2696,8 +2781,9 @@ export class DefaultClient implements Client {
     }
 
     public async selectionChanged(selection: Range): Promise<void> {
-        await this.ready;
-        return this.languageClient.sendNotification(TextEditorSelectionChangeNotification, selection);
+        console.log("selectionChanged\n" + JSON.stringify(selection, null, 4));
```
sean-mcmanus: Can you remove this?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -278,7 +280,16 @@ async function onDidChangeSettings(event: vscode.ConfigurationChangeEvent): Prom
 
 let noActiveEditorTimeout: NodeJS.Timeout | undefined;
 
-export function onDidChangeActiveTextEditor(editor?: vscode.TextEditor): void {
+async function onDidChangeTextEditorVisibleRanges(event: vscode.TextEditorVisibleRangesChangeEvent): Promise<void> {
+    if (util.isCpp(event.textEditor.document)) {
+        console.log("IN onDidChangeTextEditorVisibleRanges");
```
sean-mcmanus: Can you remove this too?

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #11748, opened 2023-12-01
Fix call hierarchy calls to being cancelled when it shouldn't be.

### LINKED ISSUE #11747: Call Hierarchy sometimes does not expand Callers Of properly when multiple are expanded quickly
We've not be able to narrow down the reason yet, but Callers Of randomly does not work.

In cases where it fails to expand an item properly, we do not see the corresponding `cpptools/callHierarchyCallsTo` in the debug log output.  That would seem to suggest the issue is on the TypeScript side.

This may be a VS Code issue, as the repro seems to correspond with not getting a call to the `CallHierarchyProvider`.  But we should rule out whether that might be due to some incorrect information we've provided.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/11747


sean-mcmanus:
There's still a bug when 2 calls to are in progress and the Cancel is used. They both get canceled, but the the progress bar for the 2nd persists and gets stuck.

sean-mcmanus:
This needs more investigation so I'm abandoning it for now.

PR #11773, opened 2023-12-12
Fix broken editor.wordBasedSuggestions setting.

sean-mcmanus:
The setting configuration is "broken" with the latest VS Code 1.85.0.

image

Without the fix:
image


bobbrow (approved):

bobbrow:
Do we need to update our VS Code min version dependency for this?

sean-mcmanus:
> Do we need to update our VS Code min version dependency for this?

Yeah, seems like we may need to -- I asked a question at https://github.com/microsoft/vscode/pull/197690 .

sean-mcmanus:
@bobbrow This causes the bug to be fixed with the latest VS Code but it breaks older VS Codes -- it seems like it's preferable to prioritize making the latest VS Code work correctly until VS Code can fix the regression?

bobbrow:
> @bobbrow This causes the bug to be fixed with the latest VS Code but it breaks older VS Codes -- it seems like it's preferable to prioritize making the latest VS Code work correctly until VS Code can fix the regression?

Yes, I would agree.  The setting doesn't work in latest. Let's check it in.

bobbrow (approved):

PR #11785, opened 2023-12-15
Perf-boost - fast query symbols

fearthecowboy:
Typescript side of the perf-boost for query symbols 
(required to let the native code work)


Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
```diff
@@ -14,8 +15,13 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
...
 
     public async provideWorkspaceSymbols(query: string, token: vscode.CancellationToken): Promise<vscode.SymbolInformation[]> {
+        if (!query) {
+            return [];
+        }
+
         const params: WorkspaceSymbolParams = {
-            query: query
+            query: query,
+            experimentEnabled: await isExperimentEnabled('CppTools1')
```
bobbrow, line 25: Just checking, is this the correct capitalization? Experiments are case-sensitive and I assumed we would call them 'cpptools1', 'cpptools2', etc.

Extension/package.json
```diff
@@ -458,11 +458,11 @@
                     },
                     "C_Cpp.intelliSenseUpdateDelay": {
                         "type": "number",
-                        "default": 2000,
+                        "default": 0,
```
bobbrow: Please revert changes to the update delay

Extension/src/LanguageServer/client.ts
```diff
@@ -1465,7 +1466,7 @@ export class DefaultClient implements Client {
             workspaceParsingPriority: workspaceSettings.workspaceParsingPriority,
             workspaceSymbols: workspaceSettings.workspaceSymbols,
             simplifyStructuredComments: workspaceSettings.simplifyStructuredComments,
-            intelliSenseUpdateDelay: workspaceSettings.intelliSenseUpdateDelay,
+            intelliSenseUpdateDelay: workspaceSettings.intelliSenseUpdateDelay || (4 - Math.min(4, os.cpus().length)) * 250 + 250,
```
bobbrow: Please revert changes to the update delay.

Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
fearthecowboy: Yes, the casing is correct (Alex was *very* specific 😄 )

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1105,7 +1105,7 @@ export class CppProperties {
         }
     }
 
-    private compileCommandsFileWatcherTimer?: NodeJS.Timer;
+    private compileCommandsFileWatcherTimer?: NodeJS.Timeout;
```
fearthecowboy, line 1111: These were actually mis-typed originally - and now that VSCode's electron is at node 18.15.0 - the interface for `Timer` and `Timeout` are slightly different, and intervals are declared as  `Timeout` and not `Timer`

Extension/package.json
```diff
@@ -6322,7 +6322,7 @@
         "@typescript-eslint/eslint-plugin": "^6.1.0",
         "@typescript-eslint/parser": "^6.1.0",
         "eslint-plugin-header": "^3.1.1",
-        "@vscode/test-electron": "^2.3.3",
+        "@vscode/test-electron": "^2.3.8",
```
fearthecowboy, line 6325: A bug in the previous version was breaking the download of the vscode test packages.

Extension/package.json
```diff
@@ -6311,7 +6311,7 @@
         "@types/minimatch": "^3.0.5",
         "@types/mkdirp": "^0.5.2",
         "@types/mocha": "^10.0.1",
-        "@types/node": "^18.11.9",
+        "@types/node": "^18.15.0",
```
fearthecowboy, line 6314: VSCode has updated to node 18.15.0 bindings

bobbrow: Does this work with the minimum version of VS Code that we support? I don't remember if these need to be in sync or not.

    "engines": {
        "vscode": "^1.67.0"
    },

fearthecowboy: Yes, in that `18.11.9` was technically newer than the node specified in vscode `1.67` too, but we don't use APIs that aren't available.  (and older node bindings break with the newer TS builds)

The difference between `18.11.9` and `18.15.0` typings is very minor (and when we compile, TS typings are gone at runtime regardless) 

So, we're no less compatible than before, but we do get to compile with greater conformance to the API now.

bobbrow (dismissed):

Extension/cpp.hint
```diff
@@ -1,2 +1,2 @@
-// https://go.microsoft.com/fwlink/?linkid=865984
-#pragma inject_default_hints
+// https://go.microsoft.com/fwlink/?linkid=865984
```
sean-mcmanus: What is the change here? Or why is there a change here.

sean-mcmanus (dismissed):
I made a comment/question

Extension/cpp.hint
sean-mcmanus: It seems like it shouldn't be changed. The file is a copy of the "source" that is in the closed source repo so it shouldn't be different from that.

Colengms (dismissed):

Extension/cpp.hint
fearthecowboy: Uh... this probably wasn't supposed to be here.

fearthecowboy: grrr. one second...

fearthecowboy: Fixed it

sean-mcmanus (approved):

PR #11845, opened 2024-01-08
fix german translation for RunCodeAnalysis

Sir2B:
German translation for running code analysis was not easy to understand


sean-mcmanus (approved):

AlexandraKemperMS (approved):
Sieht gut aus. Danke, dass Sie das bemerkt haben!

@sean-mcmanus  These translations are correct and an improvement from before.

bobbrow:
Can someone open a bug for our localization team? These files are auto-generated and this change won't persist until it's updated on their side.

sean-mcmanus:
@Sir2B Your fix is available with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.19.2

PR #11858, opened 2024-01-12
Log cpptools crash call stacks.

sean-mcmanus:
Add cpptools crash callstack reporting, currently only works on Mac (Linux is pending). cpptools-srv crashes are also a TODO.


Extension/src/LanguageServer/extension.ts
```diff
@@ -959,17 +960,64 @@ function reportMacCrashes(): void {
...
-let previousMacCrashData: string;
-let previousMacCrashCount: number = 0;
+export function watchForCrashes(crashDirectory: string): void {
+    if (process.platform !== "win32") {
+        prevCrashFile = "";
+        fs.stat(crashDirectory, (err) => {
+            const crashObject: Record<string, string> = {};
+            if (err?.code) {
+                // If the directory isn't there, we have a problem...
+                crashObject["fs.stat: err.code"] = err.code;
```
bobbrow: I'd prefer to have only alpha characters in event property names.  Can you call it "errorCode" instead?
This appears to be the same way the old mac crashes were reported but I can attest that I never queried this property.

Extension/src/LanguageServer/extension.ts
```diff
@@ -959,17 +960,64 @@ function reportMacCrashes(): void {
...
+                telemetry.logLanguageServerEvent("MacCrash", crashObject, undefined);
```
bobbrow: It shouldn't be necessary to pass in `undefined` for the optional 3rd parameter.

Extension/src/LanguageServer/extension.ts
```diff
@@ -959,17 +960,64 @@ function reportMacCrashes(): void {
...
+                    }
+                    // Wait 5 seconds to allow time for the crash log to finish being written.
+                    setTimeout(() => {
+                        fs.readFile(path.resolve(crashDirectory, filename), 'utf8', (err, data) => {
+                            void handleCrashFileRead(crashDirectory, filename, err, data);
+                        });
+                    }, 5000);
+                });
+            } catch (e) {
+                // The file watcher limit is hit (may not be possible on Mac, but just in case).
```
bobbrow, line 996: Do we need to watch for crashes, or can we just check for the existence of a crash dump file when initializing a new language server?  I thought that was how the old mac crashes worked, but it looks like I was wrong.  However, since we control/own this file and are actively deleting it after it's processed, maybe we don't need the watcher?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1062,6 +1110,37 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
...
+    }
+
+    if (data.length > 8192) { // The API has an 8k limit.
+        data = data.substring(0, 8189) + "...";
+    }
+
+    logCppCrashTelemetry(data);
+
+    await util.deleteFile(path.resolve(crashDirectory, crashFile));
+    void util.deleteDirectory(crashDirectory);
```
bobbrow: I'm not sure if we need to delete the directory if it's in the `storagePath`. But if we do, you need to `.catch()` on this in case it rejects the promise.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1062,6 +1110,37 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
...
+        data += line + "\n";
+    }
+
+    if (data.length > 8192) { // The API has an 8k limit.
+        data = data.substring(0, 8189) + "...";
+    }
+
+    logCppCrashTelemetry(data);
+
+    await util.deleteFile(path.resolve(crashDirectory, crashFile));
```
bobbrow: To be safe, we should `.catch()` on this call in case it rejects.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1062,6 +1110,37 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
     logMacCrashTelemetry(data);
 }
 
+async function handleCrashFileRead(crashDirectory: string, crashFile: string, err: NodeJS.ErrnoException | undefined | null, data: string): Promise<void> {
```
bobbrow, line 1113: `crashFile` is technically correct, but when reading this function I'm thinking of the file containing the printout of the callstacks, not the name of the executable that crashed (cpptools[-srv]).  Especially because `crashDirectory` relates to the callstacks. Could you pick a different name for this?  `crashingProcess` perhaps?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1062,6 +1110,37 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
...
+    }
+
+    const lines: string[] = data.split("\n");
+    data = crashFile + "\n";
+    const filtPath: string | null = which.sync("c++filt", { nothrow: true });
+    for (let line of lines)
+    {
+        line = line.trim();
+        if (filtPath && line.startsWith("_")) {
+            line = (await util.spawnChildProcess(filtPath, [line])).output;
```
bobbrow: If `spawnChildProcess` can fail, we should make sure that we `.catch()` the error and exit gracefully.

Extension/src/LanguageServer/extension.ts
```diff
@@ -959,17 +960,64 @@ function reportMacCrashes(): void {
...
     const crashObject: Record<string, string> = {};
     const crashCountObject: Record<string, number> = {};
     crashObject.CrashingThreadCallStack = data;
-    previousMacCrashCount = data === previousMacCrashData ? previousMacCrashCount + 1 : 0;
-    previousMacCrashData = data;
-    crashCountObject.CrashCount = previousMacCrashCount;
-    telemetry.logLanguageServerEvent("MacCrash", crashObject, crashCountObject);
+    previousCrashCount = data === previousCrashData ? previousCrashCount + 1 : 0;
+    previousCrashData = data;
+    crashCountObject.CrashCount = previousCrashCount;
```
bobbrow: So does this log 0 for CrashCount the first time?  I don't think I use this field in any of my queries today, but I didn't expect that.

bobbrow (commented):
I know this is in draft, but I'm really excited for it and had to look.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: It's not in the storage path, it's in a tmp path

Extension/src/LanguageServer/extension.ts
sean-mcmanus: It is the name of the file containing the printout of the call stacks, e.g. cpptools.txt

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I added a + 1

Extension/src/LanguageServer/extension.ts
sean-mcmanus: The plan is to add support for cpptools-srv crashes so I think the file watcher allows us to detect/send the crashes sooner and it ensures we know where to look in the temp location returned from initialize.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: done

Extension/src/LanguageServer/extension.ts
sean-mcmanus: done

Extension/src/LanguageServer/extension.ts
sean-mcmanus: removed

Extension/src/LanguageServer/extension.ts
bobbrow: Ok, yeah, we'll want the file watcher to handle cpptools-srv crashes.

Extension/src/LanguageServer/extension.ts
bobbrow: You mentioned future support for cpptools-srv.  If that crashes really fast a few times in a row will we get unique files for each crash or do we run the risk of corruption if the file is overwritten while we're trying to read it?

bobbrow: > It is the name of the file containing the printout of the call stacks, e.g. cpptools.txt

Ok, got it. I was thinking of the old telemetry that logs the process name first, I think.

sean-mcmanus: Yeah, I figured it was better to log cpptools.txt to make it easier to differentiate from cpptools-srv.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1062,6 +1110,76 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
...
 
+async function handleCrashFileRead(crashDirectory: string, crashFile: string, err: NodeJS.ErrnoException | undefined | null, data: string): Promise<void> {
+    if (err) {
+        if (err.code === "ENOENT") {
+            return; // ignore known issue
+        }
+        return logCppCrashTelemetry("readFile: " + err.code);
+    }
+
+    console.log("processing crash log");
```
sean-mcmanus: I'm not sure about this console.log yet...should I change it?

bobbrow: `console.log` doesn't show up in the Output window, so you are welcome to leave it like this if you want.

Extension/src/LanguageServer/extension.ts
```diff
@@ -922,8 +923,8 @@ function reportMacCrashes(): void {
             const crashObject: Record<string, string> = {};
             if (err?.code) {
                 // If the directory isn't there, we have a problem...
-                crashObject["fs.stat: err.code"] = err.code;
-                telemetry.logLanguageServerEvent("MacCrash", crashObject, undefined);
+                crashObject["stat: errCode"] = err.code;
```
bobbrow: Sorry if I was unclear before.  Please change `"stat: errCode"` to `"errorCode"`.  I don't like to have spaces or special characters in telemetry property names.

Extension/src/LanguageServer/extension.ts
```diff
@@ -959,17 +960,64 @@ function reportMacCrashes(): void {
...
-let previousMacCrashData: string;
-let previousMacCrashCount: number = 0;
+export function watchForCrashes(crashDirectory: string): void {
+    if (process.platform !== "win32") {
+        prevCrashFile = "";
+        fs.stat(crashDirectory, (err) => {
+            const crashObject: Record<string, string> = {};
+            if (err?.code) {
+                // If the directory isn't there, we have a problem...
+                crashObject["stat: errCode"] = err.code;
```
bobbrow: Please change from `"stat: errCode"` to `"errorCode"`

sean-mcmanus:
@bobbrow FYI, I didn't test my last commit yet.

bobbrow (dismissed):

bobbrow (dismissed):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Oh -- it does show up in the "Server" output pane, but that seems fine to me. I removed "processing crash log" though since that was not useful logging.

sean-mcmanus:
> @bobbrow FYI, I didn't test my last commit yet (logging changes).

I tested it -- fixed a bug.

Colengms (approved):

gdamiaN538 (commented):
@

Issue #11862, opened 2024-01-15
Include path settings don't trim white space


pwintz:
Type: <b>Bug</b>

I have files I want C++ to include in the directory /usr/local/include. 
When I added "/usr/local/include " (with a space at the end) to the "Include path" configuration for the C/C++ extension, a warning is shown below the box that says, "Cannot find: /usr/local/include "
The trailing white space for each line should be trimmed so that such a warning is not shown.

Extension version: 1.18.5
VS Code version: Code 1.85.1 (0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2, 2023-12-13T09:49:37.021Z)
OS version: Windows_NT x64 10.0.22621
Modes:
Remote OS version: Linux x64 5.15.133.1-microsoft-standard-WSL2

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz (8 x 2803)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|15.73GB (1.22GB free)|
|Process Argv|--crash-reporter-id 52a0bd83-cab4-40c3-bdc8-a0e3553af2a2|
|Screen Reader|no|
|VM|0%|

|Item|Value|
|---|---|
|Remote|Dev Container: Scarab-MPC Development Container|
|OS|Linux x64 5.15.133.1-microsoft-standard-WSL2|
|CPUs|11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz (8 x 2803)|
|Memory (System)|7.63GB (5.70GB free)|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vspor879:30202332
vspor708:30202333
vspor363:30204092
vstes627:30244334
vscod805cf:30301675
binariesv615:30325510
vsaa593:30376534
py29gd2263:30899288
vsclangdf:30486550
[...]

PR #11911, opened 2024-01-31
Remove whitespace if the includePath exists + minor fix

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11862

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1258,7 +1258,7 @@ export class CppProperties {
 
     private trimPathWhitespace(paths: string[] | undefined): string[] | undefined {
         if (paths === undefined) {
-            return undefined
+            return undefined;
         }
         const trimmedPaths = [];
         for (const value of paths) {
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -808,6 +808,9 @@ export class CppProperties {
                 const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
                     fastGlob.sync(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
                 resolvedGlob.push(...matches.map(s => s + suffix));
+                if (resolvedGlob.length === 0) {
+                    resolvedGlob.push(normalized);
+                }
             } else {
                 resolvedGlob.push(normalized + suffix);
             }
@@ -1208,7 +1211,7 @@ export class CppProperties {
                         // Use the active configuration as the default selected configuration to load on UI editor
                         this.settingsPanel.selectedConfigIndex = this.CurrentConfigurationIndex;
                         this.settingsPanel.createOrShow(configNames,
-                            this.configurationJson.configurations[this.settingsPanel.selectedConfigIndex],
+                            this.configurationJson.configurations[this.settingsPanel.selectedConfigIndex], this.rootUri,
                             this.getErrorsForConfigUI(this.settingsPanel.selectedConfigIndex),
                             viewColumn);
                     }
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1211,7 +1211,7 @@ export class CppProperties {
                         // Use the active configuration as the default selected configuration to load on UI editor
                         this.settingsPanel.selectedConfigIndex = this.CurrentConfigurationIndex;
                         this.settingsPanel.createOrShow(configNames,
-                            this.configurationJson.configurations[this.settingsPanel.selectedConfigIndex], this.rootUri,
+                            this.configurationJson.configurations[this.settingsPanel.selectedConfigIndex],
                             this.getErrorsForConfigUI(this.settingsPanel.selectedConfigIndex),
                             viewColumn);
                     }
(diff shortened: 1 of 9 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -808,6 +808,9 @@ export class CppProperties {
                 const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
                     fastGlob.sync(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
                 resolvedGlob.push(...matches.map(s => s + suffix));
+                if (resolvedGlob.length === 0) {
```
sean-mcmanus, line 811: What is this for? How do I get resolvedGlob to be empty?

Extension/src/LanguageServer/settingsPanel.ts
```diff
@@ -252,11 +254,30 @@ export class SettingsPanel {
         }
     }
 
+    private validatePaths(message: any): void {
+        const trimmedPaths = [];
+        for (const value of message.value.split("\n")) {
+            if (fs.existsSync(this.rootUri?.fsPath + path.sep + value.trim())) {
```
sean-mcmanus: I don't understand this -- I see rootUri be my workspace folder, but I see value being my workspace folder too, which isn't going to exist.

image

sean-mcmanus (changes requested):
I think variables need to be resolved before checking if a path exists.

Extension/src/LanguageServer/configurations.ts
browntarik: If the glob pattern doesn't resolve, it pushes an empty string into resolve glob, try any incorrect folderpath with a glob pattern 

${workspaceFolder}/* *

You will see it does not correctly count the path as invalid

Extension/src/LanguageServer/settingsPanel.ts
browntarik: Do you have an example of this code not working?

sean-mcmanus: Does the includePath property accept relative paths? I didn't think it does. But either way the path checking shouldn't add on the rootUri path if the path is already a non-relative path.

bobbrow: > Does the includePath property accept relative paths?

I was thinking it does

sean-mcmanus: An example is ${workspaceFolder} or anything with an absolute path I think.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1253,11 +1256,27 @@ export class CppProperties {
...
+            for (const value of paths) {
+                const fullPath = this.resolvePath(value);
+                if (fs.existsSync(fullPath.trim())) {
+                    trimmedPaths.push(value.trim());
+                } else {
+                    trimmedPaths.push(value);
+                }
+            }
+            return trimmedPaths;
+        }
```
sean-mcmanus, line 1271: This should return `undefined` if paths === undefined. I'd prefer a check at the start.

```typescript
if (paths === undefined) {
    return undefined;
}
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1253,11 +1256,27 @@ export class CppProperties {
         }
     }
 
+    private trimPathWhitespace(paths: string[] | undefined): string[] | undefined {
+        const trimmedPaths = [];
+        if (paths !== undefined) {
+            for (const value of paths) {
+                const fullPath = this.resolvePath(value);
+                if (fs.existsSync(fullPath.trim())) {
+                    trimmedPaths.push(value.trim());
```
sean-mcmanus: Should this also check
```typescript
&& !fs.existsSync(fullPath)) {
```
?

Extension/ui/settings.ts
```diff
@@ -233,8 +233,8 @@ class SettingsApp {
         if (this.updating) {
             return;
         }
-
         const el: HTMLInputElement = <HTMLInputElement>document.getElementById(id);
+
```
sean-mcmanus: These newline changes should be undone.

Extension/src/LanguageServer/configurations.ts
browntarik: fair

sean-mcmanus (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1253,11 +1256,28 @@ export class CppProperties {
...
     }
 
+    private trimPathWhitespace(paths: string[] | undefined): string[] | undefined {
+        if (paths === undefined) {
+            return undefined;
+        }
+        const trimmedPaths = [];
+        for (const value of paths) {
+            const fullPath = this.resolvePath(value);
+            if (fs.existsSync(fullPath.trim()) && !fs.existsSync(fullPath)) {
```
bobbrow, line 1266: It would be better to check the non-trimmed path first because in most cases the path will not have trailing whitespace. The way this is written we will almost always do two `fs.existsSync` on the same path instead of one.

Another optimization that should happen is that you should check if the path even ends with whitespace before doing any `fs.existsSync` because the validation step already does another `fs.existsSync` on this path.  So in the general case we now check the path 3 times instead of 1.

Issue #11903, opened 2024-01-28
Config UI shows incorrect error message that `compilerPath` is not found, despite picked up correctly from PATH


geissdoerfer:
### Environment

- OS and Version: Arch Linux x64 6.6.10-1-lts
- VS Code Version: 1.85.2
- C/C++ Extension Version: v1.18.5

### Bug Summary and Steps to Reproduce

Bug Summary:

I have a project opened in vscode in the directory /home/myuser/myworkspace.  arm-none-eabi-gcc is on my PATH:

```shell
myuser@mymachine ~> arm-none-eabi-gcc --version
arm-none-eabi-gcc (Arch Repository) 13.2.0
```

When specifying `"compilerPath": "arm-none-eabi-gcc"` in `c_cpp_properties.json`, the following problem is reported: _Cannot find "/home/myuser/myworkspace/arm-none-eabi-gcc"_.

Expected behavior:

Since compilerPath is specified as a relative path, I would expect that it is found on my PATH. This seems to be the intended behavior implemented with 5ca9f34769bb19263bb94a3fd6c107630e503d2c.

### Configuration and Logs

```shell
`c_cpp_properties.json`:


{
    "configurations": [
        {
            "name": "${config:operSys}",
            "includePath": [
                "${config:rioteeSdkRoot}/core/include",
            ],
            "defines": [
                "mcpu=cortex-m4",
            ],
            "compilerPath": "arm-none-eabi-gcc",
            "intelliSenseMode": "gcc-arm",
            "cStandard": "c17",
            "cppStandard": "c++14"
        }
    ],
    "version": 4
}
``
```

Colengms:
Hi @geissdoerfer . I believe the error message you are seeing is incorrect and a bug in our path validation in the Config UI. However, using a relative path should actually function correctly.

Using a simple example, using `g++` and `clang-16`, which are installed on my Linux system and accessibly via the PATH, it successfully locates `g++` and `clang-16`, without any path specified, and configures the file correctly.

We can use this issue to track correcting the invalid error message in the Config UI.

sean-mcmanus:
@geissdoerfer Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.19.3

PR #11913, opened 2024-02-01
Validate Compiler Paths Against PATH Environment Variable in Configuration UI

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11903

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1545,7 +1545,6 @@ export class CppProperties {
         if (resolvedCompilerPath === undefined) {
             resolvedCompilerPath = this.resolvePath(config.compilerPath);
         }
-        
         const settings: CppSettings = new CppSettings(this.rootUri);
         const compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(!!settings.legacyCompilerArgsBehavior, resolvedCompilerPath);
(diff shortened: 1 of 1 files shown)


Colengms (approved):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1536,9 +1536,15 @@ export class CppProperties {
 
         // Check if config name is unique.
         errors.name = this.isConfigNameUnique(config.name);
-
+        let resolvedCompilerPath: string | undefined;
         // Validate compilerPath
-        let resolvedCompilerPath: string | undefined = this.resolvePath(config.compilerPath);
+        if (config.compilerPath) {
+            resolvedCompilerPath = which.sync(config.compilerPath, { nothrow: true }) ?? undefined;
```
bobbrow, line 1542: Not sure we want to just randomly call `which` on the compilerPath without checking if it looks like a path first.  Should we be checking for path separators first?

Colengms: Looking at [the code to `which`](https://www.npmjs.com/package/which?activeTab=code), it does appear to do the right thing (not resolve against paths) if provided input with a path separator in it.  Though, I do notice that it also considers the `cwd`, which may not always be desirable (as it's unclear what we can expect the cwd to be at any given time - likely not always the current workspace folder, i.e. in multiroot).

PR #11956, opened 2024-02-13
Properly quote references to setting string values.

sean-mcmanus:
Otherwise, users could potentially mistake the value for a non-string value, such as `true` and `false`, although the json editor will complain.


browntarik (approved):

sean-mcmanus:
I'm abandoning this PR -- the VS Code editor.indentSize setting is the only setting I see that puts the setting values in `""`, so the recommended practice appears to be to refer to the setting values without `""`.

Issue #8389, opened 2021-11-12
Move use of `console.log` to C/C++ output channel


Colengms:
There are a many uses of `console.log()` that could be output to the C/C++ output channel instead.  They currently do not indicate what extension they relate to, so are mysterious to users who see them in the developer tools console output.  i.e.:

```
[Extension Host] provideCustomConfiguration
```

PR #11973, opened 2024-02-15
Silence `console.log`s

trond-snekvik:
Turns the `console.log` messages that happen regularly into output stream log messages. Leaves `console.log`s in exception handlers the way they are.

`console.log` will show up in the console window of all extension debug sessions. I'm the maintainer of [an extension](https://marketplace.visualstudio.com/items?itemName=nordic-semiconductor.nrf-connect) that integrates with the configuration provider API for the C++ extension. 

The amount of console log messages this extension generates that I have to filter out when debugging my own extension finally got annoying enough that I wanted to open a PR to try and force a change.

Note that I'm bypassing the localization mechanism for the output channel here. Not sure if this is acceptable.

Fixes most of #8389, but there are still a few `console.log`s left, so not sure if it's enough to mark it as closed.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1836,7 +1836,7 @@ export class DefaultClient implements Client {
         if (!this.configurationProvider) {
             return;
         }
-        console.log("updateCustomBrowseConfiguration");
+
         const currentProvider: CustomConfigurationProvider1 | undefined = getCustomConfigProviders().get(this.configurationProvider);
         if (!currentProvider || !currentProvider.isReady || (requestingProvider && requestingProvider.extensionId !== currentProvider.extensionId)) {
             return;
@@ -1973,7 +1973,6 @@ export class DefaultClient implements Client {
         DefaultClient.isStarted.reset();
 
         const tokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
-        console.log("provideCustomConfiguration");
 
         const params: QueryTranslationUnitSourceParams = {
             uri: docUri.toString(),
@@ -3080,10 +3079,8 @@ export class DefaultClient implements Client {
                     if (sanitized.browsePath.length === 0) {
                         sanitized.browsePath = ["${workspaceFolder}/**"];
                     }
-                    console.log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
                     break;
                 }
-                console.log("No browse configuration is available.");
                 return;
             }
 
@@ -3097,7 +3094,6 @@ export class DefaultClient implements Client {
                     if (sanitized.browsePath.length === 0) {
                         sanitized.browsePath = ["${workspaceFolder}/**"];
                     }
-                    console.log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
                     break;
                 }
                 return;
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -145,7 +145,6 @@ function sendActivationTelemetry(): void {
  */
 export async function activate(): Promise<void> {
 
-    console.log("activating extension");
     sendActivationTelemetry();
     const checkForConflictingExtensions: PersistentState<boolean> = new PersistentState<boolean>("CPP." + util.packageJson.version + ".checkForConflictingExtensions", true);
     if (checkForConflictingExtensions.Value) {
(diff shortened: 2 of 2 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -2239,7 +2239,7 @@ export class DefaultClient implements Client {
                 try {
                     promise.resolve(task ? await task() : undefined);
                 } catch (e) {
-                    console.log(e);
+                    log(e);
```
sean-mcmanus: The compiler wants this `e` to be a string -- maybe keep this as console.log?

Extension/src/LanguageServer/client.ts
```diff
@@ -3072,32 +3072,32 @@ export class DefaultClient implements Client {
...
+                    log("Received an invalid browse configuration from configuration provider.");
                 }
                 const configValue: WorkspaceBrowseConfiguration | undefined = this.lastCustomBrowseConfiguration.Value;
                 if (configValue) {
                     sanitized = configValue;
                     if (sanitized.browsePath.length === 0) {
                         sanitized.browsePath = ["${workspaceFolder}/**"];
                     }
-                    console.log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
+                    log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
```
sean-mcmanus: `, ` should be ` + `

Extension/src/LanguageServer/client.ts
```diff
@@ -3072,32 +3072,32 @@ export class DefaultClient implements Client {
...
-                console.log("Received an invalid browse configuration from configuration provider: " + JSON.stringify(sanitized));
+                log("Received an invalid browse configuration from configuration provider: " + JSON.stringify(sanitized));
                 const configValue: WorkspaceBrowseConfiguration | undefined = this.lastCustomBrowseConfiguration.Value;
                 if (configValue) {
                     sanitized = configValue;
                     if (sanitized.browsePath.length === 0) {
                         sanitized.browsePath = ["${workspaceFolder}/**"];
                     }
-                    console.log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
+                    log("Falling back to last received browse configuration: ", JSON.stringify(sanitized, null, 2));
```
sean-mcmanus: `, ` should be ` + `

sean-mcmanus (changes requested):

bobbrow:
I'm honestly not sure how critical it is to keep any/most of these messages.  I added them years ago as a cue to developers working on config providers.  Since average users generally don't need to see these, they should be logged with a higher level to be hidden by default, or simply deleted if we don't think they add value.

sean-mcmanus:
> I'm honestly not sure how critical it is to keep any/most of these messages. I added them years ago as a cue to developers working on config providers. Since average users generally don't need to see these, they should be logged with a higher level to be hidden by default, or simply deleted if we don't think they add value.

i.e. it sounds like some logDebug or equivalent should be used instead of log, so that it doesn't log these messages unless the logging_level is "Debug".

bobbrow:
> i.e. it sounds like some logDebug or equivalent should be used instead of log, so that it doesn't log these messages unless the logging_level is "Debug".

In case my last comment was unclear, I'm totally ok with deleting all of these messages.

sean-mcmanus:
> > i.e. it sounds like some logDebug or equivalent should be used instead of log, so that it doesn't log these messages unless the logging_level is "Debug".
> 
> In case my last comment was unclear, I'm totally ok with deleting all of these messages.

Sure, but it seems like `Received an invalid browse configuration from configuration provider: ` might be useful to log.

Extension/src/LanguageServer/client.ts
trond-snekvik: Kept this as-is.

Extension/src/LanguageServer/client.ts
trond-snekvik: Removed this line altogether

Extension/src/LanguageServer/client.ts
trond-snekvik: Removed this line altogether

trond-snekvik:
Thanks for the feedback. As suggested, I removed the logs instead of moving them to the output channel. Kept  the ones that report irregular behavior.

bobbrow (approved):

sean-mcmanus (approved):

Issue #12001, opened 2024-02-22
Update 1.19.4 breaks wildcard operator '*' in tasks.json code will not compile


ven-ranganathan:
### Environment

- OS and Version: macOS Sonoma 14.3.1 (23D60)
- VS Code Version: Version: 1.86.2
Commit: https://github.com/microsoft/vscode/commit/903b1e9d8990623e3d7da1df3d33db3e42d80eda
Date: 2024-02-13T19:42:12.210Z
Electron: 27.2.3
ElectronBuildId: 26908389
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin arm64 23.3.0
- C/C++ Extension Version: 1.19.4 (1.18.5 works as expected)
- If using SSH remote, specify OS of remote machine: 

### Bug Summary and Steps to Reproduce

Since this morning I can no longer use the wild card operator to compile multiple files with the same extension, this was previously working as expected since I configured it a year ago.

The same command works if I type it in the terminal manually, i.e. "clang -g *.c"

This problem is compiler agnostic, the same error shows up if I try to use gcc instead of clang

### Steps to Reproduce:

configure tasks.json to launch all files with the same extension using *.c, for example:
```
"type": "cppbuild",
"label": "C/C++: clang build multiple files",
"command": "/usr/bin/clang",
"args": [
"-fcolor-diagnostics",
"-fansi-escape-codes",
"-g",
"*.c",
],
"options": {
"cwd": "${fileDirname}"
},
"problemMatcher": [
"$gcc"
],
"group": "build",
"detail": "Task generated by Debugger."
```
### Error 

attempt to compile code using task
receive the following error:
```
Starting build...
/usr/bin/clang -fcolor-diagnostics -fansi-escape-codes -g '*.c'
clang: error: no such file or directory: '*.c'
clang: error: no input files
```
### Expected behavior: 

should compile the same as when user enters the same command in terminal manually

### Configuration and Logs

```shell
c_cpp_properties.json never changed
```

### Other Extensions

remote SSH
remote SSH editing
remote explorer
material icon theme
monokai vibrant theme
monokai night theme
python
pylance
python debugger
vscode pdf

Colengms:
[...]

PR #12003, opened 2024-02-22
Add fragments field to task.json to persist legacy command line argument behavior.

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12001

### DIFF
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -928,6 +928,7 @@
     "c_cpp.taskDefinitions.name.description": "The name of the task.",
     "c_cpp.taskDefinitions.command.description": "The path to either a compiler or script that performs compilation.",
     "c_cpp.taskDefinitions.args.description": "Additional arguments to pass to the compiler or compilation script.",
+    "c_cpp.taskDefinitions.fragments.description": "Additional arguments to pass to the compiler that are expected to receive shell processing. For example, '*.c' should expand to all .c files in the working directory.",
     "c_cpp.taskDefinitions.options.description": "Additional command options.",
     "c_cpp.taskDefinitions.options.cwd.description": "The current working directory of the executed program or script. If omitted Code's current workspace root is used.",
     "c_cpp.taskDefinitions.detail.description": "Additional details of the task.",
--- a/Extension/src/LanguageServer/cppBuildTaskProvider.ts
+++ b/Extension/src/LanguageServer/cppBuildTaskProvider.ts
@@ -394,7 +394,7 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
         const exePath: string | undefined = util.resolveVariables(util.findExePathInArgs(this.args));
         util.createDirIfNotExistsSync(exePath);
 
-        if (this.fragments.length > 0 && this.args.length == 0) {
+        if (this.fragments.length > 0 && this.args.length === 0) {
             this.args = this.fragments;
             this.args.forEach((value) => {
                 activeCommand = activeCommand + " " + value;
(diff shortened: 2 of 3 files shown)


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -188,13 +188,19 @@ export class CppBuildTaskProvider implements TaskProvider {
...
+            if (commandFragments && commandFragments.length > 0) {
+                fragments = args.concat(commandFragments);
+            }
             const cwd: string = isWindows && !isCl && !process.env.PATH?.includes(path.dirname(compilerPath)) ? path.dirname(compilerPath) : "${fileDirname}";
             const options: cp.ExecOptions | cp.SpawnOptions | undefined = { cwd: cwd };
             definition = {
                 type: CppBuildTaskProvider.CppBuildScriptType,
                 label: taskLabel,
                 command: isCl ? compilerPathBase : compilerPath,
                 args: args,
```
sean-mcmanus, line 202: pre-existing issue, but I think this can be simplified to

```
args,
fragments,
options
```

sean-mcmanus (approved):

Extension/package.nls.json
```diff
@@ -928,6 +928,7 @@
     "c_cpp.taskDefinitions.name.description": "The name of the task.",
     "c_cpp.taskDefinitions.command.description": "The path to either a compiler or script that performs compilation.",
     "c_cpp.taskDefinitions.args.description": "Additional arguments to pass to the compiler or compilation script.",
+    "c_cpp.taskDefinitions.fragments.description": "Additional arguments to pass to the compiler that are expected to receive shell processing. For example, '*.c' should expand to all .c files in the working directory.",
```
Colengms, line 931: The Windows command prompt doesn't do wildcard expansion. I'd omit the example. 

How about? :  `Command line fragments to pass to the compiler or compilation script. These arguments will receive limited shell command line processing. Multiple arguments may be specified per value.`

And maybe update `args` to, `Arguments to pass to the compiler or compilation script. These are not expected to contain shell escaping and will not receive shell processing. A single argument should be specified per value`.

(The use of `Additional` in the descriptions seems unnecessary to me.)

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -387,11 +394,29 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
         const exePath: string | undefined = util.resolveVariables(util.findExePathInArgs(this.args));
         util.createDirIfNotExistsSync(exePath);
 
-        this.args.forEach((value, index) => {
-            value = util.quoteArgument(util.resolveVariables(value));
-            activeCommand = activeCommand + " " + value;
-            this.args[index] = value;
-        });
+        if (this.fragments.length > 0 && this.args.length === 0) {
```
Colengms, line 397: Could this be simplified to? :

```
            value = util.quoteArgument(util.resolveVariables(value));
            activeCommand = activeCommand + " " + value;
            this.args[index] = value;
        });
        this.fragments.forEach((value) => {
            activeCommand = activeCommand + " " + value;
            this.args.push(value);
        });
```

Or, as I've suggested before, change `cp.spawn` to `cp.exec`, so we can just pass it `activeCommand`, which we're already building for logging purposes.  Then you could omit all of the edits of `this.args`, reducing that code to:

```
            value = util.quoteArgument(util.resolveVariables(value));
            activeCommand = activeCommand + " " + value;
        });
        this.fragments.forEach((value) => {
            activeCommand = activeCommand + " " + value;
        });
```

Colengms (changes requested):

Colengms:
I think this is the wrong way to go.  I had mentioned another approach in the original issue, to leverage the `command` field.  Doing so would align us with the behavior of VS Code and the default task provider:

https://code.visualstudio.com/docs/editor/tasks

```
{
  "label": "dir",
  "type": "shell",
  "command": "dir 'folder with spaces'"
}
```

```
{
  "label": "dir",
  "type": "shell",
  "command": "dir",
  "args": ["folder with spaces"]
}
```

Ultimately, we may want to support everything they describe there.

PR #12026, opened 2024-02-27
Update `cppbuild` task behavior for `args` and `command` fields to match VS Code's `shell` task type

### LINKED ISSUE #12001: Update 1.19.4 breaks wildcard operator '*' in tasks.json code will not compile

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12001

Reworks handling of the `args` and `command` field to exactly match the behavior documented for VS Code `shell` tasks, as documented here: https://code.visualstudio.com/docs/editor/tasks

A major portion of this was copied directly from VS Code: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/tasks/browser/terminalTaskSystem.ts

This replicates variable bugs and incomplete functionality in VS Code, such as `escape` quoting on Windows/cmd, which it does not appear to attempt. I believe we'll want to redirect any usability issues that repro with `shell` tasks, directly to VS Code, and maintain our support as identical to theirs.


sean-mcmanus (approved):

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -188,30 +192,38 @@ export class CppBuildTaskProvider implements TaskProvider {
...
                 args = args.concat(compilerArgs);
             }
-            const cwd: string = isWindows && !isCl && !process.env.PATH?.includes(path.dirname(compilerPath)) ? path.dirname(compilerPath) : "${fileDirname}";
+            const cwd: string = isWindows && !isCl && !process.env.PATH?.includes(path.dirname(compilerPathString)) ? path.dirname(compilerPathString) : "${fileDirname}";
             const options: cp.ExecOptions | cp.SpawnOptions | undefined = { cwd: cwd };
             definition = {
                 type: CppBuildTaskProvider.CppBuildScriptType,
                 label: taskLabel,
-                command: isCl ? compilerPathBase : compilerPath,
+                command: compilerPath,
```
bobbrow, line 200: Should this be `command: resolvedCompilerPath,`?

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -380,22 +392,29 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
...
+                quoting: this.command.quoting
+            };
+        }
 
         // Create the exe folder path if it doesn't exist.
         const exePath: string | undefined = util.resolveVariables(util.findExePathInArgs(this.args));
         util.createDirIfNotExistsSync(exePath);
 
         this.args.forEach((value, index) => {
-            value = util.quoteArgument(util.resolveVariables(value));
```
bobbrow, line 391: If both uses of `util.quoteArgument` are removed in this PR and we don't need that function anymore, I think we should delete the implementation.

bobbrow (approved):

Issue #10956, opened 2023-05-15
Copy declaration / definition results are not formatted when pasted


sean-mcmanus:
It gives:

![image](https://github.com/microsoft/vscode-cpptools/assets/19859882/93aab0c3-ca5c-4a72-b756-255dd11f5936)

The issue doesn't repro with Create declaration / definition, in which case the formatting is applied after the creation is done.

I think the root cause of the bug is https://github.com/microsoft/vscode-cpptools/issues/10236 .

sean-mcmanus:
This issue is reproing with Extract Function too.

sean-mcmanus:
@bobbrow It feels to me like this issue isn't high enough priority for 1.19.0 compared to the other To Do stuff (unless we get more user feedback or reports of bugs caused by this), i.e. it's at the bottom of my list unless someone else wanted to investigate it.

bobbrow:
@sean-mcmanus, sure. Should we close as a dupe of https://github.com/microsoft/vscode-cpptools/issues/10236?  I think that maybe addressing that takes care of this one.

sean-mcmanus:
@bobbrow I'm not sure. But https://github.com/microsoft/vscode-cpptools/issues/10236 seems low priority too.

sean-mcmanus:
Fixing this issue could fix https://github.com/microsoft/vscode-cpptools/issues/11675 and https://github.com/microsoft/vscode-cpptools/issues/11673 .

PR #12055, opened 2024-03-04
Add internal formatting support for Copy Declaration/Definition

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/10956


Extension/src/LanguageServer/client.ts
```diff
@@ -3452,6 +3462,25 @@ export class DefaultClient implements Client {
                     line: range.end.line
                 }
             },
+            formatParams: {
+                editorConfigSettings: {},
+                useVcFormat: cppSettings.useVcFormat(editor.document),
+                insertSpaces: editorSettings.editorInsertSpaces !== undefined ? editorSettings.editorInsertSpaces : false,
```
sean-mcmanus: I think the defaults should be `true` and `4` (the `=== undefined` case)

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -3452,6 +3462,25 @@ export class DefaultClient implements Client {
                     line: range.end.line
                 }
             },
+            formatParams: {
+                editorConfigSettings: {},
```
Colengms, line 3493: You should call getEditorConfigSettings() with the path to the file in question to read the vcFormat settings values that apply to the file in question, and populate editorConfigSettings here.

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
Colengms: It looks like you want to call `settings.useVcFormat(editor.document)` first.  If the formatting setting is set to "default", this will check if an editorConfig file is present with vcFormat entries in it.  If so, it will return true. (It'll also return true if formatting is set to vcFormat).  If true, you'll want to call `getEditorConfigSettings(editor.document.uri.fsPath)` to include relevant vcFormat settings in the request.

Extension/src/LanguageServer/client.ts
```diff
@@ -3439,10 +3449,55 @@ export class DefaultClient implements Client {
...
+        if (uri === undefined || range === undefined || editor === undefined) {
             return;
         }
 
+        let formatParams;
+        if (cppSettings.useVcFormat(editor.document))
+        {
+            const editorConfigSettings: any = getEditorConfigSettings(uri.fsPath);
+            formatParams = {
+                editorConfigSettings: {},
```
Colengms: Shouldn't `editorConfigSettings` be populated here?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
browntarik: it is, I tested this and it works.

Extension/src/LanguageServer/client.ts
```diff
@@ -3439,10 +3449,55 @@ export class DefaultClient implements Client {
             }
         }
 
-        if (uri === undefined || range === undefined) {
+        if (uri === undefined || range === undefined || editor === undefined) {
             return;
         }
 
+        let formatParams;
```
Colengms: I'm not sure why lint didn't flag this.  Variables should have types associated with them.  Could you change this to `let formatParams: FormatParams | undefined`?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: @Colengms A while back there was a change to the linter policy to allow variables without a type.

sean-mcmanus: ...but I agree that it's better to set the type unless it's obvious from the code, i.e. treat it like you would "auto" in C++.

Extension/src/LanguageServer/client.ts
```diff
@@ -358,6 +359,12 @@ export interface FormatParams extends SelectionParams {
     onChanges: boolean;
 }
 
+export interface EditorParams {
```
Colengms: Is this left over?  It doesn't appear to be referenced.

Colengms (changes requested):

sean-mcmanus (dismissed):

Colengms (approved):

Issue #12052, opened 2024-03-04
Potential listener LEAK detected


alexandruavadanii:
Type: <b>Bug</b>

1. Create a new workspace with 3 empty files (`1.c`, `2.c`, `3.c`)
2. Open all 3 empty files in 3 columns (not relevant, but makes the next step easier)
3. Click a couple of times inside each of the 3 editors (1, 2, 3, 1, 2, 3 etc.)
4. Toggle developer console and analyze the logs
5. :bug: potential listener LEAK detected, having 439 listeners already.

![image](https://github.com/microsoft/vscode-cpptools/assets/11544783/523ac295-b61c-4035-b698-0a448822ef59)

You can probably ignore the "Redundant folding ranges request received..." for this, just filter out "info" messages from the developer console.

Extension version: 1.19.4
VS Code version: Code - Insiders 1.88.0-insider (61187a0b6547df5911a0ee259f7afda46bac7ec3, 2024-03-04T05:46:35.204Z)
OS version: Windows_NT x64 10.0.18362
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz (8 x 2112)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|15.81GB (4.23GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|

</details>

alexandruavadanii:
Hi, @Colengms,
I think you got it right that the folding ranges are causing this.
I tried to bisect which vscode commit changed the behavior, but after seeing Martin's comments in https://github.com/microsoft/vscode/issues/206841, I think it's safe to assume this is caused by the doubled folding range requests coming in quick succession to vscode-cpptools. The second folding request was introduced with the sticky scroll feature, so that's why we don't see this with older vscode.

I will have to test this theory, but one possible explanation is a race condition between
[...]

PR #12063, opened 2024-03-05
Remove work-around for folding ranges issue

Colengms:
This is in the context of: https://github.com/microsoft/vscode-cpptools/issues/12052
and: https://github.com/microsoft/vscode/issues/206841

This would restore our `FoldingRangeProvider` to extremely straight-forward functionality, and not attempt to work around a possible VS Code issue with many redundant requests.  It's possible that may worsen the behavior we see when the issue repros, but would also help clarify whether the issue is in VS Code itself.

### DIFF
--- a/Extension/src/LanguageServer/Providers/foldingRangeProvider.ts
+++ b/Extension/src/LanguageServer/Providers/foldingRangeProvider.ts
@@ -3,22 +3,13 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import * as vscode from 'vscode';
-import { ManualPromise } from '../../Utility/Async/manualPromise';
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

sean-mcmanus:
@Colengms I see 1 folding range request issues per visible documunt whenever the active document changes -- the folding range requests for the inactive documents seem unnecessary though? Could that be a VS Code bug?

sean-mcmanus (changes requested):
This is not usable -- I see it sending exponentially large numbers of getFoldingRanges request after every switch document, i.e. it'll eventually start sending thousands.

sean-mcmanus:
@Colengms Actually, this doesn't fix the listener issue -- get folding range requests eventually start to fail: 
image

alexandruavadanii:
FWIW, I played a bit with this and pushed my experiment to https://github.com/alexandruavadanii/vscode-cpptools/commit/90c21e472849ee84d69bfe838a32723276c3e518.
Note that I'm no TS/JS dev and this has only been barely tested for a bit, but I don't see the initial issue any more.

Issue #11756, opened 2023-12-04
Support for `${userHome}`


paulober:
### Feature Request

You can find the official variables here: https://code.visualstudio.com/docs/editor/variables-reference

Currently the only what to make system independent configuration for `c_cpp_properties.json` is to use stuff like
`${env:HOME}/compiler/arm-none-eabi-gcc`. As you can see this isn't platform independent as it doesn't work on Windows, 
where the alternative would be `${env:USERPROFILE}/compiler/arm-none-eabi-gcc`.

So please implement support for `${userHome}` as platform independent alias for the node.js `homedir()`:
https://nodejs.org/api/os.html#oshomedir

bobbrow:
This shouldn't be hard to do. Updating tags.

sean-mcmanus:
@paulober Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.20.0

paulober:
Thanks for implementing. This really solves some issues in my projects.

PR #12071, opened 2024-03-06
Resolve userHome variable in c_cpp_properties.json

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/11756


Colengms:
Are there other variables we should add support for while we're in there?  Of the ones on `https://code.visualstudio.com/docs/editor/variables-reference`, the following also seem supportable in this context:


${workspaceFolderBasename} - the name of the folder opened in VS Code without any slashes (/)
${execPath} - the path to the running VS Code executable
${pathSeparator} - the character used by the operating system to separate components in file paths
${/} - shorthand for ${pathSeparator}

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1530,6 +1530,11 @@ export class CppProperties {
                 result = result.replace("${workspaceRoot}", this.rootUri.fsPath);
             }
         }
+
+        if (result.includes("${userHome}")) {
```
bobbrow: Please add this to the `ExtendedEnvironment` instead. (line 421-ish).

Might also want to update the regexp on line 374 of `common.ts` as well, though I'm not sure why all the variable names were added there (I think the ones after `env|config|workspaceFolder` should be deleted instead because it doesn't make sense to have a `.` or `:` after them)

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
Colengms: What about these other variables that are being hard-coded?  i.e. vcpkgroot.  Should they be moved into ExtendedEnvironment()/AdditionalEnvironment() ?

sean-mcmanus:
> Are there other variables we should add support for while we're in there? Of the ones on https://code.visualstudio.com/docs/editor/variables-reference, the following also seem supportable in this context:
> 
> ${workspaceFolderBasename} - the name of the folder opened in VS Code without any slashes (/) ${execPath} - the path to the running VS Code executable ${pathSeparator} - the character used by the operating system to separate components in file paths ${/} - shorthand for ${pathSeparator}

I think that's beyond the scope of this particular change? i.e. I would think those could/should be fixed separately unless it's trivial?

Extension/src/LanguageServer/configurations.ts
bobbrow: If vcpkgRoot is not processed on the native side, then I think it makes sense to move it to ExtendedEnvironment.  workspaceFolder needs to stay here though because resolveVariables does not touch it and it seems we need valid paths when this function completes.

bobbrow:
> > Are there other variables we should add support for while we're in there? Of the ones on https://code.visualstudio.com/docs/editor/variables-reference, the following also seem supportable in this context:
> > ${workspaceFolderBasename} - the name of the folder opened in VS Code without any slashes (/) ${execPath} - the path to the running VS Code executable ${pathSeparator} - the character used by the operating system to separate components in file paths ${/} - shorthand for ${pathSeparator}
> 
> I think that's beyond the scope of this particular change? i.e. I would think those could/should be fixed separately unless it's trivial?

They are all implemented already with the exception of `${/}`.  It's trivial to add support for variables, so it can be done now if we want.

Extension/src/LanguageServer/configurations.ts
browntarik: vcpkgRoot does not seem to have any dependency on native code so I will add it to Extended Environment

Extension/src/LanguageServer/configurations.ts
```diff
@@ -425,6 +425,10 @@ export class CppProperties {
         result["workspaceFolderBasename"] = this.rootUri ? path.basename(this.rootUri.fsPath) : "";
         result["execPath"] = process.execPath;
         result["pathSeparator"] = (os.platform() === 'win32') ? "\\" : "/";
+        result["/"] = (os.platform() === 'win32') ? "\\" : "/";
+        result["userHome"] = os.homedir();
+        result["vcpkgRoot"] = util.getVcpkgRoot();
```
bobbrow: In the case where `util.VcpkgRoot()` does not return anything, the variable should not be replaced with "".

Do something like this instead:
```ts
if (util.getVcpkgRoot()) {
    result["vcpkgRoot"] = util.getVcpkgRoot();
}
```

Extension/src/common.ts
```diff
@@ -371,7 +371,7 @@ export function resolveVariables(input: string | undefined, additionalEnvironmen
     }
 
     // Replace environment and configuration variables.
-    const regexp: () => RegExp = () => /\$\{((env|config|workspaceFolder|file|fileDirname|fileBasenameNoExtension|execPath|pathSeparator)(\.|:))?(.*?)\}/g;
+    const regexp: () => RegExp = () => /\$\{((env|config|workspaceFolder|userHome)(\.|:))?(.*?)\}/g;
```
bobbrow: `userHome` should also be removed

bobbrow (approved):

Issue #12070, opened 2024-03-06
Can't use globstar in the middle of the path in includePath


ggladilov:
### Environment

- OS and Version: macOS Sonoma Version 14.3.1
- VS Code Version: 1.87.0 (Universal)
- C/C++ Extension Version: 1.19.5

### Bug Summary and Steps to Reproduce

Bug Summary:

Path with globstar in the middle of the path doesn't work in `includePath` of `c_cpp_properties.json`. Example of usage:

```json
{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${workspaceFolder}/**",
                "/opt/homebrew/opt/**/include", <- faulty path
                "/opt/homebrew/opt/opencv/include/opencv4",
                "/opt/homebrew/opt/openvino/include"
            ],
...
    "version": 4
}
```

Per https://github.com/microsoft/vscode-cpptools/issues/723 I tried both `**` and `*`, but the result is the same:

<img width="596" alt="Screenshot 2024-03-06 at 19 39 29" src="https://github.com/microsoft/vscode-cpptools/assets/8751635/b77d9055-1eeb-41f7-b5d8-2bbcc2a21942">

Using `ls /opt/homebrew/opt/**/include` locally in external terminal shows multiple matches as expected.

Steps to reproduce:
1. Go to 'c_cpp_properties.json' file and add path with globstar in the middle to `includePath`
2. See error

Expected behavior: multiple directories as a result of globstar expansion would be added to `includePath`

### Configuration and Logs

```shell
`c_cpp_properties.json`:


{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${workspaceFolder}/**",
                "/opt/homebrew/opt/**/include",
                "/opt/homebrew/opt/opencv/include/opencv4",
                "/opt/homebrew/opt/openvino/include"
            ],
            "defines": [],
            "macFrameworkPath": [
                "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks"
            ],
            "compilerPath": "/opt/homebrew/opt/llvm/bin/clang++",
            "cStandard": "c23",
            "cppStandard": "c++23",
            "intelliSenseMode": "macos-clang-arm64",
            "compileCommands": "${command:cmake.buildDirectory}/compile_commands.json"
        }
    ],
    "version": 4
}
```

browntarik:
Thank you for reporting your issue, we will have a fix out shortly.

[...]

PR #12072, opened 2024-03-06
Eliminate squiggle from all glob configurations

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12070


Colengms:
This change prevent the squiggle.  Does that mean this scenario actually worked properly for the user, and the only issue was an unexpected squiggle in the JSON file?

browntarik:
> This change prevent the squiggle. Does that mean this scenario actually worked properly for the user, and the only issue was an unexpected squiggle in the JSON file?

It just gets rid of the squiggle

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1893,7 +1893,7 @@ export class CppProperties {
         }
 
         // Resolve and split any environment variables
-        paths = this.resolveAndSplit(paths, undefined, this.ExtendedEnvironment);
+        paths = this.resolveAndSplit(paths, undefined, this.ExtendedEnvironment, true);
```
sean-mcmanus: @browntarik I don't understand -- this is the assumeRelative parameter -- it appears you're "fixing" the squiggle by disabling squiggling for all paths (since squiggling is disabled for relative paths)?

sean-mcmanus (changes requested):

sean-mcmanus:
> > This change prevent the squiggle. Does that mean this scenario actually worked properly for the user, and the only issue was an unexpected squiggle in the JSON file?
> 
> It just gets rid of the squiggle

@browntarik I don't fully understand still -- is there a bug still with glob not working? Or is the bug just with the squiggle?

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: @browntarik Actually, I don't see any change/fix from this change? I see it exit out of resolveAndSplit before it even examines the 4th bool arg.

browntarik: You're right, it looks like the real fix is to resolve and split glob patterns before validating config paths

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1542,7 +1542,11 @@ export class CppProperties {
             // Make sure all paths result to an absolute path.
             // Do not add the root path to an unresolved env variable.
             if (!result.includes("env:") && !path.isAbsolute(result) && this.rootUri) {
-                result = path.join(this.rootUri.fsPath, result);
+
+                if (util.checkDirectoryExistsSync(path.join(this.rootUri.fsPath, result)))
+                {
```
sean-mcmanus: Can you run the formatter? I believe `{` should go on the same line.

sean-mcmanus (changes requested):
I see all error squiggles not working now.

sean-mcmanus (changes requested):
Paths ending with `*` or `**`should still squiggle -- those are not wild card expansions, but determine if a path is recursive or not. And it might be easier to check curPath instead of escapedPath, because escapedPath has the `*` escaped with `\\`, i.e. it should only skip `*` if it's not at the end of the path.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2029,7 +2029,13 @@ export class CppProperties {
             // and extend that pattern to the next quote before and next quote after it.
             const pattern: RegExp = new RegExp(`"[^"]*?(?<="|;)${escapedPath}(?="|;).*?"`, "g");
             const configMatches: string[] | null = curText.match(pattern);
-            if (configMatches && !escapedPath.includes('*')) {
+
+            let globPath: boolean = false;
+            if (curPath.includes("*") && !curPath.endsWith("*")) {
```
sean-mcmanus: Actually, it needs to be able handle paths with glob patterns in the middle and "*" at the end, in which case it shouldn't squiggle (the current code does still), i.e. the `includes(*)` should only be counted as glob pattern if the index is not at the end, i.e. maybe switch to calling some indexOf function (or whatever it's called) and checking if it's not equal to the last character.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2029,7 +2029,17 @@ export class CppProperties {
             // and extend that pattern to the next quote before and next quote after it.
             const pattern: RegExp = new RegExp(`"[^"]*?(?<="|;)${escapedPath}(?="|;).*?"`, "g");
             const configMatches: string[] | null = curText.match(pattern);
-            if (configMatches) {
+
+            let globPath: boolean = false;
+            const asteriskPosition = curPath.indexOf("*");
+            if (asteriskPosition !== -1 && asteriskPosition !== curPath.length - 1 && asteriskPosition !== curPath.length - 2) {
```
sean-mcmanus: The case "a" doesn't report a squiggle (corner case). You can fix this via moving the check `asteriskPosition !== -1` into an outer check and put the other checks inside that, i.e. the problem is `asteriskPosition === curPath.length - 2` is being checked when `asteriskPosition !== -1`.

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #12085, opened 2024-03-08
Add a command to fetch the includes graph for a TU

lukka:


Extension/src/LanguageServer/client.ts
```diff
@@ -536,6 +536,16 @@ interface DidChangeActiveEditorParams {
...
 }
 
+interface GetFileIncludeDataParams
+{
+    maxDepth: number;
+}
+
+interface IncludesGraphResult
+{
+    included_files: string[];
```
sean-mcmanus: This should be includedFiles, right?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
sean-mcmanus: ...also I don't think the interfaces match the naming in cpptools.

lukka: Fixed, thank you

Extension/src/LanguageServer/client.ts
```diff
@@ -536,6 +536,16 @@ interface DidChangeActiveEditorParams {
     selection?: Range;
 }
 
+interface GetFileIncludeDataParams
+{
+    maxDepth: number;
+}
+
+interface IncludesResult
```
sean-mcmanus: GetIncludesResult?

Extension/src/LanguageServer/client.ts
```diff
@@ -536,6 +536,16 @@ interface DidChangeActiveEditorParams {
     selection?: Range;
 }
 
+interface GetFileIncludeDataParams
```
sean-mcmanus: GetIncludesParams?

lukka: Fixed, thank you

Extension/src/LanguageServer/client.ts
lukka: fixed, thank you

sean-mcmanus (approved):

Colengms (approved):

PR #12087, opened 2024-03-08
Update changelog again for 1.19.7 (#12086)

Colengms:


sean-mcmanus (approved):

heartacker:
where can i Get the package vsix to test those unrelease feature /bugfix?

sean-mcmanus:
@heartacker When we release it on Monday. Unless you're a Microsoft internal. We generally don't release on Fridays in case there's some issue with a release over the weekend.

heartacker:
> @heartacker When we release it on Monday. Unless you're a Microsoft internal. We generally don't release on Fridays in case there's some issue with a release over the weekend.

thanks

PR #12088, opened 2024-03-09
Prevent the status bar from updating too quickly

### LINKED ISSUE #12084: Weird flashing at the bottom bar?
### Discussed in https://github.com/microsoft/vscode-cpptools/discussions/12079

In 1.19.4 and newer, the language status seems to be updating too quickly and causing a strobing effect.  We may need to set a minimum duration for how long we keep the status up.

bobbrow:
Fixes: #12084

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -630,11 +630,11 @@ class ClientModel {
         this.isInitializingWorkspace = new DataBinding<boolean>(false);
         this.isIndexingWorkspace = new DataBinding<boolean>(false);
 
-        // The following elements add a delay of 300ms before notitfying the UI that the icon can hide itself.
-        this.isParsingWorkspace = new DataBinding<boolean>(false, 300, false);
-        this.isParsingWorkspacePaused = new DataBinding<boolean>(false, 300, false);
-        this.isParsingFiles = new DataBinding<boolean>(false, 300, false);
-        this.isUpdatingIntelliSense = new DataBinding<boolean>(false, 300, false);
+        // The following elements add a delay of 500ms before notitfying the UI that the icon can hide itself.
+        this.isParsingWorkspace = new DataBinding<boolean>(false, 500, false);
+        this.isParsingWorkspacePaused = new DataBinding<boolean>(false, 500, false);
+        this.isParsingFiles = new DataBinding<boolean>(false, 500, false);
+        this.isUpdatingIntelliSense = new DataBinding<boolean>(false, 500, false);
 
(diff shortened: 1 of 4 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -3741,7 +3741,8 @@ export class DefaultClient implements Client {
             // without being opened because otherwise users may not realize that
             // the header had changed (unless they view source control differences).
             await vscode.window.showTextDocument(headerFormatUriAndRanges[0].uri, {
-                selection: headerReplaceEditRange, preserveFocus: false });
+                selection: headerReplaceEditRange, preserveFocus: false
```
bobbrow, line 3747: I didn't intend to format this file, but it looks like it needed it...

sean-mcmanus (dismissed):
FYI, the isRunningCodeAnalysis also has the potential to contribute to the C/C++ busy indicator, but from what I can tell, that always takes long enough, so it doesn't seem to need an explicit delay unless you think we should safe guard against that case anyway.

Extension/src/LanguageServer/dataBinding.ts
```diff
@@ -4,13 +4,37 @@
...
     private isActive: boolean = true;
+    private deferral?: Deferral;
 
-    constructor(value: T) {
-        this.value = value;
+    /**
+     * Bind an event to a value so that a data model can automatically update the UI when values change.
+     * Since values can change quickly and cause UI to flicker, an optional delay/trigger combination can
+     * be specified to prevent UI elements from appearing/disappearing too quickly.
+     * @param value The initial value in the binding
```
sean-mcmanus: This is missing a period while the other doc comments have them.

Colengms (changes requested):
Can we remove the existing 1s timers (in the callbacks for some of these events)?

Colengms (dismissed):

sean-mcmanus (approved):

sean-mcmanus (changes requested):
Your latest change appears to "regress" the scenario where the icons are pinned compared to 1.19.6. I think the regression is caused by the removal of the existing timers (your original change didn't repro the issue), i.e. the result is that the pinned status text changes pretty fast, compared to slowly in 1.19.6.

Should the text change be changed back to the 1s timer as 1.19.6 -- the visual result looks better to me.

sean-mcmanus (changes requested):
@bobbrow I'm seeing IntelliSense status being stuck at "Ready".

sean-mcmanus (approved):

Colengms (approved):

PR #12131, opened 2024-03-21
Add a maximum search results for goto symbol

fearthecowboy:
Updates to goto symbol experiment - add a `c_cpp.maxSymbolSearchResults` setting to cap the maximum number of results returned to the user

Needs the native code PR as well.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3079,10 +3079,7 @@
                         "scope": "machine"
                     },
                     "C_Cpp.maxSymbolSearchResults": {
-                        "type": [
-                            "integer",
-                            "null"
-                        ],
+                        "type": "integer",
                         "markdownDescription": "%c_cpp.configuration.maxSymbolSearchResults.markdownDescription%",
                         "default": 5000,
                         "minimum": 500,
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -3078,6 +3078,17 @@
                         "maximum": 65536,
                         "scope": "machine"
                     },
+                    "C_Cpp.maxSymbolSearchResults": {
+                        "type": [
+                            "integer",
+                            "null"
+                        ],
+                        "markdownDescription": "%c_cpp.configuration.maxSymbolSearchResults.markdownDescription%",
+                        "default": 5000,
+                        "minimum": 500,
+                        "maximum": 10000,
+                        "scope": "window"
+                    },
                     "C_Cpp.intelliSense.maxCachedProcesses": {
                         "type": [
                             "integer",
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -67,7 +67,7 @@
         ]
     },
     "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
-        "message": "The maximum number of results to show for Go to Symbol in Workspace. The default is `5000`.",
+        "message": "The maximum number of results to show for 'Go to Symbol in Workspace'. The default is `5000`.",
         "comment": [
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -67,7 +67,7 @@
         ]
     },
     "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
-        "message": "The maximum number of results to show for Go to Symbol. The default is 5000.",
+        "message": "The maximum number of results to show for Go to Symbol in Workspace. The default is `5000`.",
         "comment": [
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -67,7 +67,7 @@
         ]
     },
     "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
-        "message": "The maximum number of results to show for GoTo Symbol. The default is 5000.",
+        "message": "The maximum number of results to show for Go to Symbol. The default is 5000.",
         "comment": [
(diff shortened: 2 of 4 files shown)


Extension/package.nls.json
```diff
@@ -66,6 +66,12 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
+    "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
+        "message": "The maximum number of results to show for GoTo Symbol. The default is 5000.",
```
bobbrow: The command is called "Go to Symbol", so let's use that capitalization.

It doesn't look like we use markdown for this setting, so maybe a normal description is ok, but I think it's ok to leave it if you want.

Extension/package.json
```diff
@@ -3078,6 +3078,17 @@
                         "maximum": 65536,
                         "scope": "machine"
                     },
+                    "C_Cpp.maxSymbolSearchResults": {
+                        "type": [
+                            "integer",
+                            "null"
```
bobbrow: Is `null` actually ok for this setting?  I would recommend removing that.

fearthecowboy: I copied the definition from other settings - null/unspecified means 'default'. Is that not what the other settings do?

Extension/package.json
```diff
@@ -3078,6 +3078,17 @@
...
+                        ],
+                        "markdownDescription": "%c_cpp.configuration.maxSymbolSearchResults.markdownDescription%",
+                        "default": 5000,
+                        "minimum": 500,
+                        "maximum": 10000,
```
sean-mcmanus, line 3086: Can the 10k max be increased to 20000, i.e. for users with fast computers, or is that something we don't want for some reason? If the user's query hits a lot of symbol exclusions then the the initial query might need to be larger to get results.

fearthecowboy: I can't see a reason why 20000 results is something the user would **actually** want. 

What happens if you get 20000 results? You're not going to find what you're looking for. 

That's 800 pages of results in the UI.

Extension/package.json
bobbrow: null does not mean default.  The default is 5000 and VS Code will return that value if the user has not set the setting.  The way this is currently authored requires the user to actively set the value to `null` and we're saying that's valid.  I don't think it should be valid.

Extension/package.nls.json
```diff
@@ -66,6 +66,12 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
+    "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
+        "message": "The maximum number of results to show for Go to Symbol. The default is 5000.",
```
sean-mcmanus: Go to Symbol->'Go to Symbol in Workspace'

sean-mcmanus: 5000->\`5000\`

Extension/package.json
sean-mcmanus: Yeah, I think `null` is fine/expected.

fearthecowboy: I've removed `null`

Extension/package.json
sean-mcmanus: Yeah, I'm thinking about the initial query being 20k but results in a lot less. The user probably can't tell 10k vs. 20k actual results.

bobbrow: I don't think we should allow 20,000. VS Code said in the past to cap it at 10000.

sean-mcmanus: sure, I'm fine with that

Extension/package.nls.json
```diff
@@ -66,6 +66,12 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
+    "c_cpp.configuration.maxSymbolSearchResults.markdownDescription": {
+        "message": "The maximum number of results to show for Go to Symbol in Workspace. The default is `5000`.",
```
sean-mcmanus: Can you add '' single quotes around this command? e.g. `'Go to Symbol in Workspace'` That is the style I've seen most common in the VS Code setting descriptions and we have another example in our package.json.

Extension/src/LanguageServer/settings.ts
```diff
@@ -313,6 +314,7 @@ export class CppSettings extends Settings {
 
     public get maxConcurrentThreads(): number | undefined | null { return super.Section.get<number | null>("maxConcurrentThreads"); }
     public get maxMemory(): number | undefined | null { return super.Section.get<number | null>("maxMemory"); }
+    public get maxSymbolSearchResults(): number | undefined | null { return super.Section.get<number | null>("maxSymbolSearchResults"); }
```
sean-mcmanus: My preference would still be use `null` as the default, but otherwise shouldn't the `null` type be removed or is that valid still?

bobbrow: `null`/`"default"` is what we use when we want a dynamic default based on some machine or OS properties.  I don't believe we need a dynamic default here because we don't check the machine hardware for a reason to increase/decrease the default setting value.

But yes, we should remove `null` here.

fearthecowboy: removed null

bobbrow (approved):

Extension/src/LanguageServer/settings.ts
```diff
@@ -142,6 +142,7 @@ export interface SettingsParams {
     maxConcurrentThreads: number | null | undefined;
     maxCachedProcesses: number | null | undefined;
     maxMemory: number | null | undefined;
+    maxSymbolSearchResults: number | null | undefined;
```
sean-mcmanus, line 145: Is this supposed to be null?

bobbrow: no. @fearthecowboy, can you fix this one too?

PR #12135, opened 2024-03-22
Update to clang-tidy 18.1.2

Colengms:


Extension/package.json
```diff
@@ -2553,8 +2730,122 @@
                                 "cert-pos47-c",
                                 "cert-sig30-c",
                                 "cert-str34-c",
+                                "Clang Static Analyzer core.*",
```
sean-mcmanus: Are these supposed to be added? Aren't these the same as the existing clang-analyzer-*?

Colengms: You tell me.  :)  They were listed.

sean-mcmanus (approved):

PR #12148, opened 2024-03-27
Component Governance pipeline for vscode-cpptools

bobbrow:
Replacing the Classic pipeline with this one.  I need to check it in to main before I can start testing, so this is just the starting iteration.


sean-mcmanus (approved):
This fixes the Notices.txt generation?

bobbrow:
> This fixes the Notices.txt generation?

Once I get it working, yes, I hope so.🤞

Issue #12089, opened 2024-03-11
Non-existent relative path variables (i.e {workspaceFolderBasename}) do not squiggle even though it results in an incorrect path


browntarik:
![image](https://github.com/microsoft/vscode-cpptools/assets/111317156/4e18ec52-cfdf-403f-aeda-14bdd092b9d5)

{workspaceFolderBasename} does not squiggle though it expands to an incorrect path

sean-mcmanus:
This issue only seems to repro for variables that result in a relative path. I'm seeing it give a squiggle with a full path, e.g. ${workspaceFolder}.

PR #12188, opened 2024-04-10
Remove redundant variable resolution in handleSquiggles and fix error messaging for env and delimited paths under include paths in c_cpp_properties.json

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12089


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1896,7 +1896,6 @@ export class CppProperties {
         }
 
         // Resolve and split any environment variables
-        paths = this.resolveAndSplit(paths, undefined, this.ExtendedEnvironment);
         compilerPath = util.resolveVariables(compilerPath, this.ExtendedEnvironment).trim();
```
bobbrow: It may also be possible to remove the call to `util.resolveVariables` since I think it is done in `this.resolvePath`.  You'd need to keep the `trim` though.

browntarik: That's a good point, it does look like the call to resolveVariables is reundant.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1896,9 +1896,7 @@ export class CppProperties {
         }
 
         // Resolve and split any environment variables
-        paths = this.resolveAndSplit(paths, undefined, this.ExtendedEnvironment);
-        compilerPath = util.resolveVariables(compilerPath, this.ExtendedEnvironment).trim();
-        compilerPath = this.resolvePath(compilerPath);
+        compilerPath = this.resolvePath(compilerPath).trim();
```
bobbrow, line 1898: technically the `trim` goes on the input, not the output to keep the same behavior as before.  I don't know if it actually matters, but my rule of thumb is to change as little as possible to avoid unforeseen regressions.

```ts
compilerPath = this.resolvePath(compilerPath.trim());
```

bobbrow: I take that back.  This changes the behavior too, just differently.  Feel free to ignore.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1896,9 +1896,7 @@ export class CppProperties {
         }
 
         // Resolve and split any environment variables
-        paths = this.resolveAndSplit(paths, undefined, this.ExtendedEnvironment);
```
bobbrow, line 1899: Actually, we don't split the paths later down the function, so we still need to do it at some point.  We can do it in the validation loop since you'll be able to keep the original string there.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1896,9 +1896,7 @@ export class CppProperties {
         }
 
         // Resolve and split any environment variables
```
bobbrow: update the comment

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2001,6 +1999,9 @@ export class CppProperties {
                 continue;
             }
 
+            const originalPath: string = curPath;
+            curPath = this.resolveAndSplit([curPath], undefined, this.ExtendedEnvironment, true, false)[0];
```
bobbrow: If there is more than one path in the string, this looks like it throws away all but the first one.  You probably need a nested loop here.

bobbrow (changes requested):

Extension/src/LanguageServer/configurations.ts
browntarik: This is part of an existing issue that isn't handled correctly. I think the more correct thing to do is not provide squiggle handling at all for split paths/environment variables

bobbrow: If we are not handling cases with multiple paths at all, the right thing to do here is to check the length of the output array.  If it's greater than 1, then skip this path. (and add a comment why)

But if we did actually look at the first one only, the current approach is acceptable (but still would benefit from a comment).

browntarik: Unless we want to provide a fix for that. Then in that case we want to try to map expanded paths and variables back to their original values, this should probably be part of a separate fix/issue as more design decisions may need to be made.

bobbrow: You are welcome to fix it with this PR.  You have the original path and should be able to map back into it to draw the squiggle.

browntarik: This should now correctly squiggle the following while preserving existing behavior:

${workspaceFolderBasename}:
Screenshot 2024-04-11 164321

delimited paths:
Screenshot 2024-04-11 164327

paths within an environment variable:
Screenshot 2024-04-11 164334

browntarik: Just not sure what to write in the error message

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,13 +2067,29 @@ export class CppProperties {
...
                         }
-                        message = localize('cannot.find2', "Cannot find \"{0}\".", resolvedPath);
+                        // If there are incorrect paths and the first match includes "env:", split the string and check for environment variable.
+                        if (incorrectExpandedPaths.length > 0 && configMatches[0].includes("env:")) {
+                            const splitString = configMatches[0].split("env:");
+                            if (splitString[1]) {
+                                const matchResult = splitString[1].match(/\w+/);
+                                if (matchResult) {
+                                    const envVar = matchResult[0];
+                                    message = localize('cannot.find2', "Cannot find {0} in {1} environment variable.", incorrectExpandedPaths.map(s => `"${s}"`).join(', '), envVar);
```
bobbrow: This message needs a different id.  'cannot.find2' has a different string assigned already.

Consider rewording: "Cannot find {0} in environment variable: {1}"

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,13 +2067,29 @@ export class CppProperties {
...
+                                }
+                            }
+                        // If there are incorrect paths but no "env:" in the first match, generate a message indicating the paths cannot be found.
+                        } else if (incorrectExpandedPaths.length > 0) {
+                            message = localize('cannot.find2', "Cannot find {0}.", incorrectExpandedPaths.map(s => `"${s}"`).join(', '));
+                        } else {
+                            message = localize('cannot.find2', "Cannot find \"{0}\".", resolvedPath);
```
bobbrow: You can't have two different messages with the same localize id.  Localizers mess up the quotes anyway, so you should do this instead:

```ts
else {
  let badPath = "";
  if (incorrectExpandedPaths.length > 0) {
    badPath = incorrectExpandedPaths.map(s => `"${s}"`).join(', ');
  } else {
    badPath = `"${resolvedPath}"`;
  }
  message = localize('cannot.find2', "Cannot find {0}", badPath);
}
```

Or you could push `resolvedPath` into `incorrectExpandedPaths` and not have a conditional here.  I haven't reviewed the whole thing, but am wondering why you have two different variables tracking essentially the same thing.  If possible, push all invalid paths into the array so you only have to work with one variable.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2005,7 +2002,27 @@ export class CppProperties {
...
                 continue;
             }
+
+            // Cache the original value of each path to include any variables or path delimiting.
+            const originalPath: string = curPath;
+            const expandedPaths = this.resolveAndSplit([curPath], undefined, this.ExtendedEnvironment, true, false);
             let pathExists: boolean = true;
+            const incorrectExpandedPaths: string[] = [];
+
+            if (expandedPaths.length > 1) {
```
bobbrow: Can you generalize the algorithm so that you don't have to do something special for multiple paths vs. single paths?  I would think you could deal with them the same way (single path is just an array with length == 1).

bobbrow (dismissed):

Extension/src/LanguageServer/configurations.ts
browntarik: This algorithm should be the same regardless of the number of paths, thanks for pointing this out.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2005,7 +2002,27 @@ export class CppProperties {
             if (!resolvedPath) {
                 continue;
             }
+
+            // Cache the original value of each path to include any variables or path delimiting.
+            const originalPath: string = curPath;
+            const expandedPaths = this.resolveAndSplit([curPath], undefined, this.ExtendedEnvironment, true, false);
```
sean-mcmanus: Can you add a type?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,13 +2067,33 @@ export class CppProperties {
...
+                        if (incorrectExpandedPaths.length > 0 && configMatches[0].includes("env:")) {
+                            const splitString = configMatches[0].split("env:");
+                            if (splitString[1]) {
+                                const matchResult = splitString[1].match(/\w+/);
+                                if (matchResult) {
+                                    const envVar = matchResult[0];
+                                    message = localize('cannot.find3', "Cannot find {0} in environment variable: {1}.", incorrectExpandedPaths.map(s => `"${s}"`).join(', '), envVar);
+                                }
+                            }
+                        // If there are incorrect paths but no "env:" in the first match, generate a message indicating the paths cannot be found.
```
sean-mcmanus: Can you move this comment below or to the right of the `} else {`?

sean-mcmanus: Umm...but the comment doesn't match what the "else" condition 100%. Should the `else` be an `else if`?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,13 +2067,33 @@ export class CppProperties {
...
+                        } else {
+                            let badPath = "";
+                            if (incorrectExpandedPaths.length > 0) {
```
sean-mcmanus: Should this be > 1?

sean-mcmanus: Oh, it seems fine.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Oh, the comment and else seems fine.

sean-mcmanus:
This causes the valid squiggle to disappear. I'm not sure if the root issue could be hit in real world scenarios or not, but it's a regression.

image

1.20.1
image

sean-mcmanus (dismissed):
See my comment.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,13 +2071,33 @@ export class CppProperties {
...
+                                badPath = incorrectExpandedPaths.map(s => `"${s}"`).join(', ');
+                            } else {
+                                badPath = `"${resolvedPath}"`;
```
bobbrow: I don't think `resolvedPath` is a needed variable anymore.  `resolveAndSplit` and the loop on the `expandedPaths` that it returns should be doing all the stuff that we used to do with `resolvedPath` instead. It feels like there are two distinct, competing implementations in these 100 lines of code when there should be just one.  The `resolvedPath` (which is the single path case) should be one of the entries in `expandedPaths`.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: @browntarik Was this resolved? Is the comment located at the correct spot?

sean-mcmanus:
@browntarik I'm seeing this:

image

sean-mcmanus (dismissed):
See my comment above.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2001,28 +1998,39 @@ export class CppProperties {
...
+                }
+            } else {
+                continue;
             }
+
             // Normalize path separators.
-            if (path.sep === "/") {
-                resolvedPath = resolvedPath.replace(/\\/g, path.sep);
+            if (path.sep === "/" && expandedPaths.length > 0) {
+                expandedPaths[0] = expandedPaths[0].replace(/\\/g, path.sep);
```
bobbrow: Why only do this for the first one?  I feel like this should be in the loop for `expandedPaths`

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1989,6 +1985,7 @@ export class CppProperties {
 
         // Validate paths
         for (const curPath of paths) {
+
```
bobbrow: delete empty newline

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2001,28 +1998,39 @@ export class CppProperties {
                 continue;
             }
 
-            let resolvedPath: string = this.resolvePath(curPath);
-            if (!resolvedPath) {
-                continue;
-            }
+            // Cache the original value of each path to include any variables or path delimiting.
+            const originalPath: string = curPath;
```
bobbrow: I don't think you need this variable `originalPath`

Extension/src/LanguageServer/configurations.ts
bobbrow: No, the comment should be in the block that it describes.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2041,6 +2049,11 @@ export class CppProperties {
                     }
                 }
             }
+
+            if (incorrectExpandedPaths.length > 0) {
+                pathExists = false;
```
bobbrow: You could declare `pathExists` as `const` at this location in the file instead of declaring it earlier with a default, incorrect value on line 2004.
```ts
const pathExists: boolean = incorrectExpandedPaths.length === 0;
```

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,28 +2063,48 @@ export class CppProperties {
...
+                            message = localize('cannot.find2', "Cannot find {0}", badPath);
+                        }
+                        // Increment the count of non-existent paths.
                         newSquiggleMetrics.PathNonExistent++;
                     } else {
                         // Check for file versus path mismatches.
                         if ((curOffset >= forcedIncludeStart && curOffset <= forcedeIncludeEnd) ||
                                 (curOffset >= compileCommandsStart && curOffset <= compileCommandsEnd)) {
-                            if (util.checkFileExistsSync(resolvedPath)) {
+                            if (util.checkFileExistsSync(expandedPaths[0])) {
```
bobbrow: Might be good to comment that we only check the first one because multi-path is not allowed for `forcedInclude` and `compileCommands`

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,28 +2063,48 @@ export class CppProperties {
...
                                 continue;
                             }
-                            message = localize("path.is.not.a.file", "Path is not a file: {0}", resolvedPath);
+                            message = localize("path.is.not.a.file", "Path is not a file: {0}", expandedPaths[0]);
                             newSquiggleMetrics.PathNotAFile++;
                         } else {
-                            if (util.checkDirectoryExistsSync(resolvedPath)) {
+                            if (util.checkDirectoryExistsSync(expandedPaths[0])) {
```
bobbrow: I believe all expandedPaths should be checked here, not just the first one.

browntarik:
> @browntarik I'm seeing this:
> 
> image

Could this be due to your path separator setting? I am not reproing this on windows.

Extension/src/LanguageServer/configurations.ts
browntarik: You're right, I was still thinking about the old implementation.

Extension/src/LanguageServer/configurations.ts
bobbrow: You could also improve this case by checking the length of `expandedPaths` and if it's >1, have another error message.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2030,6 +2007,43 @@ export class CppProperties {
...
+            for (const [index, expandedPath] of expandedPaths.entries()) {
+                // Normalize path separators.
+                if (path.sep === "/" && expandedPaths.length > 0) {
+                    expandedPaths[index] = expandedPath.replace(/\\/g, path.sep);
+                } else {
+                    expandedPaths[index] = expandedPath.replace(/\//g, path.sep);
+                }
+            }
+
+            // Iterate through the text and apply squiggles.
```
sean-mcmanus, line 2044: Is this comment at the wrong spot? Usually "iterate" comments would occur right before a loop.

sean-mcmanus: It looks like that comment is pre-existing.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2030,6 +2007,43 @@ export class CppProperties {
             const pattern: RegExp = new RegExp(`"[^"]*?(?<="|;)${escapedPath}(?="|;).*?"`, "g");
             const configMatches: string[] | null = curText.match(pattern);
 
+            const expandedPaths: string[] = this.resolveAndSplit([curPath], undefined, this.ExtendedEnvironment, true, false);
+            const incorrectExpandedPaths: string[] = [];
+
+            if (expandedPaths.length > 0) {
+                if (this.rootUri) {
+                    for (const [index, expandedPath] of expandedPaths.entries()) {
+
```
sean-mcmanus, line 2016: Bob might complain about this blank line here. We normally don't have blank lines at the start of a block.

Extension/src/LanguageServer/configurations.ts
browntarik: image

This was an existing comment before my changes. It seems appropriate.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2030,6 +2007,43 @@ export class CppProperties {
...
+                        }
+
+                        const checkPathExists: any = util.checkPathExistsSync(expandedPaths[index], this.rootUri.fsPath + path.sep, isWindows, false);
+                        if (!checkPathExists.pathExists) {
+                            // If there are multiple paths, store any non-existing paths to squiggle later on.
+                            incorrectExpandedPaths.push(expandedPaths[index]);
+                        }
+                    }
+                }
+            } else {
```
sean-mcmanus: This is easier to read if the `continue` code is moved up.

```
if (expandedPaths.length === 0) {
    continue;

if (this.rootUri) {
```

i.e. do an early continue. That also makes the block move to the left more.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2030,6 +2007,43 @@ export class CppProperties {
...
+                continue;
+            }
+
+            const pathExists: boolean = incorrectExpandedPaths.length === 0;
+
+            for (const [index, expandedPath] of expandedPaths.entries()) {
+                // Normalize path separators.
+                if (path.sep === "/" && expandedPaths.length > 0) {
```
sean-mcmanus: This condition seems incorrect. Should it be

```
if (expandedPaths.length > 0) {
    if (path.sep === "/") {
    ...
    } else {
    ...
    }
}
```

sean-mcmanus: umm...wait, I think you can get rid of the `expandedPaths.length > 0` check because it won't be hit since it's iterating over expandedPaths.entries().

sean-mcmanus: Well...is the check supposed to be `expandedPath.length > 0`?

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: @bobbrow I don't understand, are you saying the comment is located at the correct spot or not?

sean-mcmanus: It says `generate a message indicating the paths cannot be found.` and then no code comes after it.

bobbrow: @sean-mcmanus I was answering your question that the comment was not at the correct spot and it was not resolved.
```ts
if (condition) {
  [code]

// Comment about `else` case should not be here because it's in the `if` block
} else {
  // Comment about `else` case should be here inside the `else` block.
}

File: Extension/src/LanguageServer/configurations.ts (hunk 19 above)
Comment (sean-mcmanus) [reply Extension/src/LanguageServer/configurations.ts:2044]: ok

**Review by sean-mcmanus — DISMISSED** [review]
There's a bug with `**` at the end (recursive includes). Also, make sure to test `${workspaceFolder}/*` (which makes it non-recursive for the browse.path):
image

File: Extension/src/LanguageServer/configurations.ts (hunk 22 above)
Comment (browntarik) [reply Extension/src/LanguageServer/configurations.ts]: The expandedPath.length > 0 check should have been deleted when I added the for loop

File: Extension/src/LanguageServer/configurations.ts [hunk 23]
```diff
@@ -2030,6 +2007,42 @@ export class CppProperties {
             const pattern: RegExp = new RegExp(`"[^"]*?(?<="|;)${escapedPath}(?="|;).*?"`, "g");
             const configMatches: string[] | null = curText.match(pattern);
 
+            const expandedPaths: string[] = this.resolveAndSplit([curPath], undefined, this.ExtendedEnvironment, true, true);
+            const incorrectExpandedPaths: string[] = [];
+
+            if (expandedPaths.length < 0) {
```
sean-mcmanus: <=

sean-mcmanus:
@browntarik I'm not getting squiggles with env variables? I thought that was supposed to be handled?

image

sean-mcmanus (dismissed):
.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1840,7 +1840,7 @@ export class CppProperties {
         if (this.prevSquiggleMetrics.get(currentConfiguration.name) === undefined) {
             this.prevSquiggleMetrics.set(currentConfiguration.name, { PathNonExistent: 0, PathNotAFile: 0, PathNotADirectory: 0, CompilerPathMissingQuotes: 0, CompilerModeMismatch: 0 });
         }
-        const newSquiggleMetrics: { [key: string]: number } = { PathNonExistent: 0, PathNotAFile: 0, PathNotADirectory: 0, CompilerPathMissingQuotes: 0, CompilerModeMismatch: 0 };
+        const newSquiggleMetrics: { [key: string]: number } = { PathNonExistent: 0, PathNotAFile: 0, PathNotADirectory: 0, CompilerPathMissingQuotes: 0, CompilerModeMismatch: 0, MultiplePathsNotAllowed:0 };
```
bobbrow: If you're adding a new telemetry metric, you need to also send it with the telemetry. Look around line 2158.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2050,29 +2064,58 @@ export class CppProperties {
...
                         }
-                        message = localize('cannot.find2', "Cannot find \"{0}\".", resolvedPath);
+                        let badPath = "";
+                        if (incorrectExpandedPaths.length > 0) {
+                            badPath = incorrectExpandedPaths.map(s => `"${s}"`).join(', ');
+                        } else {
+                            badPath = `"${expandedPaths[0]}"`;
+                        }
+                        message = localize('cannot.find2', "Cannot find {0}", badPath);
+                        // Increment the count of non-existent paths.
```
bobbrow: This comment can be deleted. We don't add a comment like this on the other increments.

sean-mcmanus (approved):

PR #12190, opened 2024-04-12
Crash handler update

sean-mcmanus:


Extension/src/LanguageServer/extension.ts
```diff
@@ -958,8 +958,12 @@ function reportMacCrashes(): void {
     }
 }
 
+export function usesCrashHandler(): boolean {
```
bobbrow, line 961: Is this function needed?  `watchForCrashes` checks the existence of a `crashDirectory` before doing anything, so it seems like we could just always call preIntialize to see if there is a `crashDirectory` and let the logic live in the cpptools binary.

sean-mcmanus: It avoids sending a PreInitializationRequest, which won't exist for the unsupported platforms. It's certainly possible to change, but I don't see a reason too -- it'd be a "worse".

Colengms (approved):

PR #12216, opened 2024-04-18
Fix SettingsTracking treating numerical loggingLevel as invalid.

sean-mcmanus:


Extension/src/LanguageServer/settingsTracker.ts
```diff
@@ -103,7 +103,7 @@ export class SettingsTracker {
                         return val;
                     }
                     const curEnum: any[] = curSetting["enum"];
-                    if (curEnum && curEnum.indexOf(val) === -1) {
+                    if (curEnum && curEnum.indexOf(val) === -1 && key !== "loggingLevel") {
```
bobbrow: If we do this, we also need to ensure that the value is an approved number so that we don't accidentally log PII for this setting.  Anything other than a valid enum and a valid number are "invalid" and should be treated as such.

Colengms (approved):

PR #12247, opened 2024-04-23
Improvements to crash call stack logging.

sean-mcmanus:
1. Avoid logging duplicates for repeated crashes in a row.
2. Only log if the loggingLevel is >= 1 (not None).
3. Print the call stack info to the C/C++ Crash Call Stacks output pane with a URL to report a bug.
4. Add the original crash date/time.
5. Stop truncating the printed logging to 8k characters.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1206,11 +1206,7 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, cr
         if (lines.length >= 6 && util.getNumericLoggingLevel(settings.get<string>("loggingLevel")) >= 1) {
             const out: vscode.OutputChannel = getCrashCallStacksChannel();
             out.appendLine("");
-            out.append(localize({ key: "crash.callstack.info", comment: [
-                "{0} is the process name: either cpptools or cpptools-srv. {1} is the crash date and time, e.g. 4/24/2024, 4:44:52 PM. {2} is a signal name, e.g. SIGSEGV. The call stack lines appear on a new line after the ':'" ]},
-            "The {0} process crashed on {1} from signal {2} with call stack:",
-            isCppToolsSrv ? "cpptools-srv" : "cpptools", crashDate.toLocaleString(), signalType));
-            out.appendLine(crashCallStack);
+            out.appendLine(`${isCppToolsSrv ? "cpptools-srv" : "cpptools"}\n${crashDate.toLocaleString()}\n${signalType}${crashCallStack}`);
         }
     }
 
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -17,6 +17,7 @@ import { TargetPopulation } from 'vscode-tas-client';
 import * as which from 'which';
 import { logAndReturn } from '../Utility/Async/returns';
 import * as util from '../common';
+import { Logger, getOutputChannelLogger } from '../logger';
 import { PlatformInformation } from '../platform';
 import * as telemetry from '../telemetry';
 import { Client, DefaultClient, DoxygenCodeActionCommandArguments, openFileVersions } from './client';
@@ -38,6 +39,7 @@ export const configPrefix: string = "C/C++: ";
 
 let prevMacCrashFile: string;
 let prevCppCrashFile: string;
+let prevCppCrashData: string = "";
 export let clients: ClientCollection;
 let activeDocument: vscode.TextDocument | undefined;
 let ui: LanguageStatusUI;
@@ -1126,7 +1128,8 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, er
 
     const lines: string[] = data.split("\n");
     let addressData: string = ".\n.";
-    data = (crashFile.startsWith("cpptools-srv") ? "cpptools-srv.txt" : crashFile) + "\n";
+    const isCppToolsSrv: boolean = crashFile.startsWith("cpptools-srv");
+    data = (isCppToolsSrv ? "cpptools-srv.txt" : crashFile) + "\n";
     const filtPath: string | null = which.sync("c++filt", { nothrow: true });
     const isMac: boolean = process.platform === "darwin";
     const startStr: string = isMac ? " _" : "<";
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -1199,8 +1200,21 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, er
         data = data.substring(0, 8191) + "…";
     }
 
-    console.log(`Crash call stack:\n${data}`);
-    logCppCrashTelemetry(data, addressData);
+    if (data !== prevCppCrashData) {
+        prevCppCrashData = data;
+        logCppCrashTelemetry(data, addressData);
```
bobbrow: I think we always want to log the telemetry.  If you want, you can add another property that indicates whether a report is a duplicate or not for your own filtering, but I would like to have accurate counts for this.

sean-mcmanus: ok

Extension/src/LanguageServer/extension.ts
```diff
@@ -1199,9 +1200,23 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, er
...
+    if (data !== prevCppCrashData) {
+        prevCppCrashData = data;
+
+        const settings: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("C_Cpp", null);
+        if (util.getNumericLoggingLevel(settings.get<string>("loggingLevel")) >= 1) {
+            console.log(`Crash call stack:\n${data}`);
+            if (lines.length >= 6) {
+                const out: Logger = getOutputChannelLogger();
+                out.appendLine(localize({ key: "crash.callstack.available", comment: [ "{0} is the process name: either cpptools or cpptools-srv"]},
+                    "A {0} crash call stack for use in bug reporting is available in the 'Console' tab after using the 'Toggle Developer Tools' command.",
```
Colengms: @bobbrow Would you mind if the crash stacks were logged directly to the C/C++ output log instead of console.log?  I think we'd get a lot more stacks from users that way.  Talking to Sean about it, he thought you may have had a reason we shouldn't do that?  If we do that, also, should it be Error or Debug level?

bobbrow: I didn't weigh in on the decision at all.  I'm ok with logging to the Output window with "Error" level.

bobbrow: (just don't focus the Output automatically)

sean-mcmanus: @bobbrow Okay, sounds good -- I remember talking about it a long time ago.

bobbrow: Ok. I don't remember so that means I didn't have a strong opinion.

sean-mcmanus: I put it in it's own channel to avoid the issue with it being interleaved/hidden between the list of compiler args to cpptools-srv.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1174,32 +1178,48 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, er
...
+    if (crashCallStack !== prevCppCrashCallStackData) {
+        prevCppCrashCallStackData = crashCallStack;
+
+        const settings: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("C_Cpp", null);
+        if (lines.length >= 6 && util.getNumericLoggingLevel(settings.get<string>("loggingLevel")) >= 1) {
+            const out: vscode.OutputChannel = getCrashCallStacksChannel();
+            out.appendLine("");
+            out.append(localize({ key: "crash.callstack.info", comment: [
+                "{0} is the process name: either cpptools or cpptools-srv. {1} is the crash date and time, e.g. 4/24/2024, 4:44:52 PM. {2} is a signal name, e.g. SIGSEGV. The call stack lines appear on a new line after the ':'" ]},
+            "The {0} process crashed on {1} from signal {2} with call stack:",
```
sean-mcmanus: @bobbrow @Colengms Is the "sentence" format okay or would an undescriptive version be better, with the description being in the one-time description at that top?

```txt
The crashing process name, date/time, signal, and call stack are below:

cpptools
4/24/2024, 4:44:52 PM
SIGSEGV
call_stacks...

cpptools-srv
4/25/2024, 4:44:52 PM
SIGSEGV
call_stacks2...
```

Extension/src/logger.ts
```diff
@@ -98,6 +99,15 @@ export function getDiagnosticsChannel(): vscode.OutputChannel {
     return diagnosticsChannel;
 }
 
+export function getCrashCallStacksChannel(): vscode.OutputChannel {
+    if (!crashCallStacksChannel) {
+        crashCallStacksChannel = vscode.window.createOutputChannel(localize("c.cpp.crash.call.stacks.title", "C/C++ Crash Call Stacks"));
+        crashCallStacksChannel.appendLine(localize("c.cpp.crash.call.stacks.description",
+            "The information below could be helpful to provide in a C/C++ extension bug report:"));
```
sean-mcmanus: Would mentioning `on GitHub` (and/or a `https://github.com/microsoft/vscode-cpptools/issues` URL) be better?

Extension/src/LanguageServer/extension.ts
bobbrow: This suggestion seems to make it easier for the localization team.  Either way is fine with me though.

As a general recommendation, I avoid putting punctuation like the trailing `:` in the strings (when it makes sense) so that they don't have the ability to change the character.

Extension/src/logger.ts
bobbrow: Yes, a URL is helpful.  I would also change `could` to `would` since we'd like to link crash reports to actual code as often as we can.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Sure, I'll go with the one-time description, but I'm not understanding your suggestion in regards to ":" -- they always have the potential to change that punctuation and that seems fine to me, since other languages have different punctuation guidelines.

Extension/src/logger.ts
sean-mcmanus: @bobbrow Sure, I can add the URL, which I think would allow users to click/navigate to it too. Using text "on GitHub" seems sort of superfluous given that the URL is a github.com one.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Oh, I think I'm getting rid of the trailing ":" due to re-wording anyways.

bobbrow (approved):

Colengms (approved):

PR #12355, opened 2024-06-06
Fixes to crash call stacks.

sean-mcmanus:
Fix various bugs with the crash call stack logging.
* Avoid / paths being sent (they get filtered out anyway).
* Handle a missing signal type.
* Handle failures calling c++filt.
* Handle potentially valid early frames.
* Skip ... early frames.
* Skip crash_handler/_sigtramp frames.


Colengms:
Could you provide a PR description?

sean-mcmanus:
> Could you provide a PR description?

Done

sean-mcmanus:
I might have another change for Mac-specific stuff.

Colengms (approved):

PR #12358, opened 2024-06-06
make debugging easier to inspect local var value

lukka:
This helps in inspecting the value of a local var

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1344,5 +1344,6 @@ export async function preReleaseCheck(): Promise<void> {
 }
 
 export async function getIncludes(maxDepth: number): Promise<any> {
-    return clients.ActiveClient.getIncludes(maxDepth);
+    const includes = await clients.ActiveClient.getIncludes(maxDepth);
+    return includes;
 }


Colengms (approved):

sean-mcmanus (changes requested):
This change shouldn't be required:

You should be able to set the breakpoint after the return and then look in the Variables window for the return value:

image

Does doing that work for you?

lukka:
> This change shouldn't be required:
> 
> You should be able to set the breakpoint after the return and then look in the Variables window for the return value:
> 
> image
> 
> Does doing that work for you?

It does not work for me, I had to have a local change as in this PR to be able to see the result. I am using latest  VSCode insiders.

sean-mcmanus (approved):
Ah, yeah, it's because you need to await the promise.

bobbrow:
> Ah, yeah, it's because you need to await the promise.

Wait, what?  This does not look like the correct pattern.  The caller of `getIncludes` should await, not `getIncludes` itself.  It's just going to wrap this back up in another Promise.  Just remove `async` from the signature and check that the caller is doing the `await`.

```ts
export function getIncludes(maxDepth: number): Promise<any> {
    return clients.ActiveClient.getIncludes(maxDepth);
}
```

lukka:
@bobbrow in this case we are returning always a completed Promise. Before the change, the Promise could be either resolved/unresolved, but the caller had to `await` it anyway. 
The caller is already doing the await, but it is in another extension, so it is easier for me to see the actual value in the debugger while running locally the cpptools extension.
I am open to further suggestions as well.

bobbrow:
Ok, I understand now. It was unclear to me what the purpose of this change was. This is for you (the developer) working on cpptools, not for our customers.

bobbrow:
For the future, I don't think it's common for us to commit debugging hacks (usually you would just revert this after getting the information you want), but it's fine either way.

lukka:
@bobbrow yes, I can also keep this change locally when I need it.

Issue #12384, opened 2024-06-17
Display additional Doxygen tags in tooltip documentation


MikeWeller:
### Feature Request

It looks like VSCode only shows a certain subset of Doxygen tags/commands in the tooltip documentation, e.g. `\param`, `\note` are rendered, while others are not.

In particular I would also like `@pre` (or `\pre`) (preconditions) to be visible. I see these as more important than `@note` which is currently displayed. In general it would be nice for all tags to be rendered in some way (though there may need to be a list defined somewhere to determine how multiple tags of the same name are combined into a list vs. not, for example).

Another tag I would like visible is `@see`, e.g. `@see fooNon() for a version of this method that does not frob its input`

bobbrow:
We allow you to select which tags you'd like to see in the structured comment tooltip if you'd like to see additional tags beyond the ones we chose as the defaults.

Can you try adding those tags to the `"C_Cpp.doxygen.sectionTags"` setting? It looks like we don't recognize "see" yet so no special header will be added to the tooltip, but we can use this issue as a reminder to add it.

![image](https://github.com/microsoft/vscode-cpptools/assets/12818240/0aaad08f-27d8-4158-8f40-83cc1a9d1072)
![image](https://github.com/microsoft/vscode-cpptools/assets/12818240/a6e1e596-ada7-489a-8bbe-f7d8c7e7f597)

MikeWeller:
@bobbrow ah, great! Yes this solves the issue with `pre`, so for me this is now a relatively minor issue with `see` not being recognized (and I guess there may be other official commands in https://doxygen.nl/manual/commands.html that people would want, though I haven't reviewed the list)

bobbrow:
You can also try setting `"C_Cpp.simplifyStructuredComments"` to false and it will do a more generic listing of all tags.

![image](https://github.com/microsoft/vscode-cpptools/assets/12818240/ab2b1763-70b0-405a-8de3-35d1a1428be7)

sean-mcmanus:
@MikeWeller Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.21.1

PR #12415, opened 2024-06-25
Add @see and @sa Doxygen Tags with the "See Also:" label.

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12384

The goal of this PR is to add additional section tags that will appear in the hover tooltip for Doxygen comments. Now users can use the @see and @sa doxygen tags in their code.

According the Doxygen documentation (https://www.doxygen.nl/manual/commands.html#cmdsa): "sa" is short for "See Also:" and is meant to cross reference other functions in the code. An example of this could be the following:

```cpp
/**
 * @brief Adds two integers.
 *
 * This function takes two integers as parameters and returns their sum.
 *
 * @param a First integer
 * @param b Second integer
 * @pre Sum of the integers
 * @sa subtract()
 */
int add(int a, int b) {
    return a + b;
}
```

Where \sa is referring to the subtract() function and puts it in context of the add function (see below):
image

### DIFF
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -214,7 +214,7 @@
         "text": "Returns:",
         "hint": "This label is for the return value description for a function. Usage example: 'Returns: Area of a shape.'"
     },
-    "sa_label": "See Also:",
+    "sa_label": "See also:",
     "since_label": "Since:",
     "template_parameters_label": "Template Parameters:",
     "test_label": "Test:",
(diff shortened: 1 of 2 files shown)


Extension/src/nativeStrings.json
```diff
@@ -214,6 +214,7 @@
         "text": "Returns:",
         "hint": "This label is for the return value description for a function. Usage example: 'Returns: Area of a shape.'"
     },
+    "sa_label": "See also:",
```
bobbrow, line 217: "Template Parameters" is capitalized, so this one probably should be too.  "See Also:"

browntarik: I will keep it lowercase to match existing examples of doxygen documentation 
(see: https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html).

sean-mcmanus (approved):

Issue #12413, opened 2024-06-25
[Regression] The 'Learn how to install a library for this header with vcpkg' option has disappeared


v-ericawu:
### Environment

- OS and Version: Windows/Linux/Mac OS
- VS Code Version: 1.85.0
- C/C++ Extension Version: 1.20.5

### Bug Summary and Steps to Reproduce

Bug Summary: The 'Learn how to install a library for this header with vcpkg' option has disappeared

**Steps to reproduce:**
1. Create a Test folder with .cpp file.
2. Open Test folder with VS Code.
3. Type error code in .cpp file, such as: **#include"xxxxx.h"** 
4. Hover over error squiggle and Click on 'Quick Fix'

**Expected behavior:** 'Learn how to install a library for this header with vcpkg' option appear
![image](https://github.com/microsoft/vscode-cpptools/assets/78774400/51e45fd0-58fb-46c6-bd64-ba5aad241309)

**Actual behavior:** 'Learn how to install a library for this header with vcpkg' option has disappeared from C/C++ extension 1.17.2

Latest C/C++ extension version: 1.20.5
![image](https://github.com/microsoft/vscode-cpptools/assets/78774400/26c2d252-7050-4712-8bde-85d1a08f5540)

C/C++ extension 1.17.2:
![image](https://github.com/microsoft/vscode-cpptools/assets/78774400/ad797a97-bb8d-47a9-9626-934f9f1cad51)

### Configuration and Logs

```shell
/
```

### Other Extensions

/

browntarik:
Thank you for reporting this issue. This behavior is actually by design as we only want to suggest this code action for headers that are supported by vcpkg. This does however bring to light a bug where this code action is not appearing even for such headers like `#include <zlib.h>`. 

It looks like we were incorrectly parsing our vcpkg header database. Please keep tracking this issue to be notified when a fix is checked in.

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.21.1

v-ericawu:
@sean-mcmanus Verified on C/C++ 1.21.1(pre-release) still repro this issue, so could you help to re-confirm it?
![image](https://github.com/user-attachments/assets/88c41da3-cfc4-4b21-bf2e-46b5d3740c54)

Colengms:
[...]

PR #12416, opened 2024-06-26
Strip '\r' from the end of each line when parsing the vcpkg database

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12413

We were incorrectly parsing our vcpkg header database when deciding to present the "Learn how to install a library for this header with vcpkg" code action.

The issue was due to not stripping the "\r" line endings that appear exclusively on Windows machines.

This fix will now strip the "\r\n" line endings on windows properly and if will only strip "\n" for linux/mac machines.

It should now display the following behavior on windows (see below):
Screenshot 2024-06-26 125711

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -109,7 +109,7 @@ async function lookupIncludeInVcpkg(document: vscode.TextDocument, line: number)
     if (!matches || !matches.length || !matches.groups) {
         return [];
     }
-    const missingHeader: string = matches.groups.includeFile.replace(/[\\\/]/g, '\\');
+    const missingHeader: string = matches.groups.includeFile.replace(/\//g, '\\');
 
     let portsWithHeader: string[] | undefined;
     const vcpkgDb: VcpkgDatabase = await vcpkgDbPromise;
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -113,7 +113,6 @@ async function lookupIncludeInVcpkg(document: vscode.TextDocument, line: number)
 
     let portsWithHeader: string[] | undefined;
     const vcpkgDb: VcpkgDatabase = await vcpkgDbPromise;
-    console.log("vcpkgDb: " + JSON.stringify(vcpkgDb));
     if (vcpkgDb) {
         portsWithHeader = vcpkgDb[missingHeader];
     }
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -106,7 +109,7 @@ async function lookupIncludeInVcpkg(document: vscode.TextDocument, line: number)
     if (!matches || !matches.length || !matches.groups) {
         return [];
     }
-    const missingHeader: string = matches.groups.includeFile.replace(/\//g, '\\');
+    const missingHeader: string = matches.groups.includeFile.replace(/[\\\/]/g, '\\');
```
bobbrow: I don't think this line needs to change. Do you know what was the problem here?  You just added backslash to the list of things to replace, but you'd just replace it with a backslash so that doesn't change the string.

browntarik: I see I thought the regex was slightly wrong, but I just misinterpreted how JavaScript applies backslashes.

Extension/src/LanguageServer/extension.ts
```diff
@@ -58,7 +58,11 @@ async function initVcpkgDatabase(): Promise<VcpkgDatabase> {
         const zip = new StreamZip.async({ file: util.getExtensionFilePath('VCPkgHeadersDatabase.zip') });
         try {
             const data = await zip.entryData('VCPkgHeadersDatabase.txt');
-            const lines = data.toString().split('\n');
+            // Strip '\r' and '\n' from the end of each line if they exist.
+            let lines = data.toString().split('\r\n');
+            if (lines.length === 1) {
```
bobbrow: Do we not know what the line endings are?  We should own this file and how it's formatted.  If there is a way to split with a regex, you could split on `\r?\n` instead.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Why is this needed? Isn't the changed code converting `\\` with `\\`?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: There's a `splitLines` function in `Extension/src/common.ts`.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Oh, I didn't see Bob's comment.

Extension/src/LanguageServer/extension.ts
```diff
@@ -58,6 +58,7 @@ async function initVcpkgDatabase(): Promise<VcpkgDatabase> {
         const zip = new StreamZip.async({ file: util.getExtensionFilePath('VCPkgHeadersDatabase.zip') });
         try {
             const data = await zip.entryData('VCPkgHeadersDatabase.txt');
+            // Strip '\r' and '\n' from the end of each line if they exist.
```
bobbrow: This comment seems out of place now

sean-mcmanus (approved):

PR #12470, opened 2024-07-12
Don't show warning notification when C_Cpp.getIncludes is disabled

benmcmorran:
Fixes the Copilot issue reported here https://github.com/orgs/community/discussions/131961. We only expect the command to be called by Copilot, so it should be fine to make the warning message less aggressive.

cc @lukka


Extension/src/LanguageServer/extension.ts
```diff
@@ -403,19 +403,26 @@ export function registerCommands(enabled: boolean): void {
...
         "IntelliSense-related commands cannot be executed when `C_Cpp.intelliSenseEngine` is set to `disabled`.");
-    return vscode.window.showWarningMessage(message);
+}
+
+function onDisabledCommand() {
+    return vscode.window.showWarningMessage(getDisabledCommandMessage());
+}
+
+async function onLowImpactDisabledCommand() {
+    console.log(getDisabledCommandMessage());
```
bobbrow: Do you even want to log at all? My vote would be to leave `onDisabledCommand` alone and just create your own 'no-op' handler for the disabled case.

bobbrow: (because I'm expecting that the console will be flooded with these messages by copilot and they won't be useful to anyone)

benmcmorran: Sounds good, I can simplify it.

sean-mcmanus: Yeah, I agree with Bob.

benmcmorran: Changed to no-op handler.

sean-mcmanus (approved):

benmcmorran:
@sean-mcmanus I don't have permission to actually merge this PR. Please merge when you have a chance.

bobbrow (approved):

Issue #4526, opened 2019-11-02
Launch.json creation fails with german locale


JBamberger:
Type: Debugger

**Describe the bug**

- OS and Version: Windows_NT x64 10.0.18362
- VS Code Version: 1.39.2
- C/C++ Extension Version: 0.26.1
- Other extensions you installed (and if the issue persists after disabling them): No other extensions are installed.

The extension cannot create the `launch.json` file with a debug configuration if the locale is set to German. Trying to create such a configuration results in the error message `Die Datei "launch.json" kann nicht im Ordner ".vscode" erstellt werden (Unexpected token $ in JSON at position 113).`, i.e. _The file "launch.json" cannot be created in the directory ".vscode"_. The issue is only present if the `launch.json` file doesn't exist already. If the file exists, the configuration is added but the template is no valid JSON.

**To Reproduce**
1. Set the locale to German
2. Open a C++ project without a launch.json file
3. Hit `F5`
3. Select `C++ (GDB/LLDB)` or `C++ (Windows)`

**Additional context**
If the `launch.json` file already exists the creation of new configurations works. The generated code looks as follows:
```json
{
    "name": "(Windows) Starten",
    "type": "cppvsdbg",
    "request": "launch",
    "program": "Programmnamen eingeben, z. B. "${workspaceFolder}/a.exe"",
    "args": [],
    "stopAtEntry": false,
    "cwd": "${workspaceFolder}",
    "environment": [],
    "externalConsole": false
}
```
The `program` value is no valid JSON because the quote marks are not escaped.

There are likely other strings facing the same problem. For example if I generate a configuration for `gdb` the result contains
```json
"setupCommands": [
    {
        "description": "Automatische Strukturierung und Einrückung für "gdb" aktivieren",
        "text": "-enable-pretty-printing",
        "ignoreFailures": true
    }
]
```
This also contains unescaped quote marks which cause problems.

The simplest fix would probably be to add the necessary escapes, e.g. replace `"Programmnamen eingeben, z. B. \"{0}\"",` with `"Programmnamen eingeben, z. B. \\\"{0}\\\""`, or removing the quotes entirely (as in all other languages I checked).

I'd be willing to contribute a fix but as far as I can tell, the localization files are generated. So if someone can point me to the source of the generation I could create a PR.

Colengms:
Hi @JBamberger .  Thanks for reporting this.  This looks like a bug in our recent localization work.

michelleangela:
[...]

PR #12471, opened 2024-07-12
Strip quotes from debug configurations and fix an issue with quotes in paths

bobbrow:
This should address a few CodeQL issues flagged in the repo. There was some code added to escape any double quotes that might be added from the generated debug configurations during the localization process. Since we don't really want those double quotes anyway, it should be acceptable to simply remove them.

While looking at some of the other issues flagged, I noticed that a helper function ends up thinking a quoted path is a relative path. Fixing that doesn't fix any other underlying assumptions we make about the include paths set by the user, but it does allow them to be squiggled when the true path is invalid.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -2006,6 +2006,7 @@ export class CppProperties {
             }
 
             // Escape the path string for literal use in a regular expression
+            // Need to escape any quotes to match the original text
             let escapedPath: string = curPath.replace(/"/g, '\\"');
             escapedPath = escapedPath.replace(/[-\"\/\\^$*+?.()|[\]{}]/g, '\\$&');
 
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1539,10 +1539,18 @@ export class CppProperties {
         }
 
         if (assumeRelative) {
+            let quoted = false;
+            if (result.startsWith('"') && result.endsWith('"')) {
+                quoted = true;
+                result = result.slice(1, -1);
+            }
             // Make sure all paths result to an absolute path.
             // Do not add the root path to an unresolved env variable.
             if (!result.includes("env:") && !path.isAbsolute(result) && this.rootUri) {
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1998,7 +2006,6 @@ export class CppProperties {
             }
 
             // Escape the path string for literal use in a regular expression
-            // Need to escape any quotes to match the original text
             let escapedPath: string = curPath.replace(/"/g, '\\"');
```
sean-mcmanus: What about this case?

sean-mcmanus (dismissed):
Can you get @WardenGnaw to approve too?

Extension/src/LanguageServer/configurations.ts
bobbrow: That's the one I dismissed

bobbrow: oh, but I didn't mean to remove the comment. That was left over from my poking around at it.

sean-mcmanus (approved):

bobbrow:
> Can you get @WardenGnaw to approve too?

I'll ask him, but we don't need to block on it since they're not active here.

sean-mcmanus (approved):

WardenGnaw:
It looks like these changes came from this commit: https://github.com/microsoft/vscode-cpptools/commit/b82a86e5e4b183b13c1db2e4e4149232927f7004

Which was to fix https://github.com/microsoft/vscode-cpptools/issues/4526

I would do a quick validation in a German VS Code to ensure the debug configuration is still being generated properly.

WardenGnaw (approved):

bobbrow:
> It looks like these changes came from this commit: [b82a86e](https://github.com/microsoft/vscode-cpptools/commit/b82a86e5e4b183b13c1db2e4e4149232927f7004)
> 
> Which was to fix #4526
> 
> I would do a quick validation in a German VS Code to ensure the debug configuration is still being generated properly.

I tested it by modifying the strings directly in TypeScript. It should be the same case as having the localizers put a quote in the translation.

PR #12490, opened 2024-07-18
Enable codeQL for TypeScript

bobbrow:


sean-mcmanus (approved):

sean-mcmanus:
@bobbrow It says "1 configuration not found" though.

bobbrow:
> @bobbrow It says "1 configuration not found" though.

That's why I had to do this PR. To remove the cpp configuration from the default.

bobbrow:
It seems to be looking for a configuration in main. I'll check this in and see if that replaces it and clears the error.

PR #12552, opened 2024-08-08
Support for copilot-generated summaries in quick info. (On-the-fly docs)

spebl:
This introduces a new setting `C_Cpp.onTheFlyDocsEnabled` to control the display of the option to show copilot-generated summaries in the hover tooltip. 
When `enabled`, and also authenticated with the vscode-copilot extension, the hover tooltip will display an option to generate a summary of the symbol with copilot.
The setting is defaulted to `default` which will check the feature flag control to determine if the feature should be enabled, which allows for slow rollout and the ability to rollback should any issues arise.

Updating the vscode requirement to `1.90.0` to support using copilot features with "vscode.lm".

The IntelliSense client changes supporting this feature are included in a separate PR against that repository.

Some workarounds were needed to support icon rendering in hover markdown and updating hover content dynamically which currently requires the hover to be closed and reopened. Potential fixes from proposals and unreleased patches are linked where applicable.

All feedback is very welcome!

OTF_DOCS_VSCODE


Extension/package.json
```diff
@@ -3286,6 +3286,17 @@
                         "default": false,
                         "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.markdownDescription%",
                         "scope": "application"
+                    },
+                    "C_Cpp.onTheFlyDocsEnabled": {
```
benmcmorran: Nit: I know we've moved away from using "on-the-fly docs" in public communications. The internal setting name does still use that language in VS, but users never really see the raw setting name, unlike VS Code where many users directly edit `settings.json`.

Maybe we should change the name to something like `C_Cpp.copilotHoverEnabled` instead?

Extension/package.nls.json
```diff
@@ -775,6 +775,12 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
+    "c_cpp.configuration.otfDocsEnabled.markdownDescription": {
+        "message": "If  `enabled`, the Hover tooltip will display an option to generate a summary of the symbol with copilot. If `disabled`, the option will not be displayed.",
```
benmcmorran:
```suggestion
        "message": "If  `enabled`, the Hover tooltip will display an option to generate a summary of the symbol with Copilot. If `disabled`, the option will not be displayed.",
```

I think Copilot is always capitalized for branding. It'd also be good to check if VS Code typically capitalizes "Hover" in their own documentation or not.

Extension/src/LanguageServer/extension.ts
```diff
@@ -404,6 +405,7 @@ export function registerCommands(enabled: boolean): void {
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.ExtractToMemberFunction', enabled ? () => onExtractToFunction(false, true) : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.ExpandSelection', enabled ? (r: Range) => onExpandSelection(r) : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.getIncludes', enabled ? (maxDepth: number) => getIncludes(maxDepth) : () => Promise.resolve()));
+    commandDisposables.push(vscode.commands.registerCommand('C_Cpp.OTFDocs', enabled ? onOTFDocs : onDisabledCommand));
```
benmcmorran: Same here, the command name string will technically be user visible so we should probably use something like `C_Cpp.ShowCopilotHover` instead. No need to change the rest of the code because users won't see that.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1379,3 +1381,66 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+
+    const locale = getLocaleId();
+
+    const messages = [
+        vscode.LanguageModelChatMessage
+            .User(request + locale)];
+
+    const [model] = await vscode.lm.selectChatModels({
+        vendor: 'copilot',
+        family: 'gpt-4'
```
benmcmorran: Do you know how restrictive this ends up being? Do we need a way to select a fallback model in case, for example, VS Code updates the list of supported models and we don't react in time?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1379,3 +1381,66 @@ export async function getIncludes(maxDepth: number): Promise<any> {
     const includes = await clients.ActiveClient.getIncludes(maxDepth);
     return includes;
 }
+
+async function onOTFDocs(): Promise<void> {
+    const response = await clients.ActiveClient.getOTFDocsInfo();
```
benmcmorran: Nit: `response` is a little confusing here given that we have a `chatResponse` further down in the function. Maybe something like `otfDocsInfo` instead?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1379,3 +1381,66 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+    let chatResponse: vscode.LanguageModelChatResponse | undefined;
+    try {
+        chatResponse = await model.sendRequest(
+            messages,
+            {},
+            new vscode.CancellationTokenSource().token
+        );
+    } catch (err) {
+        if (err instanceof vscode.LanguageModelError) {
+            console.log(err.message, err.code, err.cause);
```
benmcmorran, line 1457: Not sure what standard practice is in cpptools, but should we be sending some error telemetry here too?

Extension/src/LanguageServer/settings.ts
```diff
@@ -461,6 +463,33 @@ export class CppSettings extends Settings {
         return super.Section.get<boolean>("inlayHints.referenceOperator.showSpace") === true;
     }
 
+    public get otfDocsEnabled(): PromiseLike<boolean> {
+        // Check if the user has access to copilot.
+        return vscode.lm.selectChatModels({vendor: "copilot"}).then((models) => {
```
benmcmorran: This is a slightly different language model check than the one used in `onOTFDocs()`. I'd extract a shared helper function to select the language model to ensure we don't end up in a scenario where this check passes and shows the link, but then clicking the link is guaranteed to fail.

Extension/src/LanguageServer/settings.ts
```diff
@@ -461,6 +463,33 @@ export class CppSettings extends Settings {
...
+            // Check if the setting is explicitly set to enabled or disabled.
+            const setting = super.Section.get<string>("onTheFlyDocsEnabled");
```
benmcmorran: Probably doesn't matter, but we might want to check this first so we can bail early if it's `disabled` and avoid interacting with the language model API at all.

Extension/src/LanguageServer/settings.ts
```diff
@@ -461,6 +463,33 @@ export class CppSettings extends Settings {
...
+
+            if (setting === "enabled") {
+                return true;
+            }
+            if (setting === "disabled") {
+                return false;
+            }
+
+            // Check for the feature flag.
+            return isFlightEnabled("cpp.otfDocs");
```
benmcmorran: I think it's okay to have a single flight given that this is a Boolean option, but can you double check with Sinem? I know in the experimentation platform there will actually be two flights (a treatment and a control).

benmcmorran (commented):
Left a few wording comments but mostly looks good to me ✨. Leaving official sign-off for cpptools owners.

Extension/package.json
```diff
@@ -11,7 +11,7 @@
     },
     "license": "SEE LICENSE IN LICENSE.txt",
     "engines": {
-        "vscode": "^1.67.0"
+        "vscode": "^1.90.0"
```
sean-mcmanus: This will cause our extension to not be available on Mac OS 12-14, Ubuntu 16-18, and Windows 7-8 and x86. Is Bob okay with that? It might drop 10% of users or something, we could look up the exact number.

Extension/package.json
sean-mcmanus: None of our settings end with "Enabled" and "enabled" is a value of the setting (although we do have .enabled for setting groups). Can we use C_Cpp.copilotHover?

Extension/package.nls.json
sean-mcmanus: I have not seen "Hover" capitalized. I see Copilot capitalized.

sean-mcmanus:
@spebl Did you want this in our next 1.22.x or what release?

Extension/package.json
sean-mcmanus: Is there a way to make this a "soft"/optional requirement? i.e. disable the feature for the unsupported VS Code versions?

Extension/package.nls.json
sean-mcmanus: There are 2 spaces here -- maybe run the formatter?

spebl:
> @spebl Did you want this in our next 1.22.x or what release?

Not currently targeting a specific release, no need to block anything on this going in. I first wanted to get feedback on this approach, along with the upgrading of the vscode version. I'm taking a look now at how we can best keep support for the older versions while also use the new language model apis when available.

Extension/package.json
benmcmorran: If that is an issue, we might be able to work around it by leaving this as-is and then dynamically checking for the presence of the `vscode.lm` API at runtime, although I'm not sure if VS Code blocks access to newer APIs if the package.json declaration doesn't match.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Yeah, it shows on hover: 
image

Extension/package.json
bobbrow: You can try to break past the type checking by using:
```ts
const vscodelm = (vscode as any).lm;
```

Doing this removed some squiggles for me, but I did not test it to make sure it works. You would need to make sure it's not undefined before using it.

Extension/src/LanguageServer/extension.ts
spebl: I believe this is the most generic way to get the handle to gpt-4 in copilot. This is the same way as it's done in the sample from Isidor. If this ends up getting a lot of usage, it looks like gpt-3.5 and gpt-4o are also available. I think within these bundles they can update the model underneath without breaking compat.

Extension/package.json
spebl: updated naming everywhere. Thanks!

Extension/src/LanguageServer/settings.ts
spebl: Added shared model selector constant that's used for both, and can be used for any new copilot features as well.

Extension/package.json
spebl: Reverted and implemented Bob's suggestion. It's working well for me on my local machine, but will want to make sure to test against different configurations, and make sure we fail silently. The feature should never be enabled in the first place if vscode.lm isn't available.

Extension/src/LanguageServer/settings.ts
spebl: I believe we only need the one flag to check if it's enabled since we're defaulting to disabled, but I'll double check.

Extension/package.json
bobbrow: That's good news. Please make sure you test on an older version of VS Code that doesn't have the feature so we can ensure it doesn't make us crash.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1379,3 +1382,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
     const includes = await clients.ActiveClient.getIncludes(maxDepth);
     return includes;
 }
+
+// This uses several workarounds for interacting with the hover feature.
+// A proposal for dynamic hover content would help, such as the one here (https://github.com/microsoft/vscode/issues/195394)
+async function onCopilotHover(): Promise<void> {
+    if (!vscode.window.activeTextEditor) { return; }
```
bobbrow: Does our linter allow this? I expected that it would want:
```ts
if (!vscode.window.activeTextEditor) {
    return;
}
```

Extension/package.json
bobbrow: And if you miss having the types available, you could theoretically port the type definitions into our extension and use those in the LM code. I can provide some guidance if you'd like to do something like that.

Extension/src/LanguageServer/client.ts
```diff
@@ -1536,7 +1556,7 @@ export class DefaultClient implements Client {
             resetDatabase: resetDatabase,
             edgeMessagesDirectory: path.join(util.getExtensionFilePath("bin"), "messages", getLocaleId()),
             localizedStrings: localizedStrings,
-            settings: this.getAllSettings()
+            settings: await this.getAllSettings()
```
Colengms: I'm almost certain we avoided asyncs early in client creation code, as a fix for issues that occurs due to race conditions at startup.  I don't recall the exact details at the moment.  But the call to `createLanguageClient` is not awaited (it's called from a constructor, and constructors aren't async).  So, with your change, it's possible for the client to get constructed and presumably calls to `this.languageClient` to occur before that variable is properly set.  I think that's why there are no awaits there until after `languageClient` is set and everything has been hooked up.

Would it be possible to remove this async/await and deliver the setting value without awaiting to see if the flight is enabled?  Then, gate the functionality later based on whether the flight is enabled?

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -89,7 +89,17 @@ export function createProtocolFilter(): Middleware {
         provideHover: async (document, position, token, next: (document: any, position: any, token: any) => any) => clients.ActiveClient.enqueue(async () => {
             const me: Client = clients.getClientFor(document.uri);
             if (me.TrackedDocuments.has(document.uri.toString())) {
-                return next(document, position, token);
+                // Currently needed to support icons.
```
Colengms: This makes me nervous.  It looks like you're potentially hijacking any hover response, if it happens to contain an array that has a markdown string in it.

I think it would be cleaner if we switched to using a `HoverProvider` instead of hacking around the LSP hover request.  (Then you'd be more easily able to do pre/post processing of the request in TypeScript code. That may also help avoid that async/await to check if the flight is enabled, as that could happen in the a `HoverProvider`)

Extension/src/LanguageServer/client.ts
```diff
@@ -3972,6 +3998,17 @@ export class DefaultClient implements Client {
         compilerDefaults = await this.requestCompiler(path);
         DebugConfigurationProvider.ClearDetectedBuildTasks();
     }
+
+    public async showCopilotHover(content: string): Promise<ShowCopilotHoverResult> {
+        const params: ShowCopilotHoverParams = {content: content};
+        await this.ready;
+        return this.languageClient.sendRequest(ShowCopilotHoverRequest, params);
```
Colengms: It's common for multiple hover requests to be issued and get backlogged as the user moves the mouse around.  As each one is generated, the prior one is cancelled.  I think the cancellation token needs to be plumbed up here, all the way to these calls to `sendRequest`, in order for the `lsp_manager` on the native side to handle cancellation.

Cancellation will now cause an exception to be thrown from `sendRequest`, which you could catch, check for the specific cancellation error code, and handle appropriately (i.e., to throw a CancellationError exception, if that is what calling code expects).

Colengms (changes requested):

Extension/src/LanguageServer/protocolFilter.ts
Colengms: Also, I think having a hover provider would allow everything to be done in the same hover request, without needing to add additional custom LSP messages.  Instead, you could add additional fields to a primary hover request message, and do any additional work immediately before or after forwarding that to the native process using `sendRequest`, like the other provider classes.

benmcmorran:
Capturing notes from a quick discussion I had with @Colengms. He's going to go ahead and move from LSP-based hover to a `HoverProvider` which should make it easier for us to opt-in to icons and avoid async work during settings initialization. We also discussed a potential alternative using multiple `HoverProvider`s where a provider that actually displays Copilot results simply blocks until the user clicks the Copilot link from the first provider. FYI @spebl

Extension/package.json
```diff
@@ -3286,6 +3286,17 @@
                         "default": false,
                         "markdownDescription": "%c_cpp.configuration.addNodeAddonIncludePaths.markdownDescription%",
                         "scope": "application"
+                    },
+                    "C_Cpp.copilotHover": {
```
bobbrow, line 3317: This is being defined in the "Miscellaneous" section of settings. Just checking to make sure that's what you intended.

image

Extension/src/LanguageServer/extension.ts
```diff
@@ -1379,3 +1382,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+        }
+    } catch (err) {
+        return;
+    }
+
+    if (!vscode.window.activeTextEditor) { return; }
+    await vscode.window.showTextDocument(vscode.window.activeTextEditor.document, { preserveFocus: false, selection: new vscode.Selection(hoverPosition, hoverPosition) });
+
+    // Same workaround as above to force the editor to update it's content.
+    await clients.ActiveClient.showCopilotHover('$(loading~spin)');
```
Colengms: It would be nice to have a way to do this other than wiggling the mouse.

VS Code generally merges results from multiple providers.  I suspect if there are multiple HoverProviders and only one returns results (yet) it will show only one result.  If that result had the link in it, and that link 'unblocked' a secondary HoverProvider from returning an AI result, I suspect VS Code would then populate a new 'section' of the tooltip with the additional provider result.  That is just a guess, I haven't tried it.  :)

spebl: I finally got a chance to try this out, and it almost works perfectly, but unfortunately while the provider is blocking waiting for the signal, VS Code reserves hover space for it and displays "Loading..." until it's finally unblocked and the Promise resolves.

If you have any ideas on a way to not show the "Loading..." message, I think everything else should work. So far everything I've tried has hit either that same issue or the issue of the Hover content not updating after being rendered.

Colengms: Maybe we could put in a request to VS Code, to make this more perfect? :)  Since this is AI related, they might consider it a priority.  Hypothetically, if there were N hover providers, does it make sense for VS Code to show N `Loading...` messages?  Seems like just one would suffice, and additional results could appear for additional providers, once they provide results. That might be easy for them to fix. I'd suggest opening an issue against VS Code and seeing what they think.  https://github.com/microsoft/vscode/issues

Extension/src/LanguageServer/extension.ts
sean-mcmanus: @bobbrow Our formatter recently changed and allows this currently. Maybe there's some setting we need to change to alter it

bobbrow: I can look into it later. I'm not a fan of this style.

spebl: I meant to update this to match the rest of the formatting, I'll update it now. I don't have any strong preference for one or the other.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1378,3 +1381,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+
+    // Make sure the editor has focus.
+    await vscode.window.showTextDocument(vscode.window.activeTextEditor.document, { preserveFocus: false, selection: new vscode.Selection(hoverPosition, hoverPosition) });
+
+    // Workaround to force the editor to update it's content, needs to be called from another location first.
+    await vscode.commands.executeCommand('cursorMove', { to: 'right' });
+    await vscode.commands.executeCommand('editor.action.showHover', { focus: 'noAutoFocus' });
+
+    // Move back and show the correct hover.
+    await clients.ActiveClient.showCopilotHover('$(loading~spin)');
```
sean-mcmanus: The active document and active client could change after returning from an await. I think this could be leading to bugs I'm seeing when switching between documents and hovering.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1378,3 +1381,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
     const includes = await clients.ActiveClient.getIncludes(maxDepth);
     return includes;
 }
+
+// This uses several workarounds for interacting with the hover feature.
+// A proposal for dynamic hover content would help, such as the one here (https://github.com/microsoft/vscode/issues/195394)
+async function onCopilotHover(): Promise<void> {
+    if (!vscode.window.activeTextEditor) { return; }
```
sean-mcmanus: Hover is supposed to work on a document that is not the active editor.

sean-mcmanus: I see the hover disappear when Generate Copilot summary is clicked and there is no summary result.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1378,3 +1381,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+    // Prep hover with wait message and get the hover position location.
+    const copilotHoverResult = await clients.ActiveClient.showCopilotHover('$(loading~spin)');
+    const hoverPosition = new vscode.Position(copilotHoverResult.hoverPos.line, copilotHoverResult.hoverPos.character);
+
+    // Make sure the editor has focus.
+    await vscode.window.showTextDocument(vscode.window.activeTextEditor.document, { preserveFocus: false, selection: new vscode.Selection(hoverPosition, hoverPosition) });
+
+    // Workaround to force the editor to update it's content, needs to be called from another location first.
+    await vscode.commands.executeCommand('cursorMove', { to: 'right' });
+    await vscode.commands.executeCommand('editor.action.showHover', { focus: 'noAutoFocus' });
```
sean-mcmanus: If the user changes the document and cursor, it can trigger a different hover provider -- I got GitLens hover to be invoked and the line changed to some random line. It seems like it should save the original document or something.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1378,3 +1381,88 @@ export async function getIncludes(maxDepth: number): Promise<any> {
...
+    // Check if the user has access to vscode language model.
+    const vscodelm = (vscode as any).lm;
+    if (!vscodelm) { return; }
+
+    // Prep hover with wait message and get the hover position location.
+    const copilotHoverResult = await clients.ActiveClient.showCopilotHover('$(loading~spin)');
+    const hoverPosition = new vscode.Position(copilotHoverResult.hoverPos.line, copilotHoverResult.hoverPos.character);
+
+    // Make sure the editor has focus.
+    await vscode.window.showTextDocument(vscode.window.activeTextEditor.document, { preserveFocus: false, selection: new vscode.Selection(hoverPosition, hoverPosition) });
```
sean-mcmanus: If the user switches active documents, it'll move the selection to an incorrect line (i.e. the line number of the previously active document), so they'll be at some random spot in the code which may not even have C++ hover, which is why I was getting only hover from GitLens in my other comment because it moved my cursor to hover over a space that had a GitLens inlay hint on it.

spebl: No longer using the active document.

Extension/src/LanguageServer/protocolFilter.ts
spebl: Swapped to using the hover new hover provider pattern you added, it should be much more stable now, and we no longer need to make any changes to the intellisense hover.

Extension/src/LanguageServer/client.ts
spebl: I'm hoping that since we're no longer going back and forth from the lsp server, this shouldn't be an issue. I've added exception catching around the one call this still does make to the native side. Let me know if there is any additional cancellation I should be tracking still.

Extension/package.json
spebl: Unsure if that's the best place for it, but I do acknowledge that it's there for now. I'm open to moving it or making a new Copilot category if folks think that makes sense.

Extension/src/LanguageServer/extension.ts
spebl: This latest iteration is still using the multiple invocations to update the content, but swapping to using the new Hover provider has made it a lot easier to track between requests. If you or the team still have concerns about this implementation approach, I'm happy to discuss further and we can make a call on next steps.

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+    public async provideHover(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Hover | undefined> {
+        await this.client.ready;
+        const settings: CppSettings = new CppSettings(vscode.workspace.getWorkspaceFolder(document.uri)?.uri);
+        if (settings.hover === "disabled") {
+            return undefined;
+        }
+
+        if (!this.isNewHover(document, position)) {
+            if (this.ready) {
+                const contentMarkdown = new vscode.MarkdownString(`$(sparkle) Copilot\n\n${this.content}\n\n_${localize("copilot.disclaimer", "AI-generated content may be incorrect.")}_`, true);
```
sean-mcmanus: * Why does `AI-generated content may be incorrect.` need to be shown? Should that already be obvious to our target audience? Can that warning go in the GitHub Copilot docs or something?
* If we still want to show it, can it only be shown once? i.e. a persistent state could be set to stop showing it.
* If we want to show it more than once, can we show it only once per session?

sean-mcmanus (dismissed):

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
benmcmorran: This text is included as part of our Responsible AI compliance for this feature. The rules aren't entirely clear so we're erring on the side of caution by always showing the warning. While it may be acceptable to show it less frequently, given that we've already submitted our Responsible AI documentation for review it would introduce additional delays in shipping this feature.

This is roughly the same UX that we had when this feature first shipped in Visual Studio. Since then, we've been able to relax it to only showing the disclaimer in the link tooltip, but we don't have the ability to control the link tooltip in VS Code.

sean-mcmanus: Ok, FYI, I don't see the warning text currently shown for other GitHub copilot features I've tried.

sean-mcmanus: @benmcmorran You can add it to the hover link via adding ` "text"` after the command link: 
image

e.g.
```TypeScript
`](command:C_Cpp.ShowCopilotHover "${localize("copilot.disclaimer", "AI-generated content may be incorrect.")}")`
```

sean-mcmanus (changes requested):
see comment

sean-mcmanus (dismissed):
I think the putting Copilot text on a separate line is excessive. The user already saw "Copilot" text when they clicked the link.

I think the symbol is sufficient -- and on the same line:
image

versus:

image

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+                return new vscode.Hover(contentMarkdown);
+            }
+            if (this.waiting) {
+                const loadingMarkdown = new vscode.MarkdownString("$(loading~spin)", true);
```
sean-mcmanus: Should `$(sparkle)` be added here next to the loading to make it clear it's the Copilot that is loading?

benmcmorran: That seems reasonable to me.

In general, for all the AI wording comments I'm happy to ask the experts as we get further into the Responsible AI assessment process, but I don't think I'll be able to give authoritative answers right now.

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
benmcmorran: Thanks, I didn't realize we had that level of control!

spebl: I've modified the link to show the tooltip text as Sean showed above and currently removed the warning in the final output. This now matches the current VS implementation.

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
spebl: For this, as of now I've added the sparkle to the loading spinner to better indicate it's coming from the copilot feature. For now, I've left the Sparkle and Copilot header on it's own line before the content, as this is what we have in VS and I think it makes the most sense for now to aim for parity, since that's already shipped with approval. 

As Ben said, for the final wording we may get further input on best practices, but for now I've tried to match it as close to the VS wording and display as possible.

For reference, this is how it looks in VS right now:
image

sean-mcmanus: Yikes, in my opinion the VS hover is wasting too much vertical UI space (covering more lines of code unnecessarily).

sean-mcmanus (dismissed):

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
Colengms: It looks like this `sendRequest` was moved to `getRequestInfo` and doesn't have a cancellation token passed to it.  I think you'll want to capture the cancellation token from the original(?) hover request (or, maybe the subsequent hover request caused by `showCopilotContent`?), and pass that to `sendRequest` to communicate that cancellation to the native process. Even if you're not explicitly checking for cancellation (`!request.is_active()`) in your LSP message handler, this would allow the LSP manager itself to discard requests in queue that have already been cancelled and to discard the response for an operation already cancelled.

Extension/src/LanguageServer/client.ts
```diff
@@ -4009,6 +4041,19 @@ export class DefaultClient implements Client {
...
+        return this.hoverProvider;
+    }
+
+    public getCopilotHoverProvider(): CopilotHoverProvider | undefined {
+        return this.copilotHoverProvider;
+    }
+
+    public async getCopilotHoverInfo(): Promise<GetCopilotHoverInfoResult> {
+        await this.ready;
+        return this.languageClient.sendRequest(GetCopilotHoverInfoRequest, null);
```
Colengms: Similar feedback to the other call to `sendRequest`.  Assuming this happens in the context of a hover operation, could you pass the associated cancellation token?  For some of our calls to `sendRequest` that are triggered by commands, we don't get a token from VS Code, so don't pass it along.  But we should pass along cancellation tokens for any user operation that has one associated with it.

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
spebl: We don't directly have the token at call time, but we're relying on other stashed hover values so I've added similar logic for the cancellation token. Also, thanks for the heads up on the multiple send request calls added, this one isn't used anymore so I was able to remove it.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
Colengms: Could you clarify?  I would expect a hover-related operation to be conceptually cancelable/dismissible by the user.  (It looks like you currently have a bug where, if a user has moved on to another task, a slow CopilotHover request can result in the cursor being repositioned, potentially even when the user is mid-typing?).  

I think it's important that this feature handle cancellation.  The `cpptools_getCopilotHoverInfo` handler on the native side will acquire a (potentially new) TU, which may impose a hefty perf impact.  Even if the TU were already existing, if racing with edit, it may trigger a preparse.  There could be other messages in queue, delaying delivery of yours, or piling up after yours.

For a lot of users, IntelliSense is very slow. I have a project at home in which a TU can take more than a minute to perform a preparse.  If racing with edits, we don't want to waste time fully completing an operation whose result won't be used.  As soon as there are subsequent edits in queue, the preparse it performs potentially be entirely wasted.

spebl: Sorry, I think my response was extra confusing since this code location doesn't make sense anymore (was removed). 

For the other case where we're still calling to the native layer, we're not in the Copilot Hover's call stack directly, so we have to store the last hover cancellation token to use it later when called through the "onCopilotHover" command.

It looks like there's a bug with which token I was holding, so I'm fixing that now to make sure we have the most up-to-date hover cancellation token when checking it and passing it through to the native layer. 

I agree that cancellation is important, trying to handle it from the command callback function has been a little more complicated since we're currently invoking multiple follow up hovers. Hopefully we can remove that need in a future update with better platform support, which would really help streamline the cancellation.

Colengms: In my last response, I was specifically looking at calls to our `languageClient.sendRequest`, trying to ensure they all had tokens passed to them.  It looks like one of your last pushes addressed that.  I may have been looking at stale code.

One possible issue:  In `onCopilotHover`, it looks like you're passing in a token from a new (not otherwise referenced) token source:

```
        chatResponse = await model.sendRequest(
            messages,
            {},
            new vscode.CancellationTokenSource().token
        );
```

I'm guessing maybe you had to pass a token (undefined wasn't accepted?), and didn't have a valid one?

>trying to handle it from the command callback function has been a little more complicated since we're currently invoking multiple follow up hovers

Are those follow-ups all in the context of the same hover operation?  If so, could the original hover operation's token be used for all of them?

Normally, I'd suggest passing an argument when programmatically invoking the command, but it looks like you're embedding the command as text into the hover text.  Note that you can still pass an encoded argument to the command using syntax like so:

```
[Find my new Compiler](command:C_Cpp.RescanCompilers?%22walkthrough%22)",
```

Follow the call into `onRescanCompilers` for an example of how to extract it. I don't think you'd be able to pass through an object reference, but you could pass through a key that you could associate a value with in a map.

spebl: Thanks! I'll push out my updated changes soon that seems to handle the cancellation much better by holding on to the hover's cancellation token and tracking which operations are changing the editor's focus more intentionally.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1387,3 +1391,134 @@ export async function preReleaseCheck(): Promise<void> {
...
+        await reportCopilotFailure(copilotHoverProvider, hoverDocument, hoverPosition);
+        return;
+    }
+
+    await showCopilotContent(copilotHoverProvider, hoverDocument, hoverPosition, content);
+}
+
+async function reportCopilotFailure(copilotHoverProvider: CopilotHoverProvider, hoverDocument: vscode.TextDocument, hoverPosition: vscode.Position, customError?: string): Promise<void> {
+    const errorMessage = customError ? customError : "";
+    telemetry.logLanguageServerEvent("CopilotHoverError", { "ErrorMessage": errorMessage });
```
sean-mcmanus, line 1494: Could you log a separate error telemetry when there's no declaration/definition? I don't think that's really an error since it seems "by design" when used on a macro, etc. It seems like the user facing message should also not be "An error occurred while generating Copilot summary." and instead say the something that sounds more "by design" and less like an unexpected error occurred?

sean-mcmanus: image

sean-mcmanus: The telemetry and/or message could have an issue filed to follow up too -- up to your team and if they care about no definition/declaration being counted as "errors".

sean-mcmanus: Umm...but it would be easier to triage user reported "bugs" if known/"by design" issues had a different message so that we don't have to investigate the "error" (that's not really an error).

spebl: Changed this case where we have no decl/defn context to not log an error and return a more descriptive message letting the user know that it's not available. I also updated the error logging to require an error message to include in the logging, even though we'll still show the generic error occurred message to the user, but it shouldn't show up in the telemetry.

sean-mcmanus (approved):

sean-mcmanus (changes requested):
Copilot hover only works for the 1st workspace folder. If you add a 2nd+ workspace folder, then clicking Generate Copilot summary has no effect in those other workspace folders.

image

Instead of using the ActiveClient, it seems like it should use the `const client: Client = clients.getDefaultClient();`

...yeah, doing that fixes it, see the example in onDidChangeSettings (that would mean it always uses the settings from the 1st workspace folder, but that seems fine to me, since there aren't any workspace folder specific settings involved).

Extension/src/LanguageServer/Providers/CopilotHoverProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+                return new vscode.Hover(loadingMarkdown);
+            }
+        }
+
+        this.currentDocument = document;
+        this.currentPosition = position;
+        const commandString = "$(sparkle) [" + localize("generate.copilot.description", "Generate Copilot summary") + "](command:C_Cpp.ShowCopilotHover \"" + localize("copilot.disclaimer", "AI-generated content may be incorrect.") + "\")";
+        const commandMarkdown = new vscode.MarkdownString(commandString);
+        commandMarkdown.supportThemeIcons = true;
+        commandMarkdown.isTrusted = true;
```
benmcmorran: It's not a big risk in this case since there's no user-generated content in the first hover, but for security prefer setting `isTrusted` to an exact set of enabled commands instead of boolean.

```suggestion
        commandMarkdown.isTrusted = { enabledCommands: ["C_Cpp.ShowCopilotHover"] };
```

benmcmorran (dismissed):

sean-mcmanus (dismissed):

sean-mcmanus (approved):

PR #12556, opened 2024-08-09
Refactor compilerPath logic to reflect settings validation changes

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12555

Previously, default.compilerPath returned string | undefined when it was possible to return a null value in package.json. Now that it is returning null, the compilerPath logic needs to be updated to reflect this.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -995,8 +995,6 @@ export class CppProperties {
                 if (configuration.compilerPath !== undefined) {
                     configuration.compilerPath = util.resolveVariables(configuration.compilerPath, env);
                     configuration.compilerPathIsExplicit = true;
-                } else {
-                    configuration.compilerPathIsExplicit = false;
                 }
             }
 
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -992,7 +992,7 @@ export class CppProperties {
                 }
                 if (configuration.compilerPath === null) {
                     configuration.compilerPathIsExplicit = true;
-                } else if (configuration.compilerPath !== undefined ) {
+                } else if (configuration.compilerPath !== undefined) {
                     configuration.compilerPath = util.resolveVariables(configuration.compilerPath, env);
                     configuration.compilerPathIsExplicit = true;
                 } else {
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -989,12 +989,15 @@ export class CppProperties {
                 // However, if compileCommands are used and compilerPath is explicitly set, it's still necessary to resolve variables in it.
                 if (configuration.compilerPath === "${default}") {
                     configuration.compilerPath = settings.defaultCompilerPath ?? undefined;
-                    configuration.compilerPathIsExplicit = settings.defaultCompilerPath === null && configuration.compilerPath === undefined;
                 }
 
                 if (configuration.compilerPath !== undefined) {
                     configuration.compilerPath = util.resolveVariables(configuration.compilerPath, env);
                     configuration.compilerPathIsExplicit = true;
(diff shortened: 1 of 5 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -989,15 +989,12 @@ export class CppProperties {
...
                 }
-                if (configuration.compilerPath === null) {
-                    configuration.compilerPath = undefined;
-                    configuration.compilerPathIsExplicit = true;
-                } else if (configuration.compilerPath !== undefined) {
+
+                if (configuration.compilerPath !== undefined) {
                     configuration.compilerPath = util.resolveVariables(configuration.compilerPath, env);
                     configuration.compilerPathIsExplicit = true;
-                } else {
```
sean-mcmanus: I think this `else` is still required. Maybe it should be

```
} else if (settings.defaultCompilerPath === null) {
    configuration.compilerPathIsExplicit = true;
} else {
    configuration.compilerPathIsExplicit = false;
}
```

with `configuration.compilerPathIsExplicit = settings.defaultCompilerPath === null && configuration.compilerPath === undefined;` removed.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Oops, sorry for telling you to do the wrong thing, but I realized my suggestion was wrong.

There's a problem because `settings.defaultCompilerPath === null` should only be checked if `configuration.compilerPath === "${default}"` is true, so maybe just add a check for that.

sean-mcmanus: You may need to add a bool that is set to true in `if (configuration.compilerPath === "${default}") {` and check that bool.

bobbrow: You should ask @Colengms what the intent was. PR here: https://github.com/microsoft/vscode-cpptools/pull/5855/files

sean-mcmanus: @bobbrow Yeah, I think that's what the plan was before checking it in. From what I can tell, the code as originally written by Colen never executed the code in the `if` block (but I wasn't able to detect any buggy behavior as a result).

Colengms: I'm confused.  `null` is a valid value that `compilerPath` can have in a configuration.  Wouldn't the correct change here be to keep the `compilerPath === null` check (maybe remove where it gets normalized to undefined), so we are properly supporting users who set the `compilerPath` to `null` in their `c_cpp_properties.json`?  That might also require ensuring other reference to the variable properly handle null, where it was not explicitly part of the type before. (This it probably wound up being loaded as null previously, despite the type not being correct)

Colengms (changes requested):

Extension/src/LanguageServer/configurations.ts
browntarik: @Colengms I think you're approach is right, we were just unsure about the propagation of null values initially.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1598,7 +1598,7 @@ export class CppProperties {
             }
 
             // Get compiler path without arguments before checking if it exists
-            resolvedCompilerPath = compilerPathAndArgs.compilerPath;
+            resolvedCompilerPath = compilerPathAndArgs.compilerPath ?? undefined;
```
sean-mcmanus, line 1601: FYI, `| null` could be added to the type of resolvedCompilerPath, but I'm not sure if that matters.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/configurations.ts
browntarik: Thanks for catching that.

Colengms (approved):

PR #12044, opened 2024-03-01
Add context to the Copilot Chat

lukka:
@benmcmorran @spebl @hnrqbaggio FYI

PR #12577, opened 2024-08-15
Add C++ configuration as a language model tool

benmcmorran:
This is essentially the same as #12044, but adapted to use the new proposed language model tools API instead of the chat variables API. I'm leaving it in a draft statue until the API is on a path to stabilization and we've resolved issues like https://github.com/microsoft/vscode/issues/225737. I'm also guessing we'll want to add the ability to A/B experiment.

cc @lukka @sinemakinci1

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -6464,9 +6464,9 @@
             {
                 "id": "cpptools-lmtool-configuration",
                 "name": "cpp",
-                "displayName": "C/C++ configuration",
+                "displayName": "%c_cpp.languageModelTools.configuration.displayName%",
                 "canBeInvokedManually": true,
-                "userDescription": "Configuration of the active C or C++ file, like language standard version and target platform.",
+                "userDescription": "%c_cpp.languageModelTools.configuration.userDescription%",
                 "modelDescription": "For the active C or C++ file, this tool provides: the language (C, C++, or CUDA), the language standard version (such as C++11, C++14, C++17, or C++20), the compiler (such as GCC, Clang, or MSVC), the target platform (such as x86, x64, or ARM), and the target architecture (such as 32-bit or 64-bit).",
                 "icon": "$(file-code)",
                 "parametersSchema": {}
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -38,7 +38,9 @@
         "Snippets"
     ],
     "enabledApiProposals": [
-        "terminalDataWriteEvent"
+        "terminalDataWriteEvent",
+        "chatParticipant",
+        "chatVariableResolver"
     ],
     "capabilities": {
         "untrustedWorkspaces": {
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -39,8 +39,7 @@
     ],
     "enabledApiProposals": [
         "terminalDataWriteEvent",
-        "chatParticipant",
-        "chatVariableResolver"
+        "lmTools"
     ],
     "capabilities": {
         "untrustedWorkspaces": {
@@ -6459,6 +6458,18 @@
                     "title": "%c_cpp.codeActions.refactor.extract.function.title%",
                     "description": "%c_cpp.codeActions.refactor.extract.function.description%"
                 }
+            }
+        ],
+        "languageModelTools": [
+            {
+                "id": "cpptools-lmtool-configuration",
+                "name": "cpp",
+                "displayName": "C/C++ configuration",
+                "canBeInvokedManually": true,
+                "userDescription": "Configuration of the active C or C++ file, like language standard version and target platform.",
+                "modelDescription": "For the active C or C++ file, this tool provides: the language (C, C++, or CUDA), the language standard version (such as C++11, C++14, C++17, or C++20), the compiler (such as GCC, Clang, or MSVC), the target platform (such as x86, x64, or ARM), and the target architecture (such as 32-bit or 64-bit).",
+                "icon": "$(file-code)",
+                "parametersSchema": {}
             }
         ]
     },
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -541,6 +541,14 @@ interface GetIncludesResult {
     includedFiles: string[];
 }
 
+export interface ChatContextResult {
+    language: string;
+    standardVersion: string;
+    compiler: string;
+    targetPlatform: string;
+    targetArchitecture: string;
+}
+
 // Requests
 const PreInitializationRequest: RequestType<void, string, void> = new RequestType<void, string, void>('cpptools/preinitialize');
 const InitializationRequest: RequestType<CppInitializationParams, void, void> = new RequestType<CppInitializationParams, void, void>('cpptools/initialize');
@@ -560,6 +568,7 @@ const GoToDirectiveInGroupRequest: RequestType<GoToDirectiveInGroupParams, Posit
 const GenerateDoxygenCommentRequest: RequestType<GenerateDoxygenCommentParams, GenerateDoxygenCommentResult | undefined, void> = new RequestType<GenerateDoxygenCommentParams, GenerateDoxygenCommentResult, void>('cpptools/generateDoxygenComment');
 const ChangeCppPropertiesRequest: RequestType<CppPropertiesParams, void, void> = new RequestType<CppPropertiesParams, void, void>('cpptools/didChangeCppProperties');
 const IncludesRequest: RequestType<GetIncludesParams, GetIncludesResult, void> = new RequestType<GetIncludesParams, GetIncludesResult, void>('cpptools/getIncludes');
+const CppContextRequest: RequestType<void, ChatContextResult, void> = new RequestType<void, ChatContextResult, void>('cpptools/getChatContext');
 
 // Notifications to the server
 const DidOpenNotification: NotificationType<DidOpenTextDocumentParams> = new NotificationType<DidOpenTextDocumentParams>('textDocument/didOpen');
@@ -790,6 +799,7 @@ export interface Client {
     setShowConfigureIntelliSenseButton(show: boolean): void;
     addTrustedCompiler(path: string): Promise<void>;
     getIncludes(maxDepth: number): Promise<GetIncludesResult>;
+    getChatContext(): Promise<ChatContextResult | undefined>;
 }
 
 export function createClient(workspaceFolder?: vscode.WorkspaceFolder): Client {
@@ -2202,6 +2212,11 @@ export class DefaultClient implements Client {
         const params: GetIncludesParams = { maxDepth: maxDepth };
         await this.ready;
         return this.languageClient.sendRequest(IncludesRequest, params);
+    }
+
+    public async getChatContext(): Promise<ChatContextResult | undefined> {
+        await this.ready;
+        return this.languageClient.sendRequest(CppContextRequest, null);
     }
 
     /**
@@ -4084,4 +4099,5 @@ class NullClient implements Client {
     setShowConfigureIntelliSenseButton(show: boolean): void { }
     addTrustedCompiler(path: string): Promise<void> { return Promise.resolve(); }
     getIncludes(): Promise<GetIncludesResult> { return Promise.resolve({} as GetIncludesResult); }
+    getChatContext(): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
 }
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -61,7 +61,7 @@ import * as refs from './references';
 import { CppSettings, OtherSettings, SettingsParams, WorkspaceFolderSettingsParams, getEditorConfigSettings } from './settings';
 import { SettingsTracker } from './settingsTracker';
 import { ConfigurationType, LanguageStatusUI, getUI } from './ui';
-import { handleChangedFromCppToC, makeLspRange, makeVscodeLocation, makeVscodeRange } from './utils';
+import { handleChangedFromCppToC, makeLspRange, makeVscodeLocation, makeVscodeRange, withCancellation } from './utils';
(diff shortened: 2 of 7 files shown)


Extension/package.json
```diff
@@ -6458,6 +6459,18 @@
...
                 }
             }
+        ],
+        "languageModelTools": [
+            {
+                "id": "cpptools-lmtool-configuration",
+                "name": "cpp",
+                "displayName": "C/C++ configuration",
+                "canBeInvokedManually": true,
+                "userDescription": "Configuration of the active C or C++ file, like language standard version and target platform.",
```
Colengms: Not sure if it's supported here (yet?), but ultimately this should get localized.  I assume that would be similar to other `"%<desc>%"` loc strings supported in some other package.json fields.

Extension/src/LanguageServer/client.ts
```diff
@@ -2204,6 +2214,11 @@ export class DefaultClient implements Client {
         return this.languageClient.sendRequest(IncludesRequest, params);
     }
 
+    public async getChatContext(): Promise<ChatContextResult | undefined> {
+        await this.ready;
+        return this.languageClient.sendRequest(CppContextRequest, null);
```
Colengms: Not sure if applicable, but `sendRequest` supports being passed a cancellation token.  Then, if the native side cancels, you'll get a different object back.  We've been detected native-side cancellation by checking if the resulting object lacks some otherwise required field.

Extension/src/LanguageServer/extension.ts
```diff
@@ -248,6 +249,11 @@ export async function activate(): Promise<void> {
         clients.timeTelemetryCollector.setFirstFile(activeEditor.document.uri);
         activeDocument = activeEditor.document;
     }
+
+    if (util.extensionContext) {
+        const tool = vscode.lm.registerTool('cpptools-lmtool-configuration', new CppConfigurationLanguageModelTool(clients));
```
Colengms: The `client` ClientCollection is accessible as a global variable, or using the global `getClients`s, in `extension.ts`.  We do recreate client to recover after a crash, but we appear to reuse the same ClientCollection. I'd recommend using `getClients` as needed, just in case.

Colengms (changes requested):

Extension/package.json
benmcmorran: Updated to use strings from `package.nls.json` and it appears to work.

Extension/src/LanguageServer/extension.ts
benmcmorran: Thanks, changed to use `getClients()`.

Extension/src/LanguageServer/client.ts
benmcmorran: I plumbed the cancellation token through in the last push. I also added a helper to allow waiting for a promise with cancellation.

Extension/src/LanguageServer/client.ts
```diff
@@ -4084,4 +4107,5 @@ class NullClient implements Client {
     setShowConfigureIntelliSenseButton(show: boolean): void { }
     addTrustedCompiler(path: string): Promise<void> { return Promise.resolve(); }
     getIncludes(): Promise<GetIncludesResult> { return Promise.resolve({} as GetIncludesResult); }
+    getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
```
Colengms, line 4110: Should this return `Promise<ChatContextResult | undefined>` to match the other definition?

It's a bit unusual to return an undefined result field in an LSP response.  Have you confirmed that works as expected?  A while back, there was some issue that prevented that, requiring us to instead always return an object, with a reason code or optional fields.  (That may have been addressed when the `lsp_manager` was added.) If throwing CancellationError results in a better behavior than returning `"No configuration information is available for the active document."`, you might consider using server-side cancellation to indicate general failure.  Or maybe plan ahead for other failure states and provide a field for that in the result object.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    private async getContext(token: vscode.CancellationToken): Promise<string> {
+        const currentDoc = vscode.window.activeTextEditor?.document;
+        if (!currentDoc || (!util.isCpp(currentDoc) && !util.isHeaderFile(currentDoc.uri))) {
+            return 'The active document is not a C, C++, or CUDA file.';
+        }
+
+        const chatContext: ChatContextResult | undefined = await (getClients()?.ActiveClient?.getChatContext(token) ?? undefined);
+        if (!chatContext) {
+            return 'No configuration information is available for the active document.';
```
lukka, line 65: is actual content reporting absence of information needed? Could this be just an empty string?

Extension/src/LanguageServer/client.ts
benmcmorran: Thanks for pointing this out. Looks like it is indeed still an issue.

image

I noticed that there are [existing code paths like `GoToDirectiveInGroupRequest`](https://github.com/microsoft/vscode-cpptools/blob/4a7fe28adc4cca80fce9c9ece78d34ebe8fec60f/Extension/src/LanguageServer/client.ts#L559) and IncludesRequest (on the native side) that also follow this pattern, so there might need to be a broader cleanup.

Colengms: @benmcmorran I'm not sure what would surface that specific error into that UI.  Is this an exception that is propagated all the up from the call to `sendRequest`?

benmcmorran: Yep, that's bubbling all the way up [from vscode-jsonrpc](https://github.com/microsoft/vscode-languageserver-node/blob/89c6f5cb47e64964b685f42e5063a575b2c54635/jsonrpc/src/common/connection.ts#L1019).

benmcmorran: Discussed offline and there's a bug in the way errors are handled in the cpptools server today. Reported at #12591.

Colengms: LSP cancellation should now be properly reflected by an exception (containing the JSON-RPC error object) from `sendRequest`.  I think you still have an issue here if returning an undefined result, though.

sean-mcmanus:
@benmcmorran Is this supposed to be out of draft now?

benmcmorran:
@lukka is handling finalizing this PR, but yes we can take it out of draft. Ideally we'd like to get these changes in the next pre-release. Pending work:

- Only enable when the experimental features setting is enabled.
- Add runtime type checks for the API and wrap all in try/catch blocks.
- Add notification if API interactions fail.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -0,0 +1,88 @@
...
+                "compiler": chatContext.compiler,
+                "standardVersion": chatContext.standardVersion,
+                "targetPlatform": chatContext.targetPlatform,
+                "targetArchitecture": chatContext.targetArchitecture
+            });
+
+        for (const key in knownValues) {
+            const knownKey = key as keyof ChatContextResult;
+            if (knownValues[knownKey] && chatContext[knownKey])
+            {
```
sean-mcmanus, line 81: Can you run the formatter on this document? The `if` has `{` on a separate line.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/client.ts
```diff
@@ -2204,6 +2214,19 @@ export class DefaultClient implements Client {
         return this.languageClient.sendRequest(IncludesRequest, params);
     }
 
+    public async getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult | undefined> {
+        await withCancellation(this.ready, token);
+        const result = await this.languageClient.sendRequest(CppContextRequest, null, token);
```
Colengms, line 2219: Due to the changes to properly trigger exceptions when a request is cancelled, you'll need to use a try/catch block around the call to `sendRequest` now. i.e.:

        let response: ChatContextResult;
        try {
            response = await this.client.languageClient.sendRequest(CppContextRequest, null, token);
        } catch (e: any) {
            if (e instanceof ResponseError && (e.code === RequestCancelled || e.code === ServerCancelled)) {
            }
            throw e;
        }
        if (token.isCancellationRequested) {
        }

Colengms: Also, we try to explicitly provide the type when declaring a variable in TypeScript, unless otherwise explicitly stated in the same statement (i.e. `= new Typename`).  I'm not sure why lint didn't complain here.

i.e. `const result: ChatContextResult = ...`

Extension/src/LanguageServer/client.ts
Colengms: I think this comment is still open to address the original question: Should this return Promise<ChatContextResult | undefined> to match the other definition?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -0,0 +1,88 @@
...
+        return new StringLanguageModelToolResult(await this.getContext(token));
+    }
+
+    private async getContext(token: vscode.CancellationToken): Promise<string> {
+        const currentDoc = vscode.window.activeTextEditor?.document;
+        if (!currentDoc || (!util.isCpp(currentDoc) && !util.isHeaderFile(currentDoc.uri))) {
+            return 'The active document is not a C, C++, or CUDA file.';
+        }
+
+        const chatContext: ChatContextResult | undefined = await (getClients()?.ActiveClient?.getChatContext(token) ?? undefined);
```
Colengms, line 63: Is there a reason the implementation of `getChatContext` needs to live in the client object?  It seems like it might be appropriate here, whereas we're often needing to refactor things out of `client.ts` into more appropriate places.

Extension/src/LanguageServer/utils.ts
```diff
@@ -1,101 +1,108 @@
...
+            "workbench.action.openWalkthrough",
+            { category: 'ms-vscode.cpptools#cppWelcome', step: index },
+            false)
+        );
+    return;
+}
+
+const docsChangedFromCppToC: Set<string> = new Set<string>();
+
+export async function withCancellation<T>(promise: Promise<T>, token: vscode.CancellationToken): Promise<T> {
```
Colengms, line 103: Does the call to `token.onCancellationRequest` permanently modify the token passed in, to add the callback?  In general, a helper like this would be useful in some of our providers, but it seems like a bug that the callback registration is not only within the scope of this call.  We would also need to be able to call `onCancellationRequested` after calling this, to register other work we may need to do on cancellation.  I'm not sure of the consequences of multiple calls to `onCancellationRequested`.

Colengms (changes requested):

Extension/src/LanguageServer/lmTool.ts
lukka: looks good

Extension/src/LanguageServer/client.ts
lukka: @Colengms @benmcmorran  I addressed this comment in this other PR (topmost commit): https://github.com/microsoft/vscode-cpptools/pull/12685

I have no access to vscode-cpptools, hence I cant push on this PR.

Extension/src/LanguageServer/lmTool.ts
lukka: @Colengms @benmcmorran  I addressed this comment in this other PR (topmost commit): https://github.com/microsoft/vscode-cpptools/pull/12685

I have no access to vscode-cpptools, hence I cant push on this PR.

Extension/src/LanguageServer/utils.ts
lukka: @Colengms if I understand correctly, it is more about stylistic approach or future refactoring, but in principle no changes are required now in this PR. Let me know.

Extension/src/LanguageServer/client.ts
lukka: @Colengms @benmcmorran this one looks not addressed, but I thought it was reading the changes. Could you advice what changes are needed?

Extension/src/LanguageServer/lmTool.ts
lukka: @sean-mcmanus FYI the formatted is happy with `{` on the next line. I had to move this manually. The formatter used is "dbaeumer.vscode-eslint" as configured in settings.json.


@sean-mcmanus @Colengms @benmcmorran I addressed this comment in this other PR (topmost commit): #12685

I have no access to vscode-cpptools, hence I cant push on this PR.

Extension/src/LanguageServer/utils.ts
```diff
@@ -1,101 +1,108 @@
...
+    return new Promise<T>((resolve, reject) => {
+        token.onCancellationRequested(() => reject(new vscode.CancellationError()));
```
Colengms, line 105: Could you grab the Disposable that gets returned from this event registration, and dispose of it when resolving?  That way, if this function is called multiple times with the same token, multiple event registrations won't get accumulated/leaked.

Extension/src/LanguageServer/client.ts
Colengms: >> Should this return Promise<ChatContextResult | undefined> to match the other definition?

> Could you advice what changes are needed?

The line current is:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
```

But the interface declares it as:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult | undefined>;
```

I'm just suggesting updating the derived class member function declaration be consistent with the base class, like other functions here.  So:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult | undefined> { return Promise.resolve({} as ChatContextResult); }
```

Colengms:
Closing this one, as Luca has a new one here: https://github.com/microsoft/vscode-cpptools/pull/12685

Issue #12529, opened 2024-07-30
RangeError: WebAssembly.Instance(): Out of memory: Cannot allocate Wasm memory for new instance


mdeberts:
Type: <b>Bug</b>

Install C/C++ and Modern Fortran
Start remote ssh in VsCode and check "Extension Hoste (Remote)"
There is enough space in the directories to expand / save files up to 20 GB that should not be the problem.

024-07-30 15:02:24.434 [info] ExtensionService#_doActivateExtension vscode.emmet, startup: false, activationEvent: 'onLanguage'
**2024-07-30 15:02:24.435 [info] ExtensionService#_doActivateExtension ms-vscode.cpptools, startup: false, activationEvent: 'onLanguage:FortranFreeForm', root cause: fortran-lang.linter-gfortran
2024-07-30 15:02:24.770 [error] Activating extension ms-vscode.cpptools failed due to an error:
2024-07-30 15:02:24.770 [error] RangeError: WebAssembly.Instance(): Out of memory: Cannot allocate Wasm memory for new instance**
	at 1298 (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:28534:22)
	at __webpack_require__ (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:116162:42)
	at 6278 (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:40823:16)
	at __webpack_require__ (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:116162:42)
	at 2277 (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:83848:22)
	at __webpack_require__ (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:116162:42)
	at 1817 (/hpc/uhome/mdeberts/.vscode-server/extensions/ms-vscode.cpptools-1.21.4-linux-x64/dist/src/main.js:72989:20)
[...]

PR #12579, opened 2024-08-16
Remove use of `editorconfig` module to avoid dependency on `one-ini` wasm file.

Colengms:
This should address: https://github.com/microsoft/vscode-cpptools/issues/12529

The `editorconfig` module was pulling in the `one-ini` module, which included a WASM file.  Loading of WASM files appears to be a problem (see #12529 ) due to the amount of virtual memory allocated being a problem for systems imposing memory limits.

Replaces use of `editorconfig` lookup with our own code.


Extension/src/LanguageServer/settings.ts
```diff
@@ -1129,12 +1128,107 @@ function mapWrapToEditorConfig(value: string | undefined): string {
     return "never";
 }
 
+function matchesSection(filePath: string, section: string): boolean {
```
bobbrow: should we add a new editorConfig.ts file to hold this stuff?

sean-mcmanus (dismissed):

sean-mcmanus (approved):

Issue #12596, opened 2024-08-20
Remove tag parser fallback setting


bobbrow:
### Feature Request

It is time to retire the tag parser fallback feature. We should remove the setting and code related to it.

PR #12600, opened 2024-08-20
Remove usage of intellisenseEngineFallback setting.

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12596


Colengms (approved):

sean-mcmanus (commented):
Is it best practice to remove settings or should we mark it with a deprecationMessage first?

sean-mcmanus (commented):
If we don't mark it with a deprecationMessage first, then I think it will just appear grayed out and users might be confused as to why, since no message is shown. I'm not sure if that matters or not though, since we don't remove settings often.

bobbrow:
> If we don't mark it with a deprecationMessage first, then I think it will just appear grayed out and users might be confused as to why, since no message is shown. I'm not sure if that matters or not though, since we don't remove settings often.

I'm not concerned about this.

sean-mcmanus (approved):

PR #12612, opened 2024-08-22
Switch from LSP hover to HoverProvider

Colengms:
This is mainly to support Spencer's work on AI hover resultd.  Moving hover processing out of LSP and into a `HoverProvider` provides some additional flexibility.

There is a native-side PR as well.

### DIFF
--- a/Extension/src/LanguageServer/Providers/HoverProvider.ts
+++ b/Extension/src/LanguageServer/Providers/HoverProvider.ts
@@ -38,10 +38,13 @@ export class HoverProvider implements vscode.HoverProvider {
         }
         // VS Code doesn't like the raw objects returned via RPC, so we need to create proper VS Code objects here.
         const strings: vscode.MarkdownString[] = [];
-        let markdownString: vscode.MarkdownString;
         for (const element of hoverResult.contents) {
-            markdownString = element as vscode.MarkdownString;
-            strings.push(markdownString);
+            const oldMarkdownString: vscode.MarkdownString = element as vscode.MarkdownString;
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/Providers/HoverProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+            if (e instanceof ResponseError && (e.code === RequestCancelled || e.code === ServerCancelled)) {
+                throw new vscode.CancellationError();
+            }
+            throw e;
+        }
+        if (token.isCancellationRequested) {
+            throw new vscode.CancellationError();
+        }
+        // VS Code doesn't like the raw objects returned via RPC, so we need to create proper VS Code objects here.
+        const strings: (vscode.MarkdownString | vscode.MarkedString)[] = new Array<vscode.MarkdownString | vscode.MarkedString>();
```
sean-mcmanus: I think you get rid of `| vscode.MarkedString`

sean-mcmanus (changes requested):

Extension/src/LanguageServer/Providers/HoverProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+        let markdownString: vscode.MarkdownString;
+        for (const element of hoverResult.contents) {
+            markdownString = element as vscode.MarkdownString;
+            strings.push(markdownString.value);
```
sean-mcmanus: I think this should be `markdownString`.

sean-mcmanus (approved):

PR #12635, opened 2024-08-26
Log the values of editor settings that we track

bobbrow:
A follow up to the previous logging change. Let me know if this is overkill.  We don't have code that tracks changes for editor settings, so this initial proposal will always log the values.

This may or may not be desirable so I'm soliciting feedback about the path forward. Should we:
1. Always log these values?
2. Always log a subset of these values? (e.g. if we don't think we care about some of them)
3. Write code to check if the values are not the default before logging them? The right solution here might involve some refactoring in the settings tracker class or moving the code to settings.ts (more work).

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1960,6 +1960,23 @@ export class DefaultClient implements Client {
             configJson += `Modified Settings:\n${JSON.stringify(settings, null, 4)}\n`;
         }
 
+        {
+            const editorSettings = new OtherSettings(this.RootUri);
+            const settings: Record<string, any> = {};
+            settings.editorTabSize = editorSettings.editorTabSize;
+            settings.editorInsertSpaces = editorSettings.editorInsertSpaces;
+            settings.editorAutoClosingBrackets = editorSettings.editorAutoClosingBrackets;
+            settings.filesEncoding = editorSettings.filesEncoding;
+            settings.filesAssociations = editorSettings.filesAssociations;
+            settings.filesExclude = editorSettings.filesExclude;
+            settings.filesAutoSaveAfterDelay = editorSettings.filesAutoSaveAfterDelay;
+            settings.editorInlayHintsEnabled = editorSettings.editorInlayHintsEnabled;
+            settings.editorParameterHintsEnabled = editorSettings.editorParameterHintsEnabled;
+            settings.searchExclude = editorSettings.searchExclude;
+            settings.workbenchSettingsEditor = editorSettings.workbenchSettingsEditor;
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -1960,6 +1960,23 @@ export class DefaultClient implements Client {
...
+            settings.editorAutoClosingBrackets = editorSettings.editorAutoClosingBrackets;
+            settings.filesEncoding = editorSettings.filesEncoding;
+            settings.filesAssociations = editorSettings.filesAssociations;
+            settings.filesExclude = editorSettings.filesExclude;
+            settings.filesAutoSaveAfterDelay = editorSettings.filesAutoSaveAfterDelay;
+            settings.editorInlayHintsEnabled = editorSettings.editorInlayHintsEnabled;
+            settings.editorParameterHintsEnabled = editorSettings.editorParameterHintsEnabled;
+            settings.searchExclude = editorSettings.searchExclude;
+            settings.workbenchSettingsEditor = editorSettings.workbenchSettingsEditor;
+            configJson += `Tracked Editor Settings:\n${JSON.stringify(settings, null, 4)}\n`;
```
sean-mcmanus: Shouldn't "Editor" be something like "VS Code'? search.exclude isn't an editor setting for example.

sean-mcmanus (changes requested):

sean-mcmanus:
Logging all the VS Code settings, including the unmodified ones, seem fine to me -- it's already a subset of all the VS Code settings.

Extension/src/LanguageServer/client.ts
bobbrow: VS Code is the "editor." I wasn't specifically referring to the "editor" configuration section.  Using "VS Code" here seemed like it could mean all settings, including extensions.

sean-mcmanus: But I think users will interpret "editor" as the configuration section of settings and not "VS Code excluding extension settings". Maybe it could include other extension settings in the future if we have a need to check them (e.g. a cmake tools setting).

bobbrow: I don't think our users think of the configuration sections that way at all. That line of thinking is only apparent to extension developers who have some experience with the concept.

To avoid this issue, I can call it "additional settings" or "other settings" if that resolves the issue.

sean-mcmanus: The settings themselves are like "editor.tabSize": 
image
That isn't relevant/exposed to just extension developers.

But I'm fine with additional/other settings too.

sean-mcmanus (approved):

PR #12678, opened 2024-09-03
The compiler path selection control is not in sync with the textbox

### LINKED ISSUE #7427: The compilerPath option does not match the tick in front of the option in the drop-down box (Mac OS only)
Bug type: Language Service

**Describe the bug**
- OS: Mac and Mac ARM64
- C/C++ Extension Version: 1.3.0

**Steps to reproduce**

1. Open IntelliSense Configuration UI
2. Change the compiler path selection option.
3. The option does not match the tick in front of the option in the drop-down box (select any of the compiler paths, re-open the drop-down box, the tick is in the first path unchanged)
image

**Expected behavior**
The "check mark" should be in front of the corresponding compilation path.

bobbrow:
Fixes: #7427

The faux editable combobox we use to set the compilerPath property was not keeping the select control in sync with the contents of the textbox.  This change fixes it.

### DIFF
--- a/Extension/ui/settings.ts
+++ b/Extension/ui/settings.ts
@@ -109,8 +109,8 @@ class SettingsApp {
         document.getElementById(elementId.configName)?.addEventListener("change", this.onConfigNameChanged.bind(this));
         document.getElementById(elementId.configSelection)?.addEventListener("change", this.onConfigSelect.bind(this));
         document.getElementById(elementId.addConfigBtn)?.addEventListener("click", this.onAddConfigBtn.bind(this));
-        document.getElementById(elementId.addConfigOk)?.addEventListener("click", this.OnAddConfigConfirm.bind(this, true));
-        document.getElementById(elementId.addConfigCancel)?.addEventListener("click", this.OnAddConfigConfirm.bind(this, false));
+        document.getElementById(elementId.addConfigOk)?.addEventListener("click", this.onAddConfigConfirm.bind(this, true));
+        document.getElementById(elementId.addConfigCancel)?.addEventListener("click", this.onAddConfigConfirm.bind(this, false));
     }
 
     private onTabKeyDown(e: any): void {
@@ -148,7 +148,7 @@ class SettingsApp {
         this.showElement(elementId.addConfigInputDiv, true);
     }
 
-    private OnAddConfigConfirm(request: boolean): void {
+    private onAddConfigConfirm(request: boolean): void {
         this.showElement(elementId.addConfigInputDiv, false);
         this.showElement(elementId.addConfigDiv, true);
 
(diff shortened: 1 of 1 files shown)


bobbrow:
Suddenly it doesn't compile. I'm investigating why. It has to do with importing `path.normalize` which didn't repro when I was testing, but now it does. :(

Extension/ui/settings.ts
```diff
@@ -212,10 +212,24 @@ class SettingsApp {
...
+    }
 
-        // Reset selection to none
-        el.value = "";
+    // To enable custom entries, the compiler path control is a text box on top of a select control.
+    // This function ensures that the select control is updated when the text box is changed.
+    private fixKnownCompilerSelection(): void {
+        const compilerPath = <HTMLInputElement>document.getElementById(elementId.compilerPath);
+        const knownCompilers = <HTMLSelectElement>document.getElementById(elementId.knownCompilers);
+        for (let n = 0; n < knownCompilers.options.length; n++) {
```
sean-mcmanus, line 222: Can you do 

```typescript
        for (let n = 0; n < knownCompilers.options.length; n++) {
            if (compilerPath.value.toLowerCase() === knownCompilers.options[n].value.toLowerCase()) {
                knownCompilers.value = knownCompilers.options[n].value;
                return;
            }
            knownCompilers.value = '';
        }
````

File: Extension/ui/settings.ts [hunk 2]
```diff
@@ -212,10 +212,24 @@ class SettingsApp {
...
             command: "knownCompilerSelect"
         });
+    }
 
-        // Reset selection to none
-        el.value = "";
+    // To enable custom entries, the compiler path control is a text box on top of a select control.
+    // This function ensures that the select control is updated when the text box is changed.
+    private fixKnownCompilerSelection(): void {
+        const compilerPath = <HTMLInputElement>document.getElementById(elementId.compilerPath);
```
sean-mcmanus: Can you do something like

```typescript
const compilerPath = (<HTMLInputElement>document.getElementById(elementId.compilerPath)).value.toLowerCase();
```

sean-mcmanus (changes requested):

Extension/ui/settings.ts
bobbrow: I refactored this into a function after some testing revealed I needed to do it in two places, but I didn't take advantage of the fact that I could do an early return now.

I don't think I want to set the value to `''` N times during this loop though.  What I should do is move that out of the loop and let the early return handle the other case.

Extension/ui/settings.ts
bobbrow: yes, I can avoid lowercasing this N times.  Good catch.

sean-mcmanus (approved):

PR #12685, opened 2024-09-05
Add C++ configuration as a language model tool (luca)

lukka:
Addressing feedback from PR https://github.com/microsoft/vscode-cpptools/pull/12577


Extension/src/LanguageServer/utils.ts
```diff
@@ -1,101 +1,108 @@
...
+            false)
+        );
+    return;
+}
+
+const docsChangedFromCppToC: Set<string> = new Set<string>();
+
+export async function withCancellation<T>(promise: Promise<T>, token: vscode.CancellationToken): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+        token.onCancellationRequested(() => reject(new vscode.CancellationError()));
```
Colengms: Carrying this over from the other PR:

Could you grab the Disposable that gets returned from this event registration, and dispose of it when resolving? That way, if this function is called multiple times with the same token, multiple event registrations won't get accumulated/leaked.

Extension/src/LanguageServer/client.ts
```diff
@@ -4114,4 +4145,5 @@ class NullClient implements Client {
     setShowConfigureIntelliSenseButton(show: boolean): void { }
     addTrustedCompiler(path: string): Promise<void> { return Promise.resolve(); }
     getIncludes(): Promise<GetIncludesResult> { return Promise.resolve({} as GetIncludesResult); }
+    getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
```
Colengms, line 4151: (Moved this over from the other PR)

>> Should this return Promise<ChatContextResult | undefined> to match the other definition?

> Could you advice what changes are needed?

The line current is:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
```

But the interface declares it as:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult | undefined>;
```

I'm just suggesting updating the derived class member function declaration be consistent with the base class, like other functions here.  So:

```
getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult | undefined> { return Promise.resolve({} as ChatContextResult); }
```

Extension/src/LanguageServer/client.ts
```diff
@@ -2234,6 +2244,27 @@ export class DefaultClient implements Client {
...
 
+    public async getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> {
+        await withCancellation(this.ready, token);
+        let result: ChatContextResult;
+        try {
+            result = await this.languageClient.sendRequest(CppContextRequest, null, token);
+        } catch (e: any) {
+            // From <https://microsoft.github.io/language-server-protocol/specification#responseMessage>
+            // RequestCancelled = -32800, ServerCancelled = -32802,
+            if (e instanceof ResponseError && (e.code === -32800 /*RequestCancelled*/ || e.code === -32802 /*ServerCancelled*/)) {
```
Colengms: In multiple other places, we use constants for these (RequestCancelled, ServerCancelled).  Could you use those instead of hard-coding values?

Colengms (changes requested):

Extension/src/LanguageServer/client.ts
lukka: Of course, I am not that crazy to push this code further, I've done this because I could not find any ref to those and I was sure you were going to point me in the right direction. I think after I rebased onto main I can see them now :)

Extension/src/LanguageServer/client.ts
lukka: @Colengms what about dropping undefined everywhere? That is what I've actually done.

sean-mcmanus (dismissed):

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
benmcmorran: I think this is a leftover from when the native side was still returning optional values. Removing `undefined` everywhere should be the correct change.

Extension/src/LanguageServer/extension.ts
```diff
@@ -248,6 +249,11 @@ export async function activate(): Promise<void> {
         clients.timeTelemetryCollector.setFirstFile(activeEditor.document.uri);
         activeDocument = activeEditor.document;
     }
+
+    if (util.extensionContext) {
```
benmcmorran: Leaving a note so it's not forgotten -- I think we also want to gate on the experimental features setting here.

Extension/src/LanguageServer/client.ts
lukka: done

Extension/src/LanguageServer/utils.ts
lukka: done!

Extension/src/LanguageServer/extension.ts
lukka: done! Let me know if this is the expected way of doing it.

Extension/src/LanguageServer/client.ts
lukka: fixed

Colengms (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -248,6 +249,11 @@ export async function activate(): Promise<void> {
         clients.timeTelemetryCollector.setFirstFile(activeEditor.document.uri);
         activeDocument = activeEditor.document;
     }
+
+    if (util.extensionContext && new CppSettings().experimentalFeatures) {
```
lukka, line 253: @benmcmorran 
this is not great because it will throw whenever the `#cpp` tool is invoked by the user on the Chat prompt:

image


Should a `when` clause added to `languageModelTools`?

benmcmorran: Good catch, I forgot these are statically registered now. A `when` clause makes sense to me.

lukka: added a `when:` clause

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -0,0 +1,109 @@
...
+        'c23': "C23"
+    },
+    targetPlatform: {
+        'windows': 'Windows',
+        'Linux': 'Linux',
+        'macos': 'macOS'
+    }
+};
+
+class StringLanguageModelToolResult implements vscode.LanguageModelToolResult
```
sean-mcmanus: Can you run the formatter on this document?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: image

Extension/src/LanguageServer/utils.ts
```diff
@@ -1,101 +1,115 @@
...
+const docsChangedFromCppToC: Set<string> = new Set<string>();
+
+export async function withCancellation<T>(promise: Promise<T>, token: vscode.CancellationToken): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+        const disposable = token.onCancellationRequested(() => reject(new vscode.CancellationError()));
+        promise.then((value) => {
+            disposable.dispose();
+            resolve(value);
+        }, (reason) =>
+        {
```
sean-mcmanus: Can you run the formatter on this document?

image

lukka: done, fixed the settings.json to use the default vscode linter.

sean-mcmanus (dismissed):

sean-mcmanus (approved):

Extension/src/LanguageServer/lmTool.ts
lukka: yes, done

Colengms (approved):

Issue #4077, opened 2019-08-13
Switch to using setTextDocumentLanguage instead of adding to files.associations


sean-mcmanus:
The Septemember 2018 release added vscode.languages.setTextDocumentLanguage, so our extension can try using that instead of adding files to files.associations.

The previous thread is https://github.com/microsoft/vscode-cpptools/issues/722 .

sean-mcmanus:
Actually, this is problematic, because setTextDocumentLanguage is not persistent like files.associations is...

bobbrow:
@sean-mcmanus did you try this and verify that it does not persist?  We may need to re-engage with the VS Code team on this and come up with a solution that works for us.

bobbrow:
The VS Code issue on this: microsoft/vscode#78968

sean-mcmanus:
I think this needs to be spec'd out. The set document language doesn't persist. We could persist it in a "hidden" way, but then users wouldn't have any ability to set those settings. Maybe we could add a setting for users who strong don't want the files.associations in settings.json, and maybe we could change the "hidden" setting after the user manually changes the language via the status bar (not sure).

ahojnnes:
@sean-mcmanus @bobbrow I am wondering whether there is any progress on this issue? I am hitting exactly the same issue, where I don't want the C++ plugin to modify my settings, which can be achieved by `"C_Cpp.autoAddFileAssociations": false,` but then the opened STL headers through code navigation are opened as plain text. Thanks.

sean-mcmanus:
@ahojnnes We could dynamically setTextDocumentLanguage, but the problem that needs to be solved is how to decide when to call that -- did you have a proposed solution?  I believe we would need language type auto-detection for extensionless files from VS Code: https://github.com/microsoft/vscode/issues/130642

ahojnnes:
Unless I am missing something, shouldn't the language always be C++ when following headers from existing C++ code? I guess one could end up in a external C file but then C++ should still be good enough?

sean-mcmanus:
Yeah, if we're opening a file from a header, then the setTextDocumentLanguage will work, but then after a Reload Window or any other case where the header happens to be open again, the language would be reset unless the association is saved somehow.

We could fix it to call setTextDocumentLanguage on go to definition when the autoAddFileAssociaitons is set to false.

ahojnnes:
> We could fix it to call setTextDocumentLanguage on go to definition when the autoAddFileAssociaitons is set to false.

[...]

PR #12687, opened 2024-09-05
Implement a system that manages file associations without directly modifying the files.associations setting

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/4077


bobbrow:
We decided to add support just for the STL headers and disabled writing to files.associations by default.

PR #12700, opened 2024-09-10
Show the reload window message when caseSensitiveFileSupport is changed.

### LINKED ISSUE #12699: No reload window message is shown after the C_Cpp.caseSensitiveFileSupport is changed (on Windows)
I hit this while investigating the caseSensitiveFileSupport bug https://github.com/microsoft/vscode-cpptools/issues/12672 .

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/12699


Colengms (changes requested):
Would it be possible to trigger the prompt entirely on the TypeScript side?  Seems odd to plumb up a bool to the native side for the purpose of sending a message back down to the TypeScript code to call `promptForReloadWindowDueToSettingsChange`.

sean-mcmanus:
> Would it be possible to trigger the prompt entirely on the TypeScript side? Seems odd to plumb up a bool to the native side for the purpose of sending a message back down to the TypeScript code to call `promptForReloadWindowDueToSettingsChange`.

Yeah, maybe -- I was just following the previous example with experimentalFeatures. I'm not sure why that case also wasn't handled all on the TypeScript side too.

Colengms (approved):

PR #12727, opened 2024-09-17
Rework custom config requests

Colengms:

This PR removes the delivery of a custom configuration immediately before processing the didOpen, in favor of the newer mechanism of requesting each configuration as needed.  The original approach predated needing configurations for other purposes, such as Find All References.

Benefits are:
 - No need to look up and deliver configurations for files opened internally by VS Code and never active in the editor.
 - Code simplification/removal. Removed `QueryTranslationUnitSource` request.
 - Removing the potential for a config provider to stall our message queue due and delay delivery of didOpen.


Extension/src/LanguageServer/client.ts
```diff
@@ -3144,9 +3102,10 @@ export class DefaultClient implements Client {
             workspaceFolderUri: this.RootUri?.toString()
         };
 
-        if (wasRequested) {
-            void this.languageClient.sendNotification(CustomConfigurationHighPriorityNotification, params).catch(logAndReturn.undefined);
-        }
+        // We send the higher priority notification to ensure we don't deadlock if the request is blocking the queue.
+        // We send the normal priroity notification to avoid a race that could result in a redundant request when racing with
```
sean-mcmanus: priroity->priority

sean-mcmanus (dismissed):

sean-mcmanus (approved):

PR #12735, opened 2024-09-19
Migrate to Copilot related files API

benmcmorran:
This change migrates to the new API so that C++ is no longer a special case for related files. I set this up to support A/B experimentation so that we can verify that there are no regressions during the migration.

FYI @lukka @genlu @sandersn

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -266,12 +266,16 @@ export async function activate(): Promise<void> {
     if (isRelatedFilesApiEnabled) {
         const api = await getCopilotApi();
         if (util.extensionContext && api) {
-            for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
-                api.registerRelatedFilesProvider(
-                    { extensionId: util.extensionContext.extension.id, languageId },
-                    async (_uri: vscode.Uri) =>
-                        ({ entries: (await clients.ActiveClient.getIncludes(1))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [] })
-                );
+            try {
+                for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
+                    api.registerRelatedFilesProvider(
+                        { extensionId: util.extensionContext.extension.id, languageId },
+                        async (_uri: vscode.Uri) =>
+                            ({ entries: (await clients.ActiveClient.getIncludes(1))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [] })
+                    );
+                }
+            } catch {
+                console.log("Failed to register Copilot related files provider.");
             }
         }
     }
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -264,9 +264,8 @@ export async function activate(): Promise<void> {
     }
 
     if (isRelatedFilesApiEnabled) {
-        const copilotExtension = vscode.extensions.getExtension<CopilotApi>('github.copilot');
-        if (util.extensionContext && copilotExtension) {
-            const api: CopilotApi = copilotExtension.isActive ? copilotExtension.exports : await copilotExtension.activate();
+        const api = await getCopilotApi();
+        if (util.extensionContext && api) {
             for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
                 api.registerRelatedFilesProvider(
                     { extensionId: util.extensionContext.extension.id, languageId },
@@ -372,7 +371,7 @@ function onInterval(): void {
 /**
  * registered commands
  */
-export function registerCommands(enabled: boolean, isRelatedFilesApiEnabled: boolean): void {
+export function registerCommands(enabled: boolean, isRelatedFilesApiEnabled: boolean = false): void {
     commandDisposables.forEach(d => d.dispose());
     commandDisposables.length = 0;
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.SwitchHeaderSource', enabled ? onSwitchHeaderSource : onDisabledCommand));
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -254,6 +261,20 @@ export async function activate(): Promise<void> {
         const tool = vscode.lm.registerTool('cpptools-lmtool-configuration', new CppConfigurationLanguageModelTool());
         disposables.push(tool);
     }
+
+    if (await telemetry.isExperimentEnabled("CppToolsRelatedFilesApi")) {
```
benmcmorran: I've confirmed with the experimentation team that this should work even though we will be measuring telemetry from the GitHub Copilot extension, not cpptools.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
bobbrow: Instead of calling this here and in `registerCommands`, can you call it right before `registerCommands` and pass it in there so that that function doesn't need to be async?  There's probably no harm in registering a no-op in the enabled == false case.

benmcmorran:
@sean-mcmanus I'm not authorized to push to `main`. Can you merge this PR when you get a chance? Thanks.

Extension/src/LanguageServer/extension.ts
benmcmorran: Changed so that `registerCommands()` is no longer async.

Extension/src/main.ts
```diff
@@ -146,7 +146,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<CppToo
     if (shouldActivateLanguageServer) {
         await LanguageServer.activate();
     } else if (isIntelliSenseEngineDisabled) {
-        LanguageServer.registerCommands(false);
+        LanguageServer.registerCommands(false, /* isRelatedFilesApiEnabled */ false);
```
bobbrow: Another idea is that you could have this second parameter be optional and only add the registration if the variable is true. Up to you though.

Extension/src/LanguageServer/extension.ts
```diff
@@ -254,6 +262,20 @@ export async function activate(): Promise<void> {
         const tool = vscode.lm.registerTool('cpptools-lmtool-configuration', new CppConfigurationLanguageModelTool());
         disposables.push(tool);
     }
+
+    if (isRelatedFilesApiEnabled) {
+        const copilotExtension = vscode.extensions.getExtension<CopilotApi>('github.copilot');
+        if (util.extensionContext && copilotExtension) {
+            const api: CopilotApi = copilotExtension.isActive ? copilotExtension.exports : await copilotExtension.activate();
```
bobbrow: In our API's helper, we catch exceptions which can be thrown during `activate()`. You should do that too.

https://github.com/microsoft/vscode-cpptools-api/blob/5eff4c0ff7ad41ef58ba4e98bae5cc8420d2a15c/api.ts#L334C9-L340C10

Extension/src/LanguageServer/extension.ts
```diff
@@ -254,6 +262,20 @@ export async function activate(): Promise<void> {
...
+            for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
+                api.registerRelatedFilesProvider(
```
bobbrow: Since we're calling into a potentially dynamic API that doesn't have type definitions exported, I think it would be good to validate that `api.registerRelatedFilesProvider` exists before trying to call it.  A try/catch might not be a bad idea either.

Extension/src/main.ts
benmcmorran: Changed to an optional parameter.

Extension/src/LanguageServer/extension.ts
benmcmorran: Added a try-catch for accessing the API.

Extension/src/LanguageServer/extension.ts
benmcmorran: Added a try-catch for interaction with the API.

sean-mcmanus (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -408,7 +433,10 @@ export function registerCommands(enabled: boolean): void {
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.ExtractToFreeFunction', enabled ? () => onExtractToFunction(true, false) : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.ExtractToMemberFunction', enabled ? () => onExtractToFunction(false, true) : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('C_Cpp.ExpandSelection', enabled ? (r: Range) => onExpandSelection(r) : onDisabledCommand));
-    commandDisposables.push(vscode.commands.registerCommand('C_Cpp.getIncludes', enabled ? (maxDepth: number) => getIncludes(maxDepth) : () => Promise.resolve()));
+
+    if (!isRelatedFilesApiEnabled) {
```
bobbrow, line 437: I think the `!` should be removed.

bobbrow: You might also do it like this, removing the conditional block - if any other place may end up calling `getIncludes` even when the experiment is disabled: 
```ts
commandDisposables.push(vscode.commands.registerCommand('C_Cpp.getIncludes', enabled && isRelatedFilesApiEnabled ? (maxDepth: number) => getIncludes(maxDepth) : () => Promise.resolve()));
```

benmcmorran: This is intentionally negated. We plan to completely remove the `getIncludes` command if the related files API behaves as expected. For an accurate experiment, we need to make sure that the treatment group doesn't have access to the `getIncludes` command at all to catch any regressions. Unless your team plans to support the `getIncludes` command for third parties, as far as I'm aware the Copilot extension is the only consumer of this command.

Colengms (approved):

Extension/src/LanguageServer/extension.ts
bobbrow: Oh, I see. `getIncludes` is going away.  Got it.

PR #12754, opened 2024-09-23
Create stale

FreeMason9224:
Close stale issues


sean-mcmanus:
I think we may not want to auto-close "stale" issues. That might close a lot of issues that should be kept open. We already have an auto-close for stale issues that are marked with "more info needed".

PR #12773, opened 2024-09-26
Add cancellation token to react to Copilot API changes

benmcmorran:
Reacts to upstream changes in the Copilot API that add a cancellation token and make it possible to unregister a provider.

FYI @sandersn

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -36,8 +36,8 @@ import { makeLspRange, rangeEquals, showInstallCompilerWalkthrough } from './uti
 interface CopilotApi {
     registerRelatedFilesProvider(
         providerId: { extensionId: string; languageId: string },
-        callback: (uri: vscode.Uri) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
-    ): void;
+        callback: (uri: vscode.Uri, token: vscode.CancellationToken) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
+    ): vscode.Disposable;
 }
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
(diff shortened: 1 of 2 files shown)


sean-mcmanus (dismissed):

sean-mcmanus:
@benmcmorran Are you not able to squash and merge? I could try switching your permissions from Write to Maintain.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1403,7 +1403,9 @@ export async function preReleaseCheck(): Promise<void> {
 }
 
 export async function getIncludes(maxDepth: number): Promise<any> {
-    const includes = await clients.ActiveClient.getIncludes(maxDepth);
+    const tokenSource = new vscode.CancellationTokenSource();
+    const includes = await clients.ActiveClient.getIncludes(maxDepth, tokenSource.token);
```
Colengms, line 1407: We'd 'fixed' LSP cancelation such that it triggers an exception on cancellation.  However, it doesn't trigger the exception that VS Code generally expects, which would be a `CancellationError`.  I'm not sure what the caller would expect in this case.  Do you know?

Colengms (changes requested):
(See existing comment.  Waiting for clarification before approving.)

Extension/src/LanguageServer/extension.ts
benmcmorran: Thanks, I've checked and updated this to use the same VS Code semantics as the chat context request.

benmcmorran:
@Colengms I've changed the cancellation handling to follow the same pattern used by chat context. Let me know if you still have any concerns.

Colengms (approved):

benmcmorran:
@sean-mcmanus I'm still not able to squash and merge. I suspect there are extra branch polices on `main` specifically that don't include me.

image

sean-mcmanus:
> @sean-mcmanus I'm still not able to squash and merge. I suspect there are extra branch polices on `main` specifically that don't include me.

You should have permissions now.

sean-mcmanus (approved):

PR #12790, opened 2024-09-30
Update changelog for 1.22.7.

sean-mcmanus:


Colengms (approved):

Colengms (approved):

Extension/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Version 1.22.7: September 30, 2024
+### Enhancement
+* The .vsix and .js files are now signed. [#12725](https://github.com/microsoft/vscode-cpptools/issues/12725), [#12744](https://github.com/microsoft/vscode-cpptools/issues/12744)
+
+### Bug Fixes
+* Fix some Copilot-related issues. [PR #12773](https://github.com/microsoft/vscode-cpptools/pull/12773)
+* Fix an issue preventing use of a full command line in `compilerPath`. [PR #12774](https://github.com/microsoft/vscode-cpptools/pull/12774)
+* Fix an infinite loop on shutdown after changing the selected settings.
+* Fix a crash (from `insert_lines`).
```
bobbrow, line 11: People aren't going to know what 'insert_lines' means. We should remove that from the final changelog.

sean-mcmanus: @bobbrow It will be removed from the final changelog anyway, because it was a regression crash fix only present in insiders. But I think it's okay, because if a user sees "insert_lines" in their crash call stack then they'd know that is fixed with this release.

PR #12803, opened 2024-10-03
Restore the functionality to filter out C++ headers outside current workspace

lukka:
The filtering has been lost due to missing unit test for this specific case in the original repository where the functionality lived. Now the filtering has been moved here.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -20,7 +20,7 @@ import * as util from '../common';
 import { getCrashCallStacksChannel } from '../logger';
 import { PlatformInformation } from '../platform';
 import * as telemetry from '../telemetry';
-import { Client, DefaultClient, DoxygenCodeActionCommandArguments, openFileVersions } from './client';
+import { Client, DefaultClient, DoxygenCodeActionCommandArguments, GetIncludesResult, openFileVersions } from './client';
 import { ClientCollection } from './clientCollection';
 import { CodeActionDiagnosticInfo, CodeAnalysisDiagnosticIdentifiersAndUri, codeAnalysisAllFixes, codeAnalysisCodeToFixes, codeAnalysisFileToCodeActions } from './codeAnalysis';
 import { CppBuildTaskProvider } from './cppBuildTaskProvider';
@@ -282,7 +282,7 @@ export async function activate(): Promise<void> {
                     api.registerRelatedFilesProvider(
                         { extensionId: util.extensionContext.extension.id, languageId },
                         async (_uri: vscode.Uri, _context: { flags: Record<string, unknown> }, token: vscode.CancellationToken) =>
-                            ({ entries: (await clients.ActiveClient.getIncludes(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [] })
+                            ({ entries: (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [] })
                     );
                 }
             } catch {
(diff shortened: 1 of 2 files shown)


lukka:
@benmcmorran @spebl FYI

Extension/src/LanguageServer/extension.ts
```diff
@@ -1413,13 +1413,33 @@ export async function preReleaseCheck(): Promise<void> {
     }
 }
 
-export async function getIncludes(maxDepth: number): Promise<any> {
-    const tokenSource = new vscode.CancellationTokenSource();
-    const includes = await clients.ActiveClient.getIncludes(maxDepth, tokenSource.token);
-    tokenSource.dispose();
+export async function getIncludesWithCancellation(maxDepth: number, token: vscode.CancellationToken): Promise<GetIncludesResult> {
+    const includes = await clients.ActiveClient.getIncludes(maxDepth, token);
+    const wksFolder = clients.ActiveClient.RootUri
```
Colengms: I don't believe it's possible for the RootUri to not refer to the primary workspace folder.  Maybe this filtering isn't needed?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1413,13 +1413,33 @@ export async function preReleaseCheck(): Promise<void> {
...
+            clients.ActiveClient.RootUri?.toString().startsWith(folder.uri.toString())
+        )?.uri
+        : null;
+
+    if (!wksFolder) {
+        return includes;
+    }
+
+    const wksFolderString = wksFolder.toString();
+    includes.includedFiles = includes.includedFiles.filter(header => vscode.Uri.file(header).toString().startsWith(wksFolderString));
```
Colengms: I guess you're filtering out everything not under the 'primary' workspace folder?  That may filter out much more than standard headers.  It would seem to filter out all headers under other roots in a multiroot workspace.  And users may being using libraries, etc., that are not stored under their workspace folder.

Colengms (dismissed):

Extension/src/LanguageServer/extension.ts
lukka: This is the check we had, and the one we experimented with, providing positive results. In this PR the intention is to restore the status quo rather than trying a different approach, so exclusing out-of-the-folder header is by-design.

Extension/src/LanguageServer/extension.ts
lukka: Thanks! I dropped the search for the current workspace, and assume that the RootUri is the one.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1413,13 +1413,28 @@ export async function preReleaseCheck(): Promise<void> {
     }
 }
 
-export async function getIncludes(maxDepth: number): Promise<any> {
-    const tokenSource = new vscode.CancellationTokenSource();
-    const includes = await clients.ActiveClient.getIncludes(maxDepth, tokenSource.token);
-    tokenSource.dispose();
+export async function getIncludesWithCancellation(maxDepth: number, token: vscode.CancellationToken): Promise<GetIncludesResult> {
+    const includes = await clients.ActiveClient.getIncludes(maxDepth, token);
```
sean-mcmanus, line 1417: Why are you not doing this filtering at the source in the cpptools-srv process? Also, it looks like includes from other workspace folders might be missing at that location.

sean-mcmanus: The cpptools-srv side should know if a folder is a system include or not (although there's a bug in that I have a PR to fix).

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I don't think it's a different approach. It's just fixing a bug with the original approach.

It should not be filtering out headers that are included from other user workspace folders.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: I see it missing folders from my 2nd workspace folder with a complex workspace, but they exist in a simple workspace, so there might be some other reason those includes are missing.

sean-mcmanus: Maybe due to maxDepth being 1.

sean-mcmanus: Oh, never mind, I was looking at the wrong data in the debugger. I see includes from other workspace folders now.

sean-mcmanus:
@lukka In this simple example, I only see test1.h reported -- is that a bug?

image

Extension/src/LanguageServer/extension.ts
lukka: @sean-mcmanus I agree that trying excluding only the system headers is what we should experiment with in the future. As of now, we only want to restore the functionality exactly as it was.

Extension/src/LanguageServer/extension.ts
lukka: @Colengms  @sean-mcmanus we could discuss more offline about this since it would take time, but it is an interesting topic. In short, increasing dramatically the count of headers would not necessarily increase the accuracy of our algorithms (rather the opposite).
Hence while I totally agree with you that trying to exclude only the system headers is a good idea we should experiment in the future with. As of now, we only want to restore the functionality exactly as it was and as with got positive results from.

lukka:
> @lukka In this simple example, I only see test1.h reported -- is that a bug?
> 
> image

If one header is nor reported, looks like a bug indeed. Could you share a repro archive of the project you used? I'll look into it.

benmcmorran (approved):

kuchungmsft (approved):

sean-mcmanus:
> If one header is nor reported, looks like a bug indeed. Could you share a repro archive of the project you used? I'll look into it.

The repro is as seen in the screenshot. I have a source file with 2 headers and only 1 is reported.

sean-mcmanus (approved):

PR #12813, opened 2024-10-07
clang-format/tidy version comparisons fail for some builds

### LINKED ISSUE #12806: clang-format/tidy version comparisons fail for some builds
### Bug Summary 

When the user has installed their own version of clang-format or clang-tidy, but has not specified a specific path to the binary, we will compare one from the user's environment with the one we bundle in the extension to see which version is newer.

The logic currently expects the third token in the version query output to be the version number, but this is not always the case.  I think a more robust solution might attempt to convert each token to a semver until we find a valid one in case there are more or less tokens in the output.

RE: #12718

bobbrow:
Fixes: #12806

Use regular expressions to find the version number instead of token position.

### DIFF
--- a/Extension/src/LanguageServer/settings.ts
+++ b/Extension/src/LanguageServer/settings.ts
@@ -277,7 +277,6 @@ export class CppSettings extends Settings {
             if (cachedClangPath !== undefined) {
                 return cachedClangPath;
             }
-            const clangStr: string = isFormat ? this.clangFormatStr : this.clangTidyStr;
             const clangName: string = isFormat ? this.clangFormatName : this.clangTidyName;
             const setCachedClangPath: (path: string) => void = isFormat ? setCachedClangFormatPath : setCachedClangTidyPath;
             const whichPath: string | null = which.sync(clangName, { nothrow: true });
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):
It seems theoretically possible for a custom clang build to omit a patch version, but I don't think that's normal so maybe it's okay to not support that.

sean-mcmanus:
Is this for 1.22.8 or 1.23.0?

bobbrow:
> It seems theoretically possible for a custom clang build to omit a patch version, but I don't think that's normal so maybe it's okay to not support that.

I suppose we can wait to see if that's a capability people ask for.

bobbrow:
> Is this for 1.22.8 or 1.23.0?

I don't have a strong opinion. If it's not in 1.22, we may be doing a 1.23.0 pre-release shortly after.

sean-mcmanus:
> I don't have a strong opinion. If it's not in 1.22, we may be doing a 1.23.0 pre-release shortly after.

It sounds like 1.22.8 will be a pre-release still so it seems safe enough to go into that.

bobbrow:
ok. I'll merge this in now.

Issue #12807, opened 2024-10-04
Copilot getIncludes with 2 includes only returns the first


sean-mcmanus:
1. Create a new workspace folder. (I think you need the latest GitHub Copilot extensions installed too).
2. Create a test1.cpp
```cpp
#include "test1.h"
#include "test2.h"
```
3. Create a test1.h with `int test1h;` and test2.h with `int test2h;`
4. Use Ctrl+Space or something to trigger the getIncludes call.

Bug: In the debugger you can see it only gets 1 of the includes. That seems like a bug to me (although the maxDepth is 1 and I'm not sure what that means). See the screenshot:

![image](https://github.com/user-attachments/assets/afb9cea8-f8a7-4a14-8419-ae9c371d6466)

I've reproed this on 2 machines, with VS Code Insiders.

It seems possible that some subset of the valid includes are getting lost in general, but I haven't investigated...but I did notice that it seems like includes were missing for our vc project (not sure what % were missing or if the maxDepth was the cause or not).

lukka:
@sean-mcmanus was this a systematic behavior that you were seeing? I cannot repro it, headers are being provided as expected, as shown here:

![Image](https://github.com/user-attachments/assets/82a9206b-7d29-4222-8e3d-c8b8ab1fe1aa)

The list of headers is currently implemented in the native side of the extension, and if you could repro this, I'd be happy to investigate further by attaching a debugger to the machine with a repro.

@benmcmorran  @spebl @kuchungmsft

sean-mcmanus:
@lukka Oh, interesting -- it no longer repros. I don't know what changed. My repro project is the same as when it reproed. I'll let you know if it happened again.

PR #12821, opened 2024-10-08
Add Cpp Context Traits to Completions Prompt

kuchungmsft:
- Move related files code to its own module
- Refactor related files provider code to enable unit testing
- Add Cpp context traits to completions prompt

image

### DIFF
--- a/Extension/src/LanguageServer/copilotProviders.ts
+++ b/Extension/src/LanguageServer/copilotProviders.ts
@@ -1,0 +1,134 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as vscode from 'vscode';
+import * as util from '../common';
+import * as telemetry from '../telemetry';
+import { ChatContextResult, GetIncludesResult } from './client';
+import { getActiveClient } from './extension';
+
+let isRelatedFilesApiEnabled: boolean | undefined;
+
+export interface CopilotTrait {
+    name: string;
+    value: string;
+    includeInPrompt?: boolean;
+    promptTextOverride?: string;
+}
+
+export interface CopilotApi {
+    registerRelatedFilesProvider(
+        providerId: { extensionId: string; languageId: string },
+        callback: (
+            uri: vscode.Uri,
+            context: { flags: Record<string, unknown> },
+            cancellationToken: vscode.CancellationToken
+        ) => Promise<{ entries: vscode.Uri[]; traits?: CopilotTrait[] }>
+    ): Disposable;
+}
+
+export async function registerRelatedFilesProvider(): Promise<void> {
+    if (!await getIsRelatedFilesApiEnabled()) {
+        return;
+    }
+
+    const api = await getCopilotApi();
+    if (util.extensionContext && api) {
+        try {
+            for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
+                api.registerRelatedFilesProvider(
+                    { extensionId: util.extensionContext.extension.id, languageId },
+                    async (_uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken) => {
+
+                        const getIncludesHandler = async () => (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [];
+                        const getTraitsHandler = async () => {
+                            const chatContext: ChatContextResult | undefined = await (getActiveClient().getChatContext(token) ?? undefined);
+
+                            if (!chatContext) {
+                                return undefined;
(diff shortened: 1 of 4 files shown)


sean-mcmanus:
@kuchungmsft What release is this for? 1.23.0?

Extension/test/scenarios/SingleRootProject/tests/relatedFilesProvider.test.ts
```diff
@@ -0,0 +1,305 @@
...
+            id: 'test-extension-id',
+            extensionUri: vscode.Uri.parse('file:///test-extension-path'),
+            extensionPath: 'test-extension-path',
+            isActive: true,
+            packageJSON: { name: 'test-extension-name' },
+            activate: async () => { },
+            exports: mockCopilotApi,
+            extensionKind: vscode.ExtensionKind.UI
+        });
+        activeClientStub.getIncludes.resolves({ includedFiles: ['c:\\system\\include\\vector', 'c:\\system\\include\\string', 'C:\\src\\my_project\\foo.h'] });
```
sean-mcmanus: Can you add a test case with 2 headers for bug https://github.com/microsoft/vscode-cpptools/issues/12807? Or should that be added after that bug is fixed?

kuchungmsft:
> @kuchungmsft What release is this for? 1.23.0?

@benmcmorran can you please comment?

Extension/test/scenarios/SingleRootProject/tests/relatedFilesProvider.test.ts
kuchungmsft: Tried it locally but couldn't repro bug #12807, note that I mocked `Client.getIncludes` in my tests.

image

image

bobbrow:
> @kuchungmsft What release is this for? 1.23.0?

We're not taking any more changes into 1.22, so this will need to wait until 1.23.0

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -0,0 +1,136 @@
...
+                        }
+                        const getTraitsHandler = async () => {
+                            const chatContext: ChatContextResult | undefined = await (getActiveClient().getChatContext(token) ?? undefined);
+
+                            if (!chatContext) {
+                                return undefined;
+                            }
+
+                            let traits: CopilotTrait[] = [
+                                { name: "language", value: chatContext.language, includeInPrompt: true, promptTextOverride: `The language is ${chatContext.language}.` },
```
benmcmorran, line 55: Copilot already includes either a "language marker" or a "path marker" (but not both) at the top of the prompt. This particular trait may end up being redundant with the language marker, but as long as we have independent experimental control, I think it's fine to leave.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -0,0 +1,136 @@
...
+        return;
+    }
+
+    const api = await getCopilotApi();
+    if (util.extensionContext && api) {
+        try {
+            for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
+                api.registerRelatedFilesProvider(
+                    { extensionId: util.extensionContext.extension.id, languageId },
+                    async (_uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken) => {
```
benmcmorran, line 44: Doesn't necessarily need to be in this PR, but before we start the experiment, we'll need to wire up `flags` to this logic. 

If you want to do that now, it would simply be a matter of adding an if statement so that we only return traits if something like `context.flags['cpptraits']` is `true`. The exact name of the flag doesn't matter, and we'll make corresponding changes in Copilot to actually check with the experimentation service. We could also have multiple flags if we want to control traits individually.

kuchungmsft: I have that implemented like this, would that work?
image

benmcmorran: Sorry, my bad! Yes, that looks perfect.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: We can control it with a string[] flag like the following if needed, would that be ok? Thanks.
image

benmcmorran: I think that will work. We can't actually set an experimentation variable to `string[]` type in the ExP UI (only boolean, number, or string) but that shouldn't matter because there will be an translation layer in copilot-client anyway where we can for example convert a `string` to a `string[]`.

kuchungmsft: Thanks Ben for the ExP UI rules. I added the feature flags, but it's not clear to me how these two flags get passed to the context of callback calls.

kuchungmsft: I updated both my PRs, thanks.

sean-mcmanus (approved):

PR #12822, opened 2024-10-08
Add packages for copilot folks

bobbrow:
Adding new packages to CFS for PR #12821

(not for 1.22)


kuchungmsft (approved):

Extension/package.json
```diff
@@ -6540,6 +6543,7 @@
         "node-stream-zip": "^1.15.0",
         "plist": "^3.1.0",
         "posix-getopt": "^1.2.1",
+        "proxyquire": "^2.1.3",
```
bobbrow: It looks like this is only needed for tests. @kuchungmsft can you confirm?  I will move it to `devDependencies`.

kuchungmsft: Yes, it's only for testing. Thanks.

kuchungmsft: sinon is also for testing only, thanks.

kuchungmsft (approved):

sean-mcmanus (approved):

sean-mcmanus:
If this is not for 1.22.9 then it probably shouldn't be checked in yet, since we haven't merged to release yet.

bobbrow:
> If this is not for 1.22.9 then it probably shouldn't be checked in yet, since we haven't merged to release yet.

I'm holding it for 1.23.

sean-mcmanus:
I've merged for release so you could check this in now.

PR #12844, opened 2024-10-17
Support User Defines and Macro References AB Experiment

kuchungmsft:
Support A/B Compiler Arguments Traits

- Depends on https://github.com/microsoft/vscode-cpptools/pull/12979 to merge first.
- copilotcppMacroReferenceFilter: regex string to filter macro reference for telemetry.
- Telemetry related to user defines and macro references.
- Added new trait compilerUserDefines to note the relevant user defines to

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -541,6 +541,19 @@ export interface ChatContextResult {
     targetArchitecture: string;
 }
 
+export interface FileContextResult {
+    compilerArguments: string[];
+}
+
+export interface ProjectContextResult {
+    language: string;
+    standardVersion: string;
+    compiler: string;
+    targetPlatform: string;
+    targetArchitecture: string;
+    fileContext: FileContextResult;
+}
+
 // Requests
 const PreInitializationRequest: RequestType<void, string, void> = new RequestType<void, string, void>('cpptools/preinitialize');
 const InitializationRequest: RequestType<CppInitializationParams, void, void> = new RequestType<CppInitializationParams, void, void>('cpptools/initialize');
@@ -561,6 +574,7 @@ const GenerateDoxygenCommentRequest: RequestType<GenerateDoxygenCommentParams, G
 const ChangeCppPropertiesRequest: RequestType<CppPropertiesParams, void, void> = new RequestType<CppPropertiesParams, void, void>('cpptools/didChangeCppProperties');
 const IncludesRequest: RequestType<GetIncludesParams, GetIncludesResult, void> = new RequestType<GetIncludesParams, GetIncludesResult, void>('cpptools/getIncludes');
 const CppContextRequest: RequestType<void, ChatContextResult, void> = new RequestType<void, ChatContextResult, void>('cpptools/getChatContext');
+const ProjectContextRequest: RequestType<void, ProjectContextResult, void> = new RequestType<void, ProjectContextResult, void>('cpptools/getProjectContext');
 
 // Notifications to the server
 const DidOpenNotification: NotificationType<DidOpenTextDocumentParams> = new NotificationType<DidOpenTextDocumentParams>('textDocument/didOpen');
@@ -792,6 +806,7 @@ export interface Client {
     addTrustedCompiler(path: string): Promise<void>;
     getIncludes(maxDepth: number, token: vscode.CancellationToken): Promise<GetIncludesResult>;
     getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult>;
+    getProjectContext(token: vscode.CancellationToken): Promise<ProjectContextResult>;
 }
 
 export function createClient(workspaceFolder?: vscode.WorkspaceFolder): Client {
@@ -2218,6 +2233,12 @@ export class DefaultClient implements Client {
         await withCancellation(this.ready, token);
         return DefaultClient.withLspCancellationHandling(
             () => this.languageClient.sendRequest(CppContextRequest, null, token), token);
+    }
+
+    public async getProjectContext(token: vscode.CancellationToken): Promise<ProjectContextResult> {
+        await withCancellation(this.ready, token);
+        return DefaultClient.withLspCancellationHandling(
+            () => this.languageClient.sendRequest(ProjectContextRequest, null, token), token);
     }
 
     /**
@@ -4123,4 +4144,5 @@ class NullClient implements Client {
     addTrustedCompiler(path: string): Promise<void> { return Promise.resolve(); }
     getIncludes(maxDepth: number, token: vscode.CancellationToken): Promise<GetIncludesResult> { return Promise.resolve({} as GetIncludesResult); }
     getChatContext(token: vscode.CancellationToken): Promise<ChatContextResult> { return Promise.resolve({} as ChatContextResult); }
+    getProjectContext(token: vscode.CancellationToken): Promise<ProjectContextResult> { return Promise.resolve({} as ProjectContextResult); }
 }
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -543,6 +543,8 @@ export interface ChatContextResult {
 
 export interface FileContextResult {
     compilerArguments: string[];
+    compilerUserDefines: string[];
+    macroReferences: string[];
 }
 
 export interface ProjectContextResult {
--- a/Extension/src/LanguageServer/copilotProviders.ts
+++ b/Extension/src/LanguageServer/copilotProviders.ts
@@ -6,8 +6,9 @@
 
 import * as vscode from 'vscode';
 import * as util from '../common';
-import { ChatContextResult, GetIncludesResult } from './client';
+import { GetIncludesResult } from './client';
 import { getActiveClient } from './extension';
+import { getProjectContext } from './lmTool';
 
 export interface CopilotTrait {
     name: string;
@@ -38,19 +39,51 @@ export async function registerRelatedFilesProvider(): Promise<void> {
 
                         const getIncludesHandler = async () => (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [];
                         const getTraitsHandler = async () => {
-                            const chatContext: ChatContextResult | undefined = await (getActiveClient().getChatContext(token) ?? undefined);
+                            const cppContext = await getProjectContext(context, token);
 
-                            if (!chatContext) {
+                            if (!cppContext) {
                                 return undefined;
                             }
 
                             let traits: CopilotTrait[] = [
-                                { name: "language", value: chatContext.language, includeInPrompt: true, promptTextOverride: `The language is ${chatContext.language}.` },
-                                { name: "compiler", value: chatContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${chatContext.compiler}.` },
-                                { name: "standardVersion", value: chatContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${chatContext.standardVersion} language standard.` },
-                                { name: "targetPlatform", value: chatContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetPlatform}.` },
-                                { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
+                                { name: "intellisense", value: 'intellisense', includeInPrompt: true, promptTextOverride: `IntelliSense is currently configured with the following compiler information. It's best effort to reflect the active configuration, and the project may have more configurations targeting different platforms.` },
+                                { name: "intellisenseBegin", value: 'Begin', includeInPrompt: true, promptTextOverride: `Begin of IntelliSense information.` }
                             ];
+                            if (cppContext.language) {
+                                traits.push({ name: "language", value: cppContext.language, includeInPrompt: true, promptTextOverride: `The language is ${cppContext.language}.` });
+                            }
+                            if (cppContext.compiler) {
+                                traits.push({ name: "compiler", value: cppContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${cppContext.compiler}.` });
+                            }
+                            if (cppContext.standardVersion) {
+                                traits.push({ name: "standardVersion", value: cppContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${cppContext.standardVersion} language standard.` });
+                            }
+                            if (cppContext.targetPlatform) {
+                                traits.push({ name: "targetPlatform", value: cppContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${cppContext.targetPlatform}.` });
+                            }
+                            if (cppContext.targetArchitecture) {
+                                traits.push({ name: "targetArchitecture", value: cppContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${cppContext.targetArchitecture}.` });
+                            }
+                            let directAsks: string = '';
+                            if (cppContext.compilerArguments.length > 0) {
+                                // Example: JSON.stringify({'-fno-rtti': "Do not generate code using RTTI keywords."})
+                                const directAskMap: { [key: string]: string } = JSON.parse(context.flags.copilotcppCompilerArgumentDirectAskMap as string ?? '{}');
+                                const updatedArguments = cppContext.compilerArguments.filter(arg => {
+                                    if (directAskMap[arg]) {
+                                        directAsks += `${directAskMap[arg]} `;
+                                        return false;
+                                    }
+                                    return true;
+                                });
+
+                                const compilerArgumentsValue = updatedArguments.join(", ");
+                                traits.push({ name: "compilerArguments", value: compilerArgumentsValue, includeInPrompt: true, promptTextOverride: `The compiler arguments include: ${compilerArgumentsValue}.` });
+                            }
+                            if (directAsks) {
+                                traits.push({ name: "directAsks", value: directAsks, includeInPrompt: true, promptTextOverride: directAsks });
+                            }
+
+                            traits.push({ name: "intellisenseEnd", value: 'End', includeInPrompt: true, promptTextOverride: `End of IntelliSense information.` });
 
                             const excludeTraits = context.flags.copilotcppExcludeTraits as string[] ?? [];
                             traits = traits.filter(trait => !excludeTraits.includes(trait.name));
--- a/Extension/src/LanguageServer/copilotProviders.ts
+++ b/Extension/src/LanguageServer/copilotProviders.ts
@@ -79,6 +79,10 @@ export async function registerRelatedFilesProvider(): Promise<void> {
                                 const compilerArgumentsValue = updatedArguments.join(", ");
                                 traits.push({ name: "compilerArguments", value: compilerArgumentsValue, includeInPrompt: true, promptTextOverride: `The compiler arguments include: ${compilerArgumentsValue}.` });
                             }
+                            if (cppContext.compilerUserDefinesRelevant.length > 0) {
+                                const compilerUserDefinesValue = cppContext.compilerUserDefinesRelevant.join(", ");
+                                traits.push({ name: "compilerUserDefines", value: compilerUserDefinesValue, includeInPrompt: true, promptTextOverride: `These compiler command line user defines may be relevent: ${compilerUserDefinesValue}.` });
+                            }
                             if (directAsks) {
                                 traits.push({ name: "directAsks", value: directAsks, includeInPrompt: true, promptTextOverride: directAsks });
                             }
--- a/Extension/src/LanguageServer/lmTool.ts
+++ b/Extension/src/LanguageServer/lmTool.ts
@@ -66,6 +66,7 @@ export interface ProjectContext {
     targetPlatform: string;
     targetArchitecture: string;
     compilerArguments: string[];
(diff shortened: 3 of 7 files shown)


Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -59,6 +60,21 @@ export async function registerRelatedFilesProvider(): Promise<void> {
                                 { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
                             ];
 
+                            if (client instanceof DefaultClient) {
+                                const defaultClient: DefaultClient = client as DefaultClient;
+                                const configuration = defaultClient.configuration.CurrentConfiguration;
+                                if (configuration) {
+                                    const defines = configuration.defines?.join(', ') ?? undefined;
+                                    if (defines) {
+                                        traits.push({ name: "compilerDefines", value: defines, includeInPrompt: true, promptTextOverride: `This compiler defines contain: ${defines}.` });
```
sean-mcmanus: Should "contain" be "are"? Same for compiler arguments below.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/copilotProviders.ts
bobbrow: Whatever you decide, this string probably needs to be localized, right?  Or is this a prompt for the AI (e.g. not a user-facing string)?

sean-mcmanus: I assumed a prompt for the AI? In which case, it seems like it would be safer/better for the defines prompt to specifically mention that it's the "user-defined" defines so it doesn't think it includes the compiler defined ones.

sean-mcmanus: `configuration.compilerArgs` and ` configuration.defines` only has configuration info for c_cpp_properties.json, which excludes configurations from configuration providers and compile_commands.json so it's not what you want to check. Colen said there's some function you can call on cpptools to get the info I think.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -59,6 +60,21 @@ export async function registerRelatedFilesProvider(): Promise<void> {
                                 { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
                             ];
 
+                            if (client instanceof DefaultClient) {
+                                const defaultClient: DefaultClient = client as DefaultClient;
+                                const configuration = defaultClient.configuration.CurrentConfiguration;
```
Colengms: This is the 'base configuration'.  This may provide some information from some users' configurations, but a lot of users (perhaps the majority of users with non-trivial projects) do not configure the extension this way.  When configuring using a custom configuration provider (CMake Tools, Makefile tools), the 'base configuration' is effectively unused (and would be incorrect to refer to), as any file managed by those providers will get their configuration from them instead. (The base configuration is used as a fallback for files they don't handle).  Use of `compile_commands.json` to configure the extension is similar, as any file mentioned there will derived its configuration from that file.

I'd recommend extending the `cpptools_getChatContext` request to include any additional fields that might be useful, such as defines and compilerArgs.  That call will use the fully resolved configuration, so users of config providers and compile_commands.json would be supported.  However, that would be a fully resolved configuration, so could have tons of system defines. We would need to add some way for that to exclude defines retrieved from the compiler, as you probably wouldn't want those bloating the AI context.

Colengms (changes requested):

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: @kuchungmsft I think you may be able to look at the compilation_args::original_defines data for the user-defines that exclude system defines.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: @sean-mcmanus  I used 'contain` instead of `are` because the ones we provided here is subset. I have updated the implementation to get information from cpptools, please take another look, thanks.
@bobbrow this is a prompt for the AI, not a user-facing string, no need to localize. Thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Thanks @Colengms  and @sean-mcmanus, I've updated the implementation, please take a look.

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: @kuchungmsft Why is it a subset? Which defines/args are missing?

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,206 @@
...
+                compiler: 'msvc',
+                targetPlatform: 'windows',
+                targetArchitecture: 'x64'
+            }
+        });
+
+        const result = await new moduleUnderTest.CppConfigurationLanguageModelTool().invoke(mockLanguageModelToolInvocationOptions, new vscode.CancellationTokenSource().token);
+
+        ok(result['text/plain'], 'result should contain a text/plain entry');
+        ok(result['text/plain'] === 'The user is working on a c++ project. The project uses language version C++20, compiles using the MSVC compiler, targets the Windows platform, and targets the x64 architecture.');
```
sean-mcmanus: I think `c++` should be `C++`

sean-mcmanus: Oh, but this is just a test? Can the source that uses `c++` be changed?

Extension/src/LanguageServer/client.ts
```diff
@@ -539,6 +539,8 @@ export interface ChatContextResult {
     compiler: string;
     targetPlatform: string;
     targetArchitecture: string;
+    compilerArgs?: string[];
```
Colengms: I'd just like to raise a flag here about including `compilerArgs` in the context. I don't think this is going to be helpful. In some cases, this will be full of 100's of arbitrary GCC/clang warning categories being selectively enabled and disabled.  My opinion is that this would be "distracting" in an AI context, giving it many, many things to focus on (and misinterpret as relevant) that would have little to do with the request.  I would relatively strongly advise against it.  Please pass along my feedback to whoever thinks this is a good idea.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
sean-mcmanus: Copilot says:
image
so it seems better to use prompts it thinks are correct.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -52,6 +53,16 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
                             ];
 
+                            const compilerArgs = chatContext.compilerArgs?.join(' ') ?? undefined;
+                            if (compilerArgs) {
+                                traits.push({ name: "compilerArgs", value: compilerArgs, includeInPrompt: true, promptTextOverride: `The compiler command line arguments contain: ${compilerArgs}.` });
+                            }
+
+                            const compilerUserDefines = chatContext.compilerUserDefines?.join(', ') ?? undefined;
+                            if (compilerUserDefines) {
+                                traits.push({ name: "compilerUserDefines", value: compilerUserDefines, includeInPrompt: true, promptTextOverride: `The compiler command line user defines contain: ${compilerUserDefines}.` });
```
Colengms: Not added by you, but I'm just noticing this now.

"This project compiles using"
"This build targets"

We should avoid giving the AI false or misleading information. Those blurbs could easily be misinterpreted. We shouldn't tell the AI that "This project compiles...", as that could potentially incur some assumption that the code is valid as it "does compile", when it may not.

The C/C++ Extension does not include a build system. If there is an assumption that the user is "building" something, that "build" information is available, accurate and obviously applicable, as it is in VS, those are not safe assumption to make w/ the C/C++ Extension. It would probably be best not to mention "build" in the context, so as not to suggest to the AI that the user is building the code or may need assistance related to the actual build process. 

`targetPlatform` and `targetArchitecture` refer to the platform and architecture IntelliSense in configured for, not necessarily a build system. I'd suggest using something technically correct without overreaching, such as "The C++ language standard is currently set to:", and "This code may target:", or something else not potentially incorrect.

Colengms (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I think we should only include flags that we know are likely to affect the AI generated code or responses and not flags that just impact the compiler details (optimizations, warnings, etc.). And maybe each flag could be A/B tested separately.

sean-mcmanus: Also are the "defines" going to be A/B tested separately so we can determine if that improves stuff or not?

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: It seems like maybe it should be a subset. Is there a max number of defines sent that I didn't notice?

Extension/src/LanguageServer/client.ts
sean-mcmanus: And see my comment on the cpptools side.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -67,20 +67,26 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
...
+                "compiler": chatContext.compiler,
+                "standardVersion": chatContext.standardVersion,
+                "targetPlatform": chatContext.targetPlatform,
+                "targetArchitecture": chatContext.targetArchitecture
+            };
+            if (chatContext.compilerArgs) {
+                telemetryProperties["compilerArgs"] = chatContext.compilerArgs.join(' ');
+            }
+            if (chatContext.compilerUserDefines) {
+                telemetryProperties["compilerUserDefines"] = chatContext.compilerUserDefines.join(', ');
```
benmcmorran: `compilerUserDefines` and `compilerArgs` could both contain PII. To avoid privacy issues, can we only emit telemetry for values that come from an allow list of well-known defines or arguments?

benmcmorran: Discussed offline with @bobbrow. Let's drop this `compilerUserDefines` and `compilerArgs` telemetry for now. We can add it back later in a more targeted way if there are specific flags we need data on.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Thanks. Will start A/B with the statements that are already checked in, and future A/Bs to find better statements, if any.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Updated telemetry to only log the number of arguments and number of arguments being sent to copilot. Thanks.

Extension/src/LanguageServer/client.ts
kuchungmsft: Thanks. I changed the PR to better support compiler arguments A/B experiments, so we'll know better which arguments to pass to Copilot. Which arguments to pass is controlled by experimental flags.

For user defines, I think it makes more sense to just pass the ones that are referenced in current document instead of all, need to check how to do that on cpptools side.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
kuchungmsft: Thanks, fixed.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Since we'll be sending selective arguments once find out via A/Bs, will use "contain" instead of "are". Thanks.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +47,85 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+    language: string;
+    standardVersion: string;
+    compiler: string;
+    targetPlatform: string;
+    targetArchitecture: string;
+    compilerArguments: string;
+}
+
+// To be updated after A/B experiments.
+const defaultCompilerArgumentFilters: { [id: string]: RegExp | undefined } = {
```
bobbrow, line 73: nit: `undefined` means the key doesn't actually exist in the object.  This might be defined like this instead:
```ts
const defaultCompilerArgumentFilters: { [id: string]: RegExp } = {};
```

bobbrow: (though from what I can tell, this object never adds defaults from any of the compilers and will remain empty)

kuchungmsft: Updated. Will update defaults after A/B experiemnts, we'll know better on what arguments we should provide to help completions.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +47,84 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+            "targetPlatform": chatContext.targetPlatform,
+            "targetArchitecture": chatContext.targetArchitecture,
+            "compilerArgumentCount": chatContext.compilerArgs.length.toString(),
+            "filteredCompilerArgumentCount": filteredcompilerArguments.length.toString()
+        });
+
+    return {
+        language: chatContext.language,
+        standardVersion: chatContext.standardVersion,
+        compiler: chatContext.compiler,
```
sean-mcmanus: I'm worried that this will tend to generate code that only compiles for the 1 compiler/platform/architecture, and the A/B testing may be insufficient to detect that code is generated isn't compatible, because users won't realize the code doesn't compile with other targets until later on when there's a compile failure since IntelliSense won't give errors since it's only configured for the 1 target. The only reason IntelliSense only has 1 target configured is because supporting IntelliSense for multiple targets would be too expensive for performance.

According to Copilot: 
image

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -72,12 +153,7 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
...
 
-            for (const key in knownValues) {
-                const knownKey = key as keyof ChatContextResult;
-                if (knownValues[knownKey] && chatContext[knownKey]) {
-                    chatContext[knownKey] = knownValues[knownKey][chatContext[knownKey]] || chatContext[knownKey];
-                }
-            }
+            formatChatContext(chatContext);
 
             return `The user is working on a ${chatContext.language} project. The project uses language version ${chatContext.standardVersion}, compiles using the ${chatContext.compiler} compiler, targets the ${chatContext.targetPlatform} platform, and targets the ${chatContext.targetArchitecture} architecture.`;
```
sean-mcmanus: ...as mentioned in my other comment, sending `The project uses language version ${chatContext.standardVersion}, compiles using the ${chatContext.compiler} compiler, targets the ${chatContext.targetPlatform} platform, and targets the ${chatContext.targetArchitecture} architecture.`` could potentially mislead Copilot into believing the project is not cross-plat or intended to compile for multiple compilers, platforms, architectures.

sean-mcmanus: image

sean-mcmanus: image

sean-mcmanus: It seems like we should only tell Copilot what a project is intended to target if we actually know that info. The IntelliSense context just reports what IntelliSense is currently configured for in regards to real-time language service features. Users might have multiple IntelliSense configurations for multiple targets (with IntelliSense being configured for only 1 at a time for performance reasons).

kuchungmsft: Thanks Sean. This statement is only provided to support A/B experiments of Copilot Chat experience (not completions). When users type '#cpp' in the chat window, we'll send this project context to Copilot to help generating better response, if any. This is out of scope as my PR is focusing on supporting A/B to check if sending any of compiler arguments would help Copilot generating better completions. Ben already responded your email, and created a bug to track the work to make Copilot to be more aware of cross-plat context.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Here's a specific case where Copilot suggested code that only compiles with gcc/mingw, but when I switch my IntelliSense config to msvc/cl.exe it fails.

Copilot
image

gcc IntelliSense config:
image

msvc IntelliSense config:
image

kuchungmsft: @sean-mcmanus can you show me the code snippet before/after the Copilot suggestion?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +47,84 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+    targetArchitecture: string;
+    compilerArguments: string;
+}
+
+const matchNothingRegex = /(?!.*)/;
+
+// To be updated after A/B experiments, math nothing for now.
+const defaultCompilerArgumentFilters: { [id: string]: RegExp } = {
+    MSVC: matchNothingRegex,
+    Clang: matchNothingRegex,
```
Colengms: One thing to consider is that on macOS the default compiler is clang, and `/usr/bin/gcc` / `/usr/bin/g++` are actually symbolic links to clang, not gcc.  Performing a regex match on the compiler path may not be a reliable way to determine what compiler is being referred to. On the native side, we 'query' non-CL compilers (gcc or clang), and will use system defines to determine whether the compiler is clang or gcc.  (There could also be other details that could be determined based on detected system defines, that may potentially be worth surfacing to AI in the future).  I'd suggest adding plumbing on the native side to pass back the 'compiler type' (with the rest of the TU context) instead of relying on pattern matching against compiler paths.

kuchungmsft: These regex are for matching the compiler arguments to be sent to the Copilot, these are not for detecting the compiler types.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Sean, I was going to link these cross-plat related comments to the bug that Ben created this morning, it looks like you already added the comments. I'll go ahead and resolve this thread as it's out of scope for this PR. Thanks.

Extension/src/LanguageServer/lmTool.ts
Colengms: I see.  FYI, the source of "compiler" here appears to be coming from the `IntelliSenseMode` on the native side, only.  It's not quite correct to directly associate that with compiler arguments.  For instance, clang has a compiler, `clang-cl.exe`, which is clang, but accepts MSVC command line arguments (as well as some clang arguments not accepted by MSVC).  The user will get a (i.e.) `windows-clang-x64` intelliSense-mode, despite all of their arguments being CL arguments.

kuchungmsft: Thanks, Colen, I didn't know that, I think we can configure the regex to match the arguments from both scenarios for Clang so that we can cover `windows-clang-XXX` as well. So, I think we are good.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,102 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+
+        telemetryProperties["language"] = projectContext.result.language;
+        telemetryProperties["compiler"] = projectContext.result.compiler;
+        telemetryProperties["standardVersion"] = projectContext.result.standardVersion;
+        telemetryProperties["targetPlatform"] = projectContext.result.targetPlatform;
+        telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        telemetryProperties["compilerArgumentCount"] = projectContext.result.fileContext.compilerArgs.length.toString();
+        telemetryProperties["filteredCompilerArgumentCount"] = filteredcompilerArguments.length.toString();
+        telemetryProperties["compilerUserDefinesRelavantCount"] = projectContext.result.fileContext.compilerUserDefinesRelavant.length.toString();
+        telemetryProperties["targetArcompilerOriginalUserDefineCounthitecture"] = projectContext.result.fileContext.compilerOriginalUserDefineCount.toString();
```
kuchungmsft: Typo: compilerOriginalUserDefineCount

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,102 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        telemetryProperties["targetArchmacroReferenceCountitecture"] = projectContext.result.fileContext.macroReferenceCount.toString();
```
kuchungmsft: Typo: macroReferenceCount

sean-mcmanus:
@kuchungmsft FYI, you should have permissions now to create a branch in our repo instead of from a fork (you may not have had that at the time of this PR creation though).

Extension/src/LanguageServer/client.ts
```diff
@@ -541,6 +541,22 @@ export interface ChatContextResult {
     targetArchitecture: string;
 }
 
+export interface FileContextResult {
+    compilerArgs: string[];
+    compilerUserDefinesRelavant: string[];
```
sean-mcmanus: Relavant->Relevant

kuchungmsft: Fixed, thanks.

PR #12848, opened 2024-10-17
Always use related files API for Copilot completions

benmcmorran:
We've completed an A/B experiment and shown no observable difference (as expected) between the legacy `getIncludes` command and the new related files API, so we're removing the experimentation variable, the `getIncludes` command, and associated tests.

cc @microsoft/cpp-copilot

### DIFF
--- a/Extension/src/LanguageServer/copilotProviders.ts
+++ b/Extension/src/LanguageServer/copilotProviders.ts
@@ -6,11 +6,8 @@
 
 import * as vscode from 'vscode';
 import * as util from '../common';
-import * as telemetry from '../telemetry';
 import { ChatContextResult, GetIncludesResult } from './client';
 import { getActiveClient } from './extension';
-
-let isRelatedFilesApiEnabled: boolean | undefined;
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -79,12 +72,6 @@ export async function registerRelatedFilesProvider(): Promise<void> {
     }
 }
 
-export async function registerRelatedFilesCommands(commandDisposables: vscode.Disposable[], enabled: boolean): Promise<void> {
-    if (await getIsRelatedFilesApiEnabled()) {
```
benmcmorran, line 83: I realized this logic is actually backwards (we only want to register `getIncludes` when the related files API is _disabled_). However, it shouldn't affect the experiment results, because the logic was correct in v1.22.9, which most users would have been running during the experiment period.

kuchungmsft (approved):

sean-mcmanus (approved):

PR #12881, opened 2024-10-24
Values set in settings are not immediately available for use after changing them

### LINKED ISSUE #12705: Bug: Command `C/C++: Select IntelliSense Configuration... -> Use g**`
### Environment

- OS and Version: Windows 11 23H2
- VS Code Version: 1.92.2
- C/C++ Extension Version: v1.21.6
- If using SSH remote, specify OS of remote machine: wsl (Ubuntu 20.04)

Bug Summary:
If you select a compiler via `C/C++: Select IntelliSense Configuration... -> Use g**` one can notice that the updated `compilerPath` is not updated until the second time you select the same command again. It then changes `.vscode/c_cpp_properties.json` to the compiler you selected in the n-1 try.

https://github.com/user-attachments/assets/b91eeb61-edd3-4ef4-a0b9-d5738822dfda

Steps to reproduce:
1. Keep the `.vscode/c_cpp_properties.json` file open while changing the compiler via the command ``C/C++: Select IntelliSense Configuration....`
2. Notice how it does not select what you chose.

Expected behavior:
It should format the file and add a new line at EOF.

### Configuration and Logs

```shell
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "defines": [],
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "c99",
            "cppStandard": "gnu++14",
            "intelliSenseMode": "linux-gcc-x64",
            "configurationProvider": "ms-vscode.cmake-tools"
        },
        {
            "name": "CMake",
            "compileCommands": "${config:cmake.buildDirectory}/compile_commands.json",
            "configurationProvider": "ms-vscode.cmake-tools"
        }
    ],
    "version": 4
}
```

```
-------- Diagnostics - 9/4/2024, 1:10:29 PM
Version: 1.21.6
Current Configuration:
{
    "name": "Linux",
    "includePath": [
        "/home/tobias/code/c-test/**"
    ],
    "defines": [],
    "compilerPath": "/usr/bin/gcc",
    "cStandard": "c99",
    "cppStandard": "gnu++14",
    "intelliSenseMode": "linux-gcc-x64",
    "configurationProvider": "ms-vscode.cmake-tools",
    "compilerPathIsExplicit": true,
    "cStandardIsExplicit": true,
    "cppStandardIsExplicit": true,
    "intelliSenseModeIsExplicit": true,
    "compilerPathInCppPropertiesJson": "/usr/bin/gcc",
    "configurationProviderInCppPropertiesJson": "ms-vscode.cmake-tools",
    "mergeConfigurations": false,
    "browse": {
        "path": [
            "/home/tobias/code/c-test/**",
            "${workspaceFolder}"
        ],
    }
}
Custom browse configuration: 
{
    "browsePath": [
        "/home/tobias/code/c-test/src"
    ],
    "compilerPath": "/usr/bin/gcc",
    "compilerArgs": [],
    "compilerFragments": [
        "-g"
    ]
}
cpptools version (native): 1.21.6.0
Translation Unit Mappings:
[ /home/tobias/code/c-test/src/main.c - source TU]:
Translation Unit Configurations:
[ /home/tobias/code/c-test/src/main.c ]:
    Process ID: 30660
    Memory Usage: 15 MB
    Compiler Path: /usr/bin/gcc
    Includes:
    System Includes:
        /usr/lib/gcc/x86_64-linux-gnu/9/include
        /usr/local/include
        /usr/include/x86_64-linux-gnu
        /usr/include
    Standard Version: c17
    IntelliSense Mode: linux-gcc-x64
    Other Flags:
        --gcc
        --gnu_version=90400
Total Memory Usage: 15 MB

------- Workspace parsing diagnostics -------
Number of files discovered (not excluded): 4908
```

### Other Extensions

_No response_

### Additional context

_No response_

bobbrow:
Fixes: #12705

We can't use the value of a setting we changed on our side until after an `await` is processed, or we've finished handling the current event.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1046,7 +1046,7 @@ export class DefaultClient implements Client {
             if (index === paths.length - 1) {
                 action = "disable";
                 settings.defaultCompilerPath = "";
-                await this.configuration.updateCompilerPathIfSet(settings.defaultCompilerPath);
+                await this.configuration.updateCompilerPathIfSet("");
                 configurationSelected = true;
                 await this.showPrompt(sender);
                 return ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompilerPath, "disablePrompt");
@@ -1066,7 +1066,7 @@ export class DefaultClient implements Client {
                 configurationSelected = true;
                 action = "compiler browsed";
                 settings.defaultCompilerPath = result[0].fsPath;
-                await this.configuration.updateCompilerPathIfSet(settings.defaultCompilerPath);
+                await this.configuration.updateCompilerPathIfSet(result[0].fsPath);
                 void SessionState.trustedCompilerFound.set(true);
             } else {
                 configurationSelected = true;
@@ -1084,8 +1084,9 @@ export class DefaultClient implements Client {
                     return ui.ShowConfigureIntelliSenseButton(false, this, ConfigurationType.CompileCommands, showButtonSender);
                 } else {
                     action = "select compiler";
-                    settings.defaultCompilerPath = util.isCl(paths[index]) ? "cl.exe" : paths[index];
-                    await this.configuration.updateCompilerPathIfSet(settings.defaultCompilerPath);
+                    const newCompiler: string = util.isCl(paths[index]) ? "cl.exe" : paths[index];
+                    settings.defaultCompilerPath = newCompiler;
+                    await this.configuration.updateCompilerPathIfSet(newCompiler);
                     void SessionState.trustedCompilerFound.set(true);
                 }
             }


sean-mcmanus (approved):

thernstig:
@bobbrow although this most likely fixes the problem, does it also handle this?

image

The first time the IntelliSense configuration is changed in a few VS Code window, it will update the workspace settings.json as seen. This is highly unexpected I think.

The `C/C++: Select IntelliSense Configuration...` command seems to be for c_cpp_properties.json. There should be a separate command, or maybe more intuitive, a separate action after having selected `C/C++: Select IntelliSense Configuration...` to set the default IntelliSense config.

I.e. to add a new entry here that says `Set default compiler path` or something?

image

bobbrow:
@thernstig it is currently intentional that we set the setting in settings.json.  I no longer remember the discussion around why we chose this way (I want to say that the intent was to make it the default compiler but allow for overriding in a configuration in c_cpp_properties.json). We also set it in c_cpp_properties.json if there is a different value set for it because in this case setting the folder setting would have had no effect.  An alternate approach might be to set it in c_cpp_properties.json if that file exists, and in settings.json if it doesn't.

sean-mcmanus:
@thernstig If you do create an issue, if you could explain why the current design behavior is insufficient that could help.

thernstig:
@sean-mcmanus Will do! I think it is quite easy to reason around, if I opt to do this. But it is not a huge issue so not too tempted to write one since there are more important things you have at hand I am sure :)

PR #12898, opened 2024-10-29
Enable #cpp for all users

benmcmorran:
VS Code 1.95 has now shipped, and it includes the stabilization of the lmTools API. https://code.visualstudio.com/updates/v1_95#_tools-for-language-models

cc @microsoft/cpp-copilot

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -252,9 +252,14 @@ export async function activate(): Promise<void> {
         activeDocument = activeEditor.document;
     }
 
-    if (util.extensionContext && new CppSettings().experimentalFeatures) {
-        const tool = vscode.lm.registerTool('cpptools-lmtool-configuration', new CppConfigurationLanguageModelTool());
-        disposables.push(tool);
+    if (util.extensionContext) {
+        // lmTools wasn't stabilized until 1.95, but (as of October 2024)
+        // cpptools can be installed on versions of VS Code as old as 1.67.
+        const version = util.getVsCodeVersion();
+        if (version[0] > 1 || (version[0] === 1 && version[1] >= 95)) {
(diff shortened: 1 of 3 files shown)


kuchungmsft (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -252,9 +252,14 @@ export async function activate(): Promise<void> {
         activeDocument = activeEditor.document;
     }
 
-    if (util.extensionContext && new CppSettings().experimentalFeatures) {
-        const tool = vscode.lm.registerTool('cpptools-lmtool-configuration', new CppConfigurationLanguageModelTool());
-        disposables.push(tool);
+    if (util.extensionContext) {
+        // lmTools wasn't stabilized until 1.95, but (as of October 2024)
+        // cpptools can be installed on versions of VS Code as old as 1.67.
```
sean-mcmanus: Instead of hardcoding the 1.67 in the comment can you link to https://github.com/microsoft/vscode-cpptools/blob/main/Extension/package.json#L14 ?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Or mention engines.vscode in package.json.

sean-mcmanus: Otherwise, the engines.vscode could be increased and the comment might be outdated.

benmcmorran: Thanks, I updated the comment to remove the hardcoded version and link to package.json on GitHub.

sean-mcmanus (approved):

kuchungmsft (approved):

Issue #12905, opened 2024-10-29
LSP requests processed by `protocolFilter.ts` stall all other requests


Colengms:
In `protocolFilter.ts`, we are intentionally sequencing the message handlers, to ensure async calls within them can complete before the next message is processed.  That ensure our use of async doesn't cause messages to be delivered out of order (i.e. something file-specific delivered before didOpen is processed for that file). However, that appears to also be awaiting the results of request operations, causing the protocol filter to effectively stall all other requests (including those done via `languageClient.sendRequest`) whenever there is an outstanding request for one of the LSP requests processed there:

```
provideCompletionItem
resolveCompletionItem
provideSignatureHelp
provideDefinition
provideReferences
provideDocumentHighlights
provideDeclaration
```

heartacker:
good

PR #12906, opened 2024-10-29
Fix issue with requests in protocolFilter.ts causing stalls

Colengms:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12905

I managed basically remove the need to do any awaiting in protocolFilter.ts, making it unnecessary to ensure the messages are otherwise kept in sequence.  It looks like not specifying a filter at all does not cause VS Code to await responses to requests before processing the next message.

### DIFF
--- a/Extension/src/LanguageServer/protocolFilter.ts
+++ b/Extension/src/LanguageServer/protocolFilter.ts
@@ -8,6 +8,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { Middleware } from 'vscode-languageclient';
 import * as util from '../common';
+import { logAndReturn } from '../Utility/Async/returns';
 import { Client } from './client';
 import { clients } from './extension';
 import { shouldChangeFromCToCpp } from './utils';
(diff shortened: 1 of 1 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -41,62 +36,41 @@ export function createProtocolFilter(): Middleware {
                         const mappingString: string = baseFileName + "@" + document.fileName;
                         client.addFileAssociations(mappingString, "cpp");
                         client.sendDidChangeSettings();
-                        document = await vscode.languages.setTextDocumentLanguage(document, "cpp");
+                        // This will cause the file to be closed and reopened.
+                        void vscode.languages.setTextDocumentLanguage(document, "cpp");
+                        return;
```
bobbrow, line 41: Why are we returning here now? Doesn't the rest of the code need to run?

bobbrow: I missed the comment. If changing the document language closes and reopens the file, then this should be ok.  Has it always been that way?

Colengms: I believe so.

PR #12907, opened 2024-10-30
[Copilot][#cpp]Ensure only well known values are passed in to the LLM

lukka:


benmcmorran (approved):

kuchungmsft (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -75,11 +75,29 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
...
-                    chatContext[knownKey] = knownValues[knownKey][chatContext[knownKey]] || chatContext[knownKey];
+                    // Clear the value if it's not in the known values.
+                    chatContext[knownKey] = knownValues[knownKey][chatContext[knownKey]] || "";
                 }
             }
 
-            return `The user is working on a ${chatContext.language} project. The project uses language version ${chatContext.standardVersion}, compiles using the ${chatContext.compiler} compiler, targets the ${chatContext.targetPlatform} platform, and targets the ${chatContext.targetArchitecture} architecture.`;
+            let contextString = "";
+            if (chatContext.language) {
+                contextString += `The user is working on a ${chatContext.language} project. `;
```
sean-mcmanus, line 85: @lukka As mentioned elsewhere, this context is misleading -- the "language" only applies to the currently open file, which could contain multiple files of different languages. The "standardVersion" also only applies to the current file/configuration. Also, the project might be cross-plat and be intended to compile with msvc/gcc/clang/etc. but the context only reports the 1 currently configured for IntelliSense, which can incorrectly bias Copilot into generating code that won't compile with other compilers (which I've hit 2 cases of so far). Similarly, the "platform" may currently be Windows, but the project could be cross-plat and need to run on Linux/Mac too. And the "architecture" may currently be x64, but with a different configuration it could target arm64 or arm.

And users who use Copilot to generate the code that isn't compatible with the other compilers/platforms/architectures won't know until "later" when they hit a failure on those other platforms (IntelliSense is likely to not squiggle), so the reported telemetry would incorrect say it's a "good completion result" when it was not and the user had to change their code later to fix it.

I don't think we should be using the current IntelliSense configuration for this "context". I don't think we should mislead Copilot with information that we don't know to actually be accurate. Maybe we could use other info in their project or the CMake Tools extension, etc. to determine this info accurately.

Colengms: The add to this, currently the 'compiler' value appears to be coming from the IntelliSenseMode.  There is always an IntelliSense mode, but there is not always a compiler configured (for IntelliSense, irrespective of actually building code).  Users are not always building code. Unlike VS, since the extension doesn't actually have a build system, we should be careful not to imply to AI that a user is using a particular compiler to build or necessarily "building" for any particular platform or architecture, if they may not be.

"compiles using the" may imply that the code "does (successfully) compile", when it may not.

I'd suggest using terminology that is correct without overreaching. i.e. "The user has configured IntellISense to interpret this file as...", or "The user is likely targeting...", etc..

sean-mcmanus: It's also possible the user has incorrectly configured IntelliSense and is getting squiggles (i.e. the context might not actually be correct). They could be using a project that only compiles on Windows, but they might not have msvc/cl.exe installed and it could be incorrectly choosing mingw/gcc instead. If they try to ask Copilot for help configuring, and we tell Copilot it's a gcc project then the user might have harder time realizing the project is actually an msvc project.

Another example, on Windows, I had a file open that only compiled on Linux (it was in an inactive region on Windows), but Copilot was giving me suggestion on re-writing the file for Windows/cl.exe when that's not what I wanted.

Colengms: >The user is working on a ${chatContext.language} project

This implies there is a project involved (which there may not be), and that the entirety of it is using that language standard.

It would be possible to determine if a configuration comes from a configuration provider or compile_commands.json, and, if so, use verbiage that asserts assumptions about 'building' or 'compiling' (that specific file).  Whereas, without that information, it may be best to limit the context to information that is technically correct without potentially overreaching (i.e.) "This file is currently configured to be interpreted as ..."

lukka: @sean-mcmanus @Colengms thank you, I collected your feedback in the following ADO work item 2294002. 

The current PR is aimed to allowlist the values that are passed to the LLM.

sean-mcmanus:
@lukka You have permissions to "Squash and merge".

sean-mcmanus (approved):

PR #12910, opened 2024-10-30
Add telemetry for #cpp failures

benmcmorran:
Update the existing telemetry to add an "error" property if an exception occurred.

cc @microsoft/cpp-copilot

### DIFF
--- a/Extension/src/LanguageServer/lmTool.ts
+++ b/Extension/src/LanguageServer/lmTool.ts
@@ -51,6 +51,7 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
     }
 
     private async getContext(token: vscode.CancellationToken): Promise<string> {
+        const telemetryProperties: Record<string, string> = {};
         try {
             const currentDoc = vscode.window.activeTextEditor?.document;
             if (!currentDoc || (!util.isCpp(currentDoc) && !util.isHeaderFile(currentDoc.uri))) {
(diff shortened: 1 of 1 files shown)


lukka (approved):

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -83,25 +74,33 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
...
                 contextString += `The project compiles using the ${chatContext.compiler} compiler. `;
+                telemetryProperties["compiler"] = chatContext.compiler;
             }
             if (chatContext.targetPlatform) {
                 contextString += `The project targets the ${chatContext.targetPlatform} platform. `;
+                telemetryProperties["targetPlatform"] = chatContext.targetPlatform;
             }
             if (chatContext.targetArchitecture) {
                 contextString += `The project targets the ${chatContext.targetArchitecture} architecture. `;
+                telemetryProperties["targetArchitecture"] = chatContext.targetArchitecture;
```
lukka, line 93: note that this is a "breaking"  change for the values reported in telemetry (differences in casing but more importantly also in syntax, e.g. `cpp` becomes now `C++`)

kuchungmsft (approved):

Extension/src/LanguageServer/lmTool.ts
benmcmorran: Acknowledged. I'm comfortable saying that's an acceptable break given that we're not currently querying this telemetry, and it will bring the telemetry closer to what the user actually experiences. It will also give us a better signal if data is ever missing.

bobbrow: casing of event names and property names is not respected in the telemetry system. They are stored as "lower-case." Only the casing of values is preserved. The casing conventions we enforce in our code is for readability.

sean-mcmanus (approved):

Issue #12889, opened 2024-10-26
ChangeCompileCommandsNotification is sent/processed even if a configuration provider is used instead of compile commands


sean-mcmanus:
It seems like it should only be sent/processed if compile commands is actually being used for configuration.

sean-mcmanus:
If compileCommands is set, then it could still be used as a fallback if the configuration provider doesn't have a configuration for a file, but if compileCommands is not set then there's no need to send an update for one.

Colengms:
I don't think these notifications cause a perf issue.  They simply result in a dirty flag being set on the native side, causing the `compile_commands.json` to be re-parsed the next time its contents are required.

It would be nice to avoid sending redundant notifications, but I think that might need to be part of a more wholistic change, to ensure changes are properly synchronized with TU creation.

We have some work (currently on the back burner) to address https://github.com/microsoft/vscode-cpptools/issues/6931 , which may result in the current `compile_commands.json` handling being replaced (processed entirely on the TypeSide side).

PR #12911, opened 2024-10-31
Merge multiple compile_commands.json files

### LINKED ISSUE #7029: Support multiple compile commands files per project
**Type: LanguageService**

This is a feature request.

**Describe the bug**
- OS and Version: Ubuntu 20.04
- VS Code Version: 1.53.2
- C/C++ Extension Version: v1.2.1
- Other extensions you installed (and if the issue persists after disabling them): ROS (ms-iot.vscode-ros), disabling makes no difference.
- Does this issue involve using SSH remote to run the extension on a remote machine?: no

Some ROS build systems (catkin_tools, catkin_make_isolated) run CMake separately in different repositories. Thus, passing `CMAKE_EXPORT_COMPILE_COMMANDS=ON` generates n files for n projects. Merging the files manually is cumbersome, so the ideal solution would be the possibility to specify multiple compile commands files and/or support wildcards in the config.

**Expected behavior**

This configuration to be valid and all compilation databases merged:
```json
{
    "configurations": [
        {
            "compileCommands": [
              "${workspaceFolder}/build/foo_pkg/compile_commands.json",
              "${workspaceFolder}/build/bare_pkg/compile_commands.json",
              "${workspaceFolder}/build/**/compile_commands.json",
            ]
        }
    ],
}
```

yiftahw:
### Abstract
- Merge multiple `compile_commands.json` on the fly whenever they are added/changed based on a configuration saved in `.vscode/settings.json`
- closes #7029
- compilation database merging logic based on [VSCode CMake Tools PR #2029](https://github.com/microsoft/vscode-cmake-tools/pull/2029)
- glob patterns are not supported in this PR (explained why below)

### Usage  
define the following entries in `.vscode/settings.json`:   
```jsonc
// already implemented
"C_Cpp.default.compileCommands": "${workspaceFolder}/build/merged_compile_commands.json",
// new
"C_Cpp.mergeCompileCommands": { 
"sources": [
        "${workspaceFolder}/build/proj1/compile_commands.json",
        "${workspaceFolder}/build/proj2/compile_commands.json",
    ],
    "destination": "${workspaceFolder}/build/merged_compile_commands.json"
},
```
- NOTE:  `"C_Cpp.default.compileCommands"` is not changed (to a list) for backwards compatibility.

### Why are glob patterns not supported here?
- the `CMake Tools` extension supports glob patterns without consuming heavy CPU resources as it only tries to recursively search for `compile_commands.json` files when a build is finished succesfully. (one time per build process)
- the `C/C++` extension doesn't know when a build is started (or finished), it will need to recursively monitor files constantly.
  - **can** create a non-negligible and persistent CPU load if directories contain many files.
  - **will** trigger many false positive events.
### Error Handling  
- files defined in `"sources"` don't exist  
  - how would we notify the user? **(should we?)**
  - there are currently no squiggles in `settings.json`
  - logs in `C++ Configuration Errors` are used exclusively by the native language server
  - error popups seems excessive for this kind of error
- cannot create the destination folder/file (permissions, invalid path)
  - error popup seems ok in this scenario
- `compile_commands.json` can't be parsed
  - error popup seems ok in this scenario

### State Machine
```mermaid
stateDiagram-v2 
[*] --> load : settings.json loaded/changed
load : close file watchers<br>load configuration
check_dst: check if destination file exists
load --> check_dst
create : merge compile commands
check_srcs : check if any of the sources<br>is newer than dst
check_dst --> check_srcs : true
check_dst --> create : false
missingbank : periodically check if missing files are created
filewatchers : create source file watchers
check_srcs --> filewatchers: false
check_srcs --> create : true
missingbank --> filewatchers : file created
filewatchers --> missingbank : error<br>file missing
filewatchers --> create : source file changed
filewatchers --> check_dst : error<br>file watcher limit hit
create --> filewatchers : Success
create --> [*] : Error
```

## TODO:
- [ ] implement periodic file watchers on non-existant files
- [ ] finalize error handling/user notifications scheme
- [ ] some validation in `settings.ts` in `getMergeCompileCommands()`
- [ ] add tests
- [ ] remove `console.log()` prints


yiftahw:
default - no company specified) I have sole ownership of intellectual
property rights to my Submissions and I am not making Submissions in the
course of work for my employer.

‫בתאריך יום ה׳, 31 באוק׳ 2024 ב-20:33 מאת
‪microsoft-github-policy-service[bot]‬‏ ***@***.***‬‏>:‬

yiftahw:


‫בתאריך יום ה׳, 31 באוק׳ 2024 ב-20:33 מאת
‪microsoft-github-policy-service[bot]‬‏ ***@***.***‬‏>:‬

bobbrow:
Hi @yiftahw, before you get too far into this approach, what I believe we would like to see for #7029 would be a modification to the compileCommands property that accepts an array of strings (paths to the individual files to merge) instead of performing an actual merge and saving the results into another file.

The starting point is more like this (this is an incomplete code sample):
**c_cpp_properties.schema.json**
```json
                    "compileCommands": {
                        "oneOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "array",
                                "items": {
                                    "type": "string"
                                }
                            }
                        ],
                        "markdownDescription": "Full path to `compile_commands.json` file for the workspace.",
                        "descriptionHint": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered.",
                    },

```
**package.json**
```json
                    "C_Cpp.default.compileCommands": {
                        "oneOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "array",
                                "items": {
                                    "type": "string"
                                }
                            }
                        ],
                        "markdownDescription": "%c_cpp.configuration.default.compileCommands.markdownDescription%",
                        "scope": "machine-overridable"
                    },
```

Then in code, the compileCommands property/setting is treated as `string | string[]`.  To avoid having to create an additional temporary file, we would prefer to pass this new setting down to our language server as a `string[]` and then in the language server we would parse through each of the files individually and add them all into a single object containing all entries.

Unfortunately, that part of the change requires access to the closed-source portion of the code which you don't have access to.

yiftahw:
> Hi @yiftahw, before you get too far into this approach, what I believe we would like to see for #7029 would be a modification to the compileCommands property that accepts an array of strings (paths to the individual files to merge) instead of performing an actual merge and saving the results into another file.
> 
> The starting point is more like this (this is an incomplete code sample):
> **c_cpp_properties.schema.json**
> ```json
>                     "compileCommands": {
>                         "oneOf": [
>                             {
>                                 "type": "string"
>                             },
>                             {
>                                 "type": "array",
>                                 "items": {
>                                     "type": "string"
>                                 }
>                             }
>                         ],
>                         "markdownDescription": "Full path to `compile_commands.json` file for the workspace.",
>                         "descriptionHint": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered.",
>                     },
> 
> ```
> **package.json**
> ```json
>                     "C_Cpp.default.compileCommands": {
>                         "oneOf": [
>                             {
>                                 "type": "string"
>                             },
>                             {
>                                 "type": "array",
>                                 "items": {
>                                     "type": "string"
>                                 }
>                             }
>                         ],
>                         "markdownDescription": "%c_cpp.configuration.default.compileCommands.markdownDescription%",
>                         "scope": "machine-overridable"
>                     },
> ```
> 
> Then in code, the compileCommands property/setting is treated as `string | string[]`.  To avoid having to create an additional temporary file, we would prefer to pass this new setting down to our language server as a `string[]` and then in the language server we would parse through each of the files individually and add them all into a single object containing all entries.
> 
> Unfortunately, that part of the change requires access to the closed-source portion of the code which you don't have access to.

Hi @bobbrow thanks for the quick reply! Is there a plan to add this feature to the closed-source part of the code?

bobbrow:
Sorting by upvotes, this feature is on the first page of language server feature requests, though it's not terribly close to the top.  However, since a lot of the higher voted features are not currently actionable and this one shouldn't be too costly to implement, we may be able to find some time to prioritize it sooner.  I can't provide an estimate of when we'd get to it, but if you did the open source part, that could help accelerate us completing the closed source part.

Things we'd like to see on the open source part:
- [ ] updates to the c_cpp_properties.json schema and compileCommands setting to accept either a string or array of strings
- [ ] path validation for the array of strings in c_cpp_properties.json
- [ ] settings validation updates (VS Code doesn't audit the values/types of settings before sending them to us)
- [ ] file watcher support for the list of files provided by the user (I think we'd prefer not to allow glob patterns)

There could be other things I'm not thinking of right now, but we'd figure those out during PR, or perhaps you will run into them during development if you choose to help with this.

Thank you!
Bob

yiftahw:
@bobbrow I'll start fiddling with your approach and create a new draft once some work is done...  Thanks for the tips!

sean-mcmanus:
@bobbrow Yeah, the work to enable compileCommands to be an array of strings seems pretty small on the cpptools side.

yiftahw:
@bobbrow @sean-mcmanus on a different note, may I ask why do we monitor and send a `ChangeCompileCommandsNotification` to the language server on files other than the one from the selected configuration?  
I'm asking because monitoring groups of files (if multiple configurations exist) seems less trivial than a single file per configuration (although not impossible)  
the fallback implementation in `checkCompileCommands()` only monitors the selected configuration file (or the default one)  
code snippet in question:   
```ts
// partial snippet from `src/LanguageServer/configurations.ts`
// `CppProperties::updateCompileCommandsFileWatchers()`
this.configurationJson.configurations.forEach(c => {
    filePaths.add(fileSystemCompileCommandsPath);
});
filePaths.forEach((path: string) => {
    this.compileCommandsFileWatchers.push(fs.watch(path, () => {
        this.compileCommandsFileWatcherTimer = setTimeout(() => {
            this.compileCommandsFileWatcherFiles.forEach((path: string) => {
                this.onCompileCommandsChanged(path); // <- notify language server
```

sean-mcmanus:
I think that is a bug related to https://github.com/Microsoft/vscode-cpptools/issues/12889 .

yiftahw:
@sean-mcmanus just to make sure, if you think there is no benefit on notifying the language server on changes to `compile_commands.json` that we don't currently use (e.g from a different configuration index), I will file an issue and work on to fix it.  
again, the fallback case in `checkCompileCommands()` only checks the file pointed by the current configuration. (which is not so much a fallback because it's always being checked, even if file watchers don't fail)  
I think I can benefit from this change going forward with what was discussed here previously with @bobbrow

sean-mcmanus:
@yiftahw No, cpptools does not need to know that a compile_commands.json has changed if it's not set in the current configuration -- it just results in unnecessary processing. However, if a configurationProvider is set, then compileCommands may still be used as a fallback (but it doesn't need to be updated if compileCommands is not set).

yiftahw:
> @yiftahw No, cpptools does not need to know that a compile_commands.json has changed if it's not set in the current configuration -- it just results in unnecessary processing. However, if a configurationProvider is set, then compileCommands may still be used as a fallback (but it doesn't need to be updated if compileCommands is not set).

Is there a way in the open source part of the extension to know if there is some issue with the configuration provider and we need to fall back to using compile commands? (if set)

sean-mcmanus:
> Is there a way in the open source part of the extension to know if there is some issue with the configuration provider and we need to fall back to using compile commands? (if set)

You might be able to call canProvideConfiguration. The communicate with the configuration provider is all done in the open source TypeScript.

Actually, upon further review it appears my answer to your previous question was incorrect -- the current implementation does appear to rely on switching a configuration having already received a didChangeCompileCommands message, and otherwise it looks like it would continue to use the old compile commands, but that seems like the wrong implementation to me. Maybe it could be changed such that when a user switches configurations, then it can send the didChangeCompileCommands then (ideally beforehand).

yiftahw:
closing in favor of #12960

PR #12960, opened 2024-11-13
Handle multiple compile commands on client side (needs native server side support)

### LINKED ISSUE #7029: Support multiple compile commands files per project

yiftahw:
closes #7029 
- updates to the `c_cpp_properties.json` schema and `compileCommands` setting to accept either a string or array of strings.
- updates to the `settings.json` schema and `default.compileCommands` setting to accept either a string or array of strings.
- a single path with non-zero length in `c_cpp_properties.json`  or `settings.json` is placed inside an array of length 1.
- empty strings are filtered from the array of paths, a single empty string or an empty array are handled as undefined.
- file watcher support for the list of files provided by the user.
- single `compile_commands.json` is converted to a list of size 1 to consistently send an array (or undefined).
- handle squiggles for errors in `c_cpp_properties.json`.
- updated `C/C++: Edit Configurations (UI)` panel to handle multiple `compile_commands.json` paths.


Extension/c_cpp_properties.schema.json
```diff
@@ -70,9 +70,22 @@
                         ]
                     },
                     "compileCommands": {
+                        "oneOf": [
+                            {
+                                "type": "string",
+                                "default": ""
```
bobbrow: I don't believe that defaults should be defined for these properties since they are all optional

yiftahw: I added them only because when they exist, the user can clearly see that more than 1 option is available for this type:  
image  
I can revert this if it breaks a convention you wish to keep...

Extension/c_cpp_properties.schema.json
```diff
@@ -70,9 +70,22 @@
...
+                            {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "uniqueItems": true,
+                                "default": []
+                            }
+                        ],
                         "markdownDescription": "Full path to `compile_commands.json` file for the workspace.",
```
bobbrow: This description should be updated

Extension/c_cpp_properties.schema.json
bobbrow: What matters the most is that when we read the value of the property and it is not defined by the user that we read it in as undefined, not some other value.

In package.json when you set a default value, VS Code will return that value to you if the setting is not set by the user. I don't recall off the top of my head the behavior for c_cpp_properties.json. We just need to make sure this works as expected. We will have to trade off having nice IntelliSense for it if it results in undesired behavior when we read the values back.

yiftahw: Fair enough, i will double check this both in `cpp_properties.json` and in `settings.json` that we get an `undefined` value if not set.  
Also, previous logic converted empty paths to `undefined`, we can keep this behavior with empty arrays converted to `undefined` as well.

yiftahw: @bobbrow I added logic in `CppProperties.parsePropertiesFile()` to:
- filter away empty strings from the array, a single empty string or an empty array both evaluate to `undefined`.
- a single and non-empty path string will be placed in an array of length 1. 
  
let me know what you think 🙂 

an alternative to this approach that I can suggest is to allow `compileCommands` to be of type `string | string[] | undefined` and explicitly convert it to `string[] | undefined` in `updateCompileCommandsFileWatchers()` and `checkCompileCommands()` (and other places that might need it) with the same logic I implemented in `parsePropertiesFile()`.  
We can make it cleaner if we wrap the logic in a function `getCompileCommandsArray() -> string[] | undefined`

bobbrow: The approach looks good to me. I believe `parsePropertiesFile` is the right place to force the type to `string[]` so we don't ever have to worry about the ambiguity again.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -684,7 +685,7 @@ export class CppProperties {
             this.parsePropertiesFile(); // Clear out any modifications we may have made internally.
             const config: Configuration | undefined = this.CurrentConfiguration;
             if (config) {
-                config.compileCommands = path;
+                config.compileCommands = [path];
```
bobbrow, line 688: Question for my team. Do we want to prefer storing the compileCommands property as a string in the actual `c_cpp_properties.json` file, or as an array?  Either way is acceptable here because the json will be reparsed in `handleConfigurationChange` and converted back to an array.

One other thing to note is that we should change the Edit Configurations UI to make the `input` for compileCommands be a `textarea` instead, along with whatever associated changes should come with that. The answer to the question above would also guide the implementation for the storage of the data in the json file. (e.g. if we prefer string for the single entry case, we'd need to test for that when it comes via the UI and react accordingly)

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1092,11 +1093,13 @@ export class CppProperties {
             }
 
             if (configuration.compileCommands) {
-                configuration.compileCommands = this.resolvePath(configuration.compileCommands);
-                if (!this.compileCommandsFileWatcherFallbackTime.has(configuration.compileCommands)) {
-                    // Start tracking the fallback time for a new path.
-                    this.compileCommandsFileWatcherFallbackTime.set(configuration.compileCommands, new Date());
-                }
+                configuration.compileCommands = configuration.compileCommands?.map((path: string) => this.resolvePath(path));
```
bobbrow: minor: I don't believe you need to use `?.` here or on the line below because the existence of `configuration.compileCommands` was tested before entering this code block.

Extension/c_cpp_properties.schema.json
yiftahw: great!  
 just to mention, it's also used in `handleSquiggles()` because we parse `c_cpp_properties.json` there separately.

Extension/src/LanguageServer/configurations.ts
yiftahw: fixed

Extension/c_cpp_properties.schema.json
yiftahw: updated the description, feel free to suggest something different, English is not my native language.

yiftahw:
@bobbrow squiggles are now handled as well. I believe most of the work on the client side is done.  
let me know if you have more feedback.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1916,9 +1945,10 @@ export class CppProperties {
                 }
             }
         }
-        if (currentConfiguration.compileCommands) {
-            paths.push(`${currentConfiguration.compileCommands}`);
-        }
+
+        currentConfiguration.compileCommands?.forEach((file: string) => {
+            paths.push(`${file}`)
```
bobbrow: a semicolon is missing on this line.

bobbrow:
> @bobbrow squiggles are now handled as well. I believe most of the work on the client side is done.
> let me know if you have more feedback.

Thank you very much. Would you also be able to do the changes to the Edit Configurations UI in `Extension/ui/settings.*` so that multiple paths can be displayed/edited in that view?

yiftahw:
looking into it... putting the PR back to draft for now

yiftahw:
@bobbrow updated the UI panel.  
note: adding a single path in the UI will save it as an array of size 1 (similar to how `includePath` is handled)

Extension/c_cpp_properties.schema.json
yiftahw: > The approach looks good to me. I believe `parsePropertiesFile` is the right place to force the type to `string[]` so we don't ever have to worry about the ambiguity again.

Quick suggestion, might be an overkill. Instead of forcing the value to be a string array both in `parseCppProperties()` and in `handleSquiggles()`, we create a "version 5" of the properties and add a function "updateToVersion5()" which actually saves a single entry as an array in the json file.   
Again, this might be overkill, just a suggestion.

Extension/src/LanguageServer/configurations.ts
yiftahw: Quick suggestion for your first question, might be an overkill. Instead of forcing the value to be a string array both in `parsePropertiesFile()` and in `handleSquiggles()`, we create a "version 5" of the properties and add a function "updateToVersion5()" which actually saves a single entry as an array in the json file.   
Again, this might be overkill, just a suggestion.

bobbrow: We have considered making changes to the version number, but this gets disruptive because many people commit their configurations to source control. We received a lot of negative feedback in the past when version changes were more frequent so we avoid it as best we can. 

We are technically breaking versioning and assuming some risk by not incrementing the version number with this change, but the addition is very subtle. We will need to discuss this as a team and decide if this is acceptable or not.

sean-mcmanus: @bobbrow Can't it be a string or an array in the c_cpp_properties.json?

sean-mcmanus:
@yiftahw Your changes and cpptools updates to allow an array for compileCommands is available in https://github.com/microsoft/vscode-cpptools/releases/tag/v1.23.2

yiftahw:
@sean-mcmanus I downloaded the latest pre-release version and copied the binaries to my project folder, and I don't see @Colengms 's added string used
should I copy something else besides `cpptools` and `cpptools-srv` ?
```
[12/6/2024, 8:23:41 AM] "module.cxx" not found in "PLAYGROUND". 'includePath' from c_cpp_properties.json in folder 'undefined' will be used for this file instead.
```
looks like this string matches `file_not_found_in_path` instead of `file_not_found_in_path2` from #12964

Colengms:
> I don't see @Colengms 's added string used

Hi @yiftahw .  It looks like these strings' contents got reversed/mismatched.  The next insiders should correct that.  Otherwise, I believe support for use of an array of strings instead of a single string in the `compileCommands` field should not be supported. Though, it looks like some additional work will be needed to support use of a string array in the equivalent default setting (`C_Cpp.default.compileCommands`), but I don't think that should block your changes.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1441,6 +1458,13 @@ export class CppProperties {
                     }
                 }
             }
+
+            // Configuration.compileCommands is allowed to be defined as a string in the schema, but we send an array to the language server.
+            // For having a predictable behvaior, we convert it here to an array of strings.
```
sean-mcmanus: behvaior->behavior

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1790,6 +1814,11 @@ export class CppProperties {
         const configurations: ConfigurationJson = jsonc.parse(configurationsText, undefined, true) as any;
         const currentConfiguration: Configuration = configurations.configurations[this.CurrentConfigurationIndex];
 
+        // Configuration.compileCommands is allowed to be defined as a string in the schema, but we send an array to the language server.
+        // For having a predictable behvaior, we convert it here to an array of strings.
```
sean-mcmanus: behvaior->behavior

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Should this thread be resolved now?

sean-mcmanus:
@bobbrow Were you going to resolve the comments you made?

Extension/c_cpp_properties.schema.json
```diff
@@ -70,9 +70,22 @@
...
+                            },
+                            {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "uniqueItems": true,
+                                "default": []
+                            }
+                        ],
```
sean-mcmanus, line 84: Right now, if "C_Cpp.default.compileCommands" is added in the settings.json, VS Code initializes it to the invalid value of `null`. It seems like we should use

```json
                        "default": [
                            ""
                        ],
```

Or maybe just `[]`.

I don't think the "default" inside of the "oneOf" is used.

Also, it seems like this code below in `getAsArrayOfStringsOrUndefined`
```ts
        if (setting.default !== undefined) {
            console.error(`Default value for ${settingName} is expected to be undefined.`);
        }
```
should be removed if that default is added.

Also, FYI, the C_Cpp.default.browse.path also has an invalid value of `null` as the default, but we did that to prevent users from accidentally adding that default value and causing no workspace symbols to be parsed. It's possible there might be some way to eliminate that invalid default and still allow `[]` to be set explicitly, but I haven't looked into that and it's pre-existing so you don't need to worry about it.

sean-mcmanus (dismissed):

Extension/c_cpp_properties.schema.json
bobbrow: Please do not remove that check in `getAsArrayOfStringsOrUndefined`. Its purpose is to catch incorrect settings validation. I'll take a look at the PR and recommend a different approach.

bobbrow: I pushed a commit that addresses my concerns and also fixes a bug I noticed with the settings tracker not noticing when the value changes.  Please review and let me know if there are any issues.

https://github.com/microsoft/vscode-cpptools/pull/12960/commits/54af67e7d04764a85b530e0a422b83d48f3b4cd9

sean-mcmanus: ok, seems good to me

sean-mcmanus (dismissed):

Extension/c_cpp_properties.schema.json
```diff
@@ -70,9 +70,22 @@
...
+                                "type": "string",
+                                "default": ""
+                            },
+                            {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "uniqueItems": true,
+                                "default": []
```
Colengms: I believe there should not be any "default" fields in `c_cpp_properties.schema.json` at all, to avoid incorrectly indicating any value as a hard-coded 'default', as the actual defaults may have been overridden using `C_Cpp.default.*` settings.  The one existing occurrence for `enableConfigurationSquiggles` appears to have been a later addition that was in error.  Could you omit these new "default" fields?  (And maybe even remove the "default" from `enableConfigurationSquiggles` as well).

Colengms (dismissed):

Extension/c_cpp_properties.schema.json
sean-mcmanus: The defaults on "C_Cpp.default.compileCommands" are okay though -- my previous comment below was intended to be applied to package.json and not c_cpp_properties.schema.json.

yiftahw: @Colengms I added the default values to have some Intellisense because this field has 2 possible types.  
due to the logic in `CppProperties.forceCompileCommandsAsArray`, an empty string, an empty array or an array with empty strings all get demoted to `undefined`.  
This means that value from `C_Cpp.default.compileCommands` will be used instead if valid. (same logic applies to it in `CppSettings`)
image  
@bobbrow was also concerned of this change, so I'm removing it for now. if you guys change your mind about it I will re-insert this.

sean-mcmanus (approved):

Extension/c_cpp_properties.schema.json
```diff
@@ -239,7 +250,6 @@
         },
         "enableConfigurationSquiggles": {
             "type": "boolean",
-            "default": true,
```
bobbrow, line 242: Why is this default being removed?

yiftahw: [see comment by @colengms](https://github.com/microsoft/vscode-cpptools/pull/12960#discussion_r1894413496)

bobbrow: My mistake. I thought this was the setting in `package.json`.

PR #12979, opened 2024-11-19
Support Compiler Argument Trait AB Experiment

kuchungmsft:
- Depends on cpptools' update to provide ProjectContextResult.
- Send "standardVersion" trait in completion prompt by default.
- Added the following new traits
  - intelliSenseDisclaimer: compiler information disclaimer.
  - intelliSenseDisclaimerBeginning: to note the beginning of IntelliSense information.
  - compilerArguments: a list of compiler command arguments that could affect Copilot generating completions.
  - directAsks: direct asking Copilot to do something instead of providing an argument.
  - intelliSenseDisclaimerEnd: to note the end of IntelliSense information.
- A/B Experimental flags
  - copilotcppTraits: deprecated, no longer used.
  - copilotcppExcludeTraits:: deprecated, no longer used.
  - copilotcppIncludeTraits: string array to include individual trait, i.e., compilerArguments.
  - copilotcppMsvcCompilerArgumentFilter: map of regex string to absence prompt for MSVC.
  - copilotcppClangCompilerArgumentFilter: map of regex string to absence prompt for Clang.
  - copilotcppGccCompilerArgumentFilter: map of regex string to absence prompt for GCC.
  - copilotcppCompilerArgumentDirectAskMap: map of argument to prompt.

### DIFF
--- a/Extension/src/LanguageServer/copilotProviders.ts
+++ b/Extension/src/LanguageServer/copilotProviders.ts
@@ -5,9 +5,13 @@
 'use strict';
 
 import * as vscode from 'vscode';
+import { localize } from 'vscode-nls';
 import * as util from '../common';
-import { ChatContextResult, GetIncludesResult } from './client';
+import * as logger from '../logger';
+import * as telemetry from '../telemetry';
+import { GetIncludesResult } from './client';
 import { getActiveClient } from './extension';
+import { getCompilerArgumentFilterMap, getProjectContext } from './lmTool';
 
 export interface CopilotTrait {
     name: string;
@@ -34,35 +38,113 @@ export async function registerRelatedFilesProvider(): Promise<void> {
             for (const languageId of ['c', 'cpp', 'cuda-cpp']) {
                 api.registerRelatedFilesProvider(
                     { extensionId: util.extensionContext.extension.id, languageId },
-                    async (_uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken) => {
+                    async (uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken) => {
+                        const start = performance.now();
+                        const telemetryProperties: Record<string, string> = {};
+                        const telemetryMetrics: Record<string, number> = {};
+                        try {
+                            const getIncludesHandler = async () => (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [];
+                            const getTraitsHandler = async () => {
+                                const projectContext = await getProjectContext(uri, context, token);
 
-                        const getIncludesHandler = async () => (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [];
-                        const getTraitsHandler = async () => {
-                            const chatContext: ChatContextResult | undefined = await (getActiveClient().getChatContext(token) ?? undefined);
+                                if (!projectContext) {
+                                    return undefined;
+                                }
 
-                            if (!chatContext) {
-                                return undefined;
+                                let traits: CopilotTrait[] = [
+                                    { name: "intelliSenseDisclaimer", value: '', includeInPrompt: true, promptTextOverride: `IntelliSense is currently configured with the following compiler information. It reflects the active configuration, and the project may have more configurations targeting different platforms.` },
+                                    { name: "intelliSenseDisclaimerBeginning", value: '', includeInPrompt: true, promptTextOverride: `Beginning of IntelliSense information.` }
+                                ];
+                                if (projectContext.language) {
+                                    traits.push({ name: "language", value: projectContext.language, includeInPrompt: true, promptTextOverride: `The language is ${projectContext.language}.` });
+                                }
+                                if (projectContext.compiler) {
+                                    traits.push({ name: "compiler", value: projectContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${projectContext.compiler}.` });
+                                }
+                                if (projectContext.standardVersion) {
+                                    traits.push({ name: "standardVersion", value: projectContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${projectContext.standardVersion} language standard.` });
+                                }
+                                if (projectContext.targetPlatform) {
+                                    traits.push({ name: "targetPlatform", value: projectContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${projectContext.targetPlatform}.` });
+                                }
+                                if (projectContext.targetArchitecture) {
+                                    traits.push({ name: "targetArchitecture", value: projectContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${projectContext.targetArchitecture}.` });
+                                }
+
+                                if (projectContext.compiler) {
+                                    // We will process compiler arguments based on copilotcppXXXCompilerArgumentFilters and copilotcppCompilerArgumentDirectAskMap feature flags.
+                                    // The copilotcppXXXCompilerArgumentFilters are maps. The keys are regex strings for filtering and the values, if not empty,
+                                    // are the prompt text to use when no arguments are found.
+                                    // copilotcppCompilerArgumentDirectAskMap map individual matched argument to a prompt text.
+                                    // For duplicate matches, the last one will be used.
+                                    const filterMap = getCompilerArgumentFilterMap(projectContext.compiler, context);
+                                    if (filterMap !== undefined) {
+                                        const directAskMap: Record<string, string> = context.flags.copilotcppCompilerArgumentDirectAskMap ? JSON.parse(context.flags.copilotcppCompilerArgumentDirectAskMap as string) : {};
+                                        let directAsks: string = '';
+                                        const remainingArguments: string[] = [];
+
+                                        for (const key in filterMap) {
+                                            if (!key) {
+                                                continue;
+                                            }
+
+                                            const matchedArgument = projectContext.compilerArguments[key] as string;
+                                            if (matchedArgument?.length > 0) {
+                                                if (directAskMap[matchedArgument]) {
+                                                    directAsks += `${directAskMap[matchedArgument]} `;
+                                                } else {
+                                                    remainingArguments.push(matchedArgument);
+                                                }
+                                            } else if (filterMap[key]) {
+                                                // Use the prompt text in the absence of argument.
+                                                directAsks += `${filterMap[key]} `;
+                                            }
+                                        }
+
+                                        if (remainingArguments.length > 0) {
+                                            const compilerArgumentsValue = remainingArguments.join(", ");
+                                            traits.push({ name: "compilerArguments", value: compilerArgumentsValue, includeInPrompt: true, promptTextOverride: `The compiler arguments include: ${compilerArgumentsValue}.` });
+                                        }
+
+                                        if (directAsks) {
+                                            traits.push({ name: "directAsks", value: directAsks, includeInPrompt: true, promptTextOverride: directAsks });
+                                        }
+                                    }
+                                }
+
+                                traits.push({ name: "intelliSenseDisclaimerEnd", value: '', includeInPrompt: true, promptTextOverride: `End of IntelliSense information.` });
+
+                                const includeTraitsArray = context.flags.copilotcppIncludeTraits ? context.flags.copilotcppIncludeTraits as string[] : [];
+                                const includeTraits = new Set(includeTraitsArray);
+                                telemetryProperties["includeTraits"] = includeTraitsArray.join(',');
+
+                                // standardVersion trait is enabled by default.
+                                traits = traits.filter(trait => includeTraits.has(trait.name) || trait.name === 'standardVersion');
+
+                                telemetryProperties["traits"] = traits.map(trait => trait.name).join(',');
+                                return traits.length > 0 ? traits : undefined;
+                            };
+
+                            // Call both handlers in parallel
+                            const traitsPromise = getTraitsHandler();
+                            const includesPromise = getIncludesHandler();
+
+                            return { entries: await includesPromise, traits: await traitsPromise };
+                        }
+                        catch (exception) {
+                            try {
+                                const err: Error = exception as Error;
+                                logger.getOutputChannelLogger().appendLine(localize("copilot.relatedfilesprovider.error", "Error while retrieving result. Reason: {0}", err.message));
                             }
-
-                            let traits: CopilotTrait[] = [
-                                { name: "language", value: chatContext.language, includeInPrompt: true, promptTextOverride: `The language is ${chatContext.language}.` },
-                                { name: "compiler", value: chatContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${chatContext.compiler}.` },
-                                { name: "standardVersion", value: chatContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${chatContext.standardVersion} language standard.` },
-                                { name: "targetPlatform", value: chatContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetPlatform}.` },
-                                { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
-                            ];
-
-                            const excludeTraits = context.flags.copilotcppExcludeTraits as string[] ?? [];
-                            traits = traits.filter(trait => !excludeTraits.includes(trait.name));
-
-                            return traits.length > 0 ? traits : undefined;
-                        };
-
-                        // Call both handlers in parallel
-                        const traitsPromise = ((context.flags.copilotcppTraits as boolean) ?? false) ? getTraitsHandler() : Promise.resolve(undefined);
-                        const includesPromise = getIncludesHandler();
-
-                        return { entries: await includesPromise, traits: await traitsPromise };
+                            catch {
+                                // Intentionally swallow any exception.
+                            }
+                            telemetryProperties["error"] = "true";
+                            throw exception; // Throw the exception for auto-retry.
+                        } finally {
+                            telemetryMetrics['duration'] = performance.now() - start;
+                            telemetry.logCopilotEvent('RelatedFilesProvider', telemetryProperties, telemetryMetrics);
+                        }
                     }
                 );
             }
--- a/Extension/src/LanguageServer/lmTool.ts
+++ b/Extension/src/LanguageServer/lmTool.ts
@@ -9,9 +9,13 @@ import { localize } from 'vscode-nls';
 import * as util from '../common';
 import * as logger from '../logger';
 import * as telemetry from '../telemetry';
-import { ChatContextResult } from './client';
+import { ChatContextResult, ProjectContextResult } from './client';
 import { getClients } from './extension';
+import { checkDuration } from './utils';
 
+const MSVC: string = 'MSVC';
+const Clang: string = 'Clang';
+const GCC: string = 'GCC';
 const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: string } } = {
     language: {
         'c': 'C',
@@ -19,9 +23,9 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
         'cuda-cpp': 'CUDA C++'
     },
     compiler: {
-        'msvc': 'MSVC',
-        'clang': 'Clang',
-        'gcc': 'GCC'
+        'msvc': MSVC,
+        'clang': Clang,
+        'gcc': GCC
     },
     standardVersion: {
         'c++98': 'C++98',
@@ -44,6 +48,141 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
     }
 };
 
+function formatChatContext(context: ChatContextResult | ProjectContextResult): void {
+    type KnownKeys = 'language' | 'standardVersion' | 'compiler' | 'targetPlatform';
+    for (const key in knownValues) {
+        const knownKey = key as KnownKeys;
+        if (knownValues[knownKey] && context[knownKey]) {
+            // Clear the value if it's not in the known values.
+            context[knownKey] = knownValues[knownKey][context[knownKey]] || "";
+        }
+    }
+}
+
+export interface ProjectContext {
+    language: string;
+    standardVersion: string;
+    compiler: string;
+    targetPlatform: string;
+    targetArchitecture: string;
+    compilerArguments: Record<string, string>;
+}
+
+export function getCompilerArgumentFilterMap(compiler: string, context: { flags: Record<string, unknown> }): Record<string, string> | undefined {
+    // The copilotcppXXXCompilerArgumentFilters are maps.
+    // The keys are regex strings and the values, if not empty, are the prompt text to use when no arguments are found.
+    let filterMap: Record<string, string> | undefined;
+    try {
+        switch (compiler) {
+            case MSVC:
+                if (context.flags.copilotcppMsvcCompilerArgumentFilter !== undefined) {
+                    filterMap = JSON.parse(context.flags.copilotcppMsvcCompilerArgumentFilter as string);
+                }
+                break;
+            case Clang:
+                if (context.flags.copilotcppClangCompilerArgumentFilter !== undefined) {
+                    filterMap = JSON.parse(context.flags.copilotcppClangCompilerArgumentFilter as string);
+                }
+                break;
+            case GCC:
+                if (context.flags.copilotcppGccCompilerArgumentFilter !== undefined) {
+                    filterMap = JSON.parse(context.flags.copilotcppGccCompilerArgumentFilter as string);
+                }
+                break;
+        }
+    }
+    catch {
+        // Intentionally swallow any exception.
+    }
+    return filterMap;
+}
+
+function filterCompilerArguments(compiler: string, compilerArguments: string[], context: { flags: Record<string, unknown> }, telemetryProperties: Record<string, string>): Record<string, string> {
+    const filterMap = getCompilerArgumentFilterMap(compiler, context);
+    if (filterMap === undefined) {
+        return {};
+    }
+
+    const combinedArguments = compilerArguments.join(' ');
+    const result: Record<string, string> = {};
+    const filteredCompilerArguments: string[] = [];
+    for (const key in filterMap) {
+        if (!key) {
+            continue;
+        }
+        const filter = new RegExp(key, 'g');
+        const filtered = combinedArguments.match(filter);
+        if (filtered) {
+            filteredCompilerArguments.push(...filtered);
+            result[key] = filtered[filtered.length - 1];
+        }
+    }
+
+    if (filteredCompilerArguments.length > 0) {
+        // Telemetry to learn about the argument distribution. The filtered arguments are expected to be non-PII.
+        telemetryProperties["filteredCompilerArguments"] = filteredCompilerArguments.join(',');
+        telemetryProperties["filters"] = Object.keys(filterMap).filter(filter => !!filter).join(',');
+    }
+
+    return result;
+}
+
+export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken): Promise<ProjectContext | undefined> {
+    const telemetryProperties: Record<string, string> = {};
+    const telemetryMetrics: Record<string, number> = {};
+    try {
+        const projectContext = await checkDuration<ProjectContextResult | undefined>(async () => await getClients()?.ActiveClient?.getProjectContext(uri, token) ?? undefined);
+        telemetryMetrics["duration"] = projectContext.duration;
+        if (!projectContext.result) {
+            return undefined;
+        }
+
+        formatChatContext(projectContext.result);
+
+        const result: ProjectContext = {
+            language: projectContext.result.language,
+            standardVersion: projectContext.result.standardVersion,
+            compiler: projectContext.result.compiler,
+            targetPlatform: projectContext.result.targetPlatform,
+            targetArchitecture: projectContext.result.targetArchitecture,
+            compilerArguments: {}
+        };
+
+        if (projectContext.result.language) {
+            telemetryProperties["language"] = projectContext.result.language;
+        }
+        if (projectContext.result.compiler) {
+            telemetryProperties["compiler"] = projectContext.result.compiler;
+        }
+        if (projectContext.result.standardVersion) {
+            telemetryProperties["standardVersion"] = projectContext.result.standardVersion;
+        }
+        if (projectContext.result.targetPlatform) {
+            telemetryProperties["targetPlatform"] = projectContext.result.targetPlatform;
+        }
+        if (projectContext.result.targetArchitecture) {
+            telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        }
+        telemetryMetrics["compilerArgumentCount"] = projectContext.result.fileContext.compilerArguments.length;
+        result.compilerArguments = filterCompilerArguments(projectContext.result.compiler, projectContext.result.fileContext.compilerArguments, context, telemetryProperties);
+
+        return result;
+    }
+    catch (exception) {
+        try {
+            const err: Error = exception as Error;
+            logger.getOutputChannelLogger().appendLine(localize("copilot.projectcontext.error", "Error while retrieving the project context. Reason: {0}", err.message));
+        }
+        catch {
+            // Intentionally swallow any exception.
+        }
+        telemetryProperties["error"] = "true";
+        return undefined;
+    } finally {
+        telemetry.logCopilotEvent('ProjectContext', telemetryProperties, telemetryMetrics);
+    }
+}
+
 export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTool<void> {
     public async invoke(options: vscode.LanguageModelToolInvocationOptions<void>, token: vscode.CancellationToken): Promise<vscode.LanguageModelToolResult> {
         return new vscode.LanguageModelToolResult([
@@ -58,18 +197,12 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
                 return 'The active document is not a C, C++, or CUDA file.';
             }
 
-            const chatContext: ChatContextResult | undefined = await (getClients()?.ActiveClient?.getChatContext(token) ?? undefined);
+            const chatContext: ChatContextResult | undefined = await (getClients()?.ActiveClient?.getChatContext(currentDoc.uri, token) ?? undefined);
             if (!chatContext) {
                 return 'No configuration information is available for the active document.';
             }
 
-            for (const key in knownValues) {
-                const knownKey = key as keyof ChatContextResult;
-                if (knownValues[knownKey] && chatContext[knownKey]) {
-                    // Clear the value if it's not in the known values.
-                    chatContext[knownKey] = knownValues[knownKey][chatContext[knownKey]] || "";
-                }
-            }
+            formatChatContext(chatContext);
 
             let contextString = "";
             if (chatContext.language) {
@@ -100,7 +233,7 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
             telemetryProperties["error"] = "true";
             return "";
         } finally {
-            telemetry.logLanguageModelToolEvent('cpp', telemetryProperties);
+            telemetry.logCopilotEvent('Chat/Tool/cpp', telemetryProperties);
         }
     }
 
--- a/Extension/src/LanguageServer/utils.ts
+++ b/Extension/src/LanguageServer/utils.ts
@@ -112,3 +112,9 @@ export async function withCancellation<T>(promise: Promise<T>, token: vscode.Can
         });
     });
 }
+
+export async function checkDuration<T>(fn: () => Promise<T>): Promise<{ result: T; duration: number }> {
+    const start = performance.now();
+    const result = await fn();
+    return { result, duration: performance.now() - start };
+}
--- a/Extension/src/telemetry.ts
+++ b/Extension/src/telemetry.ts
@@ -123,10 +123,10 @@ export function logLanguageServerEvent(eventName: string, properties?: Record<st
     sendTelemetry();
 }
 
-export function logLanguageModelToolEvent(eventName: string, properties?: Record<string, string>, metrics?: Record<string, number>): void {
+export function logCopilotEvent(eventName: string, properties?: Record<string, string>, metrics?: Record<string, number>): void {
     const sendTelemetry = () => {
         if (experimentationTelemetry) {
-            const eventNamePrefix: string = "C_Cpp/Copilot/Chat/Tool/";
+            const eventNamePrefix: string = "C_Cpp/Copilot/";
             experimentationTelemetry.sendTelemetryEvent(eventNamePrefix + eventName, properties, metrics);
         }
     };
--- a/Extension/test/scenarios/SingleRootProject/tests/copilotProviders.test.ts
+++ b/Extension/test/scenarios/SingleRootProject/tests/copilotProviders.test.ts
@@ -9,11 +9,14 @@ import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 import * as vscode from 'vscode';
 import * as util from '../../../../src/common';
-import { ChatContextResult, DefaultClient, GetIncludesResult } from '../../../../src/LanguageServer/client';
+import { DefaultClient, GetIncludesResult } from '../../../../src/LanguageServer/client';
 import { CopilotApi, CopilotTrait } from '../../../../src/LanguageServer/copilotProviders';
 import * as extension from '../../../../src/LanguageServer/extension';
+import * as lmTool from '../../../../src/LanguageServer/lmTool';
+import { ProjectContext } from '../../../../src/LanguageServer/lmTool';
+import * as telemetry from '../../../../src/telemetry';
 
-describe('registerRelatedFilesProvider', () => {
+describe('copilotProviders Tests', () => {
     let moduleUnderTest: any;
     let mockCopilotApi: sinon.SinonStubbedInstance<CopilotApi>;
     let getActiveClientStub: sinon.SinonStub;
@@ -21,6 +24,7 @@ describe('registerRelatedFilesProvider', () => {
     let vscodeGetExtensionsStub: sinon.SinonStub;
     let callbackPromise: Promise<{ entries: vscode.Uri[]; traits?: CopilotTrait[] }> | undefined;
     let vscodeExtension: vscode.Extension<unknown>;
+    let telemetryStub: sinon.SinonStub;
 
     const includedFiles = process.platform === 'win32' ?
         ['c:\\system\\include\\vector', 'c:\\system\\include\\string', 'C:\\src\\my_project\\foo.h'] :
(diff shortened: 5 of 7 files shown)


Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -38,19 +39,51 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
-                                { name: "compiler", value: chatContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${chatContext.compiler}.` },
-                                { name: "standardVersion", value: chatContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${chatContext.standardVersion} language standard.` },
-                                { name: "targetPlatform", value: chatContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetPlatform}.` },
-                                { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
+                                { name: "intellisense", value: 'intellisense', includeInPrompt: true, promptTextOverride: `IntelliSense is currently configured with the following compiler information. It's best effort to reflect the active configuration, and the project may have more configurations targeting different platforms.` },
```
sean-mcmanus: It sounds like this is supposed to be "It's a best effort"? However, I'm confused about what "best effort" is supposed to mean. Can it just be "It reflects the active configuration"?

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -38,19 +39,51 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                                { name: "intellisenseBegin", value: 'Begin', includeInPrompt: true, promptTextOverride: `Begin of IntelliSense information.` }
```
sean-mcmanus: "Begin" should be "Beginning", i.e. "Beginning of IntelliSense information." (adjective phrase) "Begin IntelliSense information" (imperative command) would also work (I'm not sure which makes more sense in the context of the other text provided)...but it should "match" the "End of" part.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: image
notice it didn't include "Begin of".

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Will fix, thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Will fix, thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Fixed, thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Fixed, thanks.

sean-mcmanus (dismissed):
I'm approving, but I haven't been able to test it out and you should probably get another approver.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -38,19 +39,51 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
-                                { name: "compiler", value: chatContext.compiler, includeInPrompt: true, promptTextOverride: `This project compiles using ${chatContext.compiler}.` },
-                                { name: "standardVersion", value: chatContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${chatContext.standardVersion} language standard.` },
-                                { name: "targetPlatform", value: chatContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetPlatform}.` },
-                                { name: "targetArchitecture", value: chatContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${chatContext.targetArchitecture}.` }
+                                { name: "intellisense", value: 'intellisense', includeInPrompt: true, promptTextOverride: `IntelliSense is currently configured with the following compiler information. It reflects the active configuration, and the project may have more configurations targeting different platforms.` },
```
sean-mcmanus: Should this name be more descriptive? Like "intelliSenseDisclaimer"? Not sure what the value is used for.

kuchungmsft: You had concerns that providing Copilot build information derived from IntelliSense could mislead it, and you suggested that we should make it clear that the information is from IntelliSense explicitly. These `intellisense`, `intellisenseBegin`, `intellisenseEnd` are traits that we can enable to address your concerns, see the following screenshot.

image

kuchungmsft: Will change the name, thanks. `Value` is not used when `promptTextOverride` is defined. A trait with only the name and value looks like "name: value" in the prompts.

benmcmorran:
Pulling this down to test locally now and then I expect to be able to sign off.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,112 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        try {
+            logger.getOutputChannelLogger().appendLine(localize("copilot.cppcontext.error", "Error while retrieving the project context."));
+        }
+        catch {
+            // Intentionally swallow any exception.
+        }
+        telemetryProperties["error"] = "true";
+        return undefined;
+    } finally {
+        telemetry.logLanguageModelToolEvent('Completions/tool', telemetryProperties);
```
benmcmorran: Can you split this into two different events depending on if `getProjectContext()` was called for completions or called for tool (`#cpp`) use? As-is it will be hard to tell the two apart in telemetry.

Extension/src/telemetry.ts
```diff
@@ -126,7 +126,7 @@ export function logLanguageServerEvent(eventName: string, properties?: Record<st
 export function logLanguageModelToolEvent(eventName: string, properties?: Record<string, string>, metrics?: Record<string, number>): void {
```
benmcmorran: This function is misnamed if it's not actually logging under the `tool` event name any more.

Extension/src/LanguageServer/utils.ts
```diff
@@ -112,3 +112,9 @@ export async function withCancellation<T>(promise: Promise<T>, token: vscode.Can
         });
     });
 }
+
+export async function checkTime<T>(fn: () => Promise<T>): Promise<{ result: T; time: number }> {
+    const start = Date.now();
```
benmcmorran: (non-blocking) If you're expecting to use this checkTime function in very fast or performance-critical contexts, consider using [`performance.now()`](https://nodejs.org/api/perf_hooks.html#performancenow) instead of `Date.now()`. It provides sub-millisecond precision.

benmcmorran (dismissed):

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: `getProjectContext` is only used for completions. `#cpp` still uses `getContext` at line 163.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,112 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+                telemetryProperties["filteredCompilerArguments"] = filteredCompilerArguments.join(', ');
+                result.compilerArguments = filteredCompilerArguments;
+            }
+        }
+
+        return result;
+    }
+    catch {
+        try {
+            logger.getOutputChannelLogger().appendLine(localize("copilot.cppcontext.error", "Error while retrieving the project context."));
```
sean-mcmanus: @kuchungmsft I'm seeing it "randomly" (?) hit an exception here, but it's not easy for me to tell what the exception is. Can you change this to log some exception info? Should the cancel exception be caught separately? The cancel exception seems like a "by design" issue and not an error.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/utils.ts
sean-mcmanus: It currently outputs number of milliseconds. I don't think sub-millisecond precision is important in this case.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: i.e. can/should you catch `vscode.CancellationError()`? I'm not sure yet if that's an actual error/bug or not. I haven't figured out what is causing it.

sean-mcmanus: @kuchungmsft I don't understand why token.isCancellationRequested keeps getting set to true for me. It doesn't seem like it should. I don't see anything changing that would be reason to cancel.

kuchungmsft: I think Copilot Client side has a 200ms-ish timeout, did you stop at a breakpoint?

kuchungmsft: I am not able to repro. I am not hitting the exception.

sean-mcmanus: @kuchungmsft I see withLspCancellationHandling checking the cancellation token after the result has already been obtained -- should we not cancel if we already have the info? Or do we still need to cancel?

sean-mcmanus: It's from Ben's PR: https://github.com/microsoft/vscode-cpptools/pull/12773

kuchungmsft: I am sorry, I don't know. I was following the existing code. @benmcmorran  can you comment?

Extension/src/telemetry.ts
kuchungmsft: Fixed the naming, thanks.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Yeah, we can deal with this in a follow up PR -- I debugged it more and it appears to be following the existing code's pattern and the existing code has the same issues.

Extension/src/LanguageServer/utils.ts
kuchungmsft: Changed to [performance.now()](https://nodejs.org/api/perf_hooks.html#performancenow), clients can decide the precision they want to use.

sean-mcmanus (approved):

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Added the exception info, thanks.

sean-mcmanus (approved):

Extension/src/LanguageServer/lmTool.ts
benmcmorran: I'll comment on the follow up PR. Glen is correct that the cancellation is happening more frequently during debugging because a timeout is being hit on the Copilot side. We don't strictly need to cancel if we already have a result -- cancellation is a best effort process.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -38,22 +39,54 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                            }
+                            if (directAsks) {
+                                traits.push({ name: "directAsks", value: directAsks, includeInPrompt: true, promptTextOverride: directAsks });
+                            }
+
+                            traits.push({ name: "intelliSenseDisclaimerEnd", value: '', includeInPrompt: true, promptTextOverride: `End of IntelliSense information.` });
 
-                            const excludeTraits = context.flags.copilotcppExcludeTraits as string[] ?? [];
-                            traits = traits.filter(trait => !excludeTraits.includes(trait.name));
+                            const excludeTraits = new Set(context.flags.copilotcppExcludeTraits as string[] ?? []);
```
sean-mcmanus: This includes the compilerArguments currently -- when is that going to be updated/removed?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,113 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+            const err: Error = exception as Error;
+            logger.getOutputChannelLogger().appendLine(localize("copilot.projectcontext.error", "Error while retrieving the project context. Reason: {0}", err.message));
+        }
+        catch {
+            // Intentionally swallow any exception.
+        }
+        telemetryProperties["error"] = "true";
+        return undefined;
+    } finally {
+        telemetry.logLanguageModelEvent('Completions/tool', telemetryProperties);
```
sean-mcmanus: @kuchungmsft This telemetry is sent before copilotcppExcludeTraits is applied -- should all the filtering be applied before sending telemetry? Usually we want telemetry to match what was actually used.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,113 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        if (projectContext.result.standardVersion) {
+            telemetryProperties["standardVersion"] = projectContext.result.standardVersion;
+        }
+        if (projectContext.result.targetPlatform) {
+            telemetryProperties["targetPlatform"] = projectContext.result.targetPlatform;
+        }
+        if (projectContext.result.targetArchitecture) {
+            telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        }
+        telemetryProperties["compilerArgumentCount"] = projectContext.result.fileContext.compilerArguments.length.toString();
```
sean-mcmanus: Similar to my other comment, this count is before the filter is applied. I would think that the post-filtered count would be the important one.

sean-mcmanus (dismissed):
I left some comments, but I don't know if they need to be addressed or not.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: We exclude `compilerArguments` and some others via feature flag on Control Tower. See https://exp.microsoft.com/a/feature/3350270c-3f42-41af-937b-f53a170beddd?workspaceId=dbd12384-5001-4e90-a863-526321eaf233&group=/vscodeexpws/Copilot. Users would not see it when this PR gets merged.

image

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: We know post-filtered count via `filteredCompilerArguments`, I am tracking so that we can see the original argument number.

sean-mcmanus: ok

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: The telemetry is mainly to track the data we get from IntelliSense. Since we control traits via feature flag to enable or disable them, we know what traits are used. I think we are good.

marlonwq (approved):

sean-mcmanus:
@benmcmorran Were you done reviewing? I wanted to get this in on Monday.

benmcmorran (dismissed):

benmcmorran:
@sean-mcmanus Yes, changes look good to me.

kuchungmsft:
@sean-mcmanus I need to make some changes for the following requirements.
- If none of the MSVC /EHxx is provided, it means that exception handling is disabled. We need to be able to provide a prompt text to Copilot in such case.
- A/B experiments cannot share feature flags, I need to make it so that compiler argument experiment won't be affected by the feature flags(copilotcppExcludeTraits, copilotcppTraits) used in language standard trait experiment.

sean-mcmanus:
> * If none of the MSVC /EHxx is provided, it means that exception handling is disabled. We need to be able to provide a prompt text to Copilot in such case.

@kuchungmsft I don't understand -- exception handling is enabled by default with MSVC (/EHsc is the default).

kuchungmsft:
> > * If none of the MSVC /EHxx is provided, it means that exception handling is disabled. We need to be able to provide a prompt text to Copilot in such case.
> 
> @kuchungmsft I don't understand -- exception handling is enabled by default with MSVC (/EHsc is the default).

If you don't specify any /EH arguments, it's disabled, see the following.
image

sean-mcmanus:
> If you don't specify any /EH arguments, it's disabled, see the following. 

That is strange -- for some reason, I'm not getting that warning (using godbolt).

image

kuchungmsft:
> > > If you don't specify any /EH arguments, it's disabled, see the following.
> 
> Oh, well, the docs say the stack unwind doesn't work (i.e. Copilot seems incorrect), but try/catch still works: https://learn.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170
> 
> exception handling is still enabled by default, just the destructors may not be run. For code generation by Copilot, it would be equivalent to being enabled though, right?

I don't think it's enabled based on the following observation. In VS, when you set "Enable C++ exception" to No, it's basically not sending any /EH argument in the command line.
image

kuchungmsft:
> godbolt

I don't

sean-mcmanus:
> I don't think it's enabled based on the following observation. In VS, when you set "Enable C++ exception" to No, it's basically not sending any /EH argument in the command line.

That VS UI is just misleading/unideal. There is no way to disable try/catch/throw with cl.exe -- users can still build/debug/use C++ exceptions:

image

kuchungmsft:
> > I don't think it's enabled based on the following observation. In VS, when you set "Enable C++ exception" to No, it's basically not sending any /EH argument in the command line.
> 
> That VS UI is just misleading/unideal. There is no way to disable try/catch/throw with cl.exe -- users can still build/debug/use C++ exceptions:
> 
> image
> i.e. with "No" for "Enable C++ exception" -- there are still C++ exceptions enabled.

I don't know why godbolt doesn't repro the warning.


> with "No" for "Enable C++ exception"

Right, it seems that there is no way to disable the keywords per https://lopezruiz.net/2020/03/11-disabling-exception-handling-cpp.htm. Someone [claimed](https://stackoverflow.com/posts/65513682/revisions) that keywords could be disabled by the following, but I couldn't repro.

image


I think in the case when /EH flag is not present, we still want to tell Copilot that exception is disabled as that's users intend.

sean-mcmanus:
@kuchungmsft I don't think we should disable C++ exceptions. We don't know the user's intent is to disable them. The UI is from VS and we don't know the user is using VS or has seen the UI. If the user asks Copilot if C++ extensions are enabled by default, then Copilot is tell me yes, so the user's intent may be to use exceptions still, and the code compiles with exceptions.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -38,22 +39,54 @@ export async function registerRelatedFilesProvider(): Promise<void> {
 
                         const getIncludesHandler = async () => (await getIncludesWithCancellation(1, token))?.includedFiles.map(file => vscode.Uri.file(file)) ?? [];
                         const getTraitsHandler = async () => {
-                            const chatContext: ChatContextResult | undefined = await (getActiveClient().getChatContext(token) ?? undefined);
+                            const cppContext = await getProjectContext(context, token);
```
sean-mcmanus: @kuchungmsft Can you pass the uri instead of relying on the active file on the server? I don't think they can be relied on to match all the time. Same for the pre-existing getIncludes.

Or you could do that in a follow up?

kuchungmsft:
> @kuchungmsft I don't think we should disable C++ exceptions. We don't know the user's intent is to disable them. The UI is from VS and we don't know the user is using VS or has seen the UI. If the user asks Copilot if C++ extensions are enabled by default, then Copilot is tell me yes, so the user's intent may be to use exceptions still, and the code compiles with exceptions.

The problem is that if we let Copilot suggests code with those keywords, MSVC would complain "warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc" Wouldn't it better if we honor compiler's behavior here?

sean-mcmanus:
> The problem is that if we let Copilot suggests code with those keywords, MSVC would complain "warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc" Wouldn't it better if we honor compiler's behavior here?

That is just a warning. The user might have that warning disabled or there might be other situations where the warning may not appear.

kuchungmsft:
> > The problem is that if we let Copilot suggests code with those keywords, MSVC would complain "warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc" Wouldn't it better if we honor compiler's behavior here?
> 
> That is just a warning. The user might have that warning disabled or there might be other situations where the warning may not appear.
> 
> I think the goal is for Copilot to generate code that doesn't give errors, but I believe generating code that doesn't give any warnings is not a goal currently.

This PR will support both scenario and allow configuring prompt text in the control tower. Ben was suggesting that we could use softer wording, i.e., "Exception unwinding code doesn't destroy automatic C++ objects outside of try blocks that go out of scope because of an exception. Resource leaks and undefined behavior may result when a C++ exception is thrown." (quoted statement from the doc) , in the prompt as an FYI for Copilot in the case where no /EH is provided.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Fixed, thanks.

Extension/test/scenarios/SingleRootProject/tests/copilotProviders.test.ts
```diff
@@ -130,122 +137,229 @@ describe('registerRelatedFilesProvider', () => {
...
+        ok(result.traits.find((trait) => trait.name === 'directAsks')?.includeInPrompt, 'result.traits should have a direct asks trait with includeInPrompt true');
+        ok(result.traits.find((trait) => trait.name === 'directAsks')?.promptTextOverride === 'Do not generate code using RTTI keywords. Do not generate code using exception handling keywords. ', 'result.traits should have a direct ask trait with promptTextOverride');
+        ok(telemetryStub.calledOnce, 'Telemetry should be called once');
+        ok(telemetryStub.calledWithMatch('RelatedFilesProvider', sinon.match({
+            "includeTraits": 'directAsks compilerArguments',
+            'traits': 'standardVersion compilerArguments directAsks'
+        })));
+    });
+
+    it('uses only last argument from the duplicates.', async () => {
```
kuchungmsft, line 320: @sean-mcmanus this addresses the comment about duplicates arguments.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,366 @@
...
+                flags: {
+                    copilotcppMsvcCompilerArgumentFilter: 'Not a map'
+                }
+            },
+            compilerArguments: ['foo', 'bar'],
+            expectedCompilerArguments: []
+        });
+    });
+
+    it('should filter out all compilerArguments for unkonwn compilers.', async () => {
```
sean-mcmanus: unkonwn->unknown

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,366 @@
...
+        ok(result, 'result should not be undefined');
+        ok(result.language === 'C++');
+        ok(result.compiler === expectedCompiler);
+        ok(result.standardVersion === 'C++20');
+        ok(result.targetPlatform === 'Windows');
+        ok(result.targetArchitecture === 'x64');
+        ok(JSON.stringify(result.compilerArguments) === JSON.stringify(expectedCompilerArguments));
+    };
+
+    it('should log telemetry and provide cpp context properly when experimental flags are not defined.', async () => {
```
sean-mcmanus: Is this comment correct/intended? I don't see it logging anything. Should it be reworded to make that clearer?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        // Telemtry to learn about the argument distribution. The filtered arguments are expected to be non-PII.
```
sean-mcmanus: Telemtry->Telemetry

kuchungmsft: Fixed, thanks.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
kuchungmsft: Updated, thanks.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
kuchungmsft: Fixed, thanks.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
-
-                            return traits.length > 0 ? traits : undefined;
-                        };
-
-                        // Call both handlers in parallel
-                        const traitsPromise = ((context.flags.copilotcppTraits as boolean) ?? false) ? getTraitsHandler() : Promise.resolve(undefined);
-                        const includesPromise = getIncludesHandler();
-
-                        return { entries: await includesPromise, traits: await traitsPromise };
+                            catch {
```
sean-mcmanus, line 139: This breaks Copilot's auto-retry on cancel. With "main" it will retry a couple times or something, but not with the exception caught -- can you re-throw the exception?

i.e. with the return result, it's treating it like a valid result and caches that instead of re-requesting.

sean-mcmanus (changes requested):

Extension/test/scenarios/SingleRootProject/tests/copilotProviders.test.ts
sean-mcmanus: @kuchungmsft Can you clarify? How are duplicates handled? I thought we had decided to not remove duplicates.

kuchungmsft: In this test case, there are duplicate of "/std:" argument, but only the last one "/std:c++20" is used in the trait.
image

image

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: Otherwise, the context won't be valid for 2 minutes after the re-request (cache invalidation timeout).

kuchungmsft: Fixed, thank you!

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,366 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
```
sean-mcmanus, line 5: When are these tests run? Or how do I run them?

kuchungmsft: These are non unit test as they have dependency on vscode (for shared types). We don't seem to have any pipelines running these scenarios tests for some reason. I run it from command line.

`yarn test --no-sandboax --disable-updates --skip-welcome --skip-release-notes --disable-extensions --extensionDevelopmentPath=.\ --extensionTestsPath=dist\test\common\selectTests --scenario=SingleRootProject SingleRootProject`

image
image
image

I can also debug these as well.

image

image

sean-mcmanus: Ah, okay, I thought I had tried running the VSCode Tests with SingleRootProject, but I must have done something wrong prevoiusly.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,366 @@
...
+            compilerArguments: ['/RTC1', '/GR-', '/Gd', '/MDd', '/GR', '/FC', '/W3', '/EHsc', '/Zc:wchar_t', '/ZI', '/EHs-'],
+            expectedCompilerArguments: ['/GR-', '/GR', '/EHsc', '/EHs-']
+        });
+    });
+
+    it('should provide compilerArguments based on copilotcppClangCompilerArgumentFilter.', async () => {
+        await testGetProjectContext({
+            compiler: 'clang',
+            expectedCompiler: 'Clang',
+            context: { flags: { copilotcppClangCompilerArgumentFilter: '{"^(-fno-exceptions|-fno-rtti|/GR-)$": "", "^-std=.*$": ""}' } },
```
sean-mcmanus: FYI, you shouldn't ever need to have -std in the filter, because it's already provided to Copilot by the standardVersion context.

sean-mcmanus (approved):

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -0,0 +1,366 @@
...
+        ok(result.targetPlatform === 'Windows');
+        ok(result.targetArchitecture === 'x64');
+        ok(JSON.stringify(result.compilerArguments) === JSON.stringify(expectedCompilerArguments));
+    };
+
+    it('should provide compilerArguments based on copilotcppMsvcCompilerArgumentFilter.', async () => {
+        await testGetProjectContext({
+            compiler: 'msvc',
+            expectedCompiler: 'MSVC',
+            context: { flags: { copilotcppMsvcCompilerArgumentFilter: '{"^(/GR.*|/EH.*)$": ""}' } },
```
sean-mcmanus: FYI, cl.exe allows `-` or `/` so you'll want to make sure your regex accounts for either possibility. I would highly recommend sending out the filters you're using for review to help catch any other potential issues like this.

kuchungmsft: Will do.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                                    traits.push({ name: "targetArchitecture", value: cppContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${cppContext.targetArchitecture}.` });
+                                }
+
+                                if (cppContext.compiler) {
+                                    // We will process compiler arguments based on copilotcppXXXCompilerArgumentFilters and copilotcppCompilerArgumentDirectAskMap feature flags.
+                                    // The copilotcppXXXCompilerArgumentFilters are maps. The keys are regex strings for filtering and the values, if not empty,
+                                    // are the prompt text to use when no arguments are found.
+                                    // copilotcppCompilerArgumentDirectAskMap map individual matched argument to a prompt text.
+                                    // For duplicate matches, the last one will be used.
+                                    const filterMap = getCompilerArgumentFilterMap(cppContext.compiler, context);
```
sean-mcmanus: I think you'll want to filter by language too in addition to the compiler? Our c_cpp_properties.json only has 1 compilerArgs that are re-used for C and C++, but they may only be valid for C and not C++ and vice versa. Also, the cuda-cpp language/compiler accepts some different arguments that allow different formatting, e.g. `-std c++17` with space is allowed.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: Or are we relying on Copilot to figure that out? i.e. to ignore/handle C++ args correctly for C language?

sean-mcmanus: i.e. gcc may give a warning like `warning: command-line option '-fchar8_t' is valid for C++/ObjC++ but not for C`, but I don't know if Copilot could be confused and output code with char8_t?

sean-mcmanus (dismissed):

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: I am not sure I understand the first part. Are you saying that the argument list used to compile the selected file may contain invalid arguments?

sean-mcmanus: So I'm not sure if you want to check the language or not. I don't know how often Copilot can correctly handle language-specific args being provided for a different language.

I suppose if we/users notice incorrect handling of language-specific args then we could fix it later.

kuchungmsft: In case where there are spaces between argument and value, I need to think about it more. May just ignore those for now. Also, we don't plan to explore a log of arguments, just some selected ones, Sinem will share the document once ready.

kuchungmsft: I think I'll ignore the language for now as the arguments we'll be testing are likely non language-specific. Thanks.

sean-mcmanus:
@kuchungmsft FYI, clang-cl.exe isn't correctly handled -- the compiler is reported as Clang, but the args it excepts are normally msvc args like `/DFOO=1` instead of `-DFOO=1`, but it also accepts clang args via `-Xclang <arg>`...I'm not sure what priority that is, but it seems like that could be dealt with in a potential follow up.

kuchungmsft:
> @kuchungmsft FYI, clang-cl.exe isn't correctly handled -- the compiler is reported as Clang, but the args it excepts are normally msvc args like `/DFOO=1` instead of `-DFOO=1`, but it also accepts clang args via `-Xclang <arg>`...I'm not sure what priority that is, but it seems like that could be dealt with in a potential follow up.
> 
> Oh....but we're not even telling Copilot that clang-cl is being used...hmm...?

We can have regex strings to filter "/GR-", "-GR-", "-fno-rtti" for Clang compiler and map those arguments to prompt text like "Do not use dynamic_cast". We'll AB with prompt text first without raw argument values.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                                                    remainingArguments.push(matchedArguments[matchedArguments.length - 1]);
+                                                }
+                                            } else if (filterMap[key]) {
+                                                // Use the prompt text in the absence of argument.
+                                                directAsks += `${filterMap[key]} `;
+                                            }
+                                        }
+
+                                        const compilerArgumentsValue = remainingArguments.join(", ");
+                                        traits.push({ name: "compilerArguments", value: compilerArgumentsValue, includeInPrompt: true, promptTextOverride: `The compiler arguments include: ${compilerArgumentsValue}.` });
```
sean-mcmanus: @kuchungmsft Is it supposed to send this when there are no compiler args? 
image

i.e. it seems like it should check if compilerArgumentsValue is empty before pushing the trait?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        }
+        if (projectContext.result.targetArchitecture) {
+            telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        }
+        telemetryProperties["compilerArgumentCount"] = projectContext.result.fileContext.compilerArguments.length.toString();
+        // Telemetry to learn about the argument distribution. The filtered arguments are expected to be non-PII.
+        if (projectContext.result.fileContext.compilerArguments.length) {
+            const filteredCompilerArguments = filterComplierArguments(projectContext.result.compiler, projectContext.result.fileContext.compilerArguments, context);
+            if (filteredCompilerArguments.length > 0) {
+                telemetryProperties["filteredCompilerArguments"] = filteredCompilerArguments.join(' ');
```
sean-mcmanus: As I mentioned in email, I didn't understand how the comment "to learn about the argument distribution" would occur if the telemetry is filtering based on the current experiment filter. Was there supposed to be one filter for potential args to send in telemetry and another filter for the current experiment?

sean-mcmanus: Can this telemetry be moved up, after the duplicates are removed?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        if (projectContext.result.targetPlatform) {
+            telemetryProperties["targetPlatform"] = projectContext.result.targetPlatform;
+        }
+        if (projectContext.result.targetArchitecture) {
+            telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        }
+        telemetryProperties["compilerArgumentCount"] = projectContext.result.fileContext.compilerArguments.length.toString();
+        // Telemetry to learn about the argument distribution. The filtered arguments are expected to be non-PII.
+        if (projectContext.result.fileContext.compilerArguments.length) {
+            const filteredCompilerArguments = filterComplierArguments(projectContext.result.compiler, projectContext.result.fileContext.compilerArguments, context);
```
sean-mcmanus: Actually, I'm confused -- why is filterCompilerArguments being called 2 times? Can the filtering be done just 1 time?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+                break;
+        }
+    }
+    catch {
+        // Intentionally swallow any exception.
+    }
+    return filterMap;
+}
+
+function filterComplierArguments(compiler: string, compilerArguments: string[], context: { flags: Record<string, unknown> }): string[] {
```
sean-mcmanus: filterComplierArguments->filterCompilerArguments

However, as I mentioned elsewhere, I'm not sure why the filtering is done 2 times or if this method is actually needed or not.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                                    traits.push({ name: "standardVersion", value: cppContext.standardVersion, includeInPrompt: true, promptTextOverride: `This project uses the ${cppContext.standardVersion} language standard.` });
+                                }
+                                if (cppContext.targetPlatform) {
+                                    traits.push({ name: "targetPlatform", value: cppContext.targetPlatform, includeInPrompt: true, promptTextOverride: `This build targets ${cppContext.targetPlatform}.` });
+                                }
+                                if (cppContext.targetArchitecture) {
+                                    traits.push({ name: "targetArchitecture", value: cppContext.targetArchitecture, includeInPrompt: true, promptTextOverride: `This build targets ${cppContext.targetArchitecture}.` });
+                                }
+
+                                if (cppContext.compiler) {
```
sean-mcmanus: I think you can remove this `if (cppContext.compiler) {` right? From what I can tell, the value is always one of gcc/clang/msvc, right?

sean-mcmanus: Same with targetPlatform, targetArchitecture, standardVersion, and language, i.e. all the cppContext members should be filled in and non-optional/empty. (Although obviously that could be potentially improved in the future if we could capture some "unknown" values instead of defaulting to one).

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+
+                                        if (directAsks) {
+                                            traits.push({ name: "directAsks", value: directAsks, includeInPrompt: true, promptTextOverride: directAsks });
+                                        }
+                                    }
+                                }
+
+                                traits.push({ name: "intelliSenseDisclaimerEnd", value: '', includeInPrompt: true, promptTextOverride: `End of IntelliSense information.` });
+
+                                const includeTraitsArray = context.flags.copilotcppIncludeTraits as string[] ?? [];
```
bobbrow: `as string[]` is not a cast in TypeScript like it is in C#. Are you expecting this line to convert it to null if it's not a `string[]`? That won't actually happen.  `as string[]` just tells TypeScript that you know what the type is so the type checking should assume that's what it is.

Extension/src/LanguageServer/copilotProviders.ts
```diff
@@ -34,35 +38,109 @@ export async function registerRelatedFilesProvider(): Promise<void> {
...
+                                const includeTraits = new Set(includeTraitsArray);
+                                telemetryProperties["includeTraits"] = includeTraitsArray.join(' ');
```
bobbrow: Do any of the traits have spaces in them? When logging telemetry you want to make sure you have a stable delimiter.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        filteredArguments.push(...filtered);
+    }
+    return filteredArguments;
+}
+
+export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken): Promise<ProjectContext | undefined> {
+    const telemetryProperties: Record<string, string> = {};
+    try {
+        const projectContext = await checkTime<ProjectContextResult | undefined>(async () => await getClients()?.ActiveClient?.getProjectContext(uri, token) ?? undefined);
+        telemetryProperties["time"] = projectContext.time.toFixed(1);
```
bobbrow: Numerical values are generally reported in telemetry as "Metrics", not "Properties", FYI.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -44,6 +48,129 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
...
+        if (projectContext.result.standardVersion) {
+            telemetryProperties["standardVersion"] = projectContext.result.standardVersion;
+        }
+        if (projectContext.result.targetPlatform) {
+            telemetryProperties["targetPlatform"] = projectContext.result.targetPlatform;
+        }
+        if (projectContext.result.targetArchitecture) {
+            telemetryProperties["targetArchitecture"] = projectContext.result.targetArchitecture;
+        }
+        telemetryProperties["compilerArgumentCount"] = projectContext.result.fileContext.compilerArguments.length.toString();
```
bobbrow: Another property that should probably be a "metric"

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: I am following https://github.com/microsoft/vscode-cpptools/pull/12907, we'll know for sure once we get the telemetry.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: `filterCompilerArguments` gets called only once here, can you point me where the 2nds time is called?

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: I am not sure I follow, can you please elaborate on this comment?

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: The actual .filter call is done 2 times, i.e. it seems like unnecessary code duplication that makes it more confusing to follow and causes extra regex processing.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Up to the code that also calls .filter but handles the directAsksMap as well. It seems like that processing should just be done in one location, unless I'm not understanding something.

Extension/src/LanguageServer/copilotProviders.ts
sean-mcmanus: Sure, but I'm not sure why that PR added those checks. We should know for sure by looking at the cpptools side code that generates the data, and I see that it's always populated. It's more confusing if there are `if` checks for cases that are always true, because the reader is lead to believe that it might not always be true.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: However, in this case, I think the time property is specifically meant to be associated with the other data that is sent in the telemetryProperties.

sean-mcmanus: Oh, I see, you'd send it in a metric in the same telemetry event.

sean-mcmanus: However, in this case, `projectContext.time.toFixed(1);` actually returns a string and not a number, and metrics are numbers, unless you wanted to convert it back to a number.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Are you referring to this filtering.
image

The first filtering in lmTools.ts is to provide raw argument distribution in telemetry.

The second filtering in copilotProviders.ts is to handle duplicate argument as well as detecting the absence of argument so we can have a prompt.

lmTools.ts and copilotProviders.tx have different responsiblities, I am not able to filter just once and solve all problems. Does that make sense?

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Does it matter if one string is sent instead of a number? I understand that metrics would make sense if you're sending a lot of properties with number values, but is there a reason to do this for 1 property that I'm not aware of?

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: I think the value could be empty because of this in https://github.com/microsoft/vscode-cpptools/pull/12907
image

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Yeah, I still don't understand. You're sending duplicate arguments in telemetry unnecessarily. I think you can do all the processing at one spot, either processing the telemetry in getTraitsHandler or processing the directAsks in getProjectContext.

Extension/src/LanguageServer/lmTool.ts
bobbrow: You can always convert it in Kusto, but when you're aggregating a lot of data, it makes more sense for it to be stored as the correct type. It's not the end of the world if you send it as a string, but we should try to be consisent and send numbers as numbers.

Extension/src/LanguageServer/lmTool.ts
bobbrow: > Oh, I see, you'd send it in a metric in the same telemetry event.

yes.

> However, in this case, `projectContext.time.toFixed(1);` actually returns a string and not a number, and metrics are numbers, unless you wanted to convert it back to a number.

It's probably being converted to string because of the type checking for Properties. `projectContext.time` is a number.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: I updated the PR to support spaces between argument and value. Thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Fixed, thanks.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Responded in email on how to use feature flags to learn the distribtion.

kuchungmsft: Fixed the code to only filter once, thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: I updated the code, thanks.

Extension/src/LanguageServer/copilotProviders.ts
kuchungmsft: Use comma as delimiter, thanks.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Send as "Metrics" for numeric values, thanks.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Fixed, thanks.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Fixed the code to only filter once, thanks.

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Fixed the code to only filter once, thanks.

sean-mcmanus (approved):

PR #12988, opened 2024-11-22
Fix an unnecessary cancel/re-request with GitHub Copilot requests (and fix some other bugs with Copilot exception handling)

sean-mcmanus:
Follow up to https://github.com/microsoft/vscode-cpptools/pull/12773 and https://github.com/microsoft/vscode-cpptools/pull/12979 .

Fixes a GeneralError from a getProjectContext request returning undefined instead of throwing an exception (which triggers the retry).

Fixes this exception being thrown: image
and
image

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2320,11 +2320,10 @@ export class DefaultClient implements Client {
 
         // Don't cancel if the result has already been computed, because the result might still be usable
         // without needing to send an unnecessary re-request, which is the case for the callback to registerRelatedFilesProvider.
-        /*
         if (token.isCancellationRequested) {
-            throw new vscode.CancellationError();
+            return result;
+            // throw new vscode.CancellationError();
         }
-        */
 
         return result;
     }
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2321,8 +2321,7 @@ export class DefaultClient implements Client {
         // Don't cancel if the result has already been computed, because the result might still be usable
         // without needing to send an unnecessary re-request, which is the case for the callback to registerRelatedFilesProvider.
         if (token.isCancellationRequested) {
-            return result;
-            // throw new vscode.CancellationError();
+            return result; // throw new vscode.CancellationError();
         }
 
         return result;
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2228,10 +2228,24 @@ export class DefaultClient implements Client {
         await this.languageClient.sendNotification(DidOpenNotification, params);
     }
 
+    /**
+     * Copilot completion-related requests (e.g. getIncludes and getProjectContext) will have their cancellation tokens cancelled
+     * if the current request times out (showing the user completion results without context info),
+     * but the results can still be used for future requests (due to caching) so it's better to return results instead of cancelling.
+     * This is different behavior from the getChatContext, which does handle cancel requests, since the request blocks
+     * the UI results and always re-requests (no caching).
+    */
+
     public async getIncludes(maxDepth: number): Promise<GetIncludesResult> {
         const params: GetIncludesParams = { maxDepth: maxDepth };
         await this.ready;
         return this.languageClient.sendRequest(IncludesRequest, params);
+    }
+
+    public async getProjectContext(uri: vscode.Uri): Promise<ProjectContextResult> {
+        const params: TextDocumentIdentifier = { uri: uri.toString() };
+        await this.ready;
+        return this.languageClient.sendRequest(ProjectContextRequest, params);
     }
 
     public async getChatContext(uri: vscode.Uri, token: vscode.CancellationToken): Promise<ChatContextResult> {
@@ -2239,12 +2253,6 @@ export class DefaultClient implements Client {
         await withCancellation(this.ready, token);
         return DefaultClient.withLspCancellationHandling(
             () => this.languageClient.sendRequest(CppContextRequest, params, token), token);
-    }
-
-    public async getProjectContext(uri: vscode.Uri): Promise<ProjectContextResult> {
-        const params: TextDocumentIdentifier = { uri: uri.toString() };
-        await this.ready;
-        return this.languageClient.sendRequest(ProjectContextRequest, params);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


benmcmorran (dismissed):
These changes are safe, although the semantics are a bit confusing.

If cancellation is requested, that's typically an indication that the caller no longer cares about the result, so the old code was making that pattern explicit. However, it's possible that Copilot is behaving in an unexpected way here by requesting cancellation but still populating an internal cache with results that come later.

sean-mcmanus:
I was going to look at additional changes on Monday to remove some of the additional unnecessary cancelling -- and I think I'm going to create a separate function specifically for the register callback, since that appears to be a special case in regards to cancelling.

Extension/src/LanguageServer/client.ts
```diff
@@ -2318,8 +2318,10 @@ export class DefaultClient implements Client {
             }
         }
 
+        // Don't cancel if the result has already been computed, because the result might still be usable
+        // without needing to send an unnecessary re-request, which is the case for the callback to registerRelatedFilesProvider.
         if (token.isCancellationRequested) {
-            throw new vscode.CancellationError();
+            return result; // throw new vscode.CancellationError();
```
Colengms: If the caller has cancelled the request, why would it also use the result?

sean-mcmanus: The cancel it sends is unnecessary, it uses the result we send to it (for 2 minutes in its cache). If we actually, cancel, then it'll re-request 2 more times but then stop until 2 minutes (leaving it in a state without any context info). It's causing an increase in the failure rate with insiders (which added the canceling) compared to non-insiders. Luca is following up to investigate it more on the GitHub Copilot side.

lukka (dismissed):

Extension/src/LanguageServer/client.ts
```diff
@@ -2242,11 +2241,10 @@ export class DefaultClient implements Client {
             () => this.languageClient.sendRequest(CppContextRequest, params, token), token);
     }
 
-    public async getProjectContext(uri: vscode.Uri, token: vscode.CancellationToken): Promise<ProjectContextResult> {
+    public async getProjectContext(uri: vscode.Uri): Promise<ProjectContextResult> {
         const params: TextDocumentIdentifier = { uri: uri.toString() };
-        await withCancellation(this.ready, token);
```
lukka, line 2247: do we want to wait anyway for `await this.ready` ? I guess so :)

lukka: we do indeed! Closing this comment.

Extension/src/LanguageServer/client.ts
Colengms: Just to make sure this is captured here (reiterating from our discussion): This smells suspicious, and like a bug in the calling code. An operation should only be cancelled if the result is no longer needed.  That's what's cancel is for - to cancel, not deprioritize(?).  Otherwise, what's the purpose of cancellation?  Working around this here seems like a hasty work-around and not a proper fix.

sean-mcmanus: @Colengms Yes, I would agree it seems to me like a bug in the calling code, but we can't change the calling code. I believe Luca is looking into the calling code. We don't own the calling code and can't just fix it ourselves, so this works around the issue.

Colengms: As long as these don't trigger (or block on) IntelliSense/TU operations, etc., not cancelling seems like it would have a minimal impact.  But it looks like the native side is not properly setting `allow_tu_header_lookup` to false, so is incurring calls into TUs to search for headers.  And they're not setting `allow_query_translation_unit_source`, so could trigger/block rebuilding the include graph.  Since we're not removing potentially stale requests, could we set those options to ensure stale requests don't cause perf impact in those scenarios?

sean-mcmanus: @Colengms I'll talk to you offline about the IntelliSense handling. I'm cancelling this PR to so I can switch to just handling ServerCancellation (even though it's currently not possible).

sean-mcmanus: We found out ServerCancellation is still done when switching files and making edits "fast enough", which doesn't seem like an important scenario to me, but we can still ignore those ServerCancellations in case the user switches back to the file.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -127,11 +127,11 @@ function filterCompilerArguments(compiler: string, compilerArguments: string[],
     return result;
 }
 
-export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, token: vscode.CancellationToken): Promise<ProjectContext | undefined> {
+export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }): Promise<ProjectContext | undefined> {
```
Colengms, line 133: It looks like this function catches any exception, including cancellation (which it looks like your native PR changes to a general exception) and resolves it into an 'undefined' result.  Are you sure that the calling code will consider that a cancellation?  Generally, when an async request is cancelled, the proper thing to do is handle it as a cancellation, so the reject handler is called, not process it as a successful completion with an empty result. #Resolved

sean-mcmanus: @Colengms The native code doesn't change any cancels to a general exception (the previous code was not converting those types of errors to CancellationExceptions), but it does give a general exception error which is thrown by the TypeScript. At one point a few days ago with Glen's branch, I confirmed that the general errors were correctly treated as a cancel and retried, but there might have been a breaking change, so I'll re-confirm that...if that doesn't work correctly, it would think it would be a pre-existing bug from the previous PR and not related to my change unless I'm missing something, which I'll see...

sean-mcmanus: Yeah, there's a problem/bug -- getIncludes correclty throws an exception, but getProjectContext returns undefined. This looks like a bug from the previous PR -- I'll fix it by adding are re-throw which was done for the getIncludes case.

sean-mcmanus: I'm also seeing this bug from the previous PR which I'll try to fix: (which I had noticed previously but forgot to follow up on): 
image

sean-mcmanus: I've fixed the 2 bugs from the previous PR and SeverError appears to be handled correctly now (it retries).

sean-mcmanus: @Colengms Does that address the issue you were concerned about or were you referring to something else?

Colengms (approved):
:shipit:

kuchungmsft (approved):

PR #12991, opened 2024-11-22
Add instrumentation support to the typescript code

fearthecowboy:
This adds the instrumentation support to enable performance testing and monitoring.

It: 
- cleaned up some calls to logging so that we're not over-importing in typescript
- patches the logging so that log messages are sent to the collector.
- adds calls to monkey-patch vscode, the client and the providers so that we get monitoring events for as much as possible.
- if the `PERFECTO_LAUNCH_SETTINGS` environment variable is not set, everything here is a no-op
- the actual instrumentation code is loaded by the perfecto tool


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1,1278 +1,1278 @@
...
+import { promisify } from 'util';
+import * as vscode from 'vscode';
+import * as nls from 'vscode-nls';
+import * as util from '../common';
+import { isWindows } from '../constants';
+import { expandAllStrings, ExpansionOptions, ExpansionVars } from '../expand';
+import { CppBuildTask, CppBuildTaskDefinition, cppBuildTaskProvider } from '../LanguageServer/cppBuildTaskProvider';
+import { configPrefix } from '../LanguageServer/extension';
+import { CppSettings, OtherSettings } from '../LanguageServer/settings';
+import { getOutputChannel, getOutputChannelLogger, Logger, showOutputChannel } from '../logger';
```
bobbrow: It's unclear to me why this is changing. It doesn't seem related to the purpose of this PR. Can you revert this file?

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -7,7 +7,7 @@
 import * as vscode from 'vscode';
 import { localize } from 'vscode-nls';
 import * as util from '../common';
-import * as logger from '../logger';
+import { getOutputChannelLogger } from '../logger';
```
bobbrow: It's unclear to me why this is changing. It doesn't seem related to the purpose of this PR. Can you revert this file?

Extension/src/LanguageServer/references.ts
```diff
@@ -8,7 +8,7 @@ import * as vscode from 'vscode';
 import { Position, TextDocumentIdentifier } from 'vscode-languageclient';
 import * as nls from 'vscode-nls';
 import * as util from '../common';
-import * as logger from '../logger';
+import { getOutputChannel } from '../logger';
```
bobbrow: It's unclear to me why this is changing. It doesn't seem related to the purpose of this PR. Can you revert this file?

Extension/src/logger.ts
```diff
@@ -1,190 +1,194 @@
...
+export class Logger {
+    private writer: (message: string) => void;
+
+    constructor(writer: (message: string) => void) {
+        this.writer = writer;
+    }
+
+    public append(level: number, message: string): void;
+    public append(message: string): void;
+    public append(levelOrMessage: string | number, message?: string): void {
```
bobbrow: I understand why you did this, but I'm not a fan of this pattern (the optional parameter is the first arg). Since we don't want to put the loggingLevel second, I'd prefer we not change the logging API until we're ready to change it everywhere in the code.

Suggestion: Create a new API taking the level and the message and deprecate the old one. Then we can migrate the rest of the extension to it in a separate PR.

Extension/src/logger.ts
```diff
@@ -1,190 +1,194 @@
...
+        message = message || levelOrMessage as string;
+        const hasLevel = typeof levelOrMessage === 'number';
+
+        if (!hasLevel || getLoggingLevel() >= levelOrMessage) {
```
bobbrow: This being a semi-high-frequency function, perhaps the current logging level should be stored on the Logger as a field instead of having to call a function every time.

Extension/src/logger.ts
```diff
@@ -1,190 +1,194 @@
...
+    private writer: (message: string) => void;
+
+    constructor(writer: (message: string) => void) {
+        this.writer = writer;
+    }
+
+    public append(level: number, message: string): void;
+    public append(message: string): void;
+    public append(levelOrMessage: string | number, message?: string): void {
+        message = message || levelOrMessage as string;
```
bobbrow: Is it possible for `levelOrMessage` to be a number when `message` is not passed in? Better to test the inputs with the `isString` `isNumber` helpers than to assert with `as`.  (though my preference is to remove the overloads and create a new API that can forward here for logging to the output channel when the checks pass)

Extension/src/platform.ts
```diff
@@ -1,105 +1,105 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as fs from 'fs';
+import * as os from 'os';
+import * as plist from 'plist';
+import * as nls from 'vscode-nls';
+import { LinuxDistribution } from './linuxDistribution';
+import { getOutputChannelLogger, showOutputChannel } from './logger';
```
bobbrow: It's unclear to me why this is changing. It doesn't seem related to the purpose of this PR. Can you revert this file?

Extension/src/instrumentation.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // this loads the bootstrap module (global-instrumentation-support) which adds some global functions
+    if (services.launchSettings?.bootstrapModule) {
+        void require(services.launchSettings.bootstrapModule);
+    }
+}
+
+// If the instrumentation object was loaded then we can set the services from the global object
+if ((global as any).instrumentation) {
```
bobbrow: Use `isInstrumentationEnabled` here or connect it to the last block with an `else`?

Extension/src/common.ts
```diff
@@ -1,1820 +1,1817 @@
...
+    return msvcEnvVars.every((envVarName) => process.env[envVarName] !== undefined && process.env[envVarName] !== '');
+}
+
+function isIntegral(str: string): boolean {
+    const regex = /^-?\d+$/;
+    return regex.test(str);
+}
+
+export function getLoggingLevel() {
+    return getNumericLoggingLevel(vscode.workspace.getConfiguration("C_Cpp", null).get<string>("loggingLevel"));
```
bobbrow: Since we can't pull `CppSettings` into this module, maybe we need to split out the validation code because we shouldn't be reading raw settings in the extension anymore.  Apparently this file was skipped due to it not using `CppSettings`.

This isn't necessarily your problem. I'm just thinking out loud.

Extension/src/logger.ts
fearthecowboy: I didn't think storing it was a good idea, then we'd have to make sure that the stored value got updated if the value changed.

Extension/src/instrumentation.ts
fearthecowboy: fixed - however it's not actually an 'else' condition - the previous will do the global work if required, and this block wires up this module to it.

Extension/src/instrumentation.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/** sends a perf message to the monitor */
+export function sendInstrumentation(message: PerfMessage): void {
+    services.message(message);
+}
+
+/** verifies that the instrumentation is loaded into the global namespace */
+export function isInstrumentationEnabled(): boolean {
+    return !!(global as any).instrumentation;
```
sean-mcmanus: Can you explain where this global.instrumentation comes from? 
image

sean-mcmanus: Where is instrumentation loaded into the global namespace? Is that in the perfecto repo? Can you point me to the filename?

fearthecowboy: In `src/utility/vscode/global-instrumentation-support.ts` - the tool gets loaded into the process via the vscode command line switch `--extensionTestsPath=` , which bootstraps the `src/utility/vscode/initialize.ts` module which loads the `src/utility/vscode/global-instrumentation-support.ts` module. 

The `global-instrumentation-support.ts` adds some things to the global namespace so that the extension can take advantage of them without having to have them built in. (and lets most the ongoing work isolated in the tool)

Extension/src/LanguageServer/client.ts
```diff
@@ -1340,29 +1351,31 @@ export class DefaultClient implements Client {
                     this.copilotHoverProvider = new CopilotHoverProvider(this);
                     this.disposables.push(vscode.languages.registerHoverProvider(util.documentSelector, this.copilotHoverProvider));
```
sean-mcmanus: @fearthecowboy Should the Copilot hover be handled? It was added recently.

fearthecowboy: Yeah, sure. I'll add it right now.

Extension/src/instrumentation.ts
```diff
@@ -0,0 +1,61 @@
...
+if (isInstrumentationEnabled()) {
+    // instrumentation services provided by the tool
+    services.instrument = (global as any).instrumentation.instrument;
+    services.message = (global as any).instrumentation.message;
+    services.init = (global as any).instrumentation.init;
+
+    services.init(require('vscode'));
+}
+
+(globalThis as any)["_vscode_"] = require('vscode');
```
sean-mcmanus: What is this for? Does it need to be run if isInstrumentationEnabled() is false?

Extension/src/instrumentation.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/** verifies that the instrumentation is loaded into the global namespace */
+export function isInstrumentationEnabled(): boolean {
+    return !!(global as any).instrumentation;
+}
+
+// If the instrumentation object was *actually* loaded then we can set the services from the global object.
+// Having this separate ensures that this module is wired up to the global object.
+// It's not included in the previous block because if something else loads the instrumentation code first
```
sean-mcmanus: What is "the previous block"?

sean-mcmanus (changes requested):

Extension/src/instrumentation.ts
```diff
@@ -0,0 +1,61 @@
...
+// this is still needed so that *this* module is properly connected to the global object.
+if (isInstrumentationEnabled()) {
```
sean-mcmanus: Should this isInstrumentationEnabled() value be saved to a global variable and referenced in other places using the variable? Otherwise, couldn't `instrumentation` be added to the global scope later on?

sean-mcmanus (changes requested):
Shouldn't the instrument be added for `vscode.languages.registerCodeActionsProvider` in extension.ts?

sean-mcmanus:
FYI, a copilotCompletionProvider got added recently, but I don't know how you would handle that with `instrument`.

Extension/src/instrumentation.ts
fearthecowboy: Not any longer, removing.

Extension/src/instrumentation.ts
fearthecowboy: Reference to code removed.

Extension/src/instrumentation.ts
fearthecowboy: Actually, yeah, it can now.

sean-mcmanus (approved):

PR #12995, opened 2024-11-24
Main

RLCS-USD-LEAGL:
{
    "recommendations": [
        "github.copilot"
    ]
}

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "files.autoSave": "afterDelay"
+}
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,31 @@
+{
+<<<<<<< Updated upstream
```
fearthecowboy, line 2: The `<<<<<<` , `========` and `>>>>>>>` lines are not permitted.

bobbrow:
We do not intend to update the recommendations. It is also unclear why this PR merges in the contents of another unapproved PR.

PR #13000, opened 2024-11-26
Add C/C++ language support contribution point in package.json

thernstig:
Reference #4077


bobbrow:
These are actually already defined in VS Code itself.
https://github.com/microsoft/vscode/blob/6ff5e0475d8317eec3ac0dc6ffd0411c4a8cb715/extensions/cpp/package.json

The reason we have code for updating the associations is because of extensionless files (e.g. string, vector, etc)

thernstig:
@bobbrow I should have taken a look there first of course, apologize for the fuss. Closing this.

PR #13012, opened 2024-12-03
Remove Copilot completion request cancellation.

sean-mcmanus:
Follow up to https://github.com/microsoft/vscode-cpptools/pull/12773 .

The previous/abandoned PR also ignored the ServerCancellation requests that occur if a user switches files and makes edits, but I added them back: https://github.com/microsoft/vscode-cpptools/pull/12988

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -804,9 +804,9 @@ export interface Client {
     getShowConfigureIntelliSenseButton(): boolean;
     setShowConfigureIntelliSenseButton(show: boolean): void;
     addTrustedCompiler(path: string): Promise<void>;
-    getIncludes(maxDepth: number, token: vscode.CancellationToken): Promise<GetIncludesResult>;
+    getIncludes(maxDepth: number): Promise<GetIncludesResult>;
     getChatContext(uri: vscode.Uri, token: vscode.CancellationToken): Promise<ChatContextResult>;
-    getProjectContext(uri: vscode.Uri, token: vscode.CancellationToken): Promise<ProjectContextResult>;
+    getProjectContext(uri: vscode.Uri): Promise<ProjectContextResult>;
 }
 
 export function createClient(workspaceFolder?: vscode.WorkspaceFolder): Client {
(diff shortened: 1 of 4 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -2348,6 +2348,27 @@ export class DefaultClient implements Client {
         return result;
     }
 
+    // This is used to avoid processing unnecessary LSP cancel requests during a Copilot completion
```
lukka: @sean-mcmanus the exact reason is that caching is happening on copilot-client, hence even if the token is cancelled, it means that the return value will NOT be used for the prompt construction, but it will be used for caching purposes.

lukka (dismissed):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yes, I know it's not used for the currently running completion (it doesn't block Copilot), but it's cached and used on the next one. Does my comment not explain that? i.e. should I re-word my comment?

lukka: I'd omit the 2-minute detail since it is an implementation detail not located in this codebase. Instead, I'd mention that upon cancellation, the result will not be used to build the prompt (since the copilot extension gave up on waiting for a value from this provider), but the returned value is cached by the copilot extension and used on a subsequent call when available, where the cache will be hit and the call to getProjectContext will be saved.

The 2 minutes is the cache entry eviction time defined by the copilot extension.

Extension/src/LanguageServer/client.ts
```diff
@@ -2348,6 +2348,27 @@ export class DefaultClient implements Client {
...
+    private static async withoutLspCancellationHandling<T>(task: () => Promise<T>): Promise<T> {
+        let result: T;
+
+        try {
+            result = await task();
+        } catch (e: any) {
+            if (e instanceof ResponseError && e.code === ServerCancelled) {
+                // This can occur if the user switches files and makes edits fast enough.
+                // It doesn't seem like a very important scenario, so cancelling seems fine, even though
+                // not cancelling would allow the value to be cached if the user switches back to that file.
```
lukka, line 2362: FYI, the entries in the cache are per URI. The sooner an entry is computed, the better. Hence, if we have an actual result, we should always return it back regardless of the cancellation.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Ok, I tried to re-word it.

Extension/src/LanguageServer/client.ts
sean-mcmanus: @lukka This was the cancellation code in cpptools that I had removed, because you said you wanted it to stay in case there were multiple requests sent. I can certainly revive that PR.

lukka: @sean-mcmanus No need to revive that PR, I am just sharing what I had in my mind when I was readind that comment, which looks like saying that the cache is a single entry, although I may have misunderstood.

lukka (approved):

Extension/src/LanguageServer/client.ts
sean-mcmanus: Well, assuming that's what you want, I'll cancel this PR and reopen the abandoned one.

PR #13018, opened 2024-12-04
code snippet provider

lukka:


Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+import { DocumentSelector } from 'vscode-languageserver-protocol';
+import { getOutputChannelLogger, Logger } from '../logger';
+import * as telemetry from '../telemetry';
+import { getCopilotApi } from "./copilotProviders";
+import { clients } from './extension';
+import { CodeSnippet, CompletionContext, ContextProviderApiV1 } from './tmp/contextProviderV1';
+
+// An ever growing cache of completion context snippets. //?? TODO Evict old entries.
+const completionContextCache: Map<string, CodeSnippet[]> = new Map<string, CodeSnippet[]>();
+const cppDocumentSelector: DocumentSelector = [{ language: 'cpp' }, { language: 'c' }];
```
benmcmorran: Do we also want `cuda-cpp` in this list?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            if (token.isCancellationRequested) {
+                throw new CancellationError();
+            }
+            return {
+                importance: item.importance, uri: item.uri, value: item.text
+            };
+        });
+
+        completionContextCache.set(documentUri, codeSnippets);
+        const duration: number = Date.now() - startTime;
```
benmcmorran: You probably want [`performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now). It's higher precision and guaranteed to be monotonic.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+    } catch (e) {
+        const err = e as Error;
+        out.appendLine(`Copilot: getCompletionContextWithCancellation(): Error: '${err?.message}', stack '${err?.stack}`);
+
+        // //?? TODO Add telemetry for failure.
+        return [];
+    }
+}
+
+const timeBudgetFactor: number = 0.5;
```
benmcmorran: We might want to be able to experiment with this value.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            const docUri = context.documentContext.uri;
+            const cachedValue: CodeSnippet[] | undefined = completionContextCache.get(docUri.toString());
+            const snippetsPromise = getCompletionContextWithCancellation(docUri,
+                context.documentContext.offset, startTime, out, completionContextCancellation.token);
+            const [codeSnippets, kind] = await waitForCompletionWithTimeoutAndCancellation(
+                snippetsPromise, cachedValue, context.timeBudget * timeBudgetFactor, copilotAborts);
+            snippetsKind = kind;
+            // //?? TODO Add telemetry for Computed vs Cached.
+
+            return codeSnippets ?? [];
```
benmcmorran: Won't matter if we're the only provider, but we may want to experiment with lowering the importance if the snippets were from the cache.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            // //?? TODO Add telemetry for elapsed time.
+        }
+
+        return [];
+    }
+};
+
+export async function registerCopilotContextProvider(): Promise<void> {
+    try {
+        const isCustomSnippetProviderApiEnabled = await telemetry.isExperimentEnabled("CppToolsCustomSnippetsApi");
```
benmcmorran: We can't do audience filtering and set this flag at the same time. Instead, on the Copilot side you'll need to modify the construction of `activeExperiments` in `CompletionContext` to check for the experiment, then update the resolver here to respect `activeExperiments`.

Alternatively, if the intent of this check was to act as a kill-switch then you can leave it as-is (although we'll still need the `activeExperiments` plumbing). We'll just need to be sure that `CppToolsCustomSnippetsApi` is at 100% rollout for all VS Code traffic before starting the actual experiment.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    return [snippetsOrNothing, SnippetsKind.Computed];
+}
+
+// Get the completion context with a timeout and a cancellation token.
+// The cancellationToken indicates that the value should not be returned nor cached.
+async function getCompletionContextWithCancellation(documentUri: string, caretOffset: number,
+    startTime: number, out: Logger, token: vscode.CancellationToken): Promise<CodeSnippet[]> {
+    try {
+        const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
```
sean-mcmanus: I don't think need you check the active text editor. You should just use the documentUri. The user could type and then switch editors, but the user/Copilot still wants the context for the original document and the current active document shouldn't matter.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+        if (!activeEditor ||
+            activeEditor.document.uri.toString() !== vscode.Uri.parse(documentUri).toString()) {
+            return [];
+        }
+
+        const snippets = await clients.ActiveClient.getCompletionContext(activeEditor.document.uri, caretOffset, token);
```
sean-mcmanus: I think this should call `clients.getClientFor(documentUri)` instead of clients.ActiveClient.

benmcmorran:
@lukka Soon you should be able to take a dependency on https://www.npmjs.com/package/@github/copilot-language-server to get appropriate typings. The types aren't quite there yet, but https://github.com/github/copilot-client/pull/6250 is adding them.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: done!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: done!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: done!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: done, thanks!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export class CopilotCompletionContextProvider implements ContextResolver<CodeSnippet> {
+    private static readonly providerId = 'cppTools';
+    private readonly completionContextCache: Map<string, CodeSnippet[]> = new Map<string, CodeSnippet[]>();
+    private static readonly defaultCppDocumentSelector: DocumentSelector = [{ language: 'cpp' }, { language: 'c' }, { language: 'cuda-cpp' }];
+    private static readonly defaultTimeBudgetFactor: number = 0.5;
+    private completionContextCancellation = new vscode.CancellationTokenSource();
+
+    // Get the default value if the timeout expires, but throws an exception if the token is cancelled.
```
lukka: this is not accurate anymore, needs an update

Extension/src/LanguageServer/copilotContextTelemetry.ts
```diff
@@ -0,0 +1,72 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import { randomUUID } from 'crypto';
+import * as telemetry from '../telemetry';
+
+export class CopilotContextTelemetry {
```
lukka: more appropriate name should be CopilotCompletionContextTelemetry

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+            telemetry.addSnippetCount(codeSnippets?.length);
+            return codeSnippets ?? [];
+        } catch (e: any) {
+            telemetry.addError();
+            throw e;
+        } finally {
+            telemetry.addKind(codeSnippetsKind.toString());
+            const duration: number = performance.now() - startTime;
+            if (codeSnippets === undefined) {
+                out.appendLine(`Copilot: getCompletionContext(): no snkppets provided (${codeSnippetsKind.toString()}), elapsed time (ms): ${duration}`);
```
sean-mcmanus: snkppets->snippets

Extension/src/LanguageServer/copilotContextTelemetry.ts
```diff
@@ -0,0 +1,72 @@
...
+import { randomUUID } from 'crypto';
+import * as telemetry from '../telemetry';
+
+export class CopilotContextTelemetry {
+    private static readonly correlationIdKey = 'correlationId';
+    private static readonly copilotEventName = 'copilotContextProvider';
+    private readonly metrics: Record<string, number> = {};
+    private readonly properties: Record<string, string> = {};
+    private readonly id: string;
+    constructor(correlationId?: string) {
```
sean-mcmanus: There should be a blank newline before `constructor`.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: Can you round this duration to a whole number of ms? e.g. `83` instead of `83.27320000017062`.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+            this.completionContextCache.set(documentUri, codeSnippets);
+            const duration: number = performance.now() - startTime;
+            out.appendLine(`Copilot: getCompletionContextWithCancellation(): Cached in [ms]: ${duration}`);
+            telemetry.addSnippetCount(codeSnippets?.length);
+            telemetry.addCacheComputedElapsed(duration);
+
+            return codeSnippets;
+        } catch (e) {
+            const err = e as Error;
+            out.appendLine(`Copilot: getCompletionContextWithCancellation(): Error: '${err?.message}', stack '${err?.stack}`);
```
sean-mcmanus: It doesn't seem like we should output an error message on Cancel, since Cancelling seems like a non-error. Also, it seems like telemetry would want that tracked differently too.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/copilotContextTelemetry.ts
lukka: I'll add it manually Should this be enforced by whatever the formatter is? I think I have enabled ESLint and formatting the document should be the "correct" one.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+                out.appendLine(`Copilot: getCompletionContext(): no snkppets provided (${codeSnippetsKind.toString()}), elapsed time (ms): ${duration}`);
+            } else {
+                out.appendLine(`Copilot: getCompletionContext(): provided ${codeSnippets?.length} snippets (${codeSnippetsKind.toString()}), elapsed time (ms): ${duration}`);
+            }
+            telemetry.addResolvedElapsed(duration);
+            telemetry.addCacheSize(this.completionContextCache.size);
+            // //?? TODO telemetry.file();
+        }
+
+        return [];
```
sean-mcmanus: I'm seeing this "unreachable code": 

image

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+        } catch (e: any) {
+            telemetry.addError();
+            throw e;
+        } finally {
+            telemetry.addKind(codeSnippetsKind.toString());
+            const duration: number = performance.now() - startTime;
+            if (codeSnippets === undefined) {
+                out.appendLine(`Copilot: getCompletionContext(): no snkppets provided (${codeSnippetsKind.toString()}), elapsed time (ms): ${duration}`);
+            } else {
+                out.appendLine(`Copilot: getCompletionContext(): provided ${codeSnippets?.length} snippets (${codeSnippetsKind.toString()}), elapsed time (ms): ${duration}`);
```
sean-mcmanus: Are you able to log the id of the completion context request here? Or is that not valid here?

sean-mcmanus: Related to this I'm seeing "Copilot: getCompletionContext(): provided 148 snippets (gotFromCacheHit)" but then I see it make a request `LSP: (received) cpptools/getCompletionContext (id: 641)` but I don't understand why it's sending a request if it used the cache.

sean-mcmanus: Is the cache just used to return results faster and it still sends a request to update the cache for later?

Extension/src/LanguageServer/copilotContextTelemetry.ts
sean-mcmanus: I'm not sure which formatter we're using for TypeScript. We have `"editor.defaultFormatter": "vscode.typescript-language-features",` set in settings.json but we also have `"eslint.format.enable": true,` -- I've seen cases recently where the 2 formatters had different opinions and so I had to disable the vscode.typescript-language-features in order to fix the eslint issue.

I think eslint formatting is deprecated.

I'm not sure if there's a formatter rule for a newline before a member function, but feel free to add one if you know of a rule.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: As mentioned elsewhere, feel free to not log the id if it's not easily available. I'm not sure if it's worth sending it in the return message just for logging -- it depends on if there's any case where the nearest server-side message may not match the TypeScript message.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: we should come up with a sensible strategy indeed, I write down a note about it.

Extension/src/LanguageServer/copilotContextTelemetry.ts
lukka: in principle formatting should be good now, let me know if I need to apply any fixes, thanks.

lukka:
> @lukka Soon you should be able to take a dependency on https://www.npmjs.com/package/@github/copilot-language-server to get appropriate typings. The types aren't quite there yet, but [github/copilot-client#6250](https://github.com/github/copilot-client/pull/6250) is adding them.

This is done, thanks!

Extension/package.json
```diff
@@ -6573,6 +6573,7 @@
         "xml2js": "^0.6.2"
     },
     "dependencies": {
+        "@github/copilot-language-server": "^1.253.0",
```
sean-mcmanus, line 6593: @lukka There's a 1.254.0 currently -- should we switch to that one instead. Do you know how important it is to update this? I'm guessing it might be important since it's so new?

sean-mcmanus: Also, unlike most packages, there appears to be no publicly available changelog so I can't tell what's different, i.e. if anything important got fixed/changed.

lukka: There are no changes from my point of view we can keep the existing package.

Extension/src/LanguageServer/settings.ts
```diff
@@ -477,7 +477,9 @@ export class CppSettings extends Settings {
         }
         return val as string;
     }
-
+    public get copilotCompletionFeatureFlag(): string | undefined {
+        return super.Section.get<any>("copilotCompletionFeatureFlag") as string | undefined;
```
sean-mcmanus: @lukka Is this setting supposed to be added to the package.json or it intentionally a "hidden" setting?

lukka: @sean-mcmanus correct, it is not meant to be public :)

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -0,0 +1,309 @@
...
+    StaleCacheHit = 'staleCacheHit',
+    MissingCacheMiss = 'missingCacheMiss',
+    Canceled = 'canceled',
+    Unknown = 'unknown'
+}
+
+type CacheEntry = [string, CopilotCompletionContextResult];
+
+export class CopilotCompletionContextProvider implements ContextResolver<CodeSnippet> {
+    private static readonly providerId = 'cppTools';
```
sean-mcmanus: Where did you get this from? Is this supposed to be the extensionId? Our extension id is "ms-vscode.cpptools" and not "cppTools".

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: @sean-mcmanus correct, it is certainly a good idea to use the extension id as a prefix. Since we may have more than one handler, we may have a custom suffix too, ideas? @benmcmorran @kuchungmsft @spebl

sean-mcmanus (dismissed):

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -37,6 +37,14 @@ import { CppSettings } from './settings';
 import { LanguageStatusUI, getUI } from './ui';
 import { makeLspRange, rangeEquals, showInstallCompilerWalkthrough } from './utils';
 
+export interface SnippetEntry {
```
Colengms: Is there a better place for this interface? Maybe in copilotProviders.ts, if it's specific to features supported there?

Colengms: `client.ts` would be fine, if that's the only place it's used.  (`client.ts` could use cleanup. We had moved all provider classes out of it a while back, but there is still more we could do. :/ )

lukka: Yes, it has been moved to a safe place, ensuring it does not clutter the existing extension.ts content.

sean-mcmanus (dismissed):

Colengms (dismissed):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: @benmcmorran @spebl how difficult is it to grant that the kill switch is ON during the experiment? If I do not keep this switch, the provider will always be registered, which may not be desirable. In principle I'd like to shut everything off if needed, also the provider registration should be avoided in that case.

benmcmorran: If you are concerned that the code paths guarded by that flag could cause stability issues, I'd keep it, but otherwise it's fine to drop the flag.

It's not too hard to configure either way, just means we would need to technically run 2 experiments in parallel, one of which is a dummy experiment at 100% rollout that just sets the cpptools flag.

lukka: dropped the unnecessary feature flag.

benmcmorran (approved):

kuchungmsft (approved):

benmcmorran (dismissed):

kuchungmsft (dismissed):

Extension/src/LanguageServer/settings.ts
```diff
@@ -477,7 +477,9 @@ export class CppSettings extends Settings {
         }
         return val as string;
     }
-
+    public get cppCodeSnippetsFeatureNames(): string | undefined {
+        return super.Section.get<any>("cppCodeSnippetsFeatureNames") as string | undefined;
```
Colengms: Couldn't this pass `string` in place of `any` instead of using `as string | undefined`?

sean-mcmanus (approved):

PR #13047, opened 2024-12-12
Trigger reparsing when `files.encoding` setting changes

### LINKED ISSUE #10892: Changes to `files.encoding` should trigger purge/reparse of all files (without BOM) in DB
Currently, if the `files.encoding` setting is changed, a new parse-root occurs.  That's not sufficient.

The value of `files.encoding` is used to interpret files read directly from disk by the native processes (`cpptools.exe` and `cpptools-srv.exe`).  If a file does not have a BOM and is not detected to be UTF-16, it's assumed to be in the specified encoding.  Files are converted from that encoding to UTF-8.  (VS Code does the same.  It provide buffers for files open in the editor that are already converted to UTF-8).

Either the entire database needs to be reset, or encoding info about the file could be added to the database and only those files that had relied upon the default/fallback encoding could be purged/reparsed.

The last seen encoding setting value should also be persisted.  If the user changes the encoding when the C/C++ extension is not active, the database work still needs to occur.

Colengms:
Fixes: https://github.com/microsoft/vscode-cpptools/issues/10892

This is a simple fix which will reindex the database if `files.encoding` changes. It's possible to do a more granular reparse of only those files that would have been impacted by the change.  A related issue with that may be: https://github.com/microsoft/vscode/issues/209503#issuecomment-2537626506

There is a native-side change as well.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -58,7 +58,7 @@ import { DataBinding } from './dataBinding';
 import { cachedEditorConfigSettings, getEditorConfigSettings } from './editorConfig';
 import { CppSourceStr, clients, configPrefix, updateLanguageConfigurations, usesCrashHandler, watchForCrashes } from './extension';
 import { LocalizeStringParams, getLocaleId, getLocalizedString } from './localization';
-import { PersistentFolderState, PersistentWorkspaceState } from './persistentState';
+import { PersistentFolderState, PersistentState, PersistentWorkspaceState } from './persistentState';
 import { RequestCancelled, ServerCancelled, createProtocolFilter } from './protocolFilter';
 import * as refs from './references';
 import { CppSettings, OtherSettings, SettingsParams, WorkspaceFolderSettingsParams } from './settings';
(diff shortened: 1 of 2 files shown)


sean-mcmanus (dismissed):

sean-mcmanus (approved):

v-frankwang:
@Colengms  We went to verify the issue in the latest C++ extension 1.24.1 (pre-release) according to the following steps, **is the gif showing the expected result?** 

If our validation steps are not correct, **_could you share our detailed validation steps?_** 

**Verification Steps:** 

1. Create a Demo folder in PC. 
2. Open Demo with VSCode, then add a .cpp file and paste the following code: 
```
#include <iostream>
#include <vector>
#include <string>
using namespace std;
int main()
{
 vector<string> msg {"Hello", "C++", "World", "from", "VS Code", "and the C++ extension!"};
    for (const string& word : msg)
    {
        cout << word << " ";
    }
    cout << endl;
}
```
4. Run the '**C/C++: Log Diagnostics**' command in the command panel. 
5. Observe the log info in the output panel, then get current database path. 
6. Navigate to the DB path in PC. 
7. Return to VSCode, then navigate to Settings UI, search for '**Files: Encoding**' 
8. Choose other windows encoding in the workspace tab. 
9. Observe the DB file after returning to the .cpp file. 

**Verification Results**: Changes to files.encoding trigger purge/reparse of files(.BROWSE.VC&.BROWSE.VC.DB-wal) in DB path

PR #13073, opened 2024-12-19
Don't reset everyones databases

Colengms:
A follow up to: https://github.com/microsoft/vscode-cpptools/pull/13047

We shouldn't automatically reset everyone's databases with 1.23.3 due to these persistent values not being set yet.  This change considers empty value as not requiring a reset.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1385,6 +1385,9 @@ export class DefaultClient implements Client {
         let filesEncodingChanged: boolean = false;
         if (workspaceFolder) {
             const lastFilesEncoding: PersistentFolderState<string> = new PersistentFolderState<string>("CPP.lastFilesEncoding", "", workspaceFolder);
+            if (lastFilesEncoding.Value === "") {
+                lastFilesEncoding.Value = filesEncoding;
+            }
             filesEncodingChanged = lastFilesEncoding.Value !== filesEncoding;
         }
         const result: WorkspaceFolderSettingsParams = {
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -1385,6 +1385,9 @@ export class DefaultClient implements Client {
         let filesEncodingChanged: boolean = false;
         if (workspaceFolder) {
             const lastFilesEncoding: PersistentFolderState<string> = new PersistentFolderState<string>("CPP.lastFilesEncoding", "", workspaceFolder);
+            if (lastFilesEncoding.Value === "") {
+                lastFilesEncoding.Value = filesEncoding;
```
bobbrow: Could you just pass the default value in the PersistentFolderState constructor to the current filesEncoding value instead?

Extension/src/LanguageServer/client.ts
```diff
@@ -1520,6 +1523,9 @@ export class DefaultClient implements Client {
         }
         const workspaceFallbackEncoding: string = workspaceOtherSettings.filesEncoding;
         const lastWorkspaceFallbackEncoding: PersistentState<string> = new PersistentState<string>("CPP.lastWorkspaceFallbackEncoding", "");
+        if (lastWorkspaceFallbackEncoding.Value === "") {
+            lastWorkspaceFallbackEncoding.Value = workspaceFallbackEncoding;
```
bobbrow: Same here. Pass the `workspaceFallbackEncoding` as the default value to the constructor.

bobbrow (approved):

Issue #12597, opened 2024-08-20
Description of preferredPathSeparator setting is incorrect


bobbrow:
### Environment

(irrelevant)

### Bug Summary and Steps to Reproduce

We have a setting that supposedly controls the slash behavior for #include auto complete:
![image](https://github.com/user-attachments/assets/51719578-73de-4ed3-a94e-c3e825619913)

The setting description states that it is supposed to be the character used for path separators in the auto complete results. However, it doesn't appear that we add any separators for #include auto complete.  I thought that we used to, but now it seems the setting is only used for paths returned via the code actions.  I believe this setting was intended to cover more than just the #include auto complete case, but for now, I think we need to restore the behavior described in the setting and add separators as part of the #include auto complete feature.

### Configuration and Logs

```shell
n/a
```

browntarik:
Not actionable due to bug that needs to be addressed by VS Code. Until it is addressed (https://github.com/microsoft/vscode/issues/227012), we can't apply a reasonable fix.

bobbrow:
If we're not going to fix this, we should at least fix the setting's description.

PR #13082, opened 2024-12-27
Updated Description for Preferred Path Separator Setting and Refactor Its Usage

browntarik:
Addresses: https://github.com/microsoft/vscode-cpptools/issues/12597 and https://github.com/microsoft/vscode-cpptools/issues/13083


sean-mcmanus (changes requested):
I think it should just be `generated user paths`. There isn't a reason to mention the `Add #include` code action specifically.

sean-mcmanus (approved):

sean-mcmanus (approved):

Issue #12909, opened 2024-10-30
"Developer Command Prompt for VS" should not be translated in our strings


bobbrow:
We need to communicate to the Loc team that "Developer Command Prompt for VS" should not be translated.

RE: #12439

bobbrow:
I didn't look into it that much, but I think the root of this issue is that we need to add notes to the translators with our strings. However, I know that some of them are in the walkthrough or HTML and we might need to update our string scraping scripts to parse out something new to pass that information along.

sean-mcmanus:
Some of the strings got fixed, but I added "hints" to the strings saying to not translate the Developer Command Prompt but they did not follow the hints so I need to file bugs.

v-ericawu:
@sean-mcmanus Verified on C++ 1.24.1(pre-release), this issue still repro:

![Image](https://github.com/user-attachments/assets/85ef1ccd-3e05-4413-9b0c-c8f9555148e7)

bobbrow:
This should be fixed with #13561

PR #13091, opened 2024-12-31
Fix cl.exe.not.available loc.

sean-mcmanus:
Fix for one case of https://github.com/microsoft/vscode-cpptools/issues/12909 .

Remove eslint indentation check which conflicts with the TypeScript formatter (eslint no longer formats).

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -584,7 +584,9 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
     private showErrorIfClNotAvailable(_configurationLabel: string): boolean {
         if (!process.env.DevEnvDir || process.env.DevEnvDir.length === 0) {
-            void vscode.window.showErrorMessage(localize("cl.exe.not.available", "{0} build and debug is only usable when VS Code is run from the Developer Command Prompt for VS.", "cl.exe"));
+            void vscode.window.showErrorMessage(localize({ key: "cl.exe.not.available",
+                comment: ["{0} is a command option in a menu. {1} is the product name \"Developer Command Prompt for VS\"."] },
+                "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
             return true;
         }
         return false;
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -587,8 +587,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
             void vscode.window.showErrorMessage(localize({
                 key: "cl.exe.not.available",
                 comment: ["{0} is a command option in a menu. {1} is the product name \"Developer Command Prompt for VS\"."]
-            },
-                "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
+            }, "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
             return true;
         }
         return false;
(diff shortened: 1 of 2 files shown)


bobbrow (dismissed):
Fix lint and then check it in.

sean-mcmanus:
> Fix lint and then check it in.

I had already run the TypeScript formatter on the file. The linter can't be run as a formatter and conflicts with the formatter. I'm removing the eslint indentation check.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -584,7 +584,11 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
     private showErrorIfClNotAvailable(_configurationLabel: string): boolean {
         if (!process.env.DevEnvDir || process.env.DevEnvDir.length === 0) {
-            void vscode.window.showErrorMessage(localize("cl.exe.not.available", "{0} build and debug is only usable when VS Code is run from the Developer Command Prompt for VS.", "cl.exe"));
+            void vscode.window.showErrorMessage(localize({
+                key: "cl.exe.not.available",
+                comment: ["{0} is a command option in a menu. {1} is the product name \"Developer Command Prompt for VS\"."]
+            },
+                "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
```
bobbrow: Since it won't indent the lines above due to the placement of the `{`, you should move this line right after the comma or move the `{` to a new line (I saw some cases in our code that do that).

```ts
            void vscode.window.showErrorMessage(localize({
                key: "cl.exe.not.available",
                comment: ["{0} is a command option in a menu. {1} is the product name \"Developer Command Prompt for VS\"."]
            }, "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
```

or 

```ts
            void vscode.window.showErrorMessage(localize(
                {
                    key: "cl.exe.not.available",
                    comment: ["{0} is a command option in a menu. {1} is the product name \"Developer Command Prompt for VS\"."]
                },
                "{0} is only usable when VS Code is run from the {1}.", `cl.exe ${this.buildAndDebugActiveFileStr()}`, "Developer Command Prompt for VS"));
```

bobbrow: This is better than turning off linting, I think.

sean-mcmanus: @bobbrow I disagree. I think we should just rely on our TypeScript formatter instead of trying to manually adjust the formatting.

The formatting is from

settings.json
```cpp
    "[typescript]": {
        "editor.tabSize": 4,
        "editor.defaultFormatter": "vscode.typescript-language-features",
        "editor.formatOnSave": true,
```

Otherwise, whenever I save the file, it'll format and conflict with the eslint rule, so we have to temporarily disable the auto format on save, which is annoying and confuses people.

bobbrow: Either way, you should pick one of the two options I presented. The way it's currently formatted looks bad.

sean-mcmanus: Oh, well, it looks like I can adjust the formatting and it still works, but I think we should still disabling the eslint check anyway, since it has been a problem in the past when the TypeScript formatter rules conflict with the eslint rules.

bobbrow: I don't have a strong opinion on removing the eslint check if format on save is enabled.

bobbrow: ...but the problem is that some people might disable formatOnSave and now you don't have automatic verification of the formatting in the PR anymore. If you don't think that's going to be a problem, we can remove the eslint check.

sean-mcmanus: The settings.json is checked in with the formatOnSave enabled for TypeScript, so if they change it locally it's their fault and reverting to the checked in settings.json would fix the issue. The cpptools-side has the same behavior -- there's no linter for formatting. Linting for formatting is fine with me, but it would need to use the same formatter otherwise there's a conflict and the eslint formatter is deprecated/disabled.

bobbrow (approved):

Extension/src/Debugger/configurationProvider.ts
berki75: "[typescript]": {
        "editor.tabSize": 4,
        "editor.defaultFormatter": "vscode.typescript-language-features",
        "editor.formatOnSave": true,

PR #13132, opened 2025-01-08
Support C89 in Completion Prompts

kuchungmsft:
- Also send unrecognized ones in telementry for further ananlysis.

### DIFF
--- a/Extension/src/LanguageServer/lmTool.ts
+++ b/Extension/src/LanguageServer/lmTool.ts
@@ -38,7 +38,7 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
         'c++17': 'C++17',
         'c++20': 'C++20',
         'c++23': 'C++23',
-        'c90': "C90",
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/lmTool.ts
```diff
@@ -38,6 +38,7 @@ const knownValues: { [Property in keyof ChatContextResult]?: { [id: string]: str
         'c++17': 'C++17',
         'c++20': 'C++20',
         'c++23': 'C++23',
+        'c89': "C89",
         'c90': "C90",
```
bobbrow: According to [wikipedia](https://en.wikipedia.org/wiki/ANSI_C), C90 was withdrawn from the ANSI standard. We don't accept it as a version in the language server.  Just as FYI...

kuchungmsft: Removed, thanks

sean-mcmanus (approved):

PR #13143, opened 2025-01-13
Support Content Exclusion for Copilot Hover

spebl:
Get the files used for the context generation and bail before calling the LM if they're excluded.


sean-mcmanus (commented):
Should the feature be enabled now? I thought this was the only blocking issue?

Extension/src/LanguageServer/extension.ts
```diff
@@ -1430,7 +1430,16 @@ async function onCopilotHover(): Promise<void> {
 
     // Gather the content for the query from the client.
     const requestInfo = await copilotHoverProvider.getRequestInfo(hoverDocument, hoverPosition);
-    if (requestInfo.length === 0) {
+    for (const file of requestInfo.files) {
+        const fileUri = vscode.Uri.file(file);
+        if (await vscodelm.fileIsIgnored(fileUri, copilotHoverProvider.getCurrentHoverCancellationToken() ?? CancellationToken.None)) {
+            // Context is not available for this file.
+            telemetry.logLanguageServerEvent("CopilotHover", { "Message": "Copilot summary is not available for this file." });
+            await showCopilotContent(copilotHoverProvider, hoverDocument, hoverPosition, localize("copilot.hover.unavailable", "Copilot summary is not available for this file."));
```
sean-mcmanus: It seems like "this file" isn't correct from the perspective of the user, because I think they would assume it's the current file when the error/exclusion is actually applying to the file with the definition, which the user may not know what file that is. Should the error says something like "for the file containing the declaration or definition"? And/or should the location of the file be shown or should the user just use go to definition if they want to find that?

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -38,7 +38,10 @@
         "Snippets"
     ],
     "enabledApiProposals": [
-        "terminalDataWriteEvent"
+        "terminalDataWriteEvent",
+        "chatVariableResolver",
```
sean-mcmanus: I see these errors when I open the package.json file: 
>This proposal cannot be used because for this extension the product defines a fixed set of API proposals. You can test your extension but before publishing you MUST reach out to the VS Code team.

Is that an issue or what?

spebl: I've reached out to the vscode team (Isidor) to get that approval. Planning to wait for that approval before pushing.

spebl:
> Should the feature be enabled now? I thought this was the only blocking issue?

We're still waiting for final sign-off before rolling it out. @benmcmorran has been tracking the privacy approval which is the last step.

Extension/src/LanguageServer/extension.ts
spebl: Updated to be more specific about copilot exclusion so it's more obvious to the user what the cause is.
image

sean-mcmanus: I believe this localize key needs to be different from the other one below: copilot.hover.unavailable

sean-mcmanus (changes requested):

benmcmorran:
@sean-mcmanus Not sure exactly when you were planning to do the 1.23 official release, but assuming we get the approval from the VS Code team to use the proposed API soon, is it feasible to ensure this PR makes it into the release? That will put us in a position to enable for the GA audience once all sign-offs are complete.

benmcmorran:
Discussed offline with @Colengms and it sounds like that timeline may not be feasible. It's fine to hold this for the next insiders release instead.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1430,18 +1430,28 @@ async function onCopilotHover(): Promise<void> {
...
     // Gather the content for the query from the client.
     const requestInfo = await copilotHoverProvider.getRequestInfo(hoverDocument, hoverPosition);
-    if (requestInfo.length === 0) {
+    for (const file of requestInfo.files) {
+        const fileUri = vscode.Uri.file(file);
+        if (await vscodelm.fileIsIgnored(fileUri, copilotHoverProvider.getCurrentHoverCancellationToken() ?? CancellationToken.None)) {
+            // Context is not available for this file.
+            telemetry.logLanguageServerEvent("CopilotHover", { "Message": "Copilot summary is not available for definition or declaration." });
+            await showCopilotContent(copilotHoverProvider, hoverDocument, hoverPosition, localize("copilot.hover.unavailable", "Copilot summary is not available.") + "\n\n" +
+                localize("copilot.hover.excluded", "The file containing this symbol's defintion or declaration has been excluded from use with Copilot."));
```
sean-mcmanus: defintion->definition

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1430,18 +1430,28 @@ async function onCopilotHover(): Promise<void> {
 
     // Gather the content for the query from the client.
     const requestInfo = await copilotHoverProvider.getRequestInfo(hoverDocument, hoverPosition);
-    if (requestInfo.length === 0) {
+    for (const file of requestInfo.files) {
+        const fileUri = vscode.Uri.file(file);
+        if (await vscodelm.fileIsIgnored(fileUri, copilotHoverProvider.getCurrentHoverCancellationToken() ?? CancellationToken.None)) {
+            // Context is not available for this file.
+            telemetry.logLanguageServerEvent("CopilotHover", { "Message": "Copilot summary is not available for definition or declaration." });
```
sean-mcmanus: Are you sure this is what the message should be? The "for definition or declaration" doesn't seem right to me, since the summary is always for the definition or declaration.

spebl: I can update it, I wanted to make sure it was unique so we can trace the telemetry back to the cause of failure. I agree that it makes more sense to explicitly log the fact that we bailed due to content exclusion.

spebl: Updated

sean-mcmanus (changes requested):
Did the SSH build fail from these changes or is that some unrelated issue?

spebl:
> Did the SSH build fail from these changes or is that some unrelated issue?

These changes, about to push an update.

Extension/src/LanguageServer/extension.ts
```diff
@@ -1430,18 +1430,27 @@ async function onCopilotHover(): Promise<void> {
 
     // Gather the content for the query from the client.
     const requestInfo = await copilotHoverProvider.getRequestInfo(hoverDocument, hoverPosition);
-    if (requestInfo.length === 0) {
+    for (const file of requestInfo.files) {
+        const fileUri = vscode.Uri.file(file);
+        if (await vscodelm.fileIsIgnored(fileUri, copilotHoverProvider.getCurrentHoverCancellationToken() ?? CancellationToken.None)) {
```
benmcmorran: Given that this is technically still a proposed API (even if it is expected to be stable), we should wrap the call to `fileIsIgnored()` in a try/catch.

sean-mcmanus (dismissed):

Extension/src/LanguageServer/extension.ts
spebl: Added

Extension/src/LanguageServer/extension.ts
```diff
@@ -1430,18 +1430,34 @@ async function onCopilotHover(): Promise<void> {
...
+            if (await vscodelm.fileIsIgnored(fileUri, copilotHoverProvider.getCurrentHoverCancellationToken() ?? CancellationToken.None)) {
+                telemetry.logLanguageServerEvent("CopilotHover", { "Message": "Copilot summary is not available due to content exclusion." });
+                await showCopilotContent(copilotHoverProvider, hoverDocument, hoverPosition, localize("copilot.hover.unavailable", "Copilot summary is not available.") + "\n\n" +
+                    localize("copilot.hover.excluded", "The file containing this symbol's definition or declaration has been excluded from use with Copilot."));
+                return;
+            }
+        }
+    } catch (err) {
+        if (err instanceof Error) {
+            await reportCopilotFailure(copilotHoverProvider, hoverDocument, hoverPosition, err.message);
```
sean-mcmanus: Could this err.message contain PII, such as the user name and paths?

Extension/package.json
spebl: Updated to just require the one necessary proposal, and have received the approval for it. Just waiting to be put on the allow list.

Extension/src/LanguageServer/extension.ts
spebl: Good call that logging any generic error could have that issue. Perhaps something like "Copilot Hover Error: " + err.name? Should be safe from PII but it would be good to be able to track what type of errors are hit.

sean-mcmanus: Yeah, I used err.name in my PR at https://github.com/microsoft/vscode-cpptools/pull/13158 .

sean-mcmanus: I wasn't sure how to determine if vscode.LanguageModelError should be handled (which has `code`) versus Error (which only has `name`).

sean-mcmanus: The telemetry already is sent via `telemetry.logLanguageServerEvent("CopilotHoverError", { "ErrorMessage": ` so adding ` "Copilot Hover Error: "` doesn't seem to add any info.

sean-mcmanus (approved):

PR #13206, opened 2025-01-28
Fix an undefined access when an edit is done before cpptools starts

sean-mcmanus:
I reproed it with 1.23.4, but not 1.22.11, so it seems like a regression, but I don't know from what change.

image

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -1839,11 +1839,11 @@ export class DefaultClient implements Client {
         await this.languageClient.sendNotification(DidChangeTextEditorVisibleRangesNotification, params);
     }
 
-    public async onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): Promise<void> {
-        await this.ready;
+    public onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): void {
         if (util.isCpp(textDocumentChangeEvent.document)) {
             // If any file has changed, we need to abort the current rename operation
-            if (workspaceReferences.renamePending) {
+            if (workspaceReferences !== undefined // Occurs when a document changes before cpptools starts.
+                && workspaceReferences.renamePending) {
                 workspaceReferences.cancelCurrentReferenceRequest(refs.CancellationSender.User);
             }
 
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -1842,7 +1842,8 @@ export class DefaultClient implements Client {
     public onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): void {
         if (util.isCpp(textDocumentChangeEvent.document)) {
             // If any file has changed, we need to abort the current rename operation
-            if (workspaceReferences.renamePending) {
+            if (workspaceReferences !== undefined // Occurs when a document changes before cpptools starts.
+                && workspaceReferences.renamePending) {
```
sean-mcmanus, line 1846: Should I use `?.` instead?

sean-mcmanus: I think it's better with an explicit check. Otherwise, I'd have to make the comment longer.

Extension/src/LanguageServer/client.ts
```diff
@@ -1842,7 +1842,8 @@ export class DefaultClient implements Client {
     public onDidChangeTextDocument(textDocumentChangeEvent: vscode.TextDocumentChangeEvent): void {
         if (util.isCpp(textDocumentChangeEvent.document)) {
             // If any file has changed, we need to abort the current rename operation
-            if (workspaceReferences.renamePending) {
+            if (workspaceReferences !== undefined // Occurs when a document changes before cpptools starts.
```
Colengms, line 1845: It looks like `onDidChangeTextDocument` is registered in `activate` (in `extension.ts`), after clients have been created and `init` called. But `init` is async and performs an await before setting `workspaceReferences`.

Could you maybe add an `await this.ready;` instead?  It's possible all of the handlers installed (early) in `activate` should also `await this.ready;`, unless there is some reason they should run before `init` is complete.

sean-mcmanus: ok

sean-mcmanus: But I'm not sure what other handlers should have it added too -- they all seem okay to me.

Colengms (dismissed):

Colengms (dismissed):

Colengms (approved):

PR #13208, opened 2025-01-28
Update changelog and version for 1.23.5.

sean-mcmanus:


sean-mcmanus:
Does anyone see a bug fix that only reproed with 1.23.x and should be removed?

Colengms (dismissed):

Colengms (approved):

PR #13209, opened 2025-01-28
Fix a potential race between `didChange` and `didOpen`

Colengms:
Fixes a potential (probably rare?) issue where it's possible that a call to `onDidChangeTextDocument` might happen before a call to `onDidOpenTextDocument`, due to didOpen being delayed (processed by the protocol filter, after prior messages in queue). This could result the file's version in `openFileVersions` becoming incorrect (though would be corrected by subsequent edits).

`onDidOpenTextDocument` and `onDidChangeTextDocument` primarily maintain `openFileVersions`, and update UI state.  This change calls `onDidOpenTextDocument` using a synchronous handler, instead of being processed by the protocol filter, ensuring no awaiting or prior queue work will delay it, similar to what we were already doing with `onDidChangeTextDocument`.

Having `openFileVersions` kept as closely in sync as possible with the editor ensures we properly discard results for operations that shouldn't be applied if the document has since changed.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -172,12 +172,11 @@ export async function activate(): Promise<void> {
         getCustomConfigProviders().forEach(provider => void client.onRegisterCustomConfigurationProvider(provider));
     });
 
-    // These handlers for didChangeTextDocument and didOpenTextDocument are intentionally non-async and are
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
This is for 1.24.0 or 1.23.5?

sean-mcmanus (approved):

Issue #12668, opened 2024-08-30
cpptools-srv crashes sometimes during handle_initialize in uv_run on Linux with C_Cpp.intelliSenseCacheSize > 0


sean-mcmanus:
In the same scenario on Mac, I'm seeing cpptools get stuck, but it might be the same root cause (otherwise, I'll create a new issue for that later). UPDATE: That "stuck" issue is fixed now and was unrelated.

sean-mcmanus:
In certain cases (on certain machines), the crash is not rare, but repros 99.9% of the time.

sean-mcmanus:
Our repro case stopped reproing, but we're not sure if it's fixed or not.

sean-mcmanus:
@Colengms This seems to still repro (and at the same spot).

sean-mcmanus:
It's caused by an invalid mmap call, so setting `"C_Cpp.intelliSenseCacheSize": 0` should be a workaround.

PR #13253, opened 2025-02-07
Add crash log handling.

sean-mcmanus:
Add handling of logging intended to help diagnose the causes of specific crashes in cpptools or cpptools-srv.

Similar crash logging was used internally to find the root cause of the crash at https://github.com/microsoft/vscode-cpptools/issues/12668 .


sean-mcmanus:
This won't be ready till Tuesday probably.

fearthecowboy (dismissed):

Colengms (approved):

PR #13270, opened 2025-02-11
Add a changelog entry.

sean-mcmanus:


heartacker:
changelog is so large， within all versions

Colengms (approved):

PR #13278, opened 2025-02-13
Remove Compiler Argument Traits

kuchungmsft:
- Couldn't find examples of where this actually improved the prompt per ad-hoc analysis.

### DIFF
--- a/Extension/src/LanguageServer/lmTool.ts
+++ b/Extension/src/LanguageServer/lmTool.ts
@@ -72,43 +72,43 @@ export interface ProjectContext {
 
 export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, cancellationToken: vscode.CancellationToken, telemetryProperties: Record<string, string>, telemetryMetrics: Record<string, number>): Promise<ProjectContext | undefined> {
     try {
-        const projectContext = await checkDuration<ChatContextResult | undefined>(async () => await getClients()?.ActiveClient?.getChatContext(uri, cancellationToken) ?? undefined);
-        telemetryMetrics["projectContextDuration"] = projectContext.duration;
-        if (!projectContext.result) {
+        const chatContext = await checkDuration<ChatContextResult | undefined>(async () => await getClients()?.ActiveClient?.getChatContext(uri, cancellationToken) ?? undefined);
+        telemetryMetrics["projectContextDuration"] = chatContext.duration;
+        if (!chatContext.result) {
             return undefined;
         }
 
-        const originalStandardVersion = projectContext.result.standardVersion;
+        const originalStandardVersion = chatContext.result.standardVersion;
 
-        formatChatContext(projectContext.result);
+        formatChatContext(chatContext.result);
 
         const result: ProjectContext = {
-            language: projectContext.result.language,
-            standardVersion: projectContext.result.standardVersion,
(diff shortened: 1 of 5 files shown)


sean-mcmanus (changes requested):
Didn't we see that the standard version improved results? Is that already sent in a different code path?

kuchungmsft:
> Didn't we see that the standard version improved results? Is that already sent in a different code path?

Yes, language standard version trait was shipped a while ago in #12979 and it's not affected in this PR.

image

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -68,75 +68,11 @@ export interface ProjectContext {
...
-        telemetryProperties["filters"] = filters;
-    }
-
-    return result;
-}
-
-export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, telemetryProperties: Record<string, string>, telemetryMetrics: Record<string, number>): Promise<ProjectContext | undefined> {
-    try {
-        const projectContext = await checkDuration<ProjectContextResult | undefined>(async () => await getClients()?.ActiveClient?.getProjectContext(uri) ?? undefined);
+        const projectContext = await checkDuration<ChatContextResult | undefined>(async () => await getClients()?.ActiveClient?.getChatContext(uri, cancellationToken) ?? undefined);
```
sean-mcmanus: Should the `projectContext` naming be changed? You changed it from a ProjectContextResult to a ChatContextResult, but the name of the variable is still projectContext.

sean-mcmanus: I'm confused by the project context versus ChatContext naming -- are they the same thing now? Is there any way to simplify the naming?

kuchungmsft: Fixed the naming. Yes, without compiler argument information, they are identical. I have a PR to remove that API on the other side.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -68,115 +68,48 @@ export interface ProjectContext {
...
-    return result;
-}
-
-export async function getProjectContext(uri: vscode.Uri, context: { flags: Record<string, unknown> }, telemetryProperties: Record<string, string>, telemetryMetrics: Record<string, number>): Promise<ProjectContext | undefined> {
-    try {
-        const projectContext = await checkDuration<ProjectContextResult | undefined>(async () => await getClients()?.ActiveClient?.getProjectContext(uri) ?? undefined);
-        telemetryMetrics["projectContextDuration"] = projectContext.duration;
-        if (!projectContext.result) {
+        const chatContext = await checkDuration<ChatContextResult | undefined>(async () => await getClients()?.ActiveClient?.getChatContext(uri, cancellationToken) ?? undefined);
+        telemetryMetrics["projectContextDuration"] = chatContext.duration;
```
sean-mcmanus, line 76: Is this telemetry supposed to be projectContextDuration still?

sean-mcmanus (approved):

Extension/src/LanguageServer/lmTool.ts
kuchungmsft: Yes, no need to change this metric even though we call a different implementation. Thanks.

sean-mcmanus (approved):

PR #13285, opened 2025-02-15
test framework traits for Chat

lukka:


Extension/src/LanguageServer/lmTool.ts
```diff
@@ -235,7 +235,10 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
                 contextString += `The project targets the ${chatContext.targetArchitecture} architecture. `;
                 telemetryProperties["targetArchitecture"] = chatContext.targetArchitecture;
             }
-
```
sean-mcmanus, line 171: FYI, it seems strange/unexpected that this would only work if a C/C++ file is active, since I'm using #cpp in the side chat there's no indication that it only works when a C/C++ file is active in the UI, so the user doesn't know.

sean-mcmanus: image

The user has no feedback saying that #cpp is not working as expected.

sean-mcmanus: Oh, Copilot will complain about the active not being C++, so that is feedback.

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -235,7 +235,10 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
                 contextString += `The project targets the ${chatContext.targetArchitecture} architecture. `;
                 telemetryProperties["targetArchitecture"] = chatContext.targetArchitecture;
             }
-
+            if (chatContext.usedTestFrameworks?.length > 0) {
```
sean-mcmanus: I don't think `?.` is needed, right? It's not used for the other members and it seems like it would always be defined.

image

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -235,7 +235,10 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
                 contextString += `The project targets the ${chatContext.targetArchitecture} architecture. `;
                 telemetryProperties["targetArchitecture"] = chatContext.targetArchitecture;
             }
-
+            if (chatContext.usedTestFrameworks?.length > 0) {
+                contextString += `The active document contains the content of the file '${currentDoc.fileName}' which is located in a directory structure where other files are including headers for the following C++ test frameworks: ${chatContext.usedTestFrameworks.join(', ')} . `;
```
sean-mcmanus: Can this string be simplified to something like "The project uses these test frameworks: "?

sean-mcmanus: There is an unnecessary space before and after the `.`

Extension/src/LanguageServer/lmTool.ts
```diff
@@ -235,7 +235,10 @@ export class CppConfigurationLanguageModelTool implements vscode.LanguageModelTo
                 contextString += `The project targets the ${chatContext.targetArchitecture} architecture. `;
                 telemetryProperties["targetArchitecture"] = chatContext.targetArchitecture;
             }
-
+            if (chatContext.usedTestFrameworks?.length > 0) {
+                contextString += `The active document contains the content of the file '${currentDoc.fileName}' which is located in a directory structure where other files are including headers for the following C++ test frameworks: ${chatContext.usedTestFrameworks.join(', ')} . `;
+                telemetryProperties["testFrameworks"] = chatContext.usedTestFrameworks.join(',');
```
sean-mcmanus: Your join here is missing a space between the `,` -- is that intentional?

Extension/src/LanguageServer/lmTool.ts
lukka: @sean-mcmanus that is by design as of today. Should it be rethought?
@benmcmorran FYI

Extension/src/LanguageServer/lmTool.ts
lukka: @sean-mcmanus let's discuss offline about this.

lukka: @sean-mcmanus the space after the dot it is intentional as the constructions is elsewhere. The one before, either way works for me, I'll probably remove it.

Extension/src/LanguageServer/lmTool.ts
sean-mcmanus: Okay, I figured it out -- there is additional info on what `#cpp` does in the UI that I never noticed before: 
image

So, yeah, it appears "by design".

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
```diff
@@ -154,7 +154,8 @@ describe('CppConfigurationLanguageModelTool Tests', () => {
                 standardVersion: 'c++20',
                 compiler: 'msvc',
                 targetPlatform: 'windows',
-                targetArchitecture: 'x64'
+                targetArchitecture: 'x64',
+                usedTestFrameworks: ['gtest', 'catch2']
```
sean-mcmanus: Should these actually match the casing of what we send? e.g. "GTest", "Catch2"?

Extension/src/LanguageServer/lmTool.ts
lukka: You are right, the check for null/undefined is not needed. Note that all other members are being implicitly checked for null/undefined in the condition of their `if` statement.

Extension/test/scenarios/SingleRootProject/tests/lmTool.test.ts
lukka: They do not need to be, since they are arbitrary strings used in the tests. I do agree that they should have sensible names, hence fixed as suggested.

sean-mcmanus (approved):

PR #13296, opened 2025-02-20
ContextProvider resolve handler improvements

lukka:
Changes:
-new feature: add traits for C++ lang version
-telemetry: fix cancellation events.
-telemetry: more diag event for registration failure.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -22,6 +22,7 @@ import { SemanticToken, SemanticTokensProvider } from './Providers/semanticToken
 import { WorkspaceSymbolProvider } from './Providers/workspaceSymbolProvider';
 // End provider imports
 
+import { SupportedContextItem } from '@github/copilot-language-server';
 import { ok } from 'assert';
 import * as fs from 'fs';
 import * as os from 'os';
@@ -55,7 +56,7 @@ import {
 } from './codeAnalysis';
 import { Location, TextEdit, WorkspaceEdit } from './commonTypes';
 import * as configs from './configurations';
-import { CopilotCompletionContextFeatures, CopilotCompletionContextProvider, SnippetEntry } from './copilotCompletionContextProvider';
+import { CopilotCompletionContextFeatures, CopilotCompletionContextProvider } from './copilotCompletionContextProvider';
 import { DataBinding } from './dataBinding';
 import { cachedEditorConfigSettings, getEditorConfigSettings } from './editorConfig';
 import { CppSourceStr, clients, configPrefix, updateLanguageConfigurations, usesCrashHandler, watchForCrashes } from './extension';
@@ -568,11 +569,13 @@ interface FilesEncodingChanged {
 
 export interface CopilotCompletionContextResult {
     requestId: number;
-    isResultMissing: boolean;
-    snippets: SnippetEntry[];
+    areCodeSnippetsMissing: boolean;
+    snippets: SupportedContextItem[];
     translationUnitUri: string;
     caretOffset: number;
     featureFlag: CopilotCompletionContextFeatures;
+    codeSnippetsCount: number;
+    traitsCount: number;
 }
 
 export interface CopilotCompletionContextParams {
--- a/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
+++ b/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All Rights Reserved.
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-import { CodeSnippet, ContextResolver, ResolveRequest } from '@github/copilot-language-server';
+import { ContextResolver, ResolveRequest, SupportedContextItem } from '@github/copilot-language-server';
 import { randomUUID } from 'crypto';
 import * as vscode from 'vscode';
 import { DocumentSelector } from 'vscode-languageserver-protocol';
@@ -12,15 +12,8 @@ import { CopilotCompletionContextResult } from './client';
 import { CopilotCompletionContextTelemetry } from './copilotCompletionContextTelemetry';
 import { getCopilotApi } from './copilotProviders';
 import { clients } from './extension';
+import { ProjectContext } from './lmTool';
 import { CppSettings } from './settings';
-
-export interface SnippetEntry {
-    uri: string;
-    value: string;
-    startLine: number;
-    endLine: number;
-    importance: number;
-}
 
 class DefaultValueFallback extends Error {
     static readonly DefaultValue = "DefaultValue";
@@ -30,6 +23,14 @@ class DefaultValueFallback extends Error {
 class CancellationError extends Error {
     static readonly Canceled = "Canceled";
     constructor() { super(CancellationError.Canceled); }
+}
+
+class InternalCancellationError extends Error {
+    static readonly Canceled = "CpptoolsCanceled";
+    constructor() { super(InternalCancellationError.Canceled); }
+}
+
+class CopilotCancellationError extends CancellationError {
 }
 
 class CopilotContextProviderException extends Error {
@@ -43,6 +44,7 @@ class WellKnownErrors extends Error {
     }
 }
 
+// A bit mask for enabling features in the completion context.
 export enum CopilotCompletionContextFeatures {
     None = 0,
     Instant = 1,
@@ -68,7 +70,7 @@ export enum CopilotCompletionKind {
 
 type CacheEntry = [string, CopilotCompletionContextResult];
 
-export class CopilotCompletionContextProvider implements ContextResolver<CodeSnippet> {
+export class CopilotCompletionContextProvider implements ContextResolver<SupportedContextItem> {
     private static readonly providerId = 'ms-vscode.cpptools';
     private readonly completionContextCache: Map<string, CacheEntry> = new Map();
     private static readonly defaultCppDocumentSelector: DocumentSelector = [{ language: 'cpp' }, { language: 'c' }, { language: 'cuda-cpp' }];
@@ -79,16 +81,16 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
     private contextProviderDisposable: vscode.Disposable | undefined;
 
     private async waitForCompletionWithTimeoutAndCancellation<T>(promise: Promise<T>, defaultValue: T | undefined,
-        timeout: number, token: vscode.CancellationToken): Promise<[T | undefined, CopilotCompletionKind]> {
+        timeout: number, copilotToken: vscode.CancellationToken): Promise<[T | undefined, CopilotCompletionKind]> {
         const defaultValuePromise = new Promise<T>((_resolve, reject) => setTimeout(() => {
-            if (token.isCancellationRequested) {
+            if (copilotToken.isCancellationRequested) {
                 reject(new CancellationError());
             } else {
                 reject(new DefaultValueFallback());
             }
         }, timeout));
         const cancellationPromise = new Promise<T>((_, reject) => {
-            token.onCancellationRequested(() => {
+            copilotToken.onCancellationRequested(() => {
                 reject(new CancellationError());
             });
         });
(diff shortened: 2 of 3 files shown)


sean-mcmanus (approved):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -123,8 +129,19 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
             const client = clients.getClientFor(docUri);
             if (!client) { throw WellKnownErrors.clientNotFound(); }
             const getCompletionContextStartTime = performance.now();
+
+            // Start collection of project traits concurrently with the completion context computation.
+            const telemetryProperties: Record<string, string> = {};
+            const telemetryMetrics: Record<string, number> = {};
+            const projectContextPromise = (async (): Promise<ProjectContext | undefined> => {
+                const getProjectContextStartTime = performance.now();
+                const projectContext = await getProjectContext(docUri, { flags: {} }, telemetryProperties, telemetryMetrics);
```
benmcmorran: I think it's fine to pass a hardcoded `flags: {}` here, but just wanted to double check that 1) that will pass only the language standard trait and 2) we don't expect to need to experiment with any other configuration for traits before we ship a new pre-release version. I believe both of those conditions are true.

benmcmorran (dismissed):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -281,7 +303,12 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
             return copilotCompletionContext?.snippets ?? [];
         } catch (e: any) {
             if (e instanceof CancellationError) { throw e; }
-
+            // A new incoming snippet request from Copilot arrived and this request
+            // is now obsolete, give up immediately with no exception.
+            if (e instanceof InternalCancellationError) {
+                logMessage += `, (internal cancellation) `;
```
sean-mcmanus: FYI, there's a pre-existing issue with the various logMessage's ending with `:` or `: ` and then having a `, (internal cancellation)` or other `,`  message added. Ideally, it would not log `:,` or `: ,`

Extension/src/LanguageServer/client.ts
```diff
@@ -581,7 +582,7 @@ interface FilesEncodingChanged {
 export interface CopilotCompletionContextResult {
     requestId: number;
     isResultMissing: boolean;
-    snippets: SnippetEntry[];
+    snippets: SupportedContextItem[];
```
sean-mcmanus, line 573: I'm seeing the actual data still be a SnippetEntry. Does that matter? i.e. the SupportedContextItem doesn't seem to have entries like startLine.

```
export interface SnippetEntry {
    uri: string;
    value: string;
    startLine: number;
    endLine: number;
    importance: number;
}
```

sean-mcmanus (changes requested):
Should SnippetEntry be removed? It doesn't appear to be referenced anymore?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -271,7 +293,7 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
             // Handle cancellation.
             if (copilotCompletionContextKind === CopilotCompletionKind.Canceled) {
                 const duration: number = CopilotCompletionContextProvider.getRoundedDuration(resolveStartTime);
-                telemetry.addInternalCanceled(duration);
+                telemetry.addCopilotCanceled(duration);
                 throw new CancellationError();
```
sean-mcmanus: Is this cancellation supposed to appear as an error in the console? It makes it look like an error:

image

lukka: @sean-mcmanus that's correct, it looks like an error (although harmless) and it is polluting the log. The root cause is that when the Copilot extension is cancelling the request for CodeSnippets, it wants to have back an empty collection, rather than an CancelledError exception. The owners have been informed and will fix to not log an error entry for an operation intentionally cancelled.

kuchungmsft:
FYI, I have a PR opened to clean up the compiler argument traits, which may impact this PR, see https://github.com/microsoft/vscode-cpptools/pull/13278.

sean-mcmanus:
> FYI, I have a PR opened to clean up the compiler argument traits, which may impact this PR, see #13278.

Should your PR get checked in first, right?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: Should this be cached? It seems strange to re-request this on every completion. Wasn't the previous related files version cached for like 2 minutes or something like that?

kuchungmsft (dismissed):

sean-mcmanus (changes requested):
cppCodeSnippetsFeatureNames doesn't work if it's lowercased -- that seems a little error prone and unexpected, because the other C_Cpp settings values are case insensitive I believe.

sean-mcmanus (commented):
Oh, but that doesn't appear to be a regression -- it looks like I modified the setting value to be lowercase and expected it to work.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -247,7 +281,12 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
             const featureFlag: CopilotCompletionContextFeatures | undefined = await this.getEnabledFeatureFlag(context);
             telemetry.addRequestMetadata(context.documentContext.uri, context.documentContext.offset,
                 context.completionId, context.documentContext.languageId, { featureFlag, timeBudgetFactor, maxCaretDistance });
-            if (featureFlag === undefined) { return []; }
+            if (featureFlag === undefined) { throw new Error("No feature flag set!"); }
```
sean-mcmanus: Why is an error thrown and logged in telemetry as an error when no feature flag is set? Aren't some users expected to not be in the experiment and not have a flag set?

sean-mcmanus (changes requested):
getEnabledFeatureNames splits by "," and breaks if ", " with a space after the comma is used -- that seems error prone. Should it be changed to handle space after a comma?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: i.e. is it incorrect to log it as an error in telemetry?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -306,6 +363,8 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
         try {
             const copilotApi = await getCopilotApi();
             if (!copilotApi) { throw new CopilotContextProviderException("getCopilotApi() returned null."); }
+            const hasGetContextProviderAPI = "getContextProviderAPI" in copilotApi;
+            if (!hasGetContextProviderAPI) { throw new CopilotContextProviderException("getContextProviderAPI() is not available."); }
```
sean-mcmanus, line 349: Why is the existing code doing a console.warn whenever the Copilot extension is not enabled/active? That seems incorrect. We've gotten complaints about logging warning/errors in non-error conditions.

image

Shouldn't not having the Copilot extension enabled be considered a normal case?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -247,7 +281,12 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
             const featureFlag: CopilotCompletionContextFeatures | undefined = await this.getEnabledFeatureFlag(context);
             telemetry.addRequestMetadata(context.documentContext.uri, context.documentContext.offset,
                 context.completionId, context.documentContext.languageId, { featureFlag, timeBudgetFactor, maxCaretDistance });
-            if (featureFlag === undefined) { return []; }
+            if (featureFlag === undefined) { throw new Error("No feature flag set!"); }
+            // eslint-disable-next-line no-bitwise
+            if (featureFlag & CopilotCompletionContextFeatures.AlwaysFallbackToSimilarFiles) {
+                telemetry.addCompletionContextKind(CopilotCompletionKind.AlwaysFallbackToSimilarFiles);
+                throw new Error("Enable SimilarFiles");
```
sean-mcmanus: Same as the 0 flags case -- are you sure you want to log this case as an error in telemetry?

It's unusual to explicitly set expected cases as an error in telemetry.

sean-mcmanus: Users who view the Toggle Developer Tool may think it's an actual error:

image

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: image

I don't think it's a good idea to "spam" the console log with errors every time users with no feature flags set use the Enter key.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -63,12 +69,14 @@ export enum CopilotCompletionKind {
     StaleCacheHit = 'staleCacheHit',
     MissingCacheMiss = 'missingCacheMiss',
     Canceled = 'canceled',
+    FallbackToSimilarFiles = 'fallbackToSimilarFiles',
```
sean-mcmanus: Where is FallbackToSimilarFiles used?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -267,21 +306,36 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
...
+                telemetry.addCopilotCanceled(duration);
+                throw new CopilotCancellationError();
             }
             telemetry.addCompletionContextKind(copilotCompletionContextKind);
             telemetry.addResponseMetadata(copilotCompletionContext?.isResultMissing ?? true,
                 copilotCompletionContext?.snippets?.length, copilotCompletionContext?.translationUnitUri,
                 copilotCompletionContext?.caretOffset, copilotCompletionContext?.featureFlag);
+            logMessage += ` (id:${copilotCompletionContext?.requestId}) `;
+            if (!copilotCompletionContext?.snippets?.length) {
+                telemetry.addCompletionContextKind(CopilotCompletionKind.FallbackToSimilarFiles);
```
sean-mcmanus: I'm seeing it FallbackToSimilarFiles even when the feature flag is not set?

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: I see where it calls `telemetry.addCompletionContextKind(CopilotCompletionKind.FallbackToSimilarFiles);` but that is done regardless of whether the feature flag is set, so I don't see where the feature flag FallbackToSimilarFiles is used.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -267,21 +306,36 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
...
-                throw new CancellationError();
+                telemetry.addCopilotCanceled(duration);
+                throw new CopilotCancellationError();
             }
             telemetry.addCompletionContextKind(copilotCompletionContextKind);
             telemetry.addResponseMetadata(copilotCompletionContext?.isResultMissing ?? true,
                 copilotCompletionContext?.snippets?.length, copilotCompletionContext?.translationUnitUri,
                 copilotCompletionContext?.caretOffset, copilotCompletionContext?.featureFlag);
+            logMessage += ` (id:${copilotCompletionContext?.requestId}) `;
+            if (!copilotCompletionContext?.snippets?.length) {
```
sean-mcmanus: This length check appears incorrect, because it's equal to 1 if the language standard context is sent, but no actual snippets -- isn't the intent to check for 0 actual snippets?

sean-mcmanus: Note: The codeSnippetCount variable below is 0 in this case.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: yes to both questions. In particular, the flags are not present anymore since the original API being used is now gone.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Right, the error message is there because there should not be any case where the FeatureFlag is not set, i.e. in all our A/B exp flights we have it set. In any case, I dropped the error exception

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: yes, it should be logged in telemetry, but not necessarily in log. The case should not be possible, i.e. flags must always be set.
Basically the `resolve()` function is NOT called if the VSCode instance (or user) is not part of the experiment. In your case it is being called because you have it enabled in your settings.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: The idea is to trigger an error and then Copilot will consume it and fallback. I published a PR for enabling this simple mechanism, it may change depending on the feedback.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Is there anything I should fix? Let me know :)

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: You are totally right, I have to account for the traits, and fallback even if they are there. Fixed, please review.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Fixed indeed, thanks!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Yes, it does. I'll create a note to add it in a subsequent PR (or anyway asap)

kuchungmsft (dismissed):

sean-mcmanus (commented):
@lukka I keep hitting this and I'm not sure why

image

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -141,30 +180,35 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
...
+                if (projectContext?.standardVersion) {
+                    copilotCompletionContext.snippets.push({ name: "C++ language standard version", value: `This project uses the ${projectContext.standardVersion} language standard version.` });
+                }
+                telemetry.addTraitsMetadata(projectContext, telemetryProperties, telemetryMetrics);
+
+                return copilotCompletionContext;
             } else {
                 logMessage += `, result is missing`;
                 telemetry.addResponseMetadata(true);
                 return undefined;
```
sean-mcmanus: This returning undefined is a little strange because it's losing information that the caller seems to want to check.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -267,20 +316,38 @@ export class CopilotCompletionContextProvider implements ContextResolver<CodeSni
...
-                telemetry.addInternalCanceled(duration);
-                throw new CancellationError();
+                telemetry.addCopilotCanceled(duration);
+                throw new CopilotCancellationError();
             }
             telemetry.addCompletionContextKind(copilotCompletionContextKind);
             telemetry.addResponseMetadata(copilotCompletionContext?.isResultMissing ?? true,
                 copilotCompletionContext?.snippets?.length, copilotCompletionContext?.translationUnitUri,
                 copilotCompletionContext?.caretOffset, copilotCompletionContext?.featureFlag);
+            logMessage += ` (id:${copilotCompletionContext?.requestId}) `;
```
sean-mcmanus, line 303: This is odd/confusing, because `${copilotCompletionContext?.requestId` will be undefined if isResultMissing.

By returning the undefined as mention earlier, you're losing the requestId. Maybe that doesn't really matter but it makes the logging a little confusing.

sean-mcmanus (approved):

sean-mcmanus (approved):
@lukka Should the snippets be trimmed for whitespace?

image

sean-mcmanus (changes requested):
I still see unnecessary stuff sent to TypeScript like endLine. Can you remove that?

sean-mcmanus (approved):
I'm approving but it seems like some of the data sent to TypeScript could be cleaned up on the cpptools side.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Right, now I am returning the instance (not undefined), and should not miss some data like the reqId.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: should be fixed!

lukka:
> I'm approving but it seems like some of the data sent to TypeScript could be cleaned up on the cpptools side.

Thanks! I have fixed and cleanup most (if not all) the things you suggested. The only one I am not going to do is the trimming of the whitespace. That should be unnecessary, and I am going to ask to the Copilot experts before doing it. In any case, not in this PR.

Extension/src/LanguageServer/copilotCompletionContextTelemetry.ts
```diff
@@ -95,6 +96,33 @@ export class CopilotCompletionContextTelemetry {
...
+        if (statistics) {
+            for (const key in statistics) {
+                this.addMetric(`response.statistics.${key}`, statistics[key as keyof CopilotCompletionContextStatistics]);
+            }
+        }
+    }
+
+    public addCppStandardVersionMetadata(standardVersion: string, getProjectContextStartTime: number): void {
+        this.addProperty('response.cppStandardVersion', standardVersion);
+        this.addMetric('response.getProjectContextElapsedMs', Date.now() - getProjectContextStartTime);
```
sean-mcmanus: @lukka This telemetry is not being computed correctly: image

I think you want to remove `Date.now() -`.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotCompletionContextTelemetry.ts
lukka: yes, thanks for letting me know about this! Fixed now.

sean-mcmanus (approved):

PR #13303, opened 2025-02-21
Update yarn.lock (octokit/rest)

sean-mcmanus:
Update octokit.

I tested the OneLocBuild and it works.

### DIFF
--- a/.github/workflows/job-compile-and-test.yml
+++ b/.github/workflows/job-compile-and-test.yml
@@ -26,14 +26,6 @@ jobs:
         with:
           node-version: 20
 
-      - name: Delete .npmrc file (Windows)
-        if: ${{ runner.os == 'Windows' }}
-        run: Remove-Item -Path $env:USERPROFILE\.npmrc -Force
-
-      - name: Delete .npmrc file (Unix)
-        if: ${{ runner.os != 'Windows' }}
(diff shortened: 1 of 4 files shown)


sean-mcmanus:
Does anyone know why yarn is failing? It's working locally.

.github/workflows/job-compile-and-test.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Use Node.js 16
+      - name: Use Node.js 20
         uses: actions/setup-node@v3
```
lukka: could we use v4 instead of v3? Any reason to not bump it up?

sean-mcmanus: What is v4? How do you know there is a v4?

lukka: For security reason it is usually a good practice to keep up with latest version of all `@actions/`  actions, like `@actions/setup-node-v4`.  Enabling the dependabot could help in getting automated notifications and PRs to bump those.
You may read the release note to ponder whether that is needed or useful.

sean-mcmanus:
I'm still running a OneLocBuild to make sure it still works.

Colengms (approved):

PR #13306, opened 2025-02-22
Add `reduceRecursiveIncludes` setting

Colengms:
Adding as a fully-fledged setting, for now.  If we encounter issues when testing with it set to false, we could hide the setting.  Requires the corresponding PR in the language server.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -973,6 +973,12 @@
                             ]
                         },
                         "scope": "resource"
+                    },
+                    "C_Cpp.reduceRecursiveIncludes": {
+                        "type": "boolean",
+                        "default": true,
+                        "markdownDescription": "%c_cpp.configuration.reduceRecursiveIncludes.markdownDescription%",
+                        "scope": "window"
                     }
                 }
             },
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -809,6 +809,7 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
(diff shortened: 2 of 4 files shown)


Extension/package.nls.json
```diff
@@ -809,6 +809,7 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     },
+    "c_cpp.configuration.reduceRecursiveIncludes.markdownDescription": "Indicates whether recursive include paths should be reduced to only those currently referenced by the translation unit. If enabled, performance is impacted if the database needs to be populated to determine headers used. If disabled, performance may be impacted due to passing a large number of include paths to the IntelliSense process.",
```
sean-mcmanus, line 812: "enabled" and "disabled" should be in markdown/backticks like the other settings value examples.

Extension/package.json
```diff
@@ -973,6 +973,12 @@
                             ]
                         },
                         "scope": "resource"
+                    },
+                    "C_Cpp.reduceRecursiveIncludes": {
+                        "type": "boolean",
+                        "default": true,
+                        "markdownDescription": "%c_cpp.configuration.reduceRecursiveIncludes.markdownDescription%",
+                        "scope": "window"
```
sean-mcmanus, line 981: Shouldn't this have an enum like

```
                        "enum": [
                            "default",
                            "enabled",
                            "disabled"
                        ],
```

I'm not sure if "default" is needed (we could potentially use that to auto-set the default from an experiment).

sean-mcmanus: Oh -- it's a bool -- your description references "enabled"/"disabled"? Should that be `true`/`false` instead?

Extension/package.nls.json
sean-mcmanus: Or `true`/`false` as mentioned in the other thread.

PR #13324, opened 2025-02-28
Fix ${workspaceFolder} being added to browse.path if ${workspaceFolder}/* is used.

sean-mcmanus:
This was working with 1.16.3, but then some change caused the ${workspaceFolder} variable to be resolved before it checked for ${workspaceFolder} at this spot, so the regex match would always fail. This just fixes it so that the previously working case works now.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -998,10 +998,15 @@ export class CppProperties {
                 } else if (configuration.includePath) {
                     // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
                     configuration.browse.path = configuration.includePath.slice(0);
-                    if (configuration.includePath.findIndex((value: string) =>
-                        !!value.match(/^\$\{(workspaceRoot|workspaceFolder)\}(\\\*{0,2}|\/\*{0,2})?$/g)) === -1
-                    ) {
-                        configuration.browse.path.push("${workspaceFolder}");
+                    if (this.rootUri) {
+                        let normalizedRootUriFsPath: string = escapeStringRegExp(this.rootUri.fsPath.replace(/\\/g, "/"));
+                        let regexPattern = `^${normalizedRootUriFsPath}(\\*{0,2}|\/\\*{0,2})?$`;
+                        let regex = new RegExp(regexPattern, "g");
+                        if (configuration.includePath.findIndex((value: string) =>
+                            !!value.match(regex)) === -1
+                        ) {
+                            configuration.browse.path.push("${workspaceFolder}");
+                        }
                     }
                 } else {
                     configuration.browse.path = ["${workspaceFolder}"];
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -999,9 +999,9 @@ export class CppProperties {
                     // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
                     configuration.browse.path = configuration.includePath.slice(0);
                     if (this.rootUri) {
-                        let normalizedRootUriFsPath: string = escapeStringRegExp(this.rootUri.fsPath.replace(/\\/g, "/"));
-                        let regexPattern = `^${normalizedRootUriFsPath}(\\*{0,2}|\/\\*{0,2})?$`;
-                        let regex = new RegExp(regexPattern, "g");
+                        const normalizedRootUriFsPath: string = escapeStringRegExp(this.rootUri.fsPath.replace(/\\/g, "/"));
+                        const regexPattern = `^${normalizedRootUriFsPath}(\\*{0,2}|\/\\*{0,2})?$`;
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -998,10 +998,15 @@ export class CppProperties {
                 } else if (configuration.includePath) {
                     // If the user doesn't set browse.path, copy the includePath over. Make sure ${workspaceFolder} is in there though...
```
Colengms, line 999: After my current PR on the native side goes in, I want to remove this section, making it the responsibility of the native side to populate the browse paths with the include paths, if the browse path is undefined. That way, I can ensure any include paths ingested via compiler arguments get into the browse path also (if not explicitly set).  Since this would be removed soon (before the next insiders, if possible), maybe this fix isn't needed.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -998,10 +998,15 @@ export class CppProperties {
...
                     configuration.browse.path = configuration.includePath.slice(0);
-                    if (configuration.includePath.findIndex((value: string) =>
-                        !!value.match(/^\$\{(workspaceRoot|workspaceFolder)\}(\\\*{0,2}|\/\*{0,2})?$/g)) === -1
-                    ) {
-                        configuration.browse.path.push("${workspaceFolder}");
+                    if (this.rootUri) {
```
fearthecowboy, line 1001: Do we know if this behavior changed due to a change in vscode itself, or something else?

if it was a change in vscode, did we want to still make it work with `(workspaceRoot|workspaceFolder)` as well in case the newer extension was run on an older version of vscode?

bobbrow: It needs to continue to work with `${workspaceFolder}/**` because that's the default entry for includePath.

sean-mcmanus: @fearthecowboy @bobbrow The change is that this the `${workspaceFolder}` variable is resolved at this point by our own code due to calls to the resolveVariable or whatever it's called.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: @Colengms Sure, yeah, it sounds like this won't be needed.

Extension/src/LanguageServer/configurations.ts
bobbrow: `resolveVariables` does not resolve `${workspaceFolder}`. That happens in the native code.

sean-mcmanus: @bobbrow No, that changed. It resolves it in TypeScript. That's what I saw when I debugged it.

bobbrow: Where? I reviewed the implementation of `resolveVariables` before I made that comment and the behavior has not changed.

https://github.com/microsoft/vscode-cpptools/blob/db1cfda601b6bdc7463edf37d9fbcecb880920c2/Extension/src/common.ts#L429-L431

bobbrow: There is also `resolvePath` which will replace the `${workspaceFolder}` variable, but its return value is only supposed to be used as a temporary, e.g. for configuration validation and setting up watchers for compile commands.

sean-mcmanus: @bobbrow Yes, it's resolvePath, called by resolveAndSplit.

sean-mcmanus: Called by updateConfigurationPathsArray, which is called by updateServerOnFolderSettingsChange, which is the function that was modified in this PR.

sean-mcmanus: @bobbrow I recall Colen talking about changing it to not send ${workspaceFolder} to the cpptools side and resolving it on the TypeScript side instead, since it wasn't necessary keep ${workspaceFolder} around.

bobbrow: Well, I'm not sure when that changed, but I suppose if it has, then yes this code will need to be updated to deal with it. But I think it doesn't make a lot of sense to be inserting `${workspaceFolder}` at a point when `${workspaceFolder}` is expected to have been resolved away, right?  We're in a halfway state now and that causes confusion.

sean-mcmanus: @bobbrow Sure, but my PR just fixes it so that it doesn't add ${workspaceFolder} -- the pre-existing behavior already adds workspace folder. And note I've moved this to draft since Colen says one of his pending PRs would be causing this code to not be used so it might not be needed.

Colengms: The native side will resolve variables (`${workspaceRoot}`, `${workspaceFolder}`, `${vcpkgRoot}`) when ingesting browse paths (and include paths, mac framework paths, forced includes, compile commands paths, compiler path, compiler args, and the database filename) from the base configuration.  Perhaps processing was added on the TypeScript side to address needing one of these values (compiler path in a build task maybe?) with it already resolved.

PR #13334, opened 2025-03-03
fix type checking

bobbrow:
`as` does not change the type of a variable. It does nothing but tell the compiler to treat the variable as a certain type. This PR fixes some type checking issues in recently added settings as well as some other locations I noticed during the update.

### DIFF
--- a/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
+++ b/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
@@ -6,6 +6,7 @@ import { ContextResolver, ResolveRequest, SupportedContextItem } from '@github/c
 import { randomUUID } from 'crypto';
 import * as vscode from 'vscode';
 import { DocumentSelector } from 'vscode-languageserver-protocol';
+import { isNumber, isString } from '../common';
 import { getOutputChannelLogger, Logger } from '../logger';
 import * as telemetry from '../telemetry';
 import { CopilotCompletionContextResult } from './client';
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -203,7 +207,7 @@ export class CopilotCompletionContextProvider implements ContextResolver<Support
     private async fetchTimeBudgetFactor(context: ResolveRequest): Promise<number> {
         try {
             const budgetFactor = context.activeExperiments.get(CopilotCompletionContextProvider.CppCodeSnippetsTimeBudgetFactor);
-            return ((budgetFactor as number) ?? CopilotCompletionContextProvider.defaultTimeBudgetFactor) / 100.0;
```
bobbrow, line 206: This would also have been a bug if `0` was a valid value for `budgetFactor`.

fearthecowboy (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: @bobbrow the coalescing operator `??` uses the right value if left is `null` or `undefined`, but it won't in any other case, like if it was a Falsy value (i.e. `budgetFactor == 0`).

bobbrow: @lukka good to know. I was thinking it behaved the same as the `||` operator which uses the falsy rules.

Issue #13360, opened 2025-03-11
CPP IS broken


heartacker:
### Environment

版本: 1.98.1
提交: 2fc07b811f760549dab9be9d2bedd06c51dfcb9a
日期: 2025-03-10T15:38:08.854Z
浏览器: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.98.1 Chrome/132.0.6834.196 Electron/34.2.0 Safari/537.36
- If using SSH remote, specify OS of remote machine: WSL  or LOCAL

### Bug Summary and Steps to Reproduce

![Image](https://github.com/user-attachments/assets/06e4f991-5430-4191-ac92-dfb35c5b4420)

### Configuration and Logs

```shell
.
```

heartacker:
![Image](https://github.com/user-attachments/assets/01f3e0b9-d6fe-44a3-9a30-b9505a7d1c91)

sean-mcmanus:
@heartacker I don't repro the issue. I think it's a VS Code issue. Are you able to repro the error with any other extensions?

sean-mcmanus:
@heartacker I have a PR at https://github.com/microsoft/vscode-cpptools/pull/13361 that might address this, but I think it still would be a VS Code issue since we don't set `extensionKind: [ "web"]` in our package.json.

sean-mcmanus:
@heartacker I've reproed the bug. It happened after I reloaded my WSL window, maybe after VS Code Insiders updated, but it didn't repro on another machine, so I'm not sure what the repro requirements are, but this definitely seems like a VS Code bug, and https://github.com/microsoft/vscode-cpptools/pull/13361 seems like it shouldn't have any effect.

sean-mcmanus:
@heartacker I filed an issue on VS Code at https://github.com/microsoft/vscode/issues/243396 .

sean-mcmanus:
@heartacker Have you reproed the issue yet with our latest 1.24.3 release? I have not so far.

heartacker:
it works well

PR #13361, opened 2025-03-11
Set the extensionKind.

sean-mcmanus:
Possibly related to https://github.com/microsoft/vscode-cpptools/issues/13360 . The user said VS Code was trying to load our extension in the web worker extensionhost, which shouldn't be possible, but this makes it explicit that the extensionKind should be "workspace" (hopefully not "web" which seems like it might be an internal value?).

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -23,6 +23,9 @@
     },
     "homepage": "https://github.com/Microsoft/vscode-cpptools",
     "qna": "https://github.com/Microsoft/vscode-cpptools/issues",
+    "extensionKind": [
+        "workspace"
+    ],
     "keywords": [
         "C",
         "C++",


bobbrow:
"web" isn't a valid `extensionKind`. I thought you set the `browser` property for a web extension. I'm not sure what exactly this change does. I would have thought that this was the default value.

sean-mcmanus:
> "web" isn't a valid `extensionKind`. I thought you set the `browser` property for a web extension. I'm not sure what exactly this change does. I would have thought that this was the default value.

Yeah, there's no "web" explicitly defined according to the scheme version we seem to use, but the VS Code source seems to refer to "web" and "-web" as being valid values, but maybe that's not applicable (https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/extensions/common/extensionManifestPropertiesService.ts#L105), but according to Copilot, not having an extensionKind means that it can run in the web worker extension host and you set the extensionKind to include "ui" and/or "workspace" -- that could be wrong though too.

bobbrow (approved):

PR #13372, opened 2025-03-14
Add a string for browse path not found

Colengms:


sean-mcmanus (dismissed):

sean-mcmanus (commented):
Is this going to confuse anyone who doesn't have it set specifically as a browse.path? This error appears whenever a -I refers to an non-exist path, which could be in an @rsp file referenced in a compile_commands.json.

Colengms:
@sean-mcmanus  Do you have a better suggestion?  This is a browse path. There is no context at this point as to what the source of the browse path was. It could be a compile_commands arguments, a path from a custom configuration provide, an include path, an explicit browse path... anything.

bobbrow:
How about "Unable to index files from non-existent folder: {0}"

bobbrow (approved):

sean-mcmanus (approved):

PR #13377, opened 2025-03-17
Remove always-auth.

sean-mcmanus:
Fix npm warning: 

image


bobbrow:
This line is still in the instructions for CFS. Are you sure it needs to be removed now? We could wait until we upgrade to a version of npm that fails.

There is also a .npmrc in the `.github/actions` folder.

sean-mcmanus:
> This line is still in the instructions for CFS. Are you sure it needs to be removed now? We could wait until we upgrade to a version of npm that fails.
> 
> There is also a .npmrc in the `.github/actions` folder.

We could wait 11 is the current version and it says it'll be removed in 12, but it said it was removed in npm 7 in 2021 because it wasn't used: https://docs.npmjs.com/cli/v7/using-npm/changelog#v7111-2021-04-23

Colengms (dismissed):

Colengms (approved):

PR #13379, opened 2025-03-17
Filter crash telemetry data per-line so it's not completely filtered later on

sean-mcmanus:


Extension/src/LanguageServer/extension.ts
```diff
@@ -1255,12 +1256,8 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, cr
                 funcStr = funcStr.replace(/, std::allocator<std::string>/g, "");
             }
         }
-        if (funcStr.includes("/")) {
-            funcStr = "<funcForwardSlash>";
```
sean-mcmanus, line 1259: None of this `<func` cases ever occurred.

Colengms (approved):

PR #13403, opened 2025-03-22
Stop skipping the top crash stack frames.

sean-mcmanus:
The top stack frames show how much farther down the crash occurs and the addressData may be able to locate the function even if the name string isn't available.


sean-mcmanus:
It might be better if we just got rid of addressData and just replaced the "..." with the address -- I was going to look into that in a follow up PR though since it's not 100% necessary.

Colengms (approved):

PR #13408, opened 2025-03-24
Missing npmrc for other extensions

bobbrow:
I saw a warning in the last build for the extension pack that we're not linked to CFS for these extensions. These extensions don't actually download any node modules, but adding these configs should make the warnings go away.


sean-mcmanus (dismissed):
What about the always-auth deprecation: https://github.com/microsoft/vscode-cpptools/pull/13377

sean-mcmanus:
> I thought we were going to remove it in the next npm upgrade. I suppose its value doesn't matter for these projects and I could remove it now if you want.

I think it would be better removed now.

sean-mcmanus (approved):

PR #13417, opened 2025-03-25
merge rpc calls to fetch completion context

lukka:


Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -133,55 +135,43 @@ export class CopilotCompletionContextProvider implements ContextResolver<Support
             telemetry.addRequestMetadata(documentUri, caretOffset, context.completionId,
                 context.documentContext.languageId, { featureFlag: snippetsFeatureFlag });
             const docUri = vscode.Uri.parse(documentUri);
+            const getClientForTime = performance.now();
             const client = clients.getClientFor(docUri);
+            const getClientForDuration = CopilotCompletionContextProvider.getRoundedDuration(getClientForTime);
+            logMessage += ` (getClientFor elapsed:${getClientForDuration}ms) `;
```
sean-mcmanus: I'm seeing the ` ` space at the end lead to a double space.

sean-mcmanus: image

lukka: @sean-mcmanus thank, I reworked the output now, let me know!

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -191,13 +180,14 @@ export class CopilotCompletionContextProvider implements ContextResolver<Support
 
             telemetry.addError();
             if (e.message && e.stack) {
-                out.appendLine(`Copilot: getCompletionContextWithCancellation(${documentUri}: ${caretOffset}): Error: '${e.message}', stack '${e.stack}`);
+                this.logger.appendLine(`Copilot: getCompletionContextWithCancellation(${documentUri}: ${caretOffset}): Error: '${e.message}', stack '${e.stack}`);
```
sean-mcmanus: The appendLine should be appendLineAtLevel(1, ...

or a higher level if you want. 1 is Error level. Otherwise, the logging would appear when the user has None level set, which we've gotten user complaints about in the past.

Also, e.stack shouldn't be used for C/C++ logging -- we would only use that for console logging -- did you intend for console logging?

I don't think the stack would be useful in this case so it seems to me like it should be removed...oh...or is there a scenario where the stack could be useful? I was thinking we already know the location of the error based on the message itself.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
```diff
@@ -191,13 +180,14 @@ export class CopilotCompletionContextProvider implements ContextResolver<Support
...
             } else {
-                out.appendLine(`Copilot: getCompletionContextWithCancellation(${documentUri}: ${caretOffset}): Error: '${e}'`);
+                this.logger.appendLine(`Copilot: getCompletionContextWithCancellation(${documentUri}: ${caretOffset}): Error: '${e}'`);
             }
             return undefined;
         } finally {
-            out.appendLineAtLevel(6, logMessage);
+            this.logger.
+                appendLineAtLevel(6, logMessage);
```
sean-mcmanus: I created a PR to move this to 6, but on 2nd thought, the cpptools logging related to this is at 7, so do you think this should be at 7 too or did you still want it to be at the lower Debug log level?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
sean-mcmanus: If the error is from cpptools (isn't it always?) then the call stack is not helpful and is to: handle_respond in node_modules\vscode-jsonrpc\lib\common\connection.js

lukka: Right, I aligned the log-level to 7 for all Copilot related output. And dropped the callstack, not really useful indeed.

Extension/src/LanguageServer/copilotCompletionContextProvider.ts
lukka: Right, I aligned the log-level to 7 for all Copilot related output.

sean-mcmanus (approved):

Colengms (approved):

PR #13455, opened 2025-04-03
Fix some file associations

Colengms:
Aligns some hard-coded default file associations with the defaults on the native side. (There is a separate native PR for fixes there).

Fixes a bug where `.C` may be considered a C source file (uppercase implies C++).

Correctly considers `.i` a C source file, and `.ii` a C++ source file. Added missing header type: `.tlh`, `.tli`

Moves `.tcc` to the list of header files instead of source files.

Not addressed by this PR: These functions should also consider the contents of `files.assocations` (at least for source file types, as that does not distinguish between headers and sources).

### DIFF
--- a/Extension/src/common.ts
+++ b/Extension/src/common.ts
@@ -165,19 +165,19 @@ export function getVcpkgRoot(): string {
 export function isHeaderFile(uri: vscode.Uri): boolean {
     const fileExt: string = path.extname(uri.fsPath);
     const fileExtLower: string = fileExt.toLowerCase();
-    return !fileExt || [".cuh", ".hpp", ".hh", ".hxx", ".h++", ".hp", ".h", ".ii", ".inl", ".idl", ""].some(ext => fileExtLower === ext);
+    return !fileExt || [".cuh", ".hpp", ".hh", ".hxx", ".h++", ".hp", ".h", ".inl", ".ipp", ".tcc", ".tlh", ".tli", ""].some(ext => fileExtLower === ext);
 }
 
 export function isCppFile(uri: vscode.Uri): boolean {
     const fileExt: string = path.extname(uri.fsPath);
     const fileExtLower: string = fileExt.toLowerCase();
-    return (fileExt === ".C") || [".cu", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".ino", ".ipp", ".tcc"].some(ext => fileExtLower === ext);
+    return (fileExt === ".C") || [".cu", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".ii", ".ino"].some(ext => fileExtLower === ext);
 }
 
 export function isCFile(uri: vscode.Uri): boolean {
     const fileExt: string = path.extname(uri.fsPath);
     const fileExtLower: string = fileExt.toLowerCase();
-    return (fileExt === ".C") || fileExtLower === ".c";
+    return fileExt === ".c" || fileExtLower === ".i";
 }
 
 export function isCppOrCFile(uri: vscode.Uri | undefined): boolean {


fearthecowboy (approved):

bobbrow:
> Correctly considers .i a C source file,

Is this universally true? I recently preprocesed a .cpp file with MSVC and it spit out a .i for me.

Extension/src/common.ts
```diff
@@ -165,19 +165,19 @@ export function getVcpkgRoot(): string {
...
     const fileExtLower: string = fileExt.toLowerCase();
-    return (fileExt === ".C") || [".cu", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".ino", ".ipp", ".tcc"].some(ext => fileExtLower === ext);
+    return (fileExt === ".C") || [".cu", ".cpp", ".cc", ".cxx", ".c++", ".cp", ".ii", ".ino"].some(ext => fileExtLower === ext);
 }
 
 export function isCFile(uri: vscode.Uri): boolean {
     const fileExt: string = path.extname(uri.fsPath);
     const fileExtLower: string = fileExt.toLowerCase();
-    return (fileExt === ".C") || fileExtLower === ".c";
+    return fileExt === ".c" || fileExtLower === ".i";
```
sean-mcmanus, line 180: Is `.I` supposed to be a C file?

sean-mcmanus: Similarly, are all the other extension supposed to be case-insensitive?

sean-mcmanus: Should the extensions be case sensitive on Linux/Mac and case insensitive on Windows?

Colengms: ".i" and ".ii" are compiler-generated preprocessed file.  (And Bob pointed out that MSVC broke that rule and is outputting ".i" for all preprocessed output, unlike GCC and Clang).  I don't think there is any precedent uppercase ".I" have a special meaning.  ".C" is the only extension for which the case has meaning (and is enforced in `protocolFilter.ts`)

Colengms: VS Code itself considers ".i" to be a C source file.  We had it configured as a C-specific file on the native side already.

Colengms:
> > Correctly considers .i a C source file,
> 
> Is this universally true? I recently preprocessed a .cpp file with MSVC and it spit out a .i for me.

It looks like MSVC incorrectly copied the convention from GCC/clang, and doesn't properly use `.ii` for C++ sources.  However, if you look at the native PR, at the changes in `settings.h`, you'll see that we were already considering `.i` a C file. (Apparently because VS Code itself considers `.i` a C source file, which you can confirm by opening one).

Extension/src/common.ts
sean-mcmanus: But this code is treating ".I" as a C file, when on Linux it would not normally be due to case sensitivity, right?

Colengms: The file associations from VS Code itself are case insensitive.  We have this comment in the native code: 

`// VS Code file associations are case-insensitive on all OS's.`

(And some research seems to support that).

I think the right approach would be to consider all file associations as being case insensitive, with the uppercase `.C` being the only exception.

sean-mcmanus: ok

bobbrow:
> 
> It looks like MSVC incorrectly copied the convention from GCC/clang, and doesn't properly use `.ii` for C++ sources. However, if you look at the native PR, at the changes in `settings.h`, you'll see that we were already considering `.i` a C file. (Apparently because VS Code itself considers `.i` a C source file, which you can confirm by opening one).

I'm not particularly worried about it since I don't think there are any existing bugs asking for different behavior. It seems to be an uncommon scenario.

sean-mcmanus (approved):

Issue #13463, opened 2025-04-06
New Attach/preattach - waitFor feature


borjamunozf:
### Feature Request

Hello!

I've rumbling a bit for certain scenarios where the target PID we want to debug crashes so early at startup, the process ends quickly before we can debug the initialization or we are interested only in the beginning of the execution.

Sometimes this could be tricky:
- We cannot modify the source code for some reason.
- The Launch debug could be really complex to setup.
- Others 

The debuggers like GDB or LLDB has not yet fully support for this kind of stuff:
- GDB has this in the roadmap.
- LLDB has something similar but only works for MacOSX (as far as I know)

So I've working a few hours in a naive & simple waitFor approach (Linux only) with basic polling, I think it could work and be useful to handle this scenarios. 

The current idea or approach is something like this:

- New options for Attach mode: waitFor object with three properties
- Enabled: specifies to use waitFor attach support.
- Pattern: process name pattern (finds in PID detail if this pattern is included in the string process detail)
- Timeout: time to wait for process match in Ms.

Snippet:

```json
"waitFor": {
   "enabled": true,
   "pattern:" "myprocessnamepattern",
   "timeout": 10000 
}
```

Overall, something that I'm interested in -and probably others- is to "Preattach". Even if we wait for the PID and _attach_ early, if the app is huge the debugger will take time to load symbols (in our case, even using .gdbindex could take up ~15-20seconds in large C++), so we will miss again the _startup_ of the program!

The options here:
1) SIGSTOP the process by default. This would require some GDB setupCommands, for example, regarding the signal handle and so on. 
1.1) If we SIGSTOP, perhaps find a way to _execContinue_ after attaching, but I guess that this is not really doable from the Extension, right? We don't know exactly when the debugger has really completed the initialization from the Extension.

2) Allow a new json key for waitFor like stopAtEntry idea.
3) ?

I could do the PR with my current code in Draft if you consider this feature to be interesting and valuable for the extension.
Thanks!

PR #13464, opened 2025-04-07
Feat: naive waitFor preattach/attach support

borjamunozf:
Hello :)

Coming from this #13463 thread where I explain a bit the aim and context, but to sum it up:

- Support naive early attach / preattach for scenarios where the target debugee crashes before attaching, the initilization phase already took place or the aim to get full visibility of the init launch of the process.
- It's not that weird to have this kind of problems and challenges, so a _preattach_ functionality could be really useful when the Launch mode is really complex to setup or the actual code is not feasible to be modified and just put _sleep_ or pauses.

This is a naive approach I came up with, but tested with success in our C++ codebases for the company I'm working on.
- Polls the /proc (Linux support only) and waits for the pattern match PID detail.
- Sends SIGSTOP.

The SIGSTOP, as I mentioned in #13463, is something that could be optional or specified as key for the waitFor properties. but at least for decent size C++ code the symbols loading and debugger startup (be LLDB or GDB) could be slow ~15-20secs, so we need to SIGSTOP and handle that in GDB side.

I'm not a Typescript dev at all, probably there's some work and room to improvements, but I think is a decent step to propose here and discuss with you if you're up to include as functionality.

There's pending changes to apply, but I preferred to upload the current status and continue if you're interested.

Thanks!


borjamunozf:

fearthecowboy:
I like the _idea_ of this - I think we'd need to make sure this was working for all platforms.

I'm a bit nervous about how to make the polling robust as possible (and available across the platforms) - the looping around grabbing the processes is a bit of a wide-casting net (getting all the processes and then parsing the results) - we should probably make it so that we're only asking PS for the processes that are likely matches (but that complicates things with just having an expression obviously) 

We'd need another way to suspend the process on Windows (given the lack of SIGSTOP). Not impossible.

I had been thinking a while back if a script launched before debugging might be another way to go to do this - If a task was created to poll in a loop and suspend the matched process, but then we'd still need to have the editor know when to start attaching. 

I'll bring it up with some other folks and see what they think.

sean-mcmanus:
I don't think it necessarily needs to work on Windows, as long as the user can find out that it's not supported. Others might disagree though.

bobbrow:
RE: avoiding polling
It may be possible to monitor the `/proc` folder for new processes and check the `exe` entry against the program you're waiting for.

I also don't think Windows parity is required. We could support it for `cppdbg` without supporting `cppvsdbg`.

borjamunozf:
> RE: avoiding polling It may be possible to monitor the `/proc` folder for new processes and check the `exe` entry of a new processid folder against the program you're waiting for.
> 
> I also don't think Windows parity is required. We could support it for `cppdbg` without supporting `cppvsdbg`.

Indeed, the /proc polling is the approach I'm more fond of if you think Linux only support is ok, at least for now . 
I started with that approach , but after that I switch to what the others __attachers_ were using (_ps analogous_) to be as close as possible and think I mixed the comments/idea out.

bobbrow:
> Indeed, the /proc polling is the approach I'm more fond of if you think Linux only support is ok, at least for now . I started with that approach , but after that I switch to what the others __attachers_ were using (_ps analogous_) to be as close as possible and think I mixed the comments/idea out.

I think I spoke too soon. `cppdbg` is meant to work on Windows as well, and we want to have a consistent experience across all platforms, so we probably don't want to maintain 3 different implementations each doing things differently.  Polling shouldn't be too bad since it would start shortly before the process is launched. As long as there's a timeout, it should be ok.

borjamunozf:
Uploaded some changes, I had some time this weekend to get back into this.
- We use the same approach relying in the naive poll "native process listing" for Windows and Linux.
- Added a retry interval optional value as suggested by @estoykov 

For Windows support to handle the SIGSTOP could be using the _standard_ Sys internals suite program **pssuspend** [this one](https://learn.microsoft.com/en-us/sysinternals/downloads/pssuspend).

I think it could be included as offline dependency bundled into the extension, perhaps.

sean-mcmanus:
@borjamunozf We were busy with other stuff and didn't notice this PR got updated 2 weeks ago.

Extension/package.json
```diff
@@ -6566,6 +6593,7 @@
         "@types/shell-quote": "^1.7.5",
         "@types/sinon": "^17.0.3",
         "@types/tmp": "^0.2.6",
+        "@types/vscode": "^1.99.1",
```
sean-mcmanus: Why is this needed? It seems like it might be the cause of the CI failures: C:/a/vscode-cpptools/vscode-cpptools/Extension/node_modules/@types/vscode/index.d.ts(4497,3): error TS2374: Duplicate index signature for type 'string'.

Our current VS Code dependency is 1.67.0, so including a type dependency on VS Code version 1.99.1 seems incorrect.

sean-mcmanus (changes requested):
I have not finished reviewing this.

Extension/package.json
borjamunozf: Done. My fault, I think I changed because I was having some complains..

Issue #13465, opened 2025-04-07
The extension doesn't work anymore in Code - OSS


F3llFr0mTh3Sky:
I'm currently using the OSS version of VSCode, version 1.93.1 on Linux with the official C/C++ extension from the official VSCode store.
Since it got updated to version 1.24.5 on April 3, the extension doesn't loads anymore and always returns this error:
![Image](https://github.com/user-attachments/assets/e95b4685-3a05-45aa-8531-e4ce9d5671a9)

Translation:
> The C/C++ extension can only be used with Microsoft Visual Studio, Visual Studio for Mac, Visual Studio Code, Azure DevOps, Team Foundation Server and successor Microsoft products and services to develop and test your applications.

> Unable to start C/C++ language server. IntelliSense features are disabled. Error: Error: Pending response rejected since connection got disposed.

Could you confirm me it is expected behaviour that Microsoft doesn't want this extension to work anymore with Code OSS?
If it is indeed the case, why?

sean-mcmanus:
@F3llFr0mTh3Sky The bottom two error messages in your screenshot are not expected behavior, but they appear 4 minutes after the 1st message (a response regarding the 1st message is pending...).

bobbrow:
> Could you confirm me it is expected behaviour that Microsoft doesn't want this extension to work anymore with Code OSS?
If it is indeed the case, why?

Hi @F3llFr0mTh3Sky, 
Yes, what you are experiencing with the first message is expected behavior. The extension we build and publish to the VS Code Marketplace is [licensed only for products that Microsoft distributes](https://github.com/microsoft/vscode-cpptools/blob/main/RuntimeLicenses/cpptools-LICENSE.txt).  The vscode repo is open source and permission is granted to fork and release alternate distributions. Code OSS is itself is considered to be an alternate distribution without Microsoft branding and is not linked to the VS Code extension Marketplace, so it is not covered by the license.

PR #13469, opened 2025-04-07
Fix interval timer not firing causing cpptools to exit after 4 minutes

sean-mcmanus:
This is causing cpptools to exit after 4 minutes when it should not, which incorrectly leads to these messages appearing in the screenshot below, which can confuse users into thinking there's an unexpected bug, see for example https://github.com/Microsoft/vscode-cpptools/issues/13465:

image

This moves the initialization of the interval timer earlier before it could get blocked by the `await this.onDidChangeVisibleTextEditors(cppEditors);`.


benmcmorran (approved):

Colengms (approved):

Colengms (approved):

sean-mcmanus:
It sounds like there might be some follow changes as well to change how initialization and message registration are handled.

PR #13518, opened 2025-04-16
Update macos-14 to 15.

sean-mcmanus:
The mac CI was getting stuck. Not sure if this helps or not yet.


sean-mcmanus:
Doesn't seem needed yet.

PR #13521, opened 2025-04-16
Add a message explaining why we couldn't find a quote-wrapped path

### LINKED ISSUE #11955: Incorrect configuration JSON/UI warnings appear when quotes are used
Use `""` in an includePath and open the c_cpp_properties.json or the config UI.

e.g.
```json
            "includePath": [
                "${workspaceFolder}/**",
                "\"${workspaceFolder}\\new folder\""
            ],
```

image

Bug: It's giving an incorrect warning message. It seems like the "correct" include path ends up getting used though, but it seems possible there could be a case in which it could be wrong.

Not found by me, but by "other internal testing".

bobbrow:
Fixes: #11955

The bug suggests that we shouldn't show an error for paths wrapped in quotes, but we had decided to treat these strings literally since a double quote is a valid path character on some platforms.

Instead, this message improvement should signal to users that they should not place extra quotes around paths.

image


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1744,7 +1748,10 @@ export class CppProperties {
             }
 
             if (!pathExists) {
-                const message: string = localize('cannot.find', "Cannot find: {0}", resolvedPath);
+                let message: string = localize('cannot.find', "Cannot find: {0}", resolvedPath);
+                if (quotedPath) {
+                    message += '. ' + localize('wrapped.with.quotes', 'Do not add extra quotes around paths.');
```
sean-mcmanus, line 1753: Can you move the "." into the previous string? `"Cannot find: {0}."`? That prevents any issues in which the translator could add some additional punctuation that conflicts with the "." -- i.e. I've seen them add/remove periods, so 2 periods could be possible.

sean-mcmanus: Same with the case below, although in that case, it's best to keep the `\n` out of the translation string.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -2132,6 +2139,9 @@ export class CppProperties {
                             badPath = `"${expandedPaths[0]}"`;
                         }
                         message = localize('cannot.find', "Cannot find: {0}", badPath);
+                        if (incorrectExpandedPaths.length === 1 && incorrectExpandedPaths[0].match(/".*"/) !== null) {
```
sean-mcmanus: Should this be >= 1? And should it check every path in `incorrectExpandedPahts`?

Extension/src/LanguageServer/configurations.ts
bobbrow: I didn't want to add a period in the normal case. It appears to have been left out intentionally, perhaps to avoid confusion being attached to a path. Do you feel strongly about this feedback?

Extension/src/LanguageServer/configurations.ts
bobbrow: I wasn't planning on supporting that case (multiple paths in a single string), but it might be easy enough to do so.

sean-mcmanus:
Why are the 2 quotes in your screenshot: 
image

bobbrow:
> Why are the 2 quotes in your screenshot: image

That's how it's currently implemented.  We put quotes around the path for the json squiggle and then the path itself has quotes around it.

sean-mcmanus:
> That's how it's currently implemented. We put quotes around the path for the json squiggle and then the path itself has quotes around it.

But it's inconsistent with the UI message? Is that intentional?

bobbrow:
> > That's how it's currently implemented. We put quotes around the path for the json squiggle and then the path itself has quotes around it.
> 
> But it's inconsistent with the UI message? Is that intentional?

I believe it might have been done that way because we show all the paths on the same line for the tooltip, but in the UI we give each error its own line. It makes the paths easier to differentiate.

image

sean-mcmanus (dismissed):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Sure, up to you.

bobbrow: It turned out to be trivial to support this.

sean-mcmanus (approved):

PR #13531, opened 2025-04-17
Don't show bogus errors for compilerPath + args in config UI

bobbrow:
My previous PR exposed an issue with incorrect squiggles in the config UI. I'm quickly patching it to ignore these errors until a proper fix can go in.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1617,7 +1617,11 @@ export class CppProperties {
         }
         const settings: CppSettings = new CppSettings(this.rootUri);
         const compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(!!settings.legacyCompilerArgsBehavior, resolvedCompilerPath);
+
+        // compilerPath + args in the same string isn't working yet.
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -1617,7 +1617,11 @@ export class CppProperties {
         }
         const settings: CppSettings = new CppSettings(this.rootUri);
         const compilerPathAndArgs: util.CompilerPathAndArgs = util.extractCompilerPathAndArgs(!!settings.legacyCompilerArgsBehavior, resolvedCompilerPath);
+
+        // compilerPath + args in the same string isn't working yet.
+        const skipFullCommandString = !compilerPathAndArgs.compilerName && resolvedCompilerPath.indexOf(" ") >= 0;
```
sean-mcmanus: Should this be `includes(" ")`?

bobbrow: Is that preferred? I was going to use it until the doc comment for it said something about converting this to a string, so I couldn't tell if there was more overhead to use that one or not.

sean-mcmanus: I believe it is preferred. The object is already a string. The doc comment just means if it's invoked a non-string it'll try to convert it to a string.

image

sean-mcmanus (approved):

PR #13553, opened 2025-04-25
refactor the compilerPath verification to get consistent results in JSON and UI verification

bobbrow:
For 1.26


bobbrow:
Ready for review now

Extension/c_cpp_properties.schema.json
```diff
@@ -18,10 +18,7 @@
                     "compilerPath": {
                         "markdownDescription": "Full path of the compiler being used, e.g. `/usr/bin/gcc`, to enable more accurate IntelliSense.",
                         "descriptionHint": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered.",
-                        "type": [
-                            "string",
-                            "null"
-                        ]
+                        "type": "string"
```
sean-mcmanus: I still see `null` as the default? Is that correct or not?

image

sean-mcmanus (changes requested):
Didn't we want to prevent users from accidentally creating `"compilerPath: ""` in c_cpp_properties.json? I'm seeing compilerPath default to `""` when users type it in the json editor. I thought we had create some workaround to make it default to something else like null. Or maybe I'm thinking about another property? I thought we had done something similar for browse.path to prevent users from accidentally creating an empty browse.path by default.

It the type were string or null, then we could default the schema to populate it with null to prevent "" from being used by default.

Oh, it looks like null is used at the defaults for the C_Cpp.default.* equivalent settings.

sean-mcmanus (changes requested):
Something seems wrong in this case:

image

It says there are spaces and arguments, but there are not.

sean-mcmanus (changes requested):
The squiggle isn't covering the entire path:

image

Extension/c_cpp_properties.schema.json
sean-mcmanus: I think it is correct because we want to prevent `""` from being the default (and accidentally kept by users who expect the default values to be acceptable and not cause a major change in behavior).

sean-mcmanus: But as mentioned in the other message the c_cpp_properties.json version doesn't use the same defaults and it seems like it should?

bobbrow: `null` exists for the C_Cpp.default.compilerPath setting as a way to "unset" a default value set at a higher scope. I did not remove that. ~~However, `c_cpp_properties.json` does not have scopes so it does not need to support `null`.~~

bobbrow:
> Something seems wrong in this case:
> 
> image
> 
> It says there are spaces and arguments, but there are not.

If the path does not exist, and there are no quotes around the compiler, we split the string by spaces, take the first token as the compiler and the remaining tokens as the args.  I can see how this error message is confusing though. I think I can make it longer and say that it's one of two problems instead.

bobbrow:
> The squiggle isn't covering the entire path:
> 
> image

I believe that this is the current behavior in the extension (i.e. not a regression caused by my changes)

bobbrow:
> Didn't we want to prevent users from accidentally creating `"compilerPath: ""` in c_cpp_properties.json? I'm seeing compilerPath default to `""` when users type it in the json editor. I thought we had create some workaround to make it default to something else like null. Or maybe I'm thinking about another property? I thought we had done something similar for browse.path to prevent users from accidentally creating an empty browse.path by default.
> 
> It the type were string or null, then we could default the schema to populate it with null to prevent "" from being used by default.
> 
> Oh, it looks like null is used at the defaults for the C_Cpp.default.* equivalent settings.

Ok, you may be right about this for `c_cpp_properties.json`. I didn't recall why we added null for the schema. I can revert some of this. I would agree that it is preferrable to insert `null` over `""` for this property.

Extension/c_cpp_properties.schema.json
bobbrow: I just read your later comment and will retract my statement above.

bobbrow:
@sean-mcmanus I tested the compilerPath property in c_cpp_properties.json some more, and it looks like when you autocomplete `compilerPath` VS Code inserts the empty string right now (1.25.3). It looks like "null" has to be the first option in the list for it to autocomplete to `null` as we want.  I'll fix that.  CTL+Space to invoke completion after the colon does only offer null in either case.

sean-mcmanus (approved):

PR #13569, opened 2025-04-29
Support LLDB-DAP as a debugger (OSX/Linux/Windows)

fearthecowboy:
This adds the ability to debug processes via LLDB using the LLDB-DAP adapter that is bundled with LLVM's LLDB debugger.

brb


Extension/src/Debugger/configurationProvider.ts
```diff
@@ -427,13 +472,17 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
-                if (newConfig.type === DebuggerType.cppdbg) {
-                    newConfig.externalConsole = false;
-                } else {
-                    newConfig.console = "integratedTerminal";
+                switch (newConfig.type) {
+                    case DebuggerType.cppdbg:
+                        newConfig.externalConsole = false;
+                        break;
+                    case DebuggerType.cppvsdbg:
+                        newConfig.externalConsole = false;
```
bobbrow, line 482: Could you help me understand why cppvsdbg is changing here?  I was under the impression that `externalConsole` was the old, deprecated property and `console` was the newer, preferred one to set.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1128,122 +1197,95 @@ export class ConfigurationAssetProviderFactory {
...
+            new LldbDapConfigurations(this.executable)
         ];
     }
 }
 
 function convertConfigurationSnippetToCompletionItem(snippet: IConfigurationSnippet): vscode.CompletionItem {
     const item: vscode.CompletionItem = new vscode.CompletionItem(snippet.label, vscode.CompletionItemKind.Module);
 
-    item.insertText = snippet.bodyText;
+    item.insertText = JSON.stringify(snippet.body, null, '\t');
```
bobbrow, line 1280: Does VS Code not already format the JSON for us? I would hope we could avoid this if possible. Otherwise, you might need to lookup the indentation preferences for JSON files from the user.

Extension/src/Debugger/configurationProvider.ts
fearthecowboy: That is a really good question.

I .... remember running into a problem where I got an error, and now that I think about it, I can't remember why it moved me to fix it here with that. I may have made a mistake and got this backwards. lemme look into it

fearthecowboy: Ah, I see what happened. I fixed this coming from two directions and when coming back to it made the error when I converted what I was doing to a case. Weird. Fixing....

Extension/src/Debugger/lldb-dap.ts
```diff
@@ -0,0 +1,229 @@
...
+        // lldb-dap expects an object of { <name>:<value>, ... }.
+        const env: { [key: string]: string } = {};
+        for (const each of config.environment) {
+            if (typeof each.name === 'string' && typeof each.value === 'string') {
+                env[each.name] = each.value;
+            }
+        }
+        delete config.environment;
+    }
+
```
WardenGnaw, line 198: You also have `envFile` enabled in the `package.json`, you will also want to read that file and convert it into `env`.

Extension/package.json
```diff
@@ -5786,6 +5788,1589 @@
...
+                                                        "description": "%c_cpp.debuggers.sourceFileMap.sourceFileMapEntry.useForBreakpoints.description%",
+                                                        "default": true
+                                                    }
+                                                }
+                                            }
+                                        }
+                                    }
+                                ]
+                            },
+                            "logging": {
```
WardenGnaw, line 5927: The logging block should be removed, but I would recommend a way to get DAP logs to get information from the user to figure out what's going wrong in `lldb-dap`.

Extension/package.json
```diff
@@ -5786,6 +5788,1589 @@
...
+                                            {
+                                                "type": "boolean",
+                                                "description": "%c_cpp.debuggers.logging.natvisDiagnostics.description%",
+                                                "default": false
+                                            }
+                                        ]
+                                    }
+                                }
+                            },
+                            "deploySteps": {
```
WardenGnaw, line 6011: Is `deploySteps` supported?

Extension/package.json
fearthecowboy: You can set an environment variable `LLDBDAP_LOG` before running vscode and lldb-dap will dump out the log to the file you point to.

Extension/src/Debugger/lldb-dap.ts
fearthecowboy: `envFile` is already handled prior to it getting here (it loads it and converts it to the environment block

Extension/package.json
fearthecowboy: I would think so - it's handled the during `resolveDebugConfigurationWithSubstitutedVariables` - regardless of the type of debugger .

Extension/package.json
fearthecowboy: The logging block does what it does for the other debug types - which as far as I can tell is very little (ie, emitting the debugger configuration to a log channel - but again, not specific to any debug type.

WardenGnaw: there are other fields in logging too that would be no-op in the `lldb-dap` case. I would either remove it from the `package.json` definition so it would warn about it not existing.

sean-mcmanus: @WardenGnaw Should "either" be removed in your comment? I don't see an "or" mentioned.

Extension/package.json
```diff
@@ -5638,6 +5635,11 @@
                             "processId"
                         ],
                         "properties": {
+                            "program": {
+                                "type": "string",
+                                "description": "%c_cpp.debuggers.program.description%",
+                                "default": "${workspaceRoot}/program.exe"
```
sean-mcmanus, line 5641: Can you change the `workspaceRoot` references to `workspaceFolder`? `workspaceRoot` is the old variable name.

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.launchCommands.description": "Custom LLDB commands that are executed instead of launching a process.",
+    "c_cpp.debuggers.cpplldb.stopCommands.description": "LLDB commands executed each time the program stops.",
+    "c_cpp.debuggers.cpplldb.exitCommands.description": "LLDB ommands executed when the program exits.",
+    "c_cpp.debuggers.cpplldb.terminateCommands.description": "LLDB commands executed when the debugging session ends.",
+    "c_cpp.debuggers.cpplldb.enableAutoVariableSummaries.description": "Enable auto generated summaries for variables when no summaries exist for a given type. This feature can cause performance delays in large projects when viewing variables.",
+    "c_cpp.debuggers.cpplldb.displayExtendedBacktrace.description": "Enable language specific extended backtraces.",
+    "c_cpp.debuggers.cpplldb.enableSyntheticChildDebugging.description": "If a variable is displayed using a synthetic children, also display the actual contents of the variable at the end under an entry. This is useful when creating sythetic child plug-ins as it lets you see the actual contents of the variable.",
```
sean-mcmanus, line 972: sythetic->synthetic

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
     "c_cpp.debuggers.vsCodeCommand.description": "VS Code command to be invoked. Can be a command in VS Code or an active extension.",
     "c_cpp.debuggers.vsCodeCommand.command.description": "VS Code command to be invoked.",
     "c_cpp.debuggers.vsCodeCommand.args.description": "Arguments to the VS Code command.",
+    "c_cpp.debuggers.cpplldb.debuggerPath.description": "Full path to the LLDB-DAP debugger executable. If not specified, the extension will search for the debugger in the system PATH.",
+    "c_cpp.debuggers.cpplldb.debuggerArgs.description": "Additional arguments for the LLDB-DAP debugger.",
+    "c_cpp.debuggers.cpplldb.disableaslr.description": "Disable Address Space Layout Randomization (if supported).",
```
sean-mcmanus, line 956: Should this be disableASLR?

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.disablestdio.description": "Disables the standard input/output streams during debugging.",
```
sean-mcmanus, line 957: Should this be disableStdio?

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.preRunCommands.description": "LLDB commands executed just before launching/attaching, after the LLDB target has been created.",
+    "c_cpp.debuggers.cpplldb.postRunCommands.description": "LLDB commands executed after launching when it's in a stopped state prior to any automatic continuation.",
+    "c_cpp.debuggers.cpplldb.launchCommands.description": "Custom LLDB commands that are executed instead of launching a process.",
+    "c_cpp.debuggers.cpplldb.stopCommands.description": "LLDB commands executed each time the program stops.",
+    "c_cpp.debuggers.cpplldb.exitCommands.description": "LLDB ommands executed when the program exits.",
+    "c_cpp.debuggers.cpplldb.terminateCommands.description": "LLDB commands executed when the debugging session ends.",
+    "c_cpp.debuggers.cpplldb.enableAutoVariableSummaries.description": "Enable auto generated summaries for variables when no summaries exist for a given type. This feature can cause performance delays in large projects when viewing variables.",
+    "c_cpp.debuggers.cpplldb.displayExtendedBacktrace.description": "Enable language specific extended backtraces.",
```
sean-mcmanus, line 971: language specific->language-specific

Extension/package.nls.json
sean-mcmanus, line 972: It seems like `a synthetic children` should be something that matches in either singular or plural like "synthetic children" or "a synthetic child", whichever is more accurate.

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.enableSyntheticChildDebugging.description": "If a variable is displayed using a synthetic children, also display the actual contents of the variable at the end under an entry. This is useful when creating sythetic child plug-ins as it lets you see the actual contents of the variable.",
+    "c_cpp.debuggers.cpplldb.commandEscapePrefix.description": "The escape prefix character to use for executing regular LLDB commands in the Debug Console, instead of printing variables.",
+    "c_cpp.debuggers.cpplldb.customFrameFormat.description": "When specified, stack frames will have descriptions generated based on the provided format. See https://lldb.llvm.org/use/formatting.html",
+    "c_cpp.debuggers.cpplldb.customThreadFormat.description": "When specified, threads will have descriptions generated based on the provided format. See https://lldb.llvm.org/use/formatting.html",
```
sean-mcmanus, line 975: Can you locked the URIs to prevent the translators from breaking them, e.g.

```json
{
        "message": "When specified, stack frames will have descriptions generated based on the provided format. See https://lldb.llvm.org/use/formatting.html",
        "comment": [
            "{Locked=\"https://lldb.llvm.org/use/formatting.html\"}"
        ]
    }
```

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.attachCommands.description": "Custom LLDB commands that are executed instead of attaching to a process ID or to a process by name.",
+    "c_cpp.debuggers.cpplldb.waitFor.description": "Wait for the process to launch by looking for a process with a basename that matches `program`.",
```
sean-mcmanus, line 977: What is `program`?

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.coreFile.description": "Path to the core file to debug.",
+    "c_cpp.debuggers.cpplldb.gdbRemotePort.description": "TCP/IP port to attach to a remote system. Specifying both processId and port is an error.",
```
sean-mcmanus, line 979: Where is `processId` specified?

Extension/package.nls.json
```diff
@@ -951,6 +951,33 @@
...
+    "c_cpp.debuggers.cpplldb.gdbRemoteHost.description": "The hostname to connect to a remote system. The default hostname being used localhost.",
```
sean-mcmanus, line 980: Something seems incorrect with this sentence `The default hostname being used localhost` -- is it missing some words?

Extension/package.nls.json
sean-mcmanus: Also, if you spot anything else that should not be translated (i.e. a reference to a literal value) then it should also be locked.

Extension/package.nls.json
sean-mcmanus: Oh, it's just the label of the string so I'm not sure if it matters?

Extension/src/links.ts
```diff
@@ -0,0 +1,9 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+// Centralized https links that are used by the extension
+
+/** Link to documentation on how to troubleshoot LLDB-DAP. */
+export const TroubleshootingLldbDap = "https://aka.ms/vscode-cpptools/TroubleshootingLldbDap";
```
sean-mcmanus, line 9: Is this link supposed to work?

Extension/package.json
fearthecowboy: I fixed them all in the file. There were a lot of `workspaceRoot` values.

Extension/package.json
fearthecowboy: Removing.

Extension/package.nls.json
fearthecowboy: fixing

Extension/package.nls.json
fearthecowboy: The `program` field in the debug configuration

Extension/package.nls.json
fearthecowboy: It's just a label. I cap'd the A

Extension/package.nls.json
fearthecowboy: fixed

Extension/src/Debugger/debugAdapterDescriptorFactory.ts
```diff
@@ -50,3 +52,43 @@ export class CppvsdbgDebugAdapterDescriptorFactory extends AbstractDebugAdapterD
...
+        let adapter: string | undefined = session.configuration.debuggerPath || executable?.command;
+
+        if (adapter) {
+            // Verify that the path is actually valid.
+            if (!await isValidLldbDap(adapter)) {
+                adapter = await findLldbDap();
+                if (adapter) {
+                    void vscode.window.showErrorMessage(localize("debugger.not.available", "The specified LLDB-DAP debuggerPath '{0}' is not valid, falling back to {1}.", session.configuration.debuggerPath, adapter));
+                } else {
+                    void vscode.window.showErrorMessage(localize("debugger.not.available", "The specified LLDB-DAP debuggerPath '{0}' is not valid and no fallback was found.", session.configuration.debuggerPath));
```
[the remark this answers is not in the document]
fearthecowboy: fixed

Extension/package.nls.json
fearthecowboy: fixed

Extension/package.json
```diff
@@ -5786,6 +5788,1421 @@
...
+                "label": "C++ (LLDB-DAP)",
+                "languages": [
+                    "ada",
+                    "c",
+                    "cpp",
+                    "cuda-cpp",
+                    "rust"
+                ],
+                "_aiKeyComment": "Ignore 'Property aiKey is not allowed'. See https://github.com/microsoft/vscode/issues/76493",
+                "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
```
WardenGnaw, line 5803: does `lldb-dap` send telemetry events? If it doesn't, you may want to remove this.

fearthecowboy: AFAIK, this key is used by VSCode itself to send telemetry for the debugger extension (ie, `monacoworkbench/debugsessionstart`) events - which is the very code that I've written. 

Yes, this is absolutely necessary

WardenGnaw: `monacoworkbench/debugsessionstart` is from VS Code's telemetry table, they send that at https://github.com/microsoft/vscode/blob/2ba158ca779b715bd4d29d1cb6ac022b27672b26/src/vs/workbench/contrib/debug/browser/debugService.ts#L677 and does not rely on `aiKey`

This `aiKey` should be for telemetry sent from the debugger via `OutputEvent` with `category: "telemetry"`

fearthecowboy: The `aikey` for the Debugger Contribution is read in VSCode itself here in `getCustomTelemetryEndpoint()` : 

https://github.com/microsoft/vscode/blob/2ba158ca779b715bd4d29d1cb6ac022b27672b26/src/vs/workbench/contrib/debug/common/debugger.ts#L223

It would appear that `getCustomTelemetryEndpoint()` is called, and it writes at least one event here:

https://github.com/microsoft/vscode/blob/2ba158ca779b715bd4d29d1cb6ac022b27672b26/src/vs/workbench/contrib/debug/browser/debugSession.ts#L1193

I could be wrong, but I don't see any way that the lldb-dap binary would even be able to get the configuration for the debugger contribution from package.json since we never pass that to it.

This is following the same pattern that the debugger contribution points for the other debug types (`cppdbg` and `cppvsdbg` ) - cppdbg uses the MIEngine adapter, and I find no indication that MIEngine uses `aikey` at all (https://github.com/search?q=repo%3Amicrosoft%2FMIEngine%20aikey&type=code)

After reading the comments here:
https://github.com/microsoft/vscode/issues/2931
https://github.com/microsoft/vscode/issues/76493#issuecomment-507987160

I was under the impression that's why we even have the `aikey` in the debugger contribution point, otherwise we just send telemetry events ourselves with the key as normal.

fearthecowboy: ok

sean-mcmanus (changes requested):
I see this message in this status bar on startup and it doesn't seem like it should be shown because I haven't even done any debugging at all:

image

sean-mcmanus (changes requested):
I'm seeing the extension activation take a really long time sometimes. It seems like there is some sort of debugging code delaying activation. Can that be fixed or is it only an issue when debugging?

I'm seeing the language service time to hove result take 12 times as long or so -- i.e. it takes over a minute loading and then I see the Unable to find a working 'lldb-dap' adapter message.

sean-mcmanus (changes requested):
The clang and clang-cl compilers are missing when I use the Debug active file button. Can they be added back?

sean-mcmanus (changes requested):
I'm seeing debugging not working on Linux? None of the debug start options have any affect or show any error message. I thought it was working last week and it works with main.

UPDATE: It was caused by the performance issue mentioned in the other comment, i.e. it works after waiting long enough.

fearthecowboy:
> I'm seeing the extension activation take a really long time sometimes. It seems like there is some sort of debugging code delaying activation. Can that be fixed or is it only an issue when debugging?
> 
> I'm seeing the language service time to hove result take 12 times as long or so on startup -- i.e. it takes over a minute loading and then I see the Unable to find a working 'lldb-dap' adapter message.

It really shouldn't be delaying startup. Hmm. It could be going pretty aggressive tho' .. I can fix that.

Extension/src/Debugger/lldb-dap.ts
```diff
@@ -0,0 +1,229 @@
...
+                    note(localize('lldb-dap.enabled', "The {0} debugger is enabled via '{1}'", 'lldb-dap', lldbDapPath));
+                    return lldbDapPath;
+                }
+            }
+        }
+
+        if (isWindows) {
+            // If we got this far, it's not in the PATH - but worry not,
+            // we'll do a check of well known secured locations where it might be installed.
+            for (const folder of ['c:/Program Files/LLVM', 'c:/Program Files/', 'C:/program files (x86)/']) {
```
sean-mcmanus, line 110: Isn't c:/Program Files/LLVM redundant compared with c:/Program Files/?

Extension/package.json
```diff
@@ -5786,6 +5788,1419 @@
...
+                                "type": "array",
+                                "description": "%c_cpp.debuggers.args.description%",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "default": []
+                            },
+                            "type": {
+                                "type": "string",
+                                "description": "%c_cpp.debuggers.cpplldb.type.description%",
```
sean-mcmanus, line 5828: I'm seeing some warnings in the developer console:

image

sean-mcmanus (changes requested):

Extension/src/Debugger/lldb-dap.ts
```diff
@@ -0,0 +1,229 @@
...
+                if (lldbDapPath) {
+                    log(`Discovered lldb-dap binary for OSX at '${lldbDapPath}' ${Date.now() - start} msec`);
+                    note(localize('lldb-dap.enabled', "The {0} debugger is enabled via '{1}'", 'lldb-dap', lldbDapPath));
+                    return lldbDapPath;
+                }
+            }
+
+            // If we got this far, it's not in the PATH or via xcrun - but worry not,
+            // we'll do a check of well known secured locations where it might be installed.
+            for (const folder of ['/Applications', '/opt/homebrew']) {
```
sean-mcmanus, line 97: I believe these secured locations can have symlinks to unsecured locations or to executables that don't require elevated privileges to write.

sean-mcmanus (changes requested):

Extension/src/common.ts
```diff
@@ -848,6 +849,96 @@ export function isExecutable(file: string): Promise<boolean> {
...
+ * @param maxDepth The maximum depth to search (default is 4).
+ * @param results An array to store the results.
+ * @returns A promise that resolves to an array of file paths that match the filename.
+ */
+export async function searchFolder(folder: string, filename: string | RegExp, predicate?: (binary: string) => Promise<boolean>, maxDepth = 4, results = new Array<string>()): Promise<string[]> {
+    try {
+        const files: string[] = await readdir(folder);
+        const all = files.map(async (file) => {
+
+            const fullPath: string = path.resolve(folder, file);
```
sean-mcmanus, line 890: Can it avoid processing the same directory multiple times from symlinks? I'm seeing stuff like /usr/bin/X11/X11/X11/X11/X11/X11/X11/time because X11 is symlink to `.`. The depth prevents it from being infinite, but it seems like unnecessary processing. Should something like a visited set of "realpath" (resolved paths) be used to avoid duplicate processing?

sean-mcmanus (changes requested):

Extension/src/Debugger/lldb-dap.ts
```diff
@@ -0,0 +1,229 @@
...
+    if (lldbDapPath) {
+        return lldbDapPath;
+    }
+    const start = Date.now();
+    // PATH binaries take priority.
+    for (const candidate of candidates) {
+        // First, search the environment PATH for the binary.
+        lldbDapPath = await findInPath(candidate, isValidLldbDap);
+        if (lldbDapPath) {
+            log(`Discovered lldb-dap binary for OSX at '${lldbDapPath}' ${Date.now() - start} msec`);
```
sean-mcmanus, line 76: All the OSX references should be macOS.

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
When I click the Add Debug Configuration icon in the top right of a file for cpplldb, it creates a launch.json with an invalid miDebuggerPath field:

image

sean-mcmanus (changes requested):
The lldb-attach snippet that is generated is missing the "program" field. The other debugger snippet types seem to have that. Or should the schema be updated so that it doesn't show an error if program isn't needed in this case?

image

Extension/src/Debugger/lldb-dap.ts
fearthecowboy: O_o I'll filter out following symlinks entirely.

Extension/src/Debugger/lldb-dap.ts
fearthecowboy: Not entirely - While yes, it overlaps, the idea is that we want to prioritize searching that folder first. The slight bit of overlap isn't going to be detrimental

Extension/package.json
```diff
@@ -4058,11 +4059,9 @@
                                         "default": {}
                                     },
                                     "quoteArgs": {
-                                        "exceptions": {
-                                            "type": "boolean",
-                                            "description": "%c_cpp.debuggers.pipeTransport.quoteArgs.description%",
-                                            "default": true
-                                        }
+                                        "type": "boolean",
```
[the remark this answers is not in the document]
fearthecowboy: Yes - this was not proper schema - it looks like it was a copy-paste from there.

sean-mcmanus (changes requested):
List of issues to investigate/fix:
- [ ] Regression: clang compilers are not longer available from the Debug File quickpick (Windows only?).
- [ ] Regression: debugging with lldb-mi no longer works and isn't available (Windows only?). But I'm seeing lldb-mi not fully working with 1.26.1, but I see it at least start and show Debug Console logging.
- [ ] Extension host crashes?/aborts when debugging the TypeScript on Windows (with the vsix it gets stuck with an endless progress bar until Reload Window is done, stop button doesn't work).
- [ ] On Windows (not sure about Linux/Mac yet), using the Debug File only shows Windows and gdb Launch options and not lldb-dap (the bullet point above covers lldb-mi being missing), which seems like a bug to me, and unrelated to the "default" scenario where lldb-mi might be chosen previously.
- [ ] The lldb-dap versus LLDB-DAP versus LLDB references need to be reviewed/updated.
- [ ] There is no UI notification about python 3.10 needing to be installed on Windows. Where is the Install LLDB button supposed to link to?
- [ ] There is no UI notification about LLDB not being found (it's currently a C/C++ logging string at Error level).
- [ ] There is no UI notification about LLDB not being usable (it's currently a C/C++ logging string at Debug level). This might be a superset of the python 3.10 case?
- [ ] The link to https://aka.ms/vscode-cpptools/TroubleshootingLldbDap doesn't work?
- [ ] Remove (or improve?) the transient status bar message. I was seeing it disappear before I had a chance to finish reading it and I don't know of any other extensions that do that. In addition to the status bar space being limited in size and might be filled from other extensions.
- [ ] Add telemetry
  - [ ] For MIMode (not sure what exactly this involves yet).
  - [ ] For cpplldb added to launch.json.
  - [ ] For the "Install LLDB" button.
- [ ] Does waitFor, coreFile, etc. actually work?
- [ ] Are there any issues specific to Rust debugging?
- [ ] Finish code reviewing.
- [ ] Finish testing.
  - [ ] Windows
  - [ ] Linux
  - [ ] Mac
  - [ ] With and without lldb available and/or usable.
- [ ] Follow-up features for "later" (not sure when exactly yet):
  - [ ] Add an experiment to default to lldb-dap instead of lldb-mi (whenever it currently defaults to lldb-mi instead of showing the dropdown list). This might just be for the Debug button case?
  - [ ] Update the existing docs on lldb/lldb-mi for lldb-dap. It seems like any docs linked to in the code would need to be updated for V1 though.
- [ ] Pre-exising issues (I assume "won't fix" for V1):
  - [ ] https://github.com/microsoft/vscode-cpptools/issues/13655 
- [ ] VS Code bugs impacting the feature:
  - [ ] https://github.com/microsoft/vscode/issues/250116

PR #13571, opened 2025-04-30
Enable setting the VS dev environment without running from Dev Command Prompt

### LINKED ISSUE #4742: Build and debug C++ with cl without needing to start vscode from the developer prompt
image

### LINKED ISSUE #5271: Modify the terminal environment for MSVC
If I'm not mistaken this should now be possible thanks to https://github.com/microsoft/vscode/issues/46696

So it would be nice if there was a setting for the integrated terminal to be setup for MSVC without having to open VS Code through the developer command prompt or by other means.

Extension of #4178

bobbrow:
This change removes the requirement of running VS Code from a Developer Command Prompt in order to build with the Visual Studio C++ compiler on Windows.

#### Two commands are added:
* `Set Visual Studio Developer Environment` - Starts the process of picking the MSVC toolset that will be used to compile. This environment will be added to any new terminals launched from this session of VS Code. It will also be automatically applied for the Build and Debug feature. Note that the environment changes only apply to the current workspace. The command will need to be run on every workspace that needs the environment.
* `Clear Visual Studio Developer Environment` - Removes the MSVC toolset environment.

#### One setting is added:
* `C_Cpp.persistDevEnvironment` - This controls whether the developer environment is automatically applied the next time the workspace is opened. 

#### Walkthrough updates
The walkthrough is updated with a button to run the command and the instructions include the command as an alternative option to running VS Code from the Developer Command Prompt:
image

#### Play button updates
The Play button for Build and Debug Active File incorporates this command in the workflow if the developer environment is not detected. This should make the experience seamless going forward.
image

#### Terminal updates
When the command is run, any existing terminals will show a message prompting you to relaunch them so that the environment will apply to them.
image

#### Telemetry
Telemetry is added to track where the command was invoked.


Extension/package.json
```diff
@@ -3376,6 +3376,12 @@
                         "default": "default",
                         "markdownDescription": "%c_cpp.configuration.copilotHover.markdownDescription%",
                         "scope": "window"
+                    },
+                    "C_Cpp.persistDevEnvironment": {
+                        "type": "boolean",
+                        "default": true,
+                        "markdownDescription": "%c_cpp.configuration.persistDevEnvironment.markdownDescription%",
+                        "scope": "resource"
```
sean-mcmanus: This should be "window" scope, right? i.e. per-workspace and not per-workspace folder.

bobbrow: yes, I think you are right.

Extension/i18n/chs/walkthrough/devcommandprompt/open-developer-command-prompt.md.i18n.json
```diff
@@ -4,8 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 // Do not edit this file. It is machine generated.
 {
-	"walkthrough.windows.title.open.dev.command.prompt": "使用 {0} 重新启动",
```
bobbrow, line 7: I had to rename the loc string id's otherwise `translations-generate` crashes when it doesn't find an html element to match the {0}, {1} items.  The new strings don't have a substitution.  Since the properties are renamed, these ones are dead code.

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,203 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as nls from 'vscode-nls';
+import { isWindows } from '../constants';
+import { CppSettings } from './settings';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+const ERROR_NO_CONTEXT = localize('no.context.provided', 'No context provided');
```
bobbrow: When I asked Copilot to move these strings out, it chose this naming style. I think these should be camelCase

sean-mcmanus (changes requested):
Can this be fixed?

image

sean-mcmanus (changes requested):
I think it should just silently cancel instead of showing an error.

image

I think an error might be okay if it was an actual error occurrence (and without it the user may not know it failed) but not a user cancellation though (via not selecting an option in the dropdowns).

We don't share an error popup in other cancellation cases.

Same as when the user hits the "Cancel" button:

image

sean-mcmanus (changes requested):
Can you add the command name?

image

It's possible the command may not have been invoked from the developer command prompt or in a way the user is 100% sure what command is run, such as via a keyboard shortcut.

Umm...unless VS Code itself is showing that message, I haven't checked yet.

sean-mcmanus (changes requested):
Can we align the setting name more with CMake Tools?

image

i.e. C_Cpp.persistVsDeveloperEnvironment ?

i.e. "VsDeveloper" seems preferable to "Dev" since it's more explicit that it applies to Visual Studio. And "Dev" seems informal for a setting name.

sean-mcmanus (commented):
FYI, it was able to set the developer environment when I didn't have a workspace open. Not sure what effect that has.

image

sean-mcmanus (changes requested):
After setting the dev environment (with a folder open), I can't debug with cl.exe. It just builds but debugging never starts.

It seems to just run "C/C++: cl.exe build active file" and the "debug" part is not run.

i.e. the walkthrough steps aren't working for me.

bobbrow:
> After setting the dev environment (with a folder open), I can't debug with cl.exe. It just builds but debugging never starts.
> 
> It seems to just run "C/C++: cl.exe build active file" and the "debug" part is not run.
> 
> i.e. the walkthrough steps aren't working for me.

Did you make sure you copied over the `debugAdapters` folder in your local version of the extension? It's something I forgot to do when I started testing this and it confused me for a bit.

bobbrow:
> Can this be fixed?
> 
> image

Yes, I think it should have the "Preview" part of the installation name, but it's missing.

bobbrow:
> Can you add the command name?
> 
> image
> 
> It's possible the command may not have been invoked from the developer command prompt or in a way the user is 100% sure what command is run, such as via a keyboard shortcut.
> 
> Umm...unless VS Code itself is showing that message, I haven't checked yet.

Yes, I can add the command name. This is our message. The command shouldn't be visible in the Command Palette on linux/mac though. If you are seeing it there, please tell me there is a bug.

bobbrow:
> FYI, it was able to set the developer environment when I didn't have a workspace open. Not sure what effect that has.
> 
> image

This feature adds to the environment variables, so if you open a new terminal in a no-workspace window, you should see the VS devenv applied there.

sean-mcmanus:
> Did you make sure you copied over the `debugAdapters` folder in your local version of the extension? It's something I forgot to do when I started testing this and it confused me for a bit.

Yeah, I was on a new machine -- copying over the 1.25.3 vsix contents fixed it.

sean-mcmanus (changes requested):
- [x] In the Debug Open File scenario, I seem to get no dropdown to choose which VS environment to use. Is that a "by design" limitation of not having a folder open? Debugging seems to work somehow, but it shows no popup with the Apply Developer Environment option.

UPDATE: Oops, I think it was because I had previous did the Apply (last week) in the no open folder scenario -- that's "by design", right? I get the Apply option again after I do a Clear.

[bobbrow] Yes, you may also have to close any left over terminals from the previous devenv state before trying future commands too.

sean-mcmanus (changes requested):
- [x] Is it supposed to show the error "The Visual Studio Developer Environment was not applied. Please try again or run VS Code from the Developer Command Prompt for VS." after using the Debug button twice without clicking an option in the Apply Developer Environment popup? It seems unexpected, i.e. it doesn't make sense to tell the user that the dev environment wasn't applied when they never clicked a button to apply it in the first place.

[bobbrow] Interesting bug, I believe I have a fix for this.

sean-mcmanus (changes requested):
- [x] I still see the error "The Visual Studio Developer Environment was not applied. Please try again or run VS Code from the Developer Command Prompt for VS." when I hit the X/Close button in the top right of the apply dialog, but it doesn't appear when I hit the Cancel button. It seems like it shouldn't show when the user hits the X button since it's pretty obvious that it wasn't applied.

[bobbrow] I have a fix for this

sean-mcmanus (changes requested):
- [x] When the dropdown appears to select an environment, I get no "The Visual Studio Developer Environment was not applied. Please try again or run VS Code from the Developer Command Prompt for VS." when I use the Set command, but I get that error when I click the button (which is inconsistent). I think in both cases the error message should not appear since it should normally be obvious that a quick pick option was not clicked, therefore not applied. Even if there is confusion in some rare case, the user will get the Apply option when it fails the 2nd time, so it doesn't seem necessary to explicitly tell the user the environment was not set -- except potentially if there was some unexpected failure condition after the user actually did click the options.

[bobbrow] My thought process is this:
* In your first case (run the "Set VS Developer Environment" command), no message is provided because the command is the primary action and it was clear that you cancelled it by dismissing the drop down.
* In your second case (click the debug button which spawns the Apply button), a message is provided because the primary command (BuildAndDebug) can no longer continue due to a secondary operation (SetVSDevEnvironment) which took the focus but ended up being cancelled. I am changing the message to tie the consequence of the second action back to the primary action thus restoring the context back to that action.
    * I will be keeping a message for this scenario. I welcome any feedback on the message provided.

Resolution: Reworded the error message.

sean-mcmanus (changes requested):
- [x] It doesn't work if you run the Configure Default Build Task (or any steps to create a tasks.json) and then press the F5 button or the "Run and Debug" side panel button (no Apply Dev Env option appears): 
    image

    Alternate repro step are to do the Apply (generates the tasks.json), then do the Clear, and then do F5.

    The Apply popup appears with F5 if no tasks.json exists.

[bobbrow] Good bug. It uncovered another issue too. I fixed them both.

bobbrow:
I'm going to put checkboxes and responses in your messages directly, since it's somewhat difficult to track these issues outside of comments directly in the code.  I hope that's ok.

sean-mcmanus:
> I'm going to put checkboxes and responses in your messages directly, since it's somewhat difficult to track these issues outside of comments directly in the code. I hope that's ok.

Sure, you had mentioned making a comment in the code, but I wasn't referring to any particular code. I'm not sure why GitHub doesn't have a UI that allows a reply thread that is easy to use.

sean-mcmanus (changes requested):
- [x] I get no option to select the debugger type of GDB when I do F5 with no folder open. It works in 1.25.2. Using the Clear option doesn't fix it. I repro the same issue with a workspace folder open too, but you have to make sure no tasks.json/launch.son exists. Oh, the repro might require certain conditions. I don't understand why I sometimes repro it and other times I don't. UPDATE: Okay, it seems like there might be some persistent state being saved even after the Clear is used (and after a reload window). It stopped reproing when I switched from VS Code Insiders to non-Insiders, but after I ran the Apply and then Clear, it started reproing for non-Insiders.

[bobbrow] I'll mark this as resolved since it doesn't sound like it's related to my change.

But it seems to only start reproing after I apply the developer environment one time? Also, I try 1.25.2 and the bug doesn't repro, but it starts to repro when I use your branch. Oh, I repro it with main too...

UPDATE: Yeah, it seems like a bug with VS Code itself. I can't figure out how to change the debugger type after I select one for a workspace folder.

bobbrow:
I addressed your last set of feedback.

Extension/package.json
```diff
@@ -3376,6 +3376,12 @@
                         "default": "default",
                         "markdownDescription": "%c_cpp.configuration.copilotHover.markdownDescription%",
                         "scope": "window"
+                    },
+                    "C_Cpp.persistVSDeveloperEnvironment": {
```
sean-mcmanus: It looks like CMake has the `s` lower cased for `VS` -- that makes it look better in the settings UI since it treats `Vs` as a word versus `VSDeveloper` as a word.

image

Extension/package.nls.json
```diff
@@ -1025,15 +1028,15 @@
     },
     "c_cpp.walkthrough.create.cpp.file.altText": "Open a C++ file or a folder with a C++ project.",
     "c_cpp.walkthrough.command.prompt.title": {
-        "message": "Launch from the Developer Command Prompt for VS",
+        "message": "Apply the Visual Studio Developer Environment",
```
sean-mcmanus: I think `Developer Environment` should be lower cased, right? Why is it upper cased? Usually it would only be upper cased if it was a product name, but I don't think it is. Using upper case might confuse the translators too. It's not upper cased in `c_cpp.configuration.persistVSDeveloperEnvironment.markdownDescription`.

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -3541,6 +3547,16 @@
                 "category": "C/C++",
                 "icon": "$(run)"
             },
+            {
+                "command": "C_Cpp.SetVSDevEnvironment",
```
sean-mcmanus: Can all these commands be unified to `VsDeveloper` instead of `VSDev`? That makes it easier to find all the related code.

Extension/package.nls.json
bobbrow: The text for all commands should be using Title Casing.  Settings descriptions use sentence casing.

Extension/package.json
bobbrow: We don't have very good options here.  "Vs" looks more like "versus" to me but I agree it's better to split VS from Developer. (This is part of the reason why I omitted VS from the setting/command in the first iteration)

Extension/package.json
bobbrow: sure

sean-mcmanus (changes requested):
If VS Code is already launched from a particular VS developer command prompt, then the Set VS Developer Environment command is still available and runnable, but confusingly has no effect with no warning that it's not actually used, i.e. it just uses the original environment. Can that be fixed somehow or give a warning? Or is it supposed to work and switch to the newly selected developer environment?

bobbrow:
> If VS Code is already launched from a particular VS developer command prompt, then the Set VS Developer Environment command is still available and runnable, but confusingly has no effect with no warning that it's not actually used, i.e. it just uses the original environment. Can that be fixed somehow or give a warning? Or is it supposed to work and switch to the newly selected developer environment?

I would expect it to override the inherited environment. Does it not?

Extension/package.nls.json
sean-mcmanus: @bobbrow I don't understand -- `c_cpp.walkthrough.command.prompt.title` is just an arbitrary string right? You could change "command" in the string to anything. It's not actually a "command" that requires title casing. It's just a label for a string we show users in a button. Are you saying all button text should be Title cased?

sean-mcmanus:
> I would expect it to override the inherited environment. Does it not work for you? It worked for me.

It doesn't override the inherited environment. I confirm that via running `cl.exe --version` with the restarts terminal. How are you confirming that?

Extension/package.nls.json
sean-mcmanus: Also, it uses capital Developer Environment in the string `"{0} requires the Visual Studio Developer Environment.` and other messages that aren't part of a "command".

sean-mcmanus: Oh, but it does look like the recommended style for button text is to use title case, so that seems fine to me, but the other cases of "Developer Environment" that aren't in the button text should be non-title cased.

sean-mcmanus (changes requested):
When I choose x86_arm64
image

it appears to incorrectly use
 "amd64_arm64"

image

Similarly, if I choose x86 I get amd64.

It appears that the selected one is correctly used and it's only the popup info message that has the incorrect text, i.e. after choosing x86:
image

It seems to only show amd64 or amd64_arm64 -- the 1st or 2nd option in the popup, even if options 3-6 are picked.

UPDATE: This bug only seems to repro when I launch VS Code from a developer command prompt.

Extension/package.nls.json
bobbrow: > Also, it uses capital Developer Environment in the string `"{0} requires the Visual Studio Developer Environment.` and other messages that aren't part of a "command".

This one should already be fixed in the latest update.

bobbrow: > @bobbrow I don't understand -- `c_cpp.walkthrough.command.prompt.title` is just an arbitrary string right? You could change "command" in the string to anything. It's not actually a "command" that requires title casing. It's just a label for a string we show users in a button. Are you saying all button text should be Title cased?

I'm not sure where I went wrong here. For some reason I thought you were referring to the "persist" setting. In any case, I looked through the other walkthrough page titles and it looks like we're not totally consistent on the capitalization. The VS Code documentation pages also are inconsistent, but I do see enough sentence casing in the page titles that I'm fine if we standardize to that. I'll update the other strings in this PR as well.

bobbrow:
Ok, I did notice an oddity that I will need to debug. Something wrong did happen when the dev environment is already set and it does appear to fail to update all of the required variables.
image

image

I originally only checked that the compiler was coming from the correct path, but you are right that the environment is not correct. (e.g. VSINSTALLDIR is wrong in that screenshot)

bobbrow:
I have a fix for the "run from the developer command prompt" scenario in the node module. I just need to wait for it to show up in CFS so I can consume it.

https://github.com/bobbrow/node-vcvarsall/commit/5556a2851d1ae968302b2da65b7b54431560bb39

bobbrow:
The latest push should address the remaining issues.

Extension/package.nls.json
```diff
@@ -1025,15 +1028,15 @@
...
-            "{Locked=\"Developer Command Prompt for VS\"}"
+            "{Locked=\"Visual Studio\"}"
         ]
     },
     "c_cpp.walkthrough.command.prompt.description": {
-        "message": "When using the Microsoft Visual Studio C++ compiler, the C++ extension requires you to launch VS Code from the Developer Command Prompt for VS. Follow the instructions on the right to relaunch.\n[Reload Window](command:workbench.action.reloadWindow)",
+        "message": "When using the Microsoft Visual Studio C++ compiler, the Visual Studio developer environment must be present.\n\nFollow the instructions on the right to relaunch or click the button below.\n[Set Developer Environment](command:C_Cpp.SetVsDeveloperEnvironment?%22walkthrough%22)",
         "comment": [
-            "{Locked=\"Visual Studio\"} {Locked=\"C++\"} {Locked=\"VS Code\"} {Locked=\"Developer Command Prompt for VS\"} {Locked=\"\\\n[\"} {Locked=\"](command:workbench.action.reloadWindow)\"}"
+            "{Locked=\"Visual Studio\"} {Locked=\"C++\"} {Locked=\"\\\n[\"} {Locked=\"](C_Cpp.SetVsDeveloperEnvironment?%22walkthrough%22)\"}"
```
sean-mcmanus: `command:` is missing from the Locked string.

sean-mcmanus (changes requested):

Extension/package.nls.json
```diff
@@ -843,6 +845,7 @@
         ]
     },
     "c_cpp.configuration.debugShortcut.description": "Show the \"Run and Debug\" play button and \"Add Debug Configuration\" gear in the editor title bar for C++ files.",
+    "c_cpp.configuration.persistVsDeveloperEnvironment.markdownDescription": "Remember the last used Visual Studio developer environment for the current workspace. This setting is only applicable for Windows.",
```
sean-mcmanus: `markdownDescription` should be `description` because there is no markdown in the text.

Extension/package.nls.json
bobbrow: fixed

sean-mcmanus (changes requested):

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,214 @@
...
+        throw new Error(localize('something.wrong', 'Something went wrong: {0}', JSON.stringify(vars)));
+    }
+
+    const host = vars['VSCMD_ARG_HOST_ARCH'];
+    const target = vars['VSCMD_ARG_TGT_ARCH'];
+    const arch = vcvars.getArchitecture({
+        host: match(host, { 'x86': 'x86', 'x64': 'x64' }) ?? 'x64',
+        target: match(target, { 'x86': 'x86', 'x64': 'x64', 'arm64': 'ARM64', 'arm': 'ARM' }) ?? 'x64'
+    });
+    const settings = new CppSettings(vscode.workspace.workspaceFolders?.at(0)?.uri);
```
sean-mcmanus: No workspace uri should be passed to CppSettings here -- it's causing this warning in the extension host logging: `

[warning] [ms-vscode.cpptools] Accessing a window scoped configuration for a resource is not expected. To associate 'C_Cpp.persistVsDeveloperEnvironment' to a resource, define its scope to 'resource' in configuration contributions in 'package.json'

sean-mcmanus (changes requested):

Extension/package.json
```diff
@@ -3376,6 +3376,12 @@
                         "default": "default",
                         "markdownDescription": "%c_cpp.configuration.copilotHover.markdownDescription%",
                         "scope": "window"
+                    },
+                    "C_Cpp.persistVsDeveloperEnvironment": {
```
sean-mcmanus, line 3542: I have this set to false, but it's still persisting, i.e. the setting doesn't seem to be working, even when I reload window.

sean-mcmanus (changes requested):

Extension/package.json
sean-mcmanus: Oh, maybe I'm misunderstanding what the setting is supposed to do? When I reload window, is the vs developer environment supposed to be cleared on startup the next time?

bobbrow: If you have it set to false, our environment override should not be there the next time you start a session in this workspace. Are you still running in the Developer Command Prompt and maybe confusing that environment with the environment override from the Set VS Developer Environment command?

And the value of this setting is applied at the time you run the command. So maybe the feedback here is that we should notify the user that the setting will not take effect until the next time you run the command.  Or maybe there's a way for me to reset the override to use the new setting value.

bobbrow: I'll see if I can change the persist value on the fly when the user changes it.

sean-mcmanus: Ok, if the setting is true and I run the Set and then change the setting to false, it still persists, but it doesn't persist if I use the Set when the setting is false -- is that intentional? I would've expected previously called Set's to not persist when the setting changed to false without having to manually run the Clear (or Set again)

sean-mcmanus: I'm not running VS Code from the developer prompt. If the persisted Set were re-run when the user changed the setting to false, that might fix it (since that fixes it if the user manually does that).

And the same issue happens "in reverse" if the user changes the setting to true and doesn't re-run the command (the existing environment isn't persisted after reload window).

bobbrow: Changing it as a reaction to a settings change seems to work fine. So now you will get whatever behavior the setting is currently set to when you close the window.

Extension/src/LanguageServer/devcmd.ts
bobbrow: fixed

Extension/package.nls.json
bobbrow: fixed

sean-mcmanus (changes requested):
The Apply Developer Environment popup doesn't appear if the VS install for the persisted environment is changed (removed or the version changes). It just fails to find the cl.exe.

image

It seems like user could be confused why it stopped working and they may not know they need to run the Clear/Set again. Are we able to detect that condition and do the Clear ourselves?

bobbrow:
> The Apply Developer Environment popup doesn't appear if the VS install for the persisted environment is changed (removed or the version changes). It just fails to find the cl.exe.
> 
> [image]
> It seems like user could be confused why it stopped working and they may not know they need to run the Clear/Set again. Are we able to detect that condition and do the Clear ourselves?

Are you talking about when you upgrade VS and get a new toolset version for cl.exe? We can probably check for that.

sean-mcmanus:
> Are you talking about when you upgrade VS and get a new toolset version for cl.exe? We can probably check for that.

Yes upgrade or uninstall, e.g. uninstall Preview but you have non-Preview still then ideally the Apply option would appear for the non-Preview that is still installed.

sean-mcmanus (approved):

sean-mcmanus (approved):

Extension/src/LanguageServer/extension.ts
```diff
@@ -1558,3 +1561,28 @@ async function showCopilotContent(copilotHoverProvider: CopilotHoverProvider, ho
...
+        if (sender !== 'buildAndDebug') {
+            void vscode.window.showInformationMessage(`${util.extensionContext?.environmentVariableCollection.description} successfully set.`);
+        }
+    } catch (error: any) {
+        success = false;
+        if (!isWindows) {
+            throw error;
+        }
+        if (error.message !== errorOperationCancelled) {
+            void vscode.window.showErrorMessage(`Failed to apply VS developer environment: ${error.message}`);
```
sean-mcmanus: This should be `apply the VS developer environment`.

sean-mcmanus (changes requested):

sean-mcmanus (dismissed):

Colengms (dismissed):

sean-mcmanus (dismissed):

Extension/package.json
```diff
@@ -3378,6 +3378,12 @@
...
+                        "type": "boolean",
+                        "default": true,
+                        "markdownDescription": "%c_cpp.configuration.persistVsDeveloperEnvironment.description%",
```
sean-mcmanus: markdownDescription should be description (unless you intended to add some markdown, which I don't see).

Extension/package.nls.json
```diff
@@ -843,6 +845,7 @@
         ]
     },
     "c_cpp.configuration.debugShortcut.description": "Show the \"Run and Debug\" play button and \"Add Debug Configuration\" gear in the editor title bar for C++ files.",
+    "c_cpp.configuration.persistVsDeveloperEnvironment.description": "Remember the last used Visual Studio developer environment for the current workspace. This setting is only applicable for Windows.",
```
sean-mcmanus, line 854: Would using this be preferable? "c_cpp.configuration.persistVsDeveloperEnvironment.markdownDescription": "If `true`, on Windows, remember the last used Visual Studio developer environment for the current workspace."

Extension/package.json
sean-mcmanus: My comment below has a suggestion to mention "If `true`" which would change this to "%c_cpp.configuration.persistVsDeveloperEnvironment.markdownDescription%

Extension/package.nls.json
```diff
@@ -1053,15 +1056,15 @@
...
-            "{Locked=\"Developer Command Prompt for VS\"}"
+            "{Locked=\"Visual Studio\"}"
         ]
     },
     "c_cpp.walkthrough.command.prompt.description": {
-        "message": "When using the Microsoft Visual Studio C++ compiler, the C++ extension requires you to launch VS Code from the Developer Command Prompt for VS. Follow the instructions on the right to relaunch.\n[Reload Window](command:workbench.action.reloadWindow)",
+        "message": "When using the Microsoft Visual Studio C++ compiler, the Visual Studio developer environment must be present.\n\nFollow the instructions on the right to relaunch or click the button below.\n[Set Developer Environment](command:C_Cpp.SetVsDeveloperEnvironment?%22walkthrough%22)",
         "comment": [
-            "{Locked=\"Visual Studio\"} {Locked=\"C++\"} {Locked=\"VS Code\"} {Locked=\"Developer Command Prompt for VS\"} {Locked=\"\n[\"} {Locked=\"](command:workbench.action.reloadWindow)\"}"
+            "{Locked=\"Visual Studio\"} {Locked=\"C++\"} {Locked=\"\\\n[\"} {Locked=\"](command:C_Cpp.SetVsDeveloperEnvironment?%22walkthrough%22)\"}"
```
sean-mcmanus:
```
{Locked=\"\\\n[\"}
```
should be

```
{Locked=\"\n\n\"} {Locked=\"\n[\"}
```

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,238 @@
...
+
+const errorNoContext = localize('no.context.provided', 'No context provided');
+const errorNotWindows = localize('not.windows', 'The "Set Visual Studio Developer Environment" command is only available on Windows');
+const errorNoVSFound = localize('error.no.vs', 'A Visual Studio installation with the C++ compiler was not found');
+export const errorOperationCancelled = localize('operation.cancelled', 'The operation was cancelled');
+const errorNoHostsFound = localize('no.hosts', 'No hosts found');
+const configuringDevEnv = localize('config.dev.env', 'Configuring developer environment...');
+const selectVSInstallation = localize('select.vs.install', 'Select a Visual Studio installation');
+const advancedOptions = localize('advanced.options', 'Advanced options...');
+const advancedOptionsDescription = localize('advanced.options.desc', 'Select a specific host/target architecture, toolset version, etc.');
```
sean-mcmanus: Should `host/target` be `host and target` -- it's inconsistent with the usage in `select.host.target` below.

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    return true;
+}
+
+async function getVSInstallations() {
+    const installations = await vswhere.getVSInstallations({
+        all: true,
+        prerelease: true,
+        sort: true,
+        requires: ['Microsoft.VisualStudio.Component.VC.Tools.x86.x64']
```
sean-mcmanus: What about `Microsoft.VisualStudio.Component.VC.Tools.x86.x64.Spectre`?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/devcmd.ts
sean-mcmanus: Oh, the Spectre one requires the non-specter one so it doesn't seem to need to be handled.

buschtrisha77 (approved):

Extension/package.nls.json
bobbrow: The description usually implies "true", so I'll just make it "description" instead of "markdownDescription"

Extension/src/LanguageServer/devcmd.ts
bobbrow: ok

bobbrow:
~~Hmm, it seems my rebase did not go as planned.  Please wait before reviewing any changes.~~

bobbrow:
This is ready for review now.

sean-mcmanus:
> This is ready for review now.

@bobbrow Oh, sorry, I didn't notice your comment for some reason.

sean-mcmanus (dismissed):
Not sure if Colen wanted to review this.

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,238 @@
...
```
Colengms: Is this going to work on my (work) ARM64 AI Surface Laptop?  Or should `Microsoft.VisualStudio.Component.VC.Tools.ARM64` be specified there?

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,238 @@
...
+async function getHostsAndTargets(vcPath: string): Promise<vcvars.HostTarget[]> {
+    const hosts = await fs.readdir(vcPath);
+    if (hosts.length === 0) {
+        throw new Error(errorNoHostsFound);
+    }
+    const hostTargets: vcvars.HostTarget[] = [];
+    for (const host of hosts) {
+        const h = match<'x86' | 'x64' | undefined>(host.toLowerCase(), { 'hostx86': 'x86', 'hostx64': 'x64' });
+        if (!h) {
+            // skip any arm/arm64 folders because there is no arm compiler
```
Colengms: Is this accurate?  On my ARM64 work laptop, I see `bin\Hostarm64\arm64\cl.exe`.  Maybe you didn't see that because it doesn't get installed on a x64 system?

Colengms (changes requested):

Extension/src/LanguageServer/devcmd.ts
bobbrow: ~~It should. There is no native ARM64 host toolset as far as I know.~~

It seems there is a native arm64 toolset. I'll need to fix up some assumptions then.

Extension/src/LanguageServer/devcmd.ts
bobbrow: Oh, really? There's a native ARM64 toolset?

bobbrow: I guess I need access to a machine with this configuration then.

bobbrow: this is addressed now

sean-mcmanus (approved):

Extension/src/LanguageServer/devcmd.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+function vsDisplayNameWithSuffix(installation: vswhere.Installation): string {
+    const suffix = (() => {
+        if (installation.channelId.endsWith('.main')) {
+            return 'main';
+        } else if (installation.channelId.endsWith('.IntPreview')) {
+            return 'Int Preview';
+        } else if (installation.channelId.endsWith('.Preview')) {
+            return 'Preview';
```
Colengms, line 129: With 18 "Insiders", the `channelId` still ends in `.Preview`.  Maybe return "Insiders" here always? Or maybe check the version number immediately before `.Preview`, and if `18` or greater (i.e. `.18.Preview`), use "Insiders" here instead?  Not critical.

Colengms (approved):

Extension/src/LanguageServer/devcmd.ts
bobbrow: Good catch.  I will commit this as-is and follow up with another PR shortly after.

PR #13580, opened 2025-05-01
IntelliSense string updates.

sean-mcmanus:
Translations are pending in a week or so.


bobbrow (dismissed):

sean-mcmanus:
@bobbrow I updated it -- I just noticed they added new strings in the middle of the list instead of only at the end.

sean-mcmanus:
@bobbrow There's another problem -- if the updated strings have different `%` uses, then it could potentially crash and those strings need to be switched to matching English versions. i.e. I need to review all those...

sean-mcmanus:
@bobbrow And we may need to add the English EDG string copies to this GitHub repo for out-of-band translation from VS, not sure yet.

bobbrow:
> @bobbrow There's another problem -- if the updated strings (i.e. pre-existing, non-new ones) have different `%` uses, then it could potentially crash and those strings need to be switched to matching English versions. i.e. I need to review all those...

ok. Thanks for the extra attention to detail.

Colengms (approved):

sean-mcmanus:
Don't check this in until the FI gets in.

PR #13593, opened 2025-05-08
ignore speculative-edit requests

lukka:


sean-mcmanus (approved):
I don't know what a speculative/proposedEdit is supposed to be. Is that publicly documented somewhere?

Issue #13604, opened 2025-05-13
IntelliSense incorrectly matches file after setting "recursiveIncludes": {"order": "depthFirst"}


Amy-Li03:
### Environment

- OS and Version: Win11 23H2
- VS Code Version: 1.100.0
- C/C++ Extension Version: 1.25.3
- If using SSH remote, specify OS of remote machine: NO

#### Bug Summary:

My workspace have two files of the same name: common.h, one is under root folder->includeA->sub->common.h, the other is under root folder->includeA->common.h:
1. When I setting  "recursiveIncludes": {"order": "**depthFirst**"}, the intelliSense process **incorrectly** matches the file under root folder->includeA->common.h, it should match the file under root folder->includeA->sub->common.h
2. When I setting  "recursiveIncludes": {"order": "**breadthFirst**"}, the intelliSense process **correctly** matches the file under root folder->includeA->common.h

#### Steps to reproduce:
1. Make sure the version of C/C++ extension is updated to 1.25
2. Extract and open attached project with VS code
   [Order.zip](https://github.com/user-attachments/files/20184215/Order.zip)
3. Add follow code to c_cpp_properties.json and save

          "includePath": [
                "${workspaceFolder}/**"
            ],
            "recursiveIncludes": {
                "order": "depthFirst"
            }

4. Open main.cpp file and observe the intelliSense

![Image](https://github.com/user-attachments/assets/ec1c910f-ebaf-4c7c-8c2b-16d5e93b285e)

#### Expected behavior:
No intelliSense error appears

### Configuration and Logs

```shell
-------- Diagnostics - 5/13/2025, 4:38:12 PM
Version: 1.25.3
Current Configuration:
{
    "name": "Win32",
    "includePath": [
        "c:/Users/v-amyli3/Desktop/Blog/Order/**"
    ],
    "recursiveIncludes": {
        "order": "depthFirst"
    },
    "defines": [
        "_DEBUG",
        "UNICODE",
        "_UNICODE"
    ],
    "windowsSdkVersion": "10.0.26100.0",
    "cStandard": "c17",
    "cppStandard": "c++17",
    "intelliSenseMode": "windows-msvc-x64",
    "compilerPathIsExplicit": false,
    "cStandardIsExplicit": true,
    "cppStandardIsExplicit": true,
    "intelliSenseModeIsExplicit": true,
    "recursiveIncludesReduceIsExplicit": false,
    "recursiveIncludesPriorityIsExplicit": false,
[...]

PR #13607, opened 2025-05-13
Rename `depthFirst` setting value to `preOrder`

Colengms:
As pointed out by this issue: https://github.com/microsoft/vscode-cpptools/issues/13604 
`depthFirst` is not technically correct. A more accurate setting value would be `preOrder`.


sean-mcmanus (approved):

bobbrow (changes requested):
Hold on to this for a minute. I'd like to have a discussion before changing the name.

Issue #13637, opened 2025-05-25
doesn't work under custom devcontainer


icegood:
### Environment

- OS and Version:  docker 
Linux ice-home 6.14.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Sun Apr  6 15:05:05 UTC 2025 x86_64 
x86_64 x86_64 GNU/Linux
under root 
- VS Code Version:  
-Version: 1.100.2
Commit: 848b80aeb52026648a8ff9f7c45a9b0a80641e2e
Date: 2025-05-14T21:47:40.416Z
Electron: 34.5.1
ElectronBuildId: 11369351
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Linux x64 6.14.0-15-generic

- C/C++ Extension Version: 1.26.1
- If using SSH remote, specify OS of remote machine: 
rooted container

### Bug Summary and Steps to Reproduce

![Image](https://github.com/user-attachments/assets/88b48af1-3867-4181-a5f6-bf5762b04915)

Bug Summary:

Steps to reproduce:
1. prepare own cotainer. Run it as root
2. install extension under devcontainer
3. try use (say, goto definition of function)   

Expected behavior:
works

### Configuration and Logs

```shell
2025-05-25 08:17:40.416 [info] Extension host with pid 852 started
2025-05-25 08:17:40.418 [info] Lock '/root/.vscode-server/data/User/workspaceStorage/8030c29de55fbb539249111912790f17/vscode.lock': Lock acquired.
2025-05-25 08:17:40.535 [info] ExtensionService#_doActivateExtension vscode.emmet, startup: false, activationEvent: 'onLanguage'
2025-05-25 08:17:40.536 [info] ExtensionService#_doActivateExtension KylinIdeTeam.cppdebug, startup: false, activationEvent: 'onLanguage:c'
2025-05-25 08:17:40.536 [info] ExtensionService#_doActivateExtension ms-vscode.cpptools, startup: false, activationEvent: 'onLanguage:c'
2025-05-25 08:17:40.769 [info] ExtensionService#_doActivateExtension vscode.configuration-editing, startup: false, activationEvent: 'onLanguage:jsonc'
2025-05-25 08:17:40.770 [info] ExtensionService#_doActivateExtension vscode.json-language-features, startup: false, activationEvent: 'onLanguage:jsonc'
[...]

PR #13643, opened 2025-05-27
Handle multiple activation.

sean-mcmanus:
Potential fix for https://github.com/microsoft/vscode-cpptools/issues/13637


Colengms:
I added a comment to the issue.  I don't think the problem is multiple activations.

Issue #13660, opened 2025-06-02
Go to Definition (F12) will list many redundant files


gtianw:
### Environment

- OS and Version: Windows_NT x64 10.0.19045
- VS Code Version: 1.100.2 (system setup)
- C/C++ Extension Version: 1.25.3
- If using SSH remote, specify OS of remote machine: WSL2, Ubuntu20.04

### Bug Summary and Steps to Reproduce

Bug Summary:
Go to Definition will list many redundant files.
These files are not included by CMakeLists.txt or compile_commands.json.
duplicates #11887, but it is no longer active.

Steps to reproduce:
1. Download code from https://github.com/gtianw/Issue/tree/cpptools/13660
2. F12 or Ctrl+LeftClick on `target_prt`.
3. 
![Image](https://github.com/user-attachments/assets/24baf2dc-adf5-48c2-9c17-0b0ecf9efa19)

Expected behavior:
Only parse the file included by CMakeLists.txt or compile_commands.json, and directly jump to function definitions.

gtianw:
In a real project, there may be dozens of targets.

sean-mcmanus:
@gtianw It looks like this is caused by the parent folder of main.cpp being the workspaceFolder root, so both CMake Tools and compile commands are recursively parsing the workspaceFolder for symbols. I believe that is "by design" but it's possible we could implement some design change. You could potentially use

```json
    "C_Cpp.files.exclude": {
        "**/target2": true,
        "**/target3": true
    }
```

@Colengms might have additional workaround suggestions.

gtianw:
I will switch between multiple targets according to the usage scenario, and very frequently, so C_Cpp.files.exclude is not suitable.

sean-mcmanus:
@gtianw Are you able to not have the parent folder be used, i.e. by not having any source files at the root?

sean-mcmanus:
@gtianw Otherwise, did you have some sort of design change in mind (that doesn't break existing user scenarios) to prevent files at the rule from causing recursive tag parsing of all the subfolders?

gtianw:
> [@gtianw](https://github.com/gtianw) Are you able to not have the parent folder be used, i.e. by not having any source files at the root?

I tried, but the problem persists.

gtianw:
[...]

PR #13678, opened 2025-06-07
Apply mergeConfigurations for browse.path, auto-update the config, and avoid accumulating old configs

sean-mcmanus:
Fix for the configurationProvider case of https://github.com/microsoft/vscode-cpptools/issues/13660, https://github.com/microsoft/vscode-cpptools/issues/13687, and https://github.com/microsoft/vscode-cpptools/issues/13688 (the changed mergeable property isn't auto-updated and old config values are accumulated). https://github.com/microsoft/vscode-cpptools/issues/13689 is not fixed.

Also, updated the description of mergeConfigurations.

Also, removed the "string" type for mergeConfigurations, since that appears to be invalid.

mergeConfigurations currently is not used by compileCommands (so that behavior is unchanged and has browse.path merged, but other properties not merged unless there is a fallback from a file not configured from compileCommands).


Extension/package.json
```diff
@@ -841,6 +841,7 @@
                     },
                     "C_Cpp.default.mergeConfigurations": {
                         "type": "boolean",
+                        "default": false,
```
sean-mcmanus, line 844: The "false" was already implicitly being used as the default, so setting it to false here just made it explicit was already being used as the default.

Extension/src/LanguageServer/client.ts
```diff
@@ -2181,6 +2186,7 @@ export class DefaultClient implements Client {
             if (configs && configs.length > 0 && configs[0]) {
                 const fileConfiguration: configs.Configuration | undefined = this.configuration.CurrentConfiguration;
                 if (fileConfiguration?.mergeConfigurations) {
+                    configs = deepCopy(configs);
```
sean-mcmanus, line 2189: This line change fixes the duplicates/old values being accumulated: https://github.com/microsoft/vscode-cpptools/issues/13687

Extension/src/common.ts
```diff
@@ -1810,3 +1810,21 @@ export function findExePathInArgs(args: CommandString[]): string | undefined {
 export function getVsCodeVersion(): number[] {
     return vscode.version.split('.').map(num => parseInt(num, undefined));
 }
+
+export function equals(array1: string[] | undefined, array2: string[] | undefined): boolean {
```
sean-mcmanus, line 1814: TypeScript confusingly was returning false with array1 == array2 even though the contents of the arrays were identical.

Extension/package.json
sean-mcmanus: I intended to just a leave a comment -- not sure why it's adding the "Resolve Conversation" button.

Colengms (approved):

PR #13686, opened 2025-06-10
Remove old entries from changelog

### LINKED ISSUE #13628: changelog is too large to view
Image

bobbrow:
GitHub is unable to load our large changelog in the web view. This change removes entries prior to version 1.0.0


sean-mcmanus (changes requested):
Can we link to a changelog2.md with the older logs (Copilot mentioned doing that instead)? There might be value in viewing the older changelogs.

bobbrow:
> Can we link to a changelog2.md with the older logs (Copilot mentioned doing that instead)? There might be value in viewing the older changelogs.

No, I would rather not. The older change logs will be in the file history in case anyone needs to look at them for some reason.

sean-mcmanus (approved):

PR #13702, opened 2025-06-15
chore: remove Visual Studio-related files for cross-platform cleanup

Subham-KRLX:
This PR removes the `build.cmd` batch file from the `Code Samples/Fib/` folder.

Since this file is specific to Windows and Visual Studio environments, removing it helps streamline the repository for cross-platform development and makes it cleaner for non-Windows contributors.

This is part of a minor cleanup effort to reduce legacy or unused platform-specific files.


sean-mcmanus (changes requested):
I don't understand why this file is being removed. It's still referenced in Code Samples/Fib/.vscode/tasks.json for the Windows section of tasks.json.

Subham-KRLX:
Thanks for the review, @sean-mcmanus! I've now removed the reference to `build.cmd` from `tasks.json` as you pointed out. Let me know if there's anything else you'd like me to update.

Code Samples/Fib/.vscode/tasks.json
```diff
@@ -1,50 +1,41 @@
...
+        "kind": "build",
+        "isDefault": true
+      },
+      "presentation": {
+        "echo": true,
+        "reveal": "always",
+        "focus": false,
+        "panel": "shared"
+      },
+      "linux": {
```
Colengms, line 17: Hi @Subham-KRLX .  Wouldn't this prevent this code sample from working correctly on Windows (i.e. with MinGW G++) ?  I think we would want to retain the window specific arguments section.

I believe the `build.cmd` file is working around a known issue with MinGW that requires its bin directory to be in path when run, in order to properly resolve dependent DLLs.  It may be possible to move that work-around into the TypeScript code in order to get rid of `build.cmd`.  But I think we would still want to use of this code sample on Windows.

Colengms (changes requested):

Code Samples/Fib/.vscode/tasks.json
Colengms: Ah, I think the see the real issue here.  This code sample directly uses pthreads, so does not compile on Windows.  I think the ideal fix here would be for the sample to be updated to provide a Windows implementation as well.

Subham-KRLX:
@sean-mcmanus @Colengms: Understood the issues with build.cmd and MinGW. I’ll:

Revert the build.cmd removal (keep Windows support)

Propose a follow-up PR to replace pthreads with C++11 <thread> for true cross-platform support

Will that work? Or should we document the Linux/MacOS-only limitation instead?

Colengms:
> @sean-mcmanus @Colengms: Understood the issues with build.cmd and MinGW. I’ll:
> 
> Revert the build.cmd removal (keep Windows support)
> 
> Propose a follow-up PR to replace pthreads with C++11 for true cross-platform support
> 
> Will that work? Or should we document the Linux/MacOS-only limitation instead?

Reverting the `build.cmd` removal and keeping the Windows support (despite that it can't be built successfully on Windows, due to lack of pthreads) would seem to equate to simply abandoning this PR - no changes needed.  Making the sample properly cross-platform seems simple enough. Would you like to open an issue to track that?

Subham-KRLX:
Thanks all for the review—looks like the consensus is to keep build.cmd.
I’ll close this PR now.
I’ll also open a new issue to replace the sample’s pthread calls with standard C++11 <thread> for cross-platform compatibility.
Reference: #13702.

PR #13757, opened 2025-07-11
Update changelog for 1.27.0

sean-mcmanus:
I decided to keep the buggy-looking auto-generated LLVM TPN notice since the @github/copilot-language-server license is similar in style.


sean-mcmanus:
I think we might want to take another fix...

Colengms (approved):

PR #13778, opened 2025-07-22
Fix the description of debugServerPath

redstrate:
This mentions the non-existent `miDebugServerAddress`, but the correct name is actually `miDebuggerServerAddress`.


sean-mcmanus (approved):

sean-mcmanus:
@redstrate Your fix is in https://github.com/microsoft/vscode-cpptools/releases/tag/v1.27.0

PR #13809, opened 2025-08-04
Enable CG trigger on insiders branch

bobbrow:


sean-mcmanus (changes requested):
What does this fix? Is this the "non-production" branch issue? The bug I was seeing was with the "release" branch. Won't this cause the insiders branch to have the same bug? Isn't the correct fix to remove the "release" branch or is there some other fix pending?

bobbrow:
> What does this fix? Is this the "non-production" branch issue? The bug I was seeing was with the "release" branch. Won't this cause the insiders branch to have the same bug? Isn't the correct fix to remove the "release" branch or is there some other fix pending?

This doesn't fix the problem we have with release showing as non-prod, but it allows me to disable the override in the pipeline triggers and have the yaml file be the source of truth instead.

sean-mcmanus (approved):

Issue #13808, opened 2025-08-04
Intellisense is not happy when separating .h and .tpp


bhubavenski:
### Environment

OS and Version: Windows 11 Pro 24H2
VS Code Version: 1.102.1
C/C++ Extension Version: 1.26.3
If using SSH remote, specify OS of remote machine: x

### Bug Summary and Steps to Reproduce

Bug Summary: I am trying to create a template class separating its declaration in `.h` file and its implementation in `.tpp` file.  Both files have `#pragma` once and the code compiles correctly, everything works, except the intellisense . It's arguing about redefinition error.

This error happens in my `MyArray.tpp` file: 
```
redefinition of 'MyArray<T, n>'C/C++(clang-diagnostic-error)
MyArray.tpp(6, 16): previous definition is here
template<class T, size_t n> MyArray<T, n>::MyArray(std::initializer_list<T> elements)
```

Steps to reproduce:
1. Create `include/<file>.tpp and <file>.h` , create `main.cpp` 
2. write declaration and implementation code
3. `#include "<file>.tpp"` on the bottom of `<file>.h` and `#include <file>.h` on the top of `<file>.tpp`

Expected behavior: intellisense should not be confused. The exact code works well in Visual Studio

### Configuration and Logs

```shell
{
    "configurations": [
        {
            "name": "Win32",
            "includePath": [
                "${workspaceFolder}/**",
                "${vcpkgRoot}/x64-windows/include"
            ],
            "defines": [
                "_DEBUG",
                "UNICODE",
                "_UNICODE"
            ],
            "windowsSdkVersion": "10.0.26100.0",
            "compilerPath": "cl.exe",
            "cStandard": "c23",
            "cppStandard": "c++23",
            "intelliSenseMode": "windows-msvc-x64"
        }
    ],
    "version": 4
}



-------- Diagnostics - 8/4/2025, 6:16:10 PM
Version: 1.26.3
Current Configuration:
{
    "name": "Win32",
    "includePath": [
        "c:/Users/bobkata75/Desktop/sandbox-cpp/**",
        "E:/vcpkg/installed/x64-windows/include"
    ],
    "defines": [
        "_DEBUG",
        "UNICODE",
        "_UNICODE"
    ],
    "windowsSdkVersion": "10.0.26100.0",
    "compilerPath": "cl.exe",
    "cStandard": "c23",
    "cppStandard": "c++23",
    "intelliSenseMode": "windows-msvc-x64",
[...]

PR #13811, opened 2025-08-04
Handle .txx/tpp headers.

sean-mcmanus:
Add .txx/.tpp support from https://github.com/microsoft/vscode-cpptools/issues/13808 . The actual fix for the bug in the issue is actually on the cpptools side though.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2647,7 +2647,7 @@ export class DefaultClient implements Client {
             });
 
             // TODO: Handle new associations without a reload.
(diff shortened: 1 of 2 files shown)


bhubavenski:
if you can add also `.ipp` as for implementation file or better add it as an option to the user to specify in what extension his header files are ending. ( if it's possible )

sean-mcmanus:
> if you can add also `.ipp` as for implementation file or better add it as an option to the user to specify in what extension his header files are ending. ( if it's possible )

ipp is already handled.

Colengms (approved):

PR #13877, opened 2025-08-26
Register the CPP context provider with Copilot Chat as well. 

dbaeumer:
We are in the process of moving Inline completions from Copilot client to Copilot chat. For a certain time we therefore need to register the context provider with both Copilot client and Copilot chat. Which path is exercised will be determined by VS Code. So for the same [document,position] pair only one resolve request will occurr.

Registering with Copilot Chat has also the advantage that context is resolved during NES. If you want to enable this you can use the setting `"github.copilot.chat.advanced.inlineEdits.xtabProvider.languageContext.enabled"`

### DIFF
--- a/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
+++ b/Extension/src/LanguageServer/copilotCompletionContextProvider.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All Rights Reserved.
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-import { ContextResolver, ResolveRequest, SupportedContextItem } from '@github/copilot-language-server';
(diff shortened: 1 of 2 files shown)


lukka:
@dbaeumer Would you like me to test these changes locally? I’m happy to do that.
Also, regarding the use of the same context for NES: is this considered an experimental feature that’s off by default? I’m assuming so, since it looks like the setting needs to be manually enabled via:
`github.copilot.chat.advanced.inlineEdits.xtabProvider.languageContext.enabled`

dbaeumer:
@lukka I tested it locally with Copilot client & Copilot chat and ensured that you are called in both situations.

And yes `github.copilot.chat.advanced.inlineEdits.xtabProvider.languageContext.enabled` is considered experimental right now.

lukka (approved):
Thanks!

sean-mcmanus (approved):

PR #13878, opened 2025-08-26
Fix for input delays when editing c_cpp_properties.json

### LINKED ISSUE #13591: Freeze/delayed input when editing c_cpp_properties.json
### Environment

- Windows 10 Enterprise 22H2
- VS Code Version: 1.99.3
- C/C++ Extension Version: 1.25.3

Bug Summary: When I edit c_cpp_properties.json it freezes, i.e. the next input is often delayed (5 to 15 seconds). This happens after every few keystrokes. Our codebase has around 8000 C/C++ files with 2 million lines of code.

Steps to reproduce:
See summary.

Expected behavior: No input delay when editing c_cpp_properties.json.

### Configuration and Logs

```shell
{
    "configurations": [
        {
            "name": "Win32",
            "includePath": [
                "${workspaceFolder}/**/Interfaces/**",
                "${workspaceFolder}/**/BxPackaging/**",
                "${workspaceFolder}/**/DeveloperTools/**",
                "${workspaceFolder}/**/RuntimeExtensions/**",
                "${workspaceFolder}/**/Biometrics/**",
                "${workspaceFolder}/**/DataBinding/**",
                "${workspaceFolder}/**/Utility/**"
            ],
            "defines": [
                "WIN32"
            ],
            "windowsSdkVersion": "10.0.22621.0",
            "compilerPath": "cl.exe",
            "cStandard": "c11",
            "cppStandard": "c++11",
            "intelliSenseMode": "windows-msvc-x64",
            "compilerArgs": [
                "/std:c++11"
            ],
            "browse": {
                "path": [
                    "${workspaceFolder}/**/Interfaces/**",
                    "${workspaceFolder}/**/BxPackaging/**",
                    "${workspaceFolder}/**/DeveloperTools/**",
                    "${workspaceFolder}/**/RuntimeExtensions/**",
                    "${workspaceFolder}/**/Biometrics/**",
                    "${workspaceFolder}/**/DataBinding/**",
                    "${workspaceFolder}/**/Utility/**"
                ],
            }
        }
    ],
    "version": 4
}
```

### Other Extensions

_No response_

### Additional context

_No response_

bobbrow:
Fixes: #13591

Some glob operations may be more expensive than users expect. This change does the glob processing async so that other work can continue. As a bonus, I added a performance warning squiggle for when glob path resolution takes more than 10 seconds as this may be unexpected.

### DIFF
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1786,7 +1786,7 @@ export class CppProperties {
 
     private async validatePath(input: string | string[] | undefined, { isDirectory = true, assumeRelative = true, globPaths = false } = {}): Promise<string | undefined> {
         if (!input) {
-            return Promise.resolve(undefined);
+            return undefined;
         }
 
         let errorMsg: string | undefined;
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -1731,8 +1731,13 @@ export class CppProperties {
      * @returns The result of the operation and the time it took to complete in seconds.
      */
     private async timeOperation<T>(operation: () => Promise<T | undefined>): Promise<{ result: T | undefined; duration: number }> {
+        let result: T | undefined;
         const start = process.hrtime();
-        const result = await Promise.resolve(operation());
+        try {
+            result = await operation();
+        } catch {
+            // Suppress errors
+        }
         const diff = process.hrtime(start);
         const duration = diff[0] + diff[1] / 1e9; // diff[0] is in seconds, diff[1] is in nanoseconds
         return { result, duration };
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -24,6 +24,7 @@ import { PersistentFolderState } from './persistentState';
 import { CppSettings, OtherSettings } from './settings';
 import { SettingsPanel } from './settingsPanel';
 import { ConfigurationType, getUI } from './ui';
+import { Deferral } from './utils';
 import escapeStringRegExp = require('escape-string-regexp');
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
@@ -279,13 +280,13 @@ export class CppProperties {
 
         vscode.workspace.onDidChangeTextDocument((e) => {
             if (e.document.uri.fsPath === settingsPath) {
-                void this.handleSquiggles().catch(logAndReturn.undefined);
+                this.handleSquiggles(e.document);
             }
         });
 
         vscode.workspace.onDidOpenTextDocument(document => {
             if (document.uri.fsPath === settingsPath) {
-                void this.handleSquiggles().catch(logAndReturn.undefined);
+                this.handleSquiggles(document);
             }
         });
 
@@ -352,7 +353,7 @@ export class CppProperties {
 
     private onSelectionChanged(): void {
         this.selectionChanged.fire(this.CurrentConfigurationIndex);
-        void this.handleSquiggles().catch(logAndReturn.undefined);
+        this.handleSquiggles();
     }
 
     private onCompileCommandsChanged(path: string): void {
@@ -1590,8 +1591,9 @@ export class CppProperties {
             if (firstParse) {
                 // Check documents that are already open since no event will fire for them.
                 const fsPath = this.propertiesFile.fsPath;
-                if (vscode.workspace.textDocuments.some(doc => doc.uri.fsPath === fsPath)) {
-                    void this.handleSquiggles().catch(logAndReturn.undefined);
+                const document = vscode.workspace.textDocuments.find(doc => doc.uri.fsPath === fsPath);
+                if (document) {
+                    this.handleSquiggles(document);
                 }
             }
         } catch (errJS) {
@@ -1859,7 +1861,22 @@ export class CppProperties {
         return errorMsg;
     }
 
-    private async handleSquiggles(): Promise<void> {
+    private lastConfigurationVersion: number = 0;
+    private handleSquigglesDeferral: Deferral | undefined;
+
+    private handleSquiggles(doc?: vscode.TextDocument): void {
+        // When the active config changes, we don't pass the doc in since the version would not have changed.
+        if (doc?.version !== this.lastConfigurationVersion) {
+            this.lastConfigurationVersion = doc?.version ?? 0;
+            this.handleSquigglesDeferral?.cancel();
+            this.handleSquigglesDeferral = new Deferral(() => void this.handleSquigglesImpl().catch(logAndReturn.undefined), 1000);
+        }
+    }
+
+    /**
+     * Not to be called directly. Use the `handleSquiggles` method instead which will debounce the calls.
+     */
+    private async handleSquigglesImpl(): Promise<void> {
         if (!this.propertiesFile) {
             return;
         }
--- a/Extension/src/LanguageServer/configurations.ts
+++ b/Extension/src/LanguageServer/configurations.ts
@@ -286,7 +286,7 @@ export class CppProperties {
 
         vscode.workspace.onDidOpenTextDocument(document => {
             if (document.uri.fsPath === settingsPath) {
-                this.handleSquiggles(document);
+                this.handleSquiggles();
             }
         });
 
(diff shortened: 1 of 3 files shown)


Extension/src/LanguageServer/configurations.ts
```diff
@@ -280,21 +278,21 @@ export class CppProperties {
         });
 
         vscode.workspace.onDidChangeTextDocument((e) => {
-            if (e.document.uri.fsPath === settingsPath && this.isCppPropertiesJsonVisible) {
+            if (e.document.uri.fsPath === settingsPath) {
```
sean-mcmanus, line 282: I'm seeing this get called 4 times for every edit -- should we save/check the document version? Seems like it would be a pre-existing issue though, maybe a VS Code bug?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -846,7 +846,7 @@ export class CppProperties {
             if (isGlobPattern) {
                 // fastGlob silently strips non-found paths. Limit that behavior to dynamic paths only.
                 const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                    fastGlob.sync(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
+                    await fastGlob.async(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
                 resolvedGlob.push(...matches.map(s => s + suffix));
```
sean-mcmanus, line 851: I'm seeing this accumulate 6 fastGlob work tasks per edit for a big `**`...maybe that's a pre-existing issue? Shouldn't it cancel the previous ones that aren't needed?

sean-mcmanus: ...and when that occurs I don't see a timing warning.

sean-mcmanus, line 851: I'm seeing TypeScript fail with 

```
RangeError: Maximum call stack size exceeded
    at CppProperties.resolveAndSplit
```

on line 850 when this is >146k.

```TypeScript
resolvedGlob.push(...matches.map(s => s + suffix));
```

but that might be pre-existing...I haven't checked yet, but the user doesn't see any warning in that case and it seems like we want to fix that?

Extension/src/LanguageServer/configurations.ts
bobbrow: I updated the code to debounce this

Extension/src/LanguageServer/configurations.ts
bobbrow: It should be improved with the new debouncing logic, but I'll open a bug to investigate further.

bobbrow: I didn't change anything here. Are you saying that the performance warning doesn't appear in this case because of the exception?

bobbrow: I don't have your repro but pushed a change that might handle it. Could you try again?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1712,7 +1724,26 @@ export class CppProperties {
...
+     * @param operation The async operation to perform.
+     * @returns The result of the operation and the time it took to complete in seconds.
+     */
+    private async timeOperation<T>(operation: () => Promise<T | undefined>): Promise<{ result: T | undefined; duration: number }> {
+        let result: T | undefined;
+        const start = process.hrtime();
+        try {
+            result = await operation();
+        } catch {
+            // Suppress errors
```
sean-mcmanus: When this is hit, it shows the timing warning instead of an error message. Can you show an error message instead?

sean-mcmanus (changes requested):

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Maybe return a result of -1 or something and have the handler check that? Otherwise, it makes it seems like it worked but was slow, when it's really failing.

bobbrow: Sure. I had thought that showing the timing warning would be a hint that the glob was too big, but I agree the error would probably be better.

sean-mcmanus: I guess it could be warning level though, something like "Path failed to be evaluated.".

sean-mcmanus: Might be useful to track failures in telemetry too, to prioritize potentially fixes.

bobbrow: The real challenge here is that `result` is of type `T` which isn't always a string. I'll probably have to catch the error at the place this is called instead.

sean-mcmanus: I was suggesting to just show a generic error message and not something specific, i.e. just to signal it was an error instead of a normal timeout.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1834,7 +1866,22 @@ export class CppProperties {
         return errorMsg;
     }
 
-    private async handleSquiggles(): Promise<void> {
+    private lastConfigurationVersion: number = 0;
+    private handleSquigglesDeferral: Deferral | undefined;
+
+    private handleSquiggles(doc?: vscode.TextDocument): void {
+        // When the active config changes, we don't pass the doc in since the version would not have changed.
+        if (doc?.version !== this.lastConfigurationVersion) {
```
sean-mcmanus, line 1873: I'm seeing this cause no error squiggles to appear after the doc is closed/reopened.

Extension/src/LanguageServer/configurations.ts
```diff
@@ -846,7 +847,7 @@ export class CppProperties {
             if (isGlobPattern) {
                 // fastGlob silently strips non-found paths. Limit that behavior to dynamic paths only.
                 const matches: string[] = fastGlob.isDynamicPattern(normalized) ?
-                    fastGlob.sync(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
+                    await fastGlob.async(normalized, { onlyDirectories: true, cwd, suppressErrors: true, deep: 15 }) : [res];
```
sean-mcmanus, line 850: I'm still seeing this called 3 times with an edit/save. I'm not sure why.

sean-mcmanus: this.handleConfigurationChange(); is being called 2 times. The other call is from handleSquigglesImpl .

Extension/src/LanguageServer/configurations.ts
bobbrow: Yeah, that can't be done here. It needs to be done at the place this is called.

Extension/src/LanguageServer/configurations.ts
bobbrow: I wasn't fixing handleConfigurationChange in this PR. There will still be some duplicate calls.

sean-mcmanus: I think it's due to auto-save, the edit triggers one handleConfigurationChange and the save triggers another.

sean-mcmanus: Ok, I was wrong, disabling auto save, I see it trigger handleConfigurationChange  0 times on edit and 2 times on save.

sean-mcmanus: I'm guessing a format on save might be happening.

Extension/src/LanguageServer/configurations.ts
bobbrow: Sometimes I can't get `onDidOpenTextDocument` to fire, but I updated it to not pass the doc in here so that squiggles will always process on document open.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Yeah, I meant to return an error result time hear and have the logging place check that.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: I still don't see the timing squiggle though on reopen.

sean-mcmanus: Ah, it's because it was < 10 sec.

Extension/src/LanguageServer/configurations.ts
sean-mcmanus: Ok, I see it show the error now.

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #13889, opened 2025-08-29
Warn users about too many files processed in the workspace

### LINKED ISSUE #10828: Show a warning when an excessive number of C/C++ files are discovered in a workspace
The C/C++ extension may take a very long time to tag parse very large projects.  Folders containing recursive symlinks can result in a huge number of (the same) files being discovered.  Include graph generation can be very slow (and block some operations), as well as other scalability issues (some that block operations).

We've assumed users are opening individual C++ project folders as workspace folders, and perhaps using multi-root workspaces to open multiple individual project folders.  Although perhaps not the majority of users, some users may be opening the roots of very large codebases (Linux kernel, large game projects, the Windows OS repo.)  Though this can be a convenient way to navigate a large codebase with a single VS Code window, it may lead to unacceptable performance from the C/C++ extension.  It may also not be apparent why IntelliSense doesn't work reliably across all C/C++ files, or why a giant tag parser database gets generated.

Perhaps a warning could be displayed when an excessive number of C/C++ files are detected in a workspace.

bobbrow:
Fixes: #10828

For consideration. Telemetry suggests that less than 3% of workspaces have more than 100,000 source files (with 250,000 total files).  That seems like a reasonable threshold, but we could raise it if we think that's too many customers to notify.

The notification can be disabled for the workspace. The "learn more" button links to a page I wrote in the Wiki about improving the configuration (which should also be reviewed).  https://github.com/microsoft/vscode-cpptools/wiki/Optimizing-your-configuration

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2772,7 +2772,7 @@ export class DefaultClient implements Client {
 
                 const message = localize(
                     "parsing.stats.large.project",
-                    'Enumerated {0} files with {1} source files detected. You may want to consider excluding some files for better performance.',
+                    'Enumerated {0} files with {1} C/C++ source files detected. You may want to consider excluding some files for better performance.',
                     filesDiscovered,
                     parsableFiles);
                 const learnMore = localize('learn.more', 'Learn More');
@@ -2780,7 +2780,7 @@ export class DefaultClient implements Client {
 
                 // We only want to show this once per session.
                 this.excessiveFilesWarningShown = true;
-                const response = await vscode.window.showWarningMessage(message, learnMore, dontShowAgain);
+                const response = await vscode.window.showInformationMessage(message, learnMore, dontShowAgain);
 
                 if (response === dontShowAgain) {
                     showExcessiveFilesWarning.Value = false;
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2764,7 +2764,7 @@ export class DefaultClient implements Client {
         }
 
         const showExcessiveFilesWarning = new PersistentWorkspaceState<boolean>('CPP.showExcessiveFilesWarning', true);
-        if (!this.excessiveFilesWarningShown && notificationBody.event === 'ParsingStats' && showExcessiveFilesWarning.Value) {
+        if (!this.excessiveFilesWarningShown && showExcessiveFilesWarning.Value && notificationBody.event === 'ParsingStats') {
             const filesDiscovered = notificationBody.metrics?.filesDiscovered ?? 0;
             const parsableFiles = notificationBody.metrics?.parsableFiles ?? 0;
             if (filesDiscovered > 250000 || parsableFiles > 100000) {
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/client.ts
```diff
@@ -2757,10 +2757,38 @@ export class DefaultClient implements Client {
...
 
-    private logTelemetry(notificationBody: TelemetryPayload): void {
+    private excessiveFilesWarningShown: boolean = false;
+    private async logTelemetry(notificationBody: TelemetryPayload): Promise<void> {
         if (notificationBody.event === "includeSquiggles" && this.configurationProvider && notificationBody.properties) {
             notificationBody.properties["providerId"] = this.configurationProvider;
         }
+
+        const showExcessiveFilesWarning = new PersistentWorkspaceState<boolean>('CPP.showExcessiveFilesWarning', true);
+        if (!this.excessiveFilesWarningShown && notificationBody.event === 'ParsingStats' && showExcessiveFilesWarning.Value) {
```
sean-mcmanus: Can you move this before the ParsingStats string string check so it can potentially skip that?

sean-mcmanus (dismissed):

Extension/src/LanguageServer/client.ts
bobbrow: You're talking about the `showExcessiveFilesWarning.Value`?

sean-mcmanus: Yes.

sean-mcmanus: Normally, boolean/fast checks should go before string checks that require more comparisons.

bobbrow: Sure. This check grew a bit organically and I didn't think about it.

sean-mcmanus (approved):

Colengms (approved):

poleguy:
Good day, kind creators of useful free tools! 

This pop-up keeps showing up and repeatedly interrupts github copilot chat. It has a button for more help, but I couldn't see the link, and couldn't click it (because I'm running in a weird x2go session where firefox click handler is broken). By looking on github it seems it links to the wiki. 

The wiki lists three knobs that I can turn: they are all exclude settings that I must manually set. 

The other obvious option is to remove this plugin entirely... although that has a downside. :-)

I don't see anything useful in debugging how to fix the issue in the wiki. How do I find the list of 250,000+ files so I can know what's going wrong? 

It would be nice if it could propose a specific solution for me. It has the list of files, and access to a powerful computer to run an algorithm to determine a good solution, and I do not!

(I'm just dreaming here, mostly, of course.)! I did want to point out that it seems this pop-up is interacting poorly with github copilot chat.... it seemed to pop up at the same time that copilot just silently failed... twice in a row now. 

The only practical suggestion would be helping me understand how to see the list of files (or possibly worst offender directories), not just the count.

I'm not deeply invested in using this plugin. It's only installed because vs code suggested it, because it saw I had a handful of c/h files in my project.

The author of this message is either human or an advanced AI bot from the distant future that is equivalent to a human. :-)

Have a nice Tuesday.

sean-mcmanus:
@poleguy If you're using a very large workspace with only a few .c/h files and don't intend do use the C/C++ extension features, then disabling the C/C++ extension for the workspace might actually be a good idea. Otherwise, you can set C_Cpp.loggingLevel to "Debug" to get more info on what files are being parsed and it shows what directories are being used in the browse.path. You can set the browse.path to `[]` or some smaller subset or set C_Cpp.files.exclude.

The wiki docs with more info are at https://github.com/microsoft/vscode-cpptools/wiki/Optimizing-your-configuration

PR #13971, opened 2025-09-30
Fix infinite ssh config file include processing.

### LINKED ISSUE #13966: Extension not activating if remote has ssh config includes
### Environment

- OS and Version: Windows 11
- VS Code Version: 1.104.1
- C/C++ Extension Version: 1.27.7
- If using SSH remote, specify OS of remote machine: Ubuntu 24.04 LTS

Bug Summary: 
This is quite odd but I'm able to reproduce it 100%.
I'm not sure why this happens or if this issue is in vscode-cpptools or vscode-remote-ssh.
But changing the Include line as written below fixes it, so maybe it is worth checking/easy to fix.

Extension doesn't load.
No actions seems to work (formatting, go to definitions, header/cpp switch, ...).
Trying to format the whole document pops up request to install a cpp formatter.

The extension show "Activating extension..." on the status but nothing else happens.
Even "C/C++: Log Diagnostics" doesn't do anything.

Steps to reproduce:
Add

Include **/config

to remote ssh config.

Removing or changing it to (for example):

Include config.d/*

fix the issue, the extension activate and work correctly.

Expected behavior:
Extension activate correctly.

### Configuration and Logs

```shell
I cannot generate any log "C/C++: Log Diagnostics" doesn't do anything.
```

### Other Extensions

The whole vscode-remote suite is installed.
cmake-tools

### Additional context

_No response_

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/13966 .

The fix was created by Copilot, but I had to fix it up by deleting unnecessary changes. But then Copilot found a bug with the fix upon review, but it was able to create a fix.

I tested this via using the repro for the bug, plus 8 .conf files that Copilot generated (not checked in).

### DIFF
--- a/Extension/src/SSH/sshHosts.ts
+++ b/Extension/src/SSH/sshHosts.ts
@@ -105,35 +105,58 @@ function getProcessedPathKey(filePath: string): string {
     return isWindows ? normalizedPath.toLowerCase() : normalizedPath;
 }
 
-async function resolveConfigIncludes(config: Configuration, configPath: string): Promise<void> {
-    const processedIncludePaths: Set<string> = new Set<string>();
-    processedIncludePaths.add(getProcessedPathKey(configPath));
-    for (const entry of config) {
-        if (isDirective(entry) && entry.param === 'Include') {
-            let includePath: string = resolveHome(entry.value);
-            if (isWindows && !!includePath.match(/^\/[a-z]:/i)) {
-                includePath = includePath.substr(1);
-            }
-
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
Copilot found a bug/problem with this fix. It appears there are valid cases of repeated includes.

Colengms (approved):

sean-mcmanus:
I'm working on a follow up to add activation progress telemetry.

PR #13989, opened 2025-10-09
Fix some invalid assumptions for tasks

bobbrow:
The `group` property for a task can be a string or an object. This change handles the case when it's a string that weren't handled previously.


Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -292,9 +292,9 @@ export class CppBuildTaskProvider implements TaskProvider {
         if (setAsDefault) {
             rawTasksJson.tasks.forEach((task: any) => {
                 if (task.label === selectedTask?.definition.label) {
-                    task.group = { kind: "build", "isDefault": true };
-                } else if (task.group.kind && task.group.kind === "build" && task.group.isDefault && task.group.isDefault === true) {
-                    task.group = "build";
+                    task.group = { kind: "build", isDefault: true };
+                } else if (!util.isString(task.group) && task.group?.kind === "build" && task.group?.isDefault) {
+                    task.group.isDefault = false;
```
sean-mcmanus, line 297: Why was this changed from ` task.group = "build";`, to `task.group.isDefault = false;`?

bobbrow: The two are equivalent, but I think this makes it easier for them to switch it back if they don't want to keep that change we made (e.g. change a bool instead of switch back from string to object)

sean-mcmanus (approved):

PR #14002, opened 2025-10-15
put standard library file associations in package.json

### LINKED ISSUE #4077: Switch to using setTextDocumentLanguage instead of adding to files.associations

bobbrow:
Fixes: #4077

This change disables `C_Cpp.autoAddFilesAssociations` by default and lists the standard headers in `package.json` to associate them with C++. Other extensionless headers will not get special treatment, but their languageId will still be fixed when a goto definition command is handled.


sean-mcmanus (changes requested):
Can you add the missing extensionless system headers?

```
ccomplex
cfenv
cinttypes
contracts
cstdalign
cstdbool
cstdint
ctgmath
cuchar
debugging
hazard_pointer
hive
inplace_vector
linalg
rcu
simd
text_encoding
```

sean-mcmanus:
Is this for 1.28.3? Or 1.29.0? Seems like 1.29.0 might be better...I was worried about something breaking regarding the language handling.

sean-mcmanus:
Does this change potentially break something contributed by VS Code itself and/or should the change be made in VS Code itself?

sean-mcmanus:
What about the other "languages" fields like "aliases" and "configuration" -- should those be set? Does VS Code set those and/or do we inherit those if we don't set the fields?

bobbrow:
> Can you add the missing extensionless system headers?
> 

Yes. It looks like I did not include headers removed since their introduction and new C++26 headers.

sean-mcmanus:
VS Code uses

```
        "aliases": [
          "C++",
          "Cpp",
          "cpp"
        ],
        "configuration": "./language-configuration.json"
```

Do you know how to check if that is still inherited?

sean-mcmanus (commented):
Should we add the GSL library extensionless files? VS has those added (maybe others?).

sean-mcmanus (commented):
Should we generate/update this list of filenames by reading a shared file with VS which also has a similar list of extension less files? We could do that similar to the import-edge-strings script.

Or otherwise, should it manually match VS?

sean-mcmanus (approved):

bobbrow:
> Should we generate/update this list of filenames by reading a shared file with VS which also has a similar list of extension less files? We could do that similar to the import-edge-strings script.
> 
> Or otherwise, should it manually match VS?

We can consider it for the future. I'll open an issue.

> Should we add the GSL library extensionless files? VS has those added (maybe others?).

It's not really a priority for me to support every library that decided to go extensionless (eigen is another one). I think it's a bad practice to go extensionless because it is harder for tooling, but I'm biased because I work on the tooling. 😄

PR #14043, opened 2025-11-12
Retry build files task in cg.yml

sean-mcmanus:
It was generated by Copilot but it took multiple tries to get correct.


bobbrow:
Isn't there a retry property we can use? I thought we used that in our other pipelines.

sean-mcmanus:
> Isn't there a retry property we can use? I thought we used that in our other pipelines.

Copilot is saying 

```
The retryCountOnTaskFailure property is available for some tasks, but not for bash or script steps.
```

sean-mcmanus:
> > Isn't there a retry property we can use? I thought we used that in our other pipelines.
> 
> Copilot is saying
> 
> ```
> The retryCountOnTaskFailure property is available for some tasks, but not for bash or script steps.
> ```

Oh, but I see it used in our other scripts? Maybe Copilot is wrong? Or could our existing scripts just be ignoring that retry?

UPDATE: I believe Copilot may have used the script because that allows a timed sleep between retries which the retryCountOnTaskFaiure doesn't do...however, I don't know if this failure case requires a sleep or not.

UPDATE: Next time when Copilot generates unexpected code, I should ask it why instead of just changing it since it might have a reason such as this case. I hit that issue before in another case too.

bobbrow (dismissed):

sean-mcmanus:
@bobbrow I switched by back to the script to allow a sleep between retries since otherwise I suspect it will just fail repeatedly like we hit in another case where we had to use a similar loop with a sleep.

UPDATE: Okay, I finally got it to work using retry on the task and a sleep in the script -- I hit a problem with the `\` not working from the variable expansion in the bash script that I had to add a couple lines to fix.

Colengms (approved):

bobbrow (approved):

Issue #8998, opened 2022-03-09
C_Cpp.commentContinuationPatterns only recognized for follow-up line


sreischl:
Bug type: C_Cpp.commentContinuationPatterns 

**Describe the bug**
- OS and Version: Windows 10 (20H2, Build 19042.1526)
- VS Code Version: 1.65.0
- C/C++ Extension Version: v1.8.4

Self introduced comment continuation pattern is recognized. But only for one follow-up line. 

E.g.: 
```
    "C_Cpp.commentContinuationPatterns": [
      "/**",
      "/*!",
      "//'"
    ]
```

Starting with this sequence, it works as expected...: 
```
/*!<enter>
 * 
```

But continuing with: 
```
/*!<enter>
 * asdf<enter>
 
```

The asterisk is getting lost. 

This behavior does not occur with the standard pattern of '/**'. 

**Steps to reproduce**

1. introduce a custom pattern in settings.json
2. use the multi-line comment as described above

**Expected behavior**
The asterisk should arise in the second follow-up line, too. Just as the default pattern does. 

**Code sample and logs**
```
/**
 * \internal
 * <- this example works perfectly
 * 
 */
void bar(void)
{

}

/*!
 * asdf
 <- typing <enter> after the 'asdf' above leaves us out missing the asterisk

void foo(void)
{

}
```

**Screenshots**
-

**Additional context**
-

sean-mcmanus:
The implementation is at https://github.com/microsoft/vscode-cpptools/blob/main/Extension/src/LanguageServer/languageConfig.ts#L240 .

sreischl:
What does this information help me?

sean-mcmanus:
It would help anyone who wanted to fix it.

kosude:
Any updates on this? I'm getting a similar issue (difference is it doesn't work for the follow-up line either!)

sean-mcmanus:
Fixed with https://github.com/microsoft/vscode-cpptools/releases/tag/v1.30.0 (pre-release).

PR #14074, opened 2025-12-02
Fix custom comment continuation for multiple follow-up lines

dinhtam2c:
Fix microsoft/vscode-cpptools#8998

### DIFF
--- a/Extension/src/LanguageServer/languageConfig.ts
+++ b/Extension/src/LanguageServer/languageConfig.ts
@@ -239,8 +239,9 @@ export function getLanguageConfig(languageId: string): vscode.LanguageConfigurat
 }
 
 export function getLanguageConfigFromPatterns(languageId: string, patterns?: (string | CommentPattern)[]): vscode.LanguageConfiguration {
-    const beginPatterns: string[] = []; // avoid duplicate rules
-    const continuePatterns: string[] = []; // avoid duplicate rules
+    const beginPatterns: string[] = []; // avoid duplicate begin rules
+    const continuePatterns: string[] = []; // avoid duplicate continue rules
+    const endPatterns: string[] = []; // avoid duplicate end rules
     let duplicates: boolean = false;
     let beginRules: vscode.OnEnterRule[] = [];
     let continueRules: vscode.OnEnterRule[] = [];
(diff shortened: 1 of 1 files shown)


dinhtam2c:

Extension/src/LanguageServer/languageConfig.ts
```diff
@@ -258,15 +259,13 @@ export function getLanguageConfigFromPatterns(languageId: string, patterns?: (st
         } else {
             duplicates = true;
         }
-        if (continuePatterns.indexOf(c.continue) < 0) {
```
bobbrow, line 261: The check here exists to avoid registering duplicate rules and with this removed, there is a unit test that fails when the duplicate rules are registered. It's possible the test has an incorrect assumption and needs to be fixed, but I didn't see an update there.

bobbrow: The test is in `test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts`, BTW

dinhtam2c: Pushed a fix for that

Extension/src/LanguageServer/languageConfig.ts
```diff
@@ -258,14 +260,17 @@ export function getLanguageConfigFromPatterns(languageId: string, patterns?: (st
...
+        if (continuePatterns.indexOf(`${c.begin}\0${c.continue}`)) {
```
bobbrow: should it be comparing with `< 0`?

Extension/src/LanguageServer/languageConfig.ts
bobbrow: thanks

Extension/src/LanguageServer/languageConfig.ts
dinhtam2c: yes. my bad

bobbrow (approved):
Thank you for this fix!

bobbrow:
I will follow up with an additional unit test to cover the scenario we missed.

bobbrow:
I didn't notice this was targeting the wrong branch. It needs to go to main first. 1.29 is currently prepping for release in that branch and this will go into 1.30.

dinhtam2c:
Do I need to create a new PR?

bobbrow:
> Do I need to create a new PR?

No, I did it already. #14076

PR #14076, opened 2025-12-02
Fix custom comment continuation for multiple follow-up lines

### LINKED ISSUE #8998: C_Cpp.commentContinuationPatterns only recognized for follow-up line

bobbrow:
Fixes: #8998

Cherry-picks the original change from the release branch. It should have gone to main first.
Also adds a unit test.

### DIFF
--- a/Extension/src/LanguageServer/languageConfig.ts
+++ b/Extension/src/LanguageServer/languageConfig.ts
@@ -55,7 +55,8 @@ function getMLSplitAfterPattern(): string {
 
 function getMLPreviousLinePattern(insert: string): string | undefined {
     if (insert.startsWith("/*")) {
-        return `(?=^(\\s*(\\/\\*\\*|\\*)).*)(?=(?!(\\s*\\*\\/)))`;
+        const match: string = escape(insert);
+        return `(?=^(\\s*(${match}|\\*)).*)(?=(?!(\\s*\\*\\/)))`;
     }
     return undefined;
 }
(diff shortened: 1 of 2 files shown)


sean-mcmanus:
@bobbrow There's a bug whenever one begin pattern is a prefix of another, e.g. if `//` and `///` or `/*` and `/*!" are used, it causes the "continue" of the non-matching, shorter pattern to be used.

This works correctly,
```json
    "C_Cpp.commentContinuationPatterns": [
        {
            "begin": "/*!",
            "continue": " ! "
        }
    ],
```

but adding a "/*" breaks it (uses `*` as the continuation instead):

```json
    "C_Cpp.commentContinuationPatterns": [
        "/*",
        {
            "begin": "/*!",
            "continue": " ! "
        }
    ],
```

similarly, "//" is used as the continue for "///".

I could file a new bug for that (it's a pre-existing bug and not a regression from this PR).

UPDATE: I moved it to https://github.com/microsoft/vscode-cpptools/issues/14081 (with a fix at https://github.com/microsoft/vscode-cpptools/pull/14082)

sean-mcmanus (approved):

PR #14108, opened 2025-12-14
feat: Support program parameter in attach configurations (#14046)

### LINKED ISSUE #14046: Support "program" for {"request": "attach"} instead of only "processId"
### Feature Request

Currently 
```json
{
    "name": "(Windows) Launch",
    "type": "cppvsdbg",
    "request": "launch",
    "program": "myprogram.exe",
},
```
is used to launch (and attach to) a new process for debugging, while
```json
{
    "name": "(Windows) Attach",
    "type": "cppvsdbg",
    "request": "attach",
    "processId": "${command:pickProcess}",
},
```
is used to attach to a process that is already running.

The former is great, the latter is incredibly inconvenient, since you often know exactly which process to attach to, but have to type it every single time. LLDB ([`vadimcn.vscode-lldb`](https://github.com/vadimcn/codelldb)) supports
```json
{
    "name": "(LLDB) Attach",
    "type": "lldb",
    "request": "attach",
    "program": "myprogram.exe",
},
```
which will simply look for a running program called `myprogram.exe`, and attach to it.

---

A related issue is #1904, where they asked to add `"processName"` to the schema. I think they were under the impression that cpptools supported (they probably found the name in something like [this Stack Overflow question](https://stackoverflow.com/questions/60440765/how-can-i-attach-to-a-specific-process-in-visual-studio-code)).

Also perhaps relevant is #1272, though that applies to GDB and specifically asks to make `"program"` optional.

Subham-KRLX:
Fixes #14046

Add support for program parameter in attach configurations to attach to processes by executable name, similar to LLDB extension.

Example:

json
{
    "type": "cppvsdbg",
    "request": "attach",
    "program": "myprogram.exe"
}
Automatically finds matching process, prompts if multiple matches. Backward compatible with existing processId configurations.


sean-mcmanus:
@Subham-KRLX It looks like there are compile errors: src/Debugger/configurationProvider.ts(361,40): error TS2339: Property 'findProcessByProgramName' does not exist on type 'DebugConfigurationProvider'.
src/Debugger/configurationProvider.ts(1340,19): error TS6133: 'findProcessByProgramName' is declared but its value is never read.

Subham-KRLX:
@sean-mcmanus We've resolved both compilation errors by moving the [findProcessByProgramName](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) method to the correct location inside the [DebugConfigurationProvider](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) class. The fix has been committed and pushed to the [feature/attach-by-program-name](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) branch. The TypeScript compilation errors are now resolved.

Som1Lse:
@sean-mcmanus I just want to add that I am also really interested in this feature.

It seems to have been sitting for over a month waiting for a review, so I'm bumping this in case people forgot.

sean-mcmanus (changes requested):
Why are there Code Sample changes? Can you remove those? Aren't those changes in a separate PR already?

Why is there an MIEngine_Debug subproject added? Can that be removed?

Subham-KRLX:
@sean-mcmanus I have cleaned up the PR by reverting the unrelated Code Samples changes and removing the MIEngine_Debug subproject. The PR now only contains the relevant changes for supporting the program parameter in attach configurations.

sean-mcmanus:
> @sean-mcmanus I have cleaned up the PR by reverting the unrelated Code Samples changes and removing the MIEngine_Debug subproject. The PR now only contains the relevant changes for supporting the program parameter in attach configurations.

@Subham-KRLX I still see the Code Sample changes, not reverted.

Subham-KRLX:
@sean-mcmanus I have thoroughly reverted all unrelated Code Samples changes by syncing with upstream/main and removed the MIEngine_Debug subproject.

Extension/package.nls.json
```diff
@@ -931,7 +931,13 @@
...
         ]
     },
-    "c_cpp.debuggers.symbolSearchPath.description": "Semicolon separated list of directories to use to search for symbol (that is, pdb) files. Example: \"c:\\dir1;c:\\dir2\".",
+    "c_cpp.debuggers.program.attach.markdownDescription": {
+        "message": "Optional full path to program executable. When specified, the debugger will search for a running process matching this executable name and attach to it. If multiple processes match, a selection prompt will be shown. Use either `program` or `processId`, not both.",
+        "comment": [
+            "{Locked=\"`program`\"} {Locked=\"`processId`\"}"
+        ]
+    },
+    "c_cpp.debuggers.symbolSearchPath.description": "Semicolon separated list of directories to use to search for .so files. Example: \"c:\\dir1;c:\\dir2\".",
```
sean-mcmanus: On Windows with MSVC it uses .pdb files and not so files, so the description could be something like `search for symbols (that is, pdb or .so files).`

Extension/package.nls.json
```diff
@@ -931,7 +931,13 @@
             "{Locked=\"`${command:pickProcess}`\"}"
         ]
     },
-    "c_cpp.debuggers.symbolSearchPath.description": "Semicolon separated list of directories to use to search for symbol (that is, pdb) files. Example: \"c:\\dir1;c:\\dir2\".",
+    "c_cpp.debuggers.program.attach.markdownDescription": {
+        "message": "Optional full path to program executable. When specified, the debugger will search for a running process matching this executable name and attach to it. If multiple processes match, a selection prompt will be shown. Use either `program` or `processId`, not both.",
```
sean-mcmanus: Should `executable name` be `executable path`?

sean-mcmanus: `path to program executable` should be `path to the program executable`.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -354,13 +355,24 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         // Pick process if process id is empty
         if (config.request === "attach" && !config.processId) {
             let processId: string | undefined;
-            if (config.pipeTransport || config.useExtendedRemote) {
-                const remoteAttachPicker: RemoteAttachPicker = new RemoteAttachPicker();
-                processId = await remoteAttachPicker.ShowAttachEntries(config);
+
+            // If program is specified, try to find matching process by name
```
sean-mcmanus: `matching process by name`->`the matching process by name.`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -354,13 +355,24 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+                processId = await this.findProcessByProgramName(config.program, config, token);
+                if (!processId) {
+                    void logger.getOutputChannelLogger().showErrorMessage(`No running process found matching "${config.program}".`);
+                    return undefined;
+                }
             } else {
-                const attachItemsProvider: AttachItemsProvider = NativeAttachItemsProviderFactory.Get();
-                const attacher: AttachPicker = new AttachPicker(attachItemsProvider);
-                processId = await attacher.ShowAttachEntries(token);
+                // Show process picker if no program specified
```
sean-mcmanus: `Show process picker if no program specified`->`Show the process picker if no program is specified.`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+
+    private async findProcessByProgramName(
+        programPath: string,
+        config: CppDebugConfiguration,
+        token?: vscode.CancellationToken
+    ): Promise<string | undefined> {
+        const programBaseName: string = path.basename(programPath);
+        let processes: AttachItem[];
+
+        // Get process list using the same logic as interactive attach
```
sean-mcmanus: `Get process list using the same logic as interactive attach`->`Get the process list using the same logic as interactive attach.`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+
+        if (matchingProcesses.length === 0) {
+            return undefined;
+        } else if (matchingProcesses.length === 1) {
+            void logger.getOutputChannelLogger().appendLine(
+                `Found process "${matchingProcesses[0].label}" with PID ${matchingProcesses[0].id}`
+            );
+            return matchingProcesses[0].id;
+        } else {
+            // Multiple matches - let user choose
```
sean-mcmanus: `let user choose`->`let the user choose.`

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -354,13 +355,24 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
             let processId: string | undefined;
-            if (config.pipeTransport || config.useExtendedRemote) {
-                const remoteAttachPicker: RemoteAttachPicker = new RemoteAttachPicker();
-                processId = await remoteAttachPicker.ShowAttachEntries(config);
+
+            // If program is specified, try to find matching process by name
+            if (config.program) {
+                processId = await this.findProcessByProgramName(config.program, config, token);
+                if (!processId) {
+                    void logger.getOutputChannelLogger().showErrorMessage(`No running process found matching "${config.program}".`);
```
sean-mcmanus: These user facing strings should be passed to the localize function with some unique key, see the example such as `                    void logger.getOutputChannelLogger().showErrorMessage(localize('missing.properties.copyFile', '"host", "files", and "targetDir" are required in {0} steps.', isScp ? 'SCP' : 'rsync'));`

I saw a spot in this file where pre-existing code appears to be missing a localize call (we can fix that ourselves later).

sean-mcmanus:
@Subham-KRLX I've started reviewing this. I think I can finish tomorrow.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -354,13 +355,24 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
         if (config.request === "attach" && !config.processId) {
             let processId: string | undefined;
-            if (config.pipeTransport || config.useExtendedRemote) {
-                const remoteAttachPicker: RemoteAttachPicker = new RemoteAttachPicker();
-                processId = await remoteAttachPicker.ShowAttachEntries(config);
+
+            // If program is specified, try to find matching process by name
+            if (config.program) {
+                processId = await this.findProcessByProgramName(config.program, config, token);
+                if (!processId) {
```
sean-mcmanus: It seems like this would cause a failure for users who currently have processName set to an invalid value, which is the default.

Can you instead fall back to the process picker if `findProcessByProgramName` doesn't return a valid processId and don't print any error message?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         }
         return true;
     }
+
+    private async findProcessByProgramName(
+        programPath: string,
+        config: CppDebugConfiguration,
+        token?: vscode.CancellationToken
+    ): Promise<string | undefined> {
+        const programBaseName: string = path.basename(programPath);
```
sean-mcmanus: Can you first check if the programPath is a valid path? Otherwise, invalid paths like "this is an invalid path.exe" will use "path.exe" as the programBaseName.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        let processes: AttachItem[];
+
+        // Get process list using the same logic as interactive attach
+        if (config.pipeTransport || config.useExtendedRemote) {
```
sean-mcmanus: Can you move this check to the very beginning before "program" is checked and don't show any error? Otherwise, those users would start to get an error.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            );
+            return undefined;
+        } else {
+            const attachItemsProvider: AttachItemsProvider = NativeAttachItemsProviderFactory.Get();
+            processes = await attachItemsProvider.getAttachItems(token);
+        }
+
+        // Find processes matching the program name
+        const matchingProcesses: AttachItem[] = processes.filter(p => {
+            const processName: string = p.label.toLowerCase();
```
sean-mcmanus: The toLowerCase should only be used on Windows where the paths are case insensitive.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            const targetName: string = programBaseName.toLowerCase();
```
sean-mcmanus: This should be moved out of the lambda and into a local variable at the out scope to avoid unnecessarily calling it more than once.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            // Match if the process name exactly matches or starts with the target name
+            return processName === targetName || processName.startsWith(targetName);
```
sean-mcmanus: Can you remove this startsWith to prevent the an unintended process from being accidentally attached to?

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: And you could use something like

```
                    if (isWindows) {
                        targetName = targetName.endsWith(".exe") ? targetName : (targetName + ".exe");
                    }
```

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: Maybe call `checkExecutableWithoutExtensionExists`.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        if (matchingProcesses.length === 0) {
+            return undefined;
+        } else if (matchingProcesses.length === 1) {
+            void logger.getOutputChannelLogger().appendLine(
+                `Found process "${matchingProcesses[0].label}" with PID ${matchingProcesses[0].id}`
+            );
+            return matchingProcesses[0].id;
+        } else {
+            // Multiple matches - let user choose
+            void logger.getOutputChannelLogger().appendLine(
```
sean-mcmanus: This logging is unnecessary -- users will see the quick pick UI and not see the logging.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -1157,6 +1169,50 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+            const targetName: string = programBaseName.toLowerCase();
+            // Match if the process name exactly matches or starts with the target name
+            return processName === targetName || processName.startsWith(targetName);
+        });
+
+        if (matchingProcesses.length === 0) {
+            return undefined;
+        } else if (matchingProcesses.length === 1) {
+            void logger.getOutputChannelLogger().appendLine(
+                `Found process "${matchingProcesses[0].label}" with PID ${matchingProcesses[0].id}`
```
sean-mcmanus: There's no need to log here.

sean-mcmanus (changes requested):

Extension/package.nls.json
Subham-KRLX: Fixed in [4046187](https://github.com/Subham-KRLX/vscode-cpptools/commit/4046187c) - Updated to "search for symbol (that is, pdb or .so) files"

Extension/package.nls.json
Subham-KRLX: Changed "executable name" to "executable path"

Subham-KRLX: Changed to "path to the program executable"

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: Updated comment to "try to find the matching process by name"

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: Error message removed entirely; now falls back to process picker silently instead of showing an error

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: Now falls back to interactive process picker when program match fails, no error shown

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: Added checkExecutableWithoutExtensionExists validation before processing path

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: Case-insensitive matching is now restricted to Windows only using the isWindows check.

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: targetName computation moved outside the filter lambda with .exe suffix logic for Windows included.

Extension/src/Debugger/configurationProvider.ts
Subham-KRLX: @sean-mcmanus I have addressed all review items in commit [4046187](https://github.com/Subham-KRLX/vscode-cpptools/commit/4046187c). This update restricts case-insensitivity to Windows, adds path validation, ensures exact process matching, and implements a silent fallback to the process picker. I have also removed all internal logging and corrected the documentation strings in 
package.nls.json.

sean-mcmanus:
Sorry, I've been busy with other stuff. I'll try to get to this tomorrow.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -354,13 +355,23 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
-                processId = await attacher.ShowAttachEntries(token);
+
+            // If program is specified, try to find the matching process by name
+            if (config.program) {
+                processId = await this.findProcessByProgramName(config.program, config, token);
+            }
+            
+            // Fall back to process picker if program wasn't specified or didn't match
+            if (!processId) {
+                // Show the process picker if no program is specified
```
sean-mcmanus: This comment should be removed. The comment on line 364 appears to replace it.

Extension/package.json
```diff
@@ -4770,15 +4770,8 @@
                     "attach": {
                         "type": "object",
                         "default": {},
-                        "required": [
-                            "program"
-                        ],
+                        "required": [],
```
sean-mcmanus: I am seeing "program" being required for GDB debugging.

Image

sean-mcmanus: Can you put it back as a requirement or why should it be removed as a requirement? I think it can stay as a required field.

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
Also, there are some linter errors: Trailing spaces not allowed

Extension/package.json
Subham-KRLX: The "required": [] is correct  both program and processId are optional. Users can specify either or neither (falls back to process picker). The implementation in configurationProvider.ts(lines 356-382) handles all cases.

sean-mcmanus: @Subham-KRLX  Yeah, but gdb debugging fails if there's no program in the launch.json...I think MIEngine might read that. The gdb error happens after the debugger starts.

sean-mcmanus (changes requested):
I think the "program" field is required for gdb/lldb attach debugging. Are you able to debug without the "program" field?

Subham-KRLX:
@sean-mcmanus You are correct - restored `program` as required. MIEngine needs it for debug symbols even in attach mode. Fixed in 709194d6.

sean-mcmanus (approved):

sean-mcmanus:
@Subham-KRLX It seems good to me, but I'm not sure if anyone else on our team wanted to review it. I might be able to get into 1.31.0 for next Thursday.

PR #14157, opened 2026-02-02
Update eslint from 8 to 9.

sean-mcmanus:
This was mostly done by Copilot Agent mode...well, the initial edit...I had to fix a bunch of stuff it did.


sean-mcmanus:
There are issues, as metioned by the Copilot reviewer.

Extension/eslint.config.js
```diff
@@ -0,0 +1,144 @@
...
+        __dirname: "readonly",
+        exports: "readonly",
+      },
+    },
+    plugins: {
+      "@typescript-eslint": tseslint,
+      jsdoc,
+      import: eslintImport,
+    },
+    rules: {
```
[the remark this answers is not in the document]
sean-mcmanus: formatting linter rules were removed with eslint 9.

Extension/src/LanguageServer/client.ts
```diff
@@ -2074,7 +2076,7 @@ export class DefaultClient implements Client {
             if (!hasCompleted) {
                 hasCompleted = true;
                 if (currentProvider.version >= Version.v2) {
-                    await this.resumeParsing();
+                    this.resumeParsing();
```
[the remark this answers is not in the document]
sean-mcmanus: no

Extension/src/LanguageServer/client.ts
```diff
@@ -3112,16 +3114,18 @@ export class DefaultClient implements Client {
      */
     public activate(): void {
         this.model.activate();
-        void this.resumeParsing().catch(logAndReturn.undefined);
+        this.resumeParsing();
```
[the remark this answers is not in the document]
sean-mcmanus: no

Extension/.scripts/clean.ts
```diff
@@ -18,25 +18,28 @@ export async function main() {
 }
 
 export async function all() {
-    await rimraf(...(await getModifiedIgnoredFiles()).filter(each => !each.includes('node_modules')));
+    await rimraf(...(await getModifiedIgnoredFiles()).filter((each): each is string => each !== undefined && !each.includes('node_modules')));
```
bobbrow, line 21: Why does it want `each is string` annotations on the functions? `getModifiedIgnoredFiles` returns `string[]` so there's no doubt concerning its type.

Extension/.scripts/clean.ts
```diff
@@ -18,25 +18,28 @@ export async function main() {
...
+            return null;
+        }
         return {
             filename: stats.isDirectory() ? cyan(`${each}${sep}**`) : brightGreen(`${each}`),
             date: stats.mtime.toLocaleDateString().replace(/\b(\d)\//g, '0$1\/'),
             time: stats.mtime.toLocaleTimeString().replace(/^(\d)\:/g, '0$1:'),
             modified: stats.mtime
         };
-    }));
+    })).then(results => results.filter((each): each is NonNullable<typeof each> => each !== null));
```
bobbrow: Please don't combine `await` and `.then`.  Use two `awaits`.

Extension/.scripts/common.ts
```diff
@@ -258,7 +262,7 @@ export function position(text: string) {
     return gray(`${text}`);
 }
 
-export async function assertAnyFolder(oneOrMoreFolders: string | string[], errorMessage?: string): Promise<string> {
+export async function assertAnyFolder(oneOrMoreFolders: string | string[], errorMessage?: string): Promise<string | undefined> {
```
bobbrow, line 265: I think the problem here is that the `process.exit(1)` should not be inside the `if (errorMessage)` block. It should be the last line of the function. Then you shouldn't need a `| defined` here.

Extension/.scripts/common.ts
```diff
@@ -275,7 +279,7 @@ export async function assertAnyFolder(oneOrMoreFolders: string | string[], error
     }
 }
 
-export async function assertAnyFile(oneOrMoreFiles: string | string[], errorMessage?: string): Promise<string> {
+export async function assertAnyFile(oneOrMoreFiles: string | string[], errorMessage?: string): Promise<string | undefined> {
```
bobbrow, line 282: Same comment as above.

Extension/.scripts/clean.ts
sean-mcmanus: It doesn't work with the linter. It still thinks `all` can have null types unless it's in the .then.

sean-mcmanus: Oh, but it works if it create a new variable instead of re-assigning the same variable.

Extension/.scripts/clean.ts
sean-mcmanus: Hmm...but it gives a linter error if I remove it.

sean-mcmanus: Copilot thinks the code is good. And other fixes it says are not as good.

`This will create a TypeScript error because the filter won't narrow the type from [(string | undefined)[]`

sean-mcmanus: Also, it's `getModifiedIgnoredFiles(): Promise<(string | undefined)[]>`

Extension/.scripts/common.ts
sean-mcmanus: Ah, okay.

Extension/.scripts/clean.ts
sean-mcmanus: Oh, but the code is "better" with the .then.

sean-mcmanus: Well, I guess I can use 2 vars if you want.

sean-mcmanus:
Does anyone understand why the build is failing?

UPDATE: It's from the process.exit in the assert.

Hmm..it builds locally. I'm not sure what assert is failing in the lab...there's no output shown?

Extension/.scripts/common.ts
sean-mcmanus: @bobbrow It turns out, the original code is written intentionally to not actually exit if there is no error message, with the intent that the caller will handle the assert in their own way based on checking the return value (otherwise, our existing code fails to build due to the exit).

Colengms (dismissed):

sean-mcmanus:
@bobbrow Were all your concerns addressed?

bobbrow:
> @bobbrow Were all your concerns addressed?

I haven't had the time to go through very many files yet.

Extension/src/LanguageServer/client.ts
```diff
@@ -1779,10 +1779,12 @@ export class DefaultClient implements Client {
         return { wasShutdown: initializeResult.shouldShutdown };
     }
 
-    public async sendDidChangeSettings(): Promise<void> {
+    public sendDidChangeSettings(): void {
         // Send settings json to native side
-        await this.ready;
-        await this.languageClient.sendNotification(DidChangeSettingsNotification, this.getAllSettings());
+        void (async () => {
```
bobbrow: ew.  Why is this better?  Can we disable whatever this rule is?

sean-mcmanus: I'll check.

sean-mcmanus: It's this one https://typescript-eslint.io/rules/no-misused-promises/ .

Extension/eslint.config.js
```diff
@@ -0,0 +1,136 @@
+const js = require("@eslint/js");
```
bobbrow, line 1: If you're changing file names, please make sure .vscodeignore is up to date so we don't include this in the shipping package.

Extension/src/LanguageServer/client.ts
bobbrow: We were already enabling no-misused-promises. The docs don't seem to match the changes I'm seeing in several of these functions. Where does the actual problem surface?  Where this function is called?

sean-mcmanus: I can change the function to return a Promise<void> and have the callers call await or void.

sean-mcmanus: It looks like the issue is that the Client functions return void instead of `Promise<void>`, so it looks like I can just fix that instead, same with the other cases like this.

sean-mcmanus: I fixed the return types of the Client functions. We want to keep no-misused-promises since it helps catch bugs.

Extension/eslint.config.js
sean-mcmanus: done

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -370,21 +371,23 @@ class CustomBuildTaskTerminal implements Pseudoterminal {
     constructor(private command: string | util.IQuotedString, private args: (string | util.IQuotedString)[], private options: cp.ExecOptions | undefined, private buildOptions: BuildOptions) {
     }
 
-    async open(_initialDimensions: TerminalDimensions | undefined): Promise<void> {
-        if (this.buildOptions.taskUsesActiveFile && !util.isCppOrCFile(window.activeTextEditor?.document.uri)) {
-            this.writeEmitter.fire(localize("cannot.build.non.cpp", 'Cannot build and debug because the active file is not a C or C++ source file.') + this.endOfLine);
-            this.closeEmitter.fire(-1);
-            return;
-        }
+    open(_initialDimensions: TerminalDimensions | undefined): void {
```
bobbrow: Since the interface requires a void return, maybe we can create an async helper method that uses the nicer syntax instead of this IIFE pattern. I might feel more strongly about this than you, so if you don't want to do it in this PR, I can follow up with a change later.

bobbrow (dismissed):
I left one optional comment

Extension/src/LanguageServer/clientCollection.ts
```diff
@@ -177,7 +177,7 @@ export class ClientCollection {
                 // read for new workspace folders. Send them up before creating new
                 // clients, so the native process is already aware of all workspace
                 // folders before receiving new CppProperties for new folders.
-                this.defaultClient.sendDidChangeSettings();
+                await this.defaultClient.sendDidChangeSettings();
```
Colengms: It might have been intentional that there was no await here.  An await would allow other things to happen in parallel, potentially allowing other LSP messages through before the call to `createClient` a few lines down.  I don't think awaiting `sendDidChangeSettings` is required.  It's sending a notification as soon as `ready`.  It might be safest to preserve the original behavior and use `void` instead of `await`.

Extension/src/LanguageServer/protocolFilter.ts
```diff
@@ -36,7 +36,7 @@ export function createProtocolFilter(): Middleware {
                             const baseFileName: string = path.basename(document.fileName);
                             const mappingString: string = baseFileName + "@" + document.fileName;
                             client.addFileAssociations(mappingString, "cpp");
-                            client.sendDidChangeSettings();
+                            await client.sendDidChangeSettings();
```
Colengms: This should not be awaited here.  In the protocol handler, a call to await can results in other messages being processed, resulting in messages out of order (i.e. file-related requests before the `didOpen` is processed).

Colengms (changes requested):

Extension/src/LanguageServer/clientCollection.ts
sean-mcmanus: Sure, I can change it back to void, but comment makes it sound like it expected the settings to be send before hand though: `send them up before creating new clients`.

Extension/src/LanguageServer/protocolFilter.ts
sean-mcmanus: I did that because otherwise it seemed like the `setTextDocumentLanguage` could happen after the sendDidChangeSettings, but I can change it back to void.

bobbrow (approved):

Colengms (approved):

PR #14204, opened 2026-02-20
Fix/symbol search scope

### LINKED ISSUE #14200: Workspace symbol search with scope::variable may not work after symbols are deleted and then added back
I think it used to work. `scope::` works, but then it fails after any text is typed after that. This might be a regression from the "fast symbol search" work, since I see another bug report on this at https://github.com/microsoft/vscode-cpptools/issues/2484 saying it was fixed.

Image

Image

This also seems like it might be causing Copilot to have trouble finding the definition for scope::var symbols.

Subham-KRLX:
Workspace symbol search with scope::variable (e.g., std::vector) fails after Typing . This regression occurred when the experimentEnabled flag was removed, causing the language server to revert to a legacy path that handles scoped queries incorrectly.
Solution Restores the modern "fast symbol search" path by unconditionally passing experimentEnabled: true. This path correctly parses and filters scoped symbol queries.

Impact
Fixes #14200
Restores proper C++ symbol filtering for users and AI tools like Copilot.

### DIFF
--- a/Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
+++ b/Extension/src/LanguageServer/Providers/workspaceSymbolProvider.ts
@@ -22,7 +22,8 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
         }
 
         const params: WorkspaceSymbolParams = {
-            query: query
+            query: query,
+            experimentEnabled: true
         };
 
         let symbols: LocalizeSymbolInformation[];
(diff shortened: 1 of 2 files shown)


sean-mcmanus (changes requested):
I see it not working still. Also, the cpptools side has the input as a string and doesn't actually check any experimentEnabled bool. I'll debug the cpptools process to see what's going on. I don't think this can be fixed via the TypeScript.

sean-mcmanus:
Closing this. The linked bug was already fixed.

PR #14243, opened 2026-03-05
Add @vscode/vsce.

sean-mcmanus:
This is to avoid downloading @vsce/vscode during the vsix build.


bobbrow:
Is this needed for compliance or something? It adds tons of new dependencies to the lock file.

bobbrow (approved):

PR #14278, opened 2026-03-16
Stop using yarn with extension pack

sean-mcmanus:


Colengms:
Why not use yarn?  It's generally superior to npm - much faster.  (There is no description explaining why.)

sean-mcmanus:
> Why not use yarn? It's generally superior to npm - much faster. (There is no description explaining why.)

There's no yarn.lock checked in. The package-lock.json is checked in. The ExtensionPack never used yarn. It was incorrectly using yarn just for the @vscode/vsce. The speed doesn't matter in this case.

Colengms (approved):

PR #14281, opened 2026-03-16
Add internal silent find-all-references command

carsonRadtke:
Introduce an internal C_Cpp.FindAllReferences command that issues the existing cpptools/findAllReferences request without joining the workspaceReferences single-flight cancellation path.

Extract the shared request and confirmed-location mapping logic from FindAllReferencesProvider so the existing vscode.executeReferenceProvider flow and the new silent command use the same request translation and cancellation handling for server-side cancel responses.

Keep the interactive provider behavior unchanged: user-invoked references still cancel prior work, reset reference progress state, and update the ReferencesManager UI. The new command resolves the owning client from the target URI and returns locations without progress UI, preview notifications, or references panel updates, enabling concurrent silent callers such as Copilot.

Is part of a fix to ADO:2826103 and ADO:2741397 along with internal changes to the cpptools server and the devtools extension.

### DIFF
--- a/Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
+++ b/Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
@@ -11,6 +11,50 @@ import { CancellationSender, ReferenceInfo, ReferenceType, ReferencesParams, Ref
 const FindAllReferencesRequest: RequestType<ReferencesParams, ReferencesResult, void> =
     new RequestType<ReferencesParams, ReferencesResult, void>('cpptools/findAllReferences');
 
+export interface FindAllReferencesResult {
+    referencesResult: ReferencesResult;
+    locations: vscode.Location[];
+}
+
+function convertConfirmedReferencesToLocations(referencesResult: ReferencesResult): vscode.Location[] {
+    const locationsResult: vscode.Location[] = [];
+    referencesResult.referenceInfos.forEach((referenceInfo: ReferenceInfo) => {
+        if (referenceInfo.type === ReferenceType.Confirmed) {
+            const uri: vscode.Uri = vscode.Uri.file(referenceInfo.file);
+            const range: vscode.Range = new vscode.Range(referenceInfo.position.line, referenceInfo.position.character,
+                referenceInfo.position.line, referenceInfo.position.character + referencesResult.text.length);
+            locationsResult.push(new vscode.Location(uri, range));
+        }
+    });
+    return locationsResult;
+}
+
+export async function sendFindAllReferencesRequest(client: DefaultClient, uri: vscode.Uri, position: vscode.Position, token: vscode.CancellationToken): Promise<FindAllReferencesResult | undefined> {
+    const params: ReferencesParams = {
+        newName: "",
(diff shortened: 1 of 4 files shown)


sean-mcmanus:
Should call hierarchy be handled too? I haven't checked yet if the issue repros with that or not, but it might.

UPDATE: Yeah, I tried it out (without this change) and I see Call Hierarchy Callers of cancel Copilot's Find All References. I see the same thing with Rename. Both of those share the same underlying implementation.

Does this change fix those cases too?

I also see #GetSymbolCallHierarchy_CppTools get canceled by a user invoked Find All References/etc..

Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
```diff
@@ -10,72 +10,83 @@ import { getLocalizedString, getLocalizedSymbolScope } from '../localization';
...
-                    } else {
-                        detail = getLocalizedSymbolScope("private", detail);
-                    }
-                } else if (symbol.scope === SymbolScope.Protected) {
-                    if (detail.length === 0) {
-                        detail = "protected";
-                    } else {
-                        detail = getLocalizedSymbolScope("protected", detail);
-                    }
+function getChildrenSymbols(symbols: LocalizeDocumentSymbol[]): vscode.DocumentSymbol[] {
```
sean-mcmanus: @carsonRadtke I thought C/C++ DevTools only used workspace symbols and not document symbols? Why is document symbols changing instead of workspace symbols?

Extension/src/LanguageServer/Providers/findAllReferencesProvider.ts
sean-mcmanus, line 67: I'm seeing DevTools hit this cancel. Isn't it supposed to not issue a cancel?

sean-mcmanus: Oh...I'm guessing I need an updated DevTools that calls a different command.

Extension/src/LanguageServer/Providers/documentSymbolProvider.ts
sean-mcmanus: I see DevTools issue workspace symbol requests and not document symbol requests.

sean-mcmanus: I have found potential document symbol requests, but it doesn't seem like they can be cancelled by an additional user/DevTools invoked request?

sean-mcmanus: i.e. I think the token cancellation for a request only applies to that particular request and it doesn't affect other requests (only the FAR-related requests cancel previous requests when new ones are requested)...unless I'm misunderstanding something.

sean-mcmanus: It appears you fixed that.

sean-mcmanus (approved):

sean-mcmanus (approved):

sean-mcmanus (approved):

Issue #14211, opened 2026-02-24
Under Remote-SSH, cpptools is causing the whole vscode interface to be laggy.


kuroa-me:
### Environment

- OS and Version: MacOS 26.3 (25D125)
- VS Code Version: 1.109.5 (Universal)
- C/C++ Extension Version: 1.30.5
- If using SSH remote, specify OS of remote machine: Debian 6.12.73-1 (2026-02-17) x86_64 GNU/Linux

### Bug Summary and Steps to Reproduce

Bug Summary:
The whole UI is laggy after click anywhere in an opened C file or terminal, feels like the frontend is blocked by the cpptools. Disabling the cpptools or open a workspace without C/C++ files make the lag go away.
My guess is the LSP communication is causing the problem.

Steps to reproduce:
1. Open any workspace through remote-ssh that have C file.
2. Open the C file.
3. Click anywhere in the main editor area or terminal.
4. Whole UI stops responding.

Expected behavior:
Work without lag.

### Configuration and Logs

```shell
c_cpp_properties.json

{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "forcedInclude": [
                "${workspaceFolder}/ebpf/ebpf.c"
            ],
            "compilerPath": "/usr/bin/clang",
            "cStandard": "c17",
            "cppStandard": "c++17",
            "intelliSenseMode": "linux-clang-x64",
            "configurationProvider": "ms-vscode.makefile-tools",
            "defines": [
                "_GNU_SOURCE"
            ]
        }
    ],
    "version": 4
}

C/C++: Log Diagnostics

-------- Diagnostics - 2/24/2026, 12:55:31 PM
Version: 1.30.5
Current Configuration:
{
    "name": "Linux",
    "includePath": [
        "/home/kuro/Repositio/udpgrm/**"
    ],
    "forcedInclude": [
        "/home/kuro/Repositio/udpgrm/ebpf/ebpf.c"
    ],
    "compilerPath": "/usr/bin/clang",
    "cStandard": "c17",
    "cppStandard": "c++17",
    "intelliSenseMode": "linux-clang-x64",
    "configurationProvider": "ms-vscode.makefile-tools",
    "defines": [
        "_GNU_SOURCE"
    ],
    "compilerPathIsExplicit": true,
    "cStandardIsExplicit": true,
    "cppStandardIsExplicit": true,
    "intelliSenseModeIsExplicit": true,
[...]

Issue #14168, opened 2026-02-06
Version 1.30.4 has a very, very, very serious memory leak and CPU usage problem.


LittleFatHero:
### Environment

- OS and Version: windows 11 25H2
- VS Code Version: 1.109
- C/C++ Extension Version: 1.30.4
- If using SSH remote, specify OS of remote machine: Ubuntu 20.04

### Bug Summary and Steps to Reproduce

Bug Summary:
I open a aosp（Android Open Source Project）， It keep  doing IntelliSense and parsing workspace for a very long time, It never ended, lasting more than 5 hours. Most importantly, when it does this, VS Code becomes severely laggy and almost unusable. In VS Code, a thread was consuming over 5 GB of memory. When I closed the C++ plugin, the thread consuming 5 GB of memory disappeared. This proves that the C++ plugin has a serious memory leak issue; this thread was also consuming over 15% of CPU.

Steps to reproduce:
1.open vscode;
2. connect to remote machine ubuntu 20.04;
3. open aosp project；
4. open a C source file;

Expected behavior:
VS Code runs smoothly with low memory usage. Even with 5GB of memory used and over 15% CPU usage, VS Code shouldn't lag.

### Configuration and Logs

```shell
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "type": "object",
    "definitions": {
        "configurations": {
            "type": "array",
            "items": {
                "type": "object",
                "required": [
                    "name"
                ],
                "properties": {
                    "name": {
                        "markdownDescription": "Configuration identifier. `Mac`, `Linux`, and `Win32` are special identifiers for configurations that will be auto-selected on those platforms, but the identifier can be anything.",
                        "descriptionHint": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered.",
                        "type": "string"
                    },
                    "compilerPath": {
                        "markdownDescription": "Full path of the compiler being used, e.g. `/usr/bin/gcc`, to enable more accurate IntelliSense.",
                        "descriptionHint": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered.",
                        "type": [
                            "null",
                            "string"
                        ]
                    },
                    "compilerArgs": {
[...]

PR #14316, opened 2026-03-24
Set `C_Cpp.copilotHover` to `disabled` by default.

sean-mcmanus:
The user at https://github.com/microsoft/vscode-cpptools/issues/14211#issuecomment-3963811644 reported this fixed their issue, also a couple users at https://github.com/microsoft/vscode-cpptools/issues/14168#issuecomment-4086899225 .

### DIFF
--- a/Extension/CHANGELOG.md
+++ b/Extension/CHANGELOG.md
@@ -1,5 +1,9 @@
 # C/C++ for Visual Studio Code Changelog
 
+## Version 1.32.0: March 25, 2026
+### Bug Fix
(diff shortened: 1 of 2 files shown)


Colengms (approved):

sean-mcmanus:
There is already a flighting variable, so that might be usable instead: CppCopilotHoverDisabled .

Also, it seems to me like this should get into "release" and not "insiders" if the flighting can't be used.

sean-mcmanus:
We're looking into adding more telemetry instead...

Issue #14341, opened 2026-04-01
Code Actions are enabled by default


luc0x61:
### Environment

- OS and Version: Windows 11 Pro 25H2
- VS Code Version: 1.114.0
- C/C++ Extension Version: 1.31.4
- If using SSH remote, specify OS of remote machine: n.a.

### Bug Summary and Steps to Reproduce

Bug Summary:

Code Actions is enabled by default, without any option to be disabled in settings UI

Steps to reproduce:
1. Open/create a .c/.h file with an enum definition like the following

<img width="283" height="124" alt="Image" src="https://github.com/user-attachments/assets/3a95b5ce-3c1c-4aed-b676-173ef6d1226b" />

2. Left click on a defined enumerator
3. The light bulb icon of Code Actions will appear, and will be enabled to be interacted with (menu)

Expected behavior:

- Processing step 2. there shall not be a resulting icon/menu as in step 3.

### Configuration and Logs

```shell
From settings.json:

    "C_Cpp.vcFormat.indent.preserveWithinParentheses": true,
    "C_Cpp.vcFormat.indent.withinParentheses": "alignToParenthesis",
    "C_Cpp.vcFormat.space.pointerReferenceAlignment": "center",
    "C_Cpp.doxygen.generateOnType": false,
    "C_Cpp.default.defines": [
        "__TMS320C28XX__"
    ],
    "C_Cpp.copilotHover": "disabled",
    "C_Cpp.default.compilerPath": "C:\\mingw64\\bin\\gcc.exe",
    "C_Cpp.default.enableConfigurationSquiggles": false,
    "C_Cpp.codeAnalysis.clangTidy.codeAction.showDocumentation": false,
```

sean-mcmanus:
@luc0x61 This looks like it's from the Generate Doxygen comment feature. It looks like a setting to disable that was never implemented.

sean-mcmanus:
@luc0x61 We have a PR at https://github.com/microsoft/vscode-cpptools/pull/14342 that allows disabing the Generate doxygen comment code actions -- is that good enough or were there more code actions you wanted to be able to disable?

Colengms:
@luc0x61 Are you asking for a way to disable all code actions, or doxygen code actions specifically, while others remain enabled?

It looks like there is a general `editor.lightbulb.enabled` setting that can disable display of all code action lightbulbs.

[...]

PR #14342, opened 2026-04-01
Add setting doxygen.generateOnCodeAction.

sean-mcmanus:
Enables a fix/workaround for https://github.com/microsoft/vscode-cpptools/issues/14341


Colengms:
While this would disable the feature the user used as an example, aren't they asking for all code actions to be disable-able, not just this one?

sean-mcmanus:
> While this would disable the feature the user used as an example, aren't they asking for all code actions to be disable-able, not just this one?

I don't know. I asked the user. If there are more code actions that make sense to disable, we could potentially add settings for those too. I'm not sure a "global" setting needs to be added.

However, I think we would want this setting regardless of the user's feedback regarding other code actions or their defaults.

bobbrow (approved):

PR #14351, opened 2026-04-03
Run without debugging

### LINKED ISSUE #1201: Support "Run without debugging"
I'm running VSCode 1.17.02 on Ubuntu 16.04 x64 and cpptools 0.14.

Could you please add information about how to launch an application w/o debugging (CTRL+F5) to the VSCode C++ docs and maybe post a complete tasks.json / launch.json example here? I've read a couple of issues now and can't quite make launching work...
I use make to build my projects. I put
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch",
            "runtimeExecutable": "MYEXECUTABLE",
            "program": "${workspaceRoot}/MYOUTPUT.FILE",
            "preLaunchTask": "Make"
            }
    ]
}
```
into launch.json. My tasks.json looks like this:
```
{
    "version": "2.0.0",
    "tasks": [
        {
            "taskName": "Make",
            "type": "shell",
            "command": "make",
            "args": [""],
            "group": "build",
            "presentation": {
                "reveal": "always",
                "panel": "dedicated"
            },
            "problemMatcher": {
                "owner": "cpp",
                "fileLocation": ["relative", "${workspaceRoot}"],
                "pattern": {
                    "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
                    "file": 1,
                    "line": 2,
                    "column": 3,
                    "severity": 4,
                    "message": 5
                }
            }
        }
    ]
}
```
The binary file is built successfully via CTRL+SHIFT+B, but when launching via CTRL+F5 I get the message "No executable targets are available". I guess vscode does not know what output file make generates...

bobbrow:
Fixes: #1201

Supports both the Play button in the editor (run active file) and configurations specified in launch.json.  macOS support needs to be tested - I don't currently have access to a machine.

*Co-authored by GitHub Copilot*

### DIFF
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -147,7 +147,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
                 Telemetry.logDebuggerEvent(DebuggerEvent.debugPanel, { "debugType": DebugType.debug, "configSource": folder ? ConfigSource.workspaceFolder : ConfigSource.singleFile, "configMode": ConfigMode.noLaunchConfig, "cancelled": "true", "succeeded": "true" });
                 return undefined; // aborts debugging silently
             } else {
-                const noDebug = config.noDebug ?? false; // preserve the noDebug value from the config if it exists.
+                const noDebug = config.noDebug ?? false; // Preserve the noDebug value from the config if it exists.
                 // Currently, we expect only one debug config to be selected.
                 console.assert(configs.length === 1, "More than one debug config is selected.");
                 config = configs[0];
--- a/Extension/src/Debugger/configurationProvider.ts
+++ b/Extension/src/Debugger/configurationProvider.ts
@@ -147,12 +147,14 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
                 Telemetry.logDebuggerEvent(DebuggerEvent.debugPanel, { "debugType": DebugType.debug, "configSource": folder ? ConfigSource.workspaceFolder : ConfigSource.singleFile, "configMode": ConfigMode.noLaunchConfig, "cancelled": "true", "succeeded": "true" });
                 return undefined; // aborts debugging silently
             } else {
+                const noDebug = config.noDebug ?? false; // preserve the noDebug value from the config if it exists.
                 // Currently, we expect only one debug config to be selected.
                 console.assert(configs.length === 1, "More than one debug config is selected.");
                 config = configs[0];
                 // Keep track of the entry point where the debug config has been selected, for telemetry purposes.
                 config.debuggerEvent = DebuggerEvent.debugPanel;
                 config.configSource = folder ? ConfigSource.workspaceFolder : ConfigSource.singleFile;
+                config.noDebug = noDebug;
             }
         }
 
--- a/Extension/src/Debugger/configurations.ts
+++ b/Extension/src/Debugger/configurations.ts
@@ -97,7 +97,7 @@ function createLaunchString(name: string, type: string, executable: string): str
 "stopAtEntry": false,
 "cwd": "$\{fileDirname\}",
 "environment": [],
-${ type === "cppdbg" ? `"externalConsole": false` : `"console": "externalTerminal"` }
+${ type === "cppdbg" ? `"externalConsole": false` : `"console": "integratedTerminal"` }
 `;
 }
 
--- a/Extension/src/Debugger/configurations.ts
+++ b/Extension/src/Debugger/configurations.ts
@@ -97,7 +97,7 @@ function createLaunchString(name: string, type: string, executable: string): str
 "stopAtEntry": false,
 "cwd": "$\{fileDirname\}",
 "environment": [],
-${ type === "cppdbg" ? `"externalConsole": false` : `"console": "integratedTerminal"` }
+${type === "cppdbg" ? `"externalConsole": false` : `"console": "internalConsole"`}
 `;
 }
 
@@ -164,7 +164,7 @@ export class MIConfigurations extends Configuration {
 \t${indentJsonString(createLaunchString(name, this.miDebugger, this.executable))},
 \t"MIMode": "${this.MIMode}"{0}{1}
 }`, [this.miDebugger === "cppdbg" && os.platform() === "win32" ? `,${os.EOL}\t"miDebuggerPath": "/path/to/gdb"` : "",
-            this.additionalProperties ? `,${os.EOL}\t${indentJsonString(this.additionalProperties)}` : ""]);
+        this.additionalProperties ? `,${os.EOL}\t${indentJsonString(this.additionalProperties)}` : ""]);
 
         return {
             "label": configPrefix + name,
@@ -182,7 +182,7 @@ export class MIConfigurations extends Configuration {
 \t${indentJsonString(createAttachString(name, this.miDebugger, this.executable))}
 \t"MIMode": "${this.MIMode}"{0}{1}
 }`, [this.miDebugger === "cppdbg" && os.platform() === "win32" ? `,${os.EOL}\t"miDebuggerPath": "/path/to/gdb"` : "",
-            this.additionalProperties ? `,${os.EOL}\t${indentJsonString(this.additionalProperties)}` : ""]);
+        this.additionalProperties ? `,${os.EOL}\t${indentJsonString(this.additionalProperties)}` : ""]);
 
         return {
             "label": configPrefix + name,
--- a/Extension/src/Debugger/runWithoutDebuggingAdapter.ts
+++ b/Extension/src/Debugger/runWithoutDebuggingAdapter.ts
@@ -7,7 +7,11 @@ import * as cp from 'child_process';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
+import * as nls from 'vscode-nls';
 import { sessionIsWsl } from '../common';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize = nls.loadMessageBundle();
 
 /**
  * A minimal inline Debug Adapter that runs the target program directly without a debug adapter
@@ -117,9 +121,46 @@ export class RunWithoutDebuggingAdapter implements vscode.DebugAdapter {
         } else if (platform === 'linux' && sessionIsWsl()) {
             cp.spawn('/mnt/c/Windows/System32/cmd.exe', ['/c', 'start', 'bash', '-c', `${cmdLine};read -p 'Press enter to continue...'`], { env, detached: true, stdio: 'ignore' }).unref();
         } else { // platform === 'linux'
-            cp.spawn('bash', ['-c', `${cmdLine};read -p 'Press enter to continue...'`], { cwd, env, detached: true, stdio: 'ignore' }).unref();
+            this.launchLinuxExternalTerminal(cmdLine, cwd, env);
         }
         this.sendEvent('terminated');
+    }
+
+    /**
+     * On Linux, find and launch an available terminal emulator to run the command.
+     */
+    private launchLinuxExternalTerminal(cmdLine: string, cwd: string | undefined, env: NodeJS.ProcessEnv): void {
+        const bashCmd = `${cmdLine}; echo; read -p 'Press enter to continue...'`;
+        const bashArgs = ['bash', '-c', bashCmd];
+
+        // Terminal emulators in order of preference, with the correct flag style for each.
+        const candidates: { cmd: string; buildArgs: () => string[] }[] = [
+            { cmd: 'x-terminal-emulator', buildArgs: () => ['-e', ...bashArgs] },
+            { cmd: 'gnome-terminal', buildArgs: () => ['-e', ...bashArgs] },
+            { cmd: 'konsole', buildArgs: () => ['-e', ...bashArgs] },
+            { cmd: 'xterm', buildArgs: () => ['-e', ...bashArgs] }
+        ];
+
+        // Honor the $TERMINAL environment variable if set.
+        const terminalEnv = process.env['TERMINAL'];
+        if (terminalEnv) {
+            candidates.unshift({ cmd: terminalEnv, buildArgs: () => ['-e', ...bashArgs] });
+        }
+
+        for (const candidate of candidates) {
+            try {
+                const result = cp.spawnSync('which', [candidate.cmd], { stdio: 'pipe' });
+                if (result.status === 0) {
+                    cp.spawn(candidate.cmd, candidate.buildArgs(), { cwd, env, detached: true, stdio: 'ignore' }).unref();
+                    return;
+                }
+            } catch {
+                continue;
+            }
+        }
+
+        const message = localize('no.terminal.emulator', 'No terminal emulator found. Please set the $TERMINAL environment variable to your terminal emulator of choice, or install one of the following: x-terminal-emulator, gnome-terminal, konsole, xterm.');
+        vscode.window.showErrorMessage(message);
     }
 
     /**
--- a/Extension/src/Debugger/runWithoutDebuggingAdapter.ts
+++ b/Extension/src/Debugger/runWithoutDebuggingAdapter.ts
@@ -7,6 +7,7 @@ import * as cp from 'child_process';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
+import { sessionIsWsl } from '../common';
 
 /**
  * A minimal inline Debug Adapter that runs the target program directly without a debug adapter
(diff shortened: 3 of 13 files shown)


Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -0,0 +1,226 @@
...
+     * Launch the program in an external terminal. We do not keep track of this terminal or the spawned process.
+     */
+    private launchExternalTerminal(program: string, args: string[], cwd: string | undefined, env: NodeJS.ProcessEnv): void {
+        const quotedArgs: string[] = [program, ...args].map(a => this.quoteArg(a));
+        const cmdLine: string = quotedArgs.join(' ');
+        const platform: string = os.platform();
+        if (platform === 'win32') {
+            cp.spawn('cmd.exe', ['/c', 'start', 'cmd.exe', '/K', cmdLine], { cwd, env, detached: true, stdio: 'ignore' }).unref();
+        } else if (platform === 'darwin') {
+            cp.spawn('osascript', ['-e', `tell application "Terminal" to do script "${this.escapeQuotes(cmdLine)}"`], { cwd, env, detached: true, stdio: 'ignore' }).unref();
```
bobbrow, line 154: This needs to be tested.

sean-mcmanus: @bobbrow I've tested this code path on mac (it worked).

bobbrow: @sean-mcmanus cool! Does it work for paths with spaces and args with spaces too?

Extension/src/Debugger/configurations.ts
```diff
@@ -97,7 +97,7 @@ function createLaunchString(name: string, type: string, executable: string): str
 "stopAtEntry": false,
 "cwd": "$\{fileDirname\}",
 "environment": [],
-${ type === "cppdbg" ? `"externalConsole": false` : `"console": "externalTerminal"` }
+${ type === "cppdbg" ? `"externalConsole": false` : `"console": "integratedTerminal"` }
```
sean-mcmanus: Why is this changing?

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -147,12 +147,14 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
                 Telemetry.logDebuggerEvent(DebuggerEvent.debugPanel, { "debugType": DebugType.debug, "configSource": folder ? ConfigSource.workspaceFolder : ConfigSource.singleFile, "configMode": ConfigMode.noLaunchConfig, "cancelled": "true", "succeeded": "true" });
                 return undefined; // aborts debugging silently
             } else {
+                const noDebug = config.noDebug ?? false; // preserve the noDebug value from the config if it exists.
```
sean-mcmanus: Comments ends with a period but doesn't start with a capital letter (versus the case on the next line).

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
sean-mcmanus: @bobbrow Yeah, it works with folders/files with spaces in the path, but what do you mean by "args with spaces"? i.e. I see the args be an array of strings. Do you mean like `[ "one two", "three four" ]`?

bobbrow: Yes. That's what I mean.

Extension/src/Debugger/configurations.ts
bobbrow: cppdbg is "internal" by default, so I believe that cppvsdbg should be "internal" by default as well.

Extension/src/Debugger/configurationProvider.ts
bobbrow: Fixed

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
sean-mcmanus: Yeah, that works.

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -0,0 +1,308 @@
...
+            this.sendEvent('exited', { exitCode: 1 });
+            this.sendEvent('terminated');
+        });
+        this.childProcess.on('exit', (code: number | null) => {
+            this.sendEvent('exited', { exitCode: code ?? 0 });
+            this.sendEvent('terminated');
+        });
+    }
+
+    private escapeQuotes(arg: string): string {
```
Colengms, line 200: New functions that escape (only) quotes or add quotes, are a red flag for me.  There are special rules for quoting that differ between shells/platforms.

I'm not sure if it's suitable for your purposes, but some of those rules appear to be encoded into `buildShellCommandLine` (which appears to be used for launching build tasks).

Taken at face value, a helper function named "quoteArg" is likely inherently problematic.  Does it deal with existing quoted or escaped characters properly?  Should it be named/commented to enforce the rule that it only be called on strings we generate that are known not to have escaping already (i.e. user input, or parsed from an existing command line).  Does it deal with Windows weird rules for quote escaping?

Colengms (changes requested):

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
bobbrow: @Colengms 
`escapeQuotes` will need to stay because of how the macOS launcher works (see line 133), but I can use `buildShellCommandLine` for building the command line instead of `quoteArgs` (line 127-128) if that is better. Since I don't have the "original" command, do I just pass empty string?

Looking at the one usage of `buildShellCommandLine` in the source code, it appears that originalCommand and command should be swapped...?

bobbrow:
@WardenGnaw had some feedback about dealing with some of the other properties in the launch.json config and shared a commit in MIEngine where these things might have been handled.  The cases where the config asks for something other than a local launch of a program should be handled in some way (either fail the launch session or fall back to regular debugging mode).

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
bobbrow: Switched to `buildShellCommandLine`

Extension/src/Debugger/configurations.ts
sean-mcmanus: But why was it set to externalTerminal in the 1st place?

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
```diff
@@ -0,0 +1,306 @@
...
+                clearTimeout(timeoutHandle);
+                timeoutHandle = undefined;
+            }
+            trackerRegistration?.dispose();
+            trackerRegistration = undefined;
+        }
+    };
+}
+
+suite('Run Without Debugging Integration Test', function (): void {
```
Colengms: Should this be with other tests somewhere (other than common.ts)?

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    if (isLinux) {
+        const result = await runProcess('g++', ['-std=c++17', '-g', sourcePath, '-o', outputPath], workspacePath);
+        assert.strictEqual(result.code, 0, `g++ compilation failed. stdout: ${result.stdout}\nstderr: ${result.stderr}`);
+        return;
+    }
+
+    assert.fail(`Unsupported test platform: ${process.platform}`);
+}
+
+async function createBreakpointAtReturnStatement(sourceUri: vscode.Uri): Promise<vscode.SourceBreakpoint> {
```
Colengms: Maybe debugger related helpers should go in Debugger\utils.ts ?

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    const promise = vscode.commands.executeCommand('C_Cpp.SetVsDeveloperEnvironment', 'test');
+    const timer = setInterval(() => {
+        void vscode.commands.executeCommand('workbench.action.acceptSelectedQuickOpenItem');
+    }, 1000);
+    await promise;
+    clearInterval(timer);
+    assert.strictEqual(util.hasMsvcEnvironment(), true, 'MSVC environment not set correctly.');
+}
+
+async function compileProgram(workspacePath: string, sourcePath: string, outputPath: string): Promise<void> {
```
Colengms: What about other needed compiler args?

We also kick off builds in `cppBuildTaskProvider.ts` .  It considers some other details there, such as other compiler args and ensuring ansi color coding is used in the output (which we could also get if we fixed https://github.com/microsoft/vscode-cpptools/issues/9681 , so we could use node-pty ).

Extension/src/Debugger/configurations.ts
sean-mcmanus: The default is still internalConsole in the package.json.

sean-mcmanus: Oh, yeah, that matches the package.json default.

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
```diff
@@ -0,0 +1,306 @@
...
+
+interface TrackerController {
+    state: TrackerState;
+    lastEvent: Promise<'stopped' | 'exited'>;
+    dispose(): void;
+}
+
+function runProcess(command: string, args: string[], cwd: string, env?: NodeJS.ProcessEnv): Promise<ProcessResult> {
+    return new Promise((resolve, reject) => {
+        const child = cp.spawn(command, args, { cwd, env });
```
Colengms: If you're quoting args and using `spawn`, you'll probably want to pass `shell: true`. Default would be false, and it wouldn't expect there to be escaping/quoting.  We also have `spawnChildProcess` and `execChildProcess`.  There is also direct use of `exec` after the call to `buildShellCommandLine` in `cppBuildTaskProvider.ts`.

I remember dealing with the whole shell-escaping can of worms once upon a time, and at least one additional spawn codepath (In Garrett's Process.ts) has snuck in since.  I believe the right expectation is:
- If we receive a command line from a user, that is a shell concept and must already contain shell quoting/escaping if required. We shouldn't add/remove quoting/escaping if we need to pass it on as a shell command line.  If we need to parse it into individual args, we need to remove the quoting/escaping.
- If we receive individual arguments (in an array), that is not a shell concept, and those arguments should not contain shell quoting/escaping.  If we need to pass it on a command line through a shell, we need to add quoting/escaping.

Colengms (changes requested):

Extension/src/Debugger/configurations.ts
sean-mcmanus: i.e. yeah, this seems like a pre-existing bug.

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
bobbrow: There aren't any other needed compiler args. This is a very simple test and compilation is just one of the steps needed to set it up. I wasn't planning to set up tasks, etc. Do you feel that would be necessary?

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
bobbrow: This helper is kind of specific to the test. I suppose a more generic helper could be written, but it doesn't seem to be needed quite yet. I can revisit this if we start adding lots of debugger tests.

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
bobbrow: Not sure I follow. What are you asking here? You want me to put this with the language server test scenarios?

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
bobbrow: Right now, I don't forsee the need for that level of support. This is only used for building the code and the inputs are hard coded. I could move this into the `compileProgram` function if that helps solidify that this isn't meant to be a general purpose helper method.

Extension/src/Debugger/configurations.ts
bobbrow: Oh, this should be 'internalConsole', shouldn't it? I want the default for "run without debugging" to be 'integratedTerminal', not 'internalConsole' though. I wonder if we should not allow 'internalConsole' at all for "run without debugging". (e.g. switch out 'internalConsole' for 'integratedTerminal' whenever we see it)

@WardenGnaw do you have an opinion on this?

WardenGnaw: If it's not backed by MIEngine, `integratedTerminal` makes sense. 

`internalConsole` is good if there's different types of `OutputEvents` and VS Code can colorize it.

> I wonder if we should not allow 'internalConsole' at all for "run without debugging". 

`cppvsdbg` supports `noDebug`, so I wouldnt change it for that debug type.

bobbrow: @WardenGnaw the reason I ask is because most people will likely reuse the same debugger config for "run without debugging." For debugging they probably want internalConsole, but for noDebug, I would think they'd expect output to show up in the integratedTerminal. So I expect the config to say 'internalConsole', but we could map it to 'integratedTerminal' for the noDebug case.

Colengms (dismissed):

Extension/src/Debugger/configurations.ts
bobbrow: I am removing internalConsole as an option for run without debugging.

sean-mcmanus (changes requested):
I'm seeing Debug work, but Run not work on mac. I'm still investigating...

Seems like it fails when there's a space in the workspace folder name that has the executable (that case was working earlier).

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -0,0 +1,282 @@
...
+        });
+        this.terminal.show(true);
+
+        const shellIntegration: vscode.TerminalShellIntegration | undefined =
+            this.terminal.shellIntegration ?? await this.waitForShellIntegration(this.terminal, 3000);
+
+        // Not all terminals support shell integration. If it's not available, we'll just send the command as text though we won't be able to monitor its execution.
+        if (shellIntegration) {
+            this.monitorIntegratedTerminal(this.terminal);
+            if (program.includes(' ')) {
```
sean-mcmanus: Should you use the same check that VS Code does for the args `const wrapInQuotes = !arg.match(/["'`]/) && arg.match(/\s/); `?

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -108,6 +108,10 @@ export class RunWithoutDebuggingAdapter implements vscode.DebugAdapter {
         // Not all terminals support shell integration. If it's not available, we'll just send the command as text though we won't be able to monitor its execution.
         if (shellIntegration) {
             this.monitorIntegratedTerminal(this.terminal);
+            if (program.includes(' ')) {
```
Colengms: I opened a bug against vscode here: https://github.com/microsoft/vscode/issues/308890

If fixed, wrapping in quotes may break in the future.

sean-mcmanus: @Colengms I assume they would have a similar check for pre-existing quotes that they currently have for pre-existing quotes in args, so it wouldn't break, right?

Colengms: There is no safe, "it may or may not have shell quoting/escaping" scenario.  If they add quotes, it will be double-quoted with the internal quotes considered literal, which would not successfully match the file in the filesystem

sean-mcmanus: @Colengms But is the quoting that VS Code does for the args okay for the "program"?

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
bobbrow: Sure

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
bobbrow: We'll just have to react to a change if VS Code fixes that bug.

bobbrow:
> I'm seeing Debug work, but Run not work on mac. I'm still investigating...
> 
> Seems like it fails when there's a space in the workspace folder name that has the executable (that case was working earlier).

@sean-mcmanus 
The latest commit flushes out the remaining issues on Windows, Linux, and WSL.  You can try running the new tests on macOS now. They cover spacing in the program name and args for both internal and external terminals.

sean-mcmanus (approved):

sean-mcmanus (approved):

PR #14354, opened 2026-04-07
Fix npx @vscode/vsce.

sean-mcmanus:
Follow up to https://github.com/microsoft/vscode-cpptools/pull/14346

These are the changes recommended by Copilot from the failures it was hitting during testing.

Tested with the extension-pack and cg vsix.


sean-mcmanus:
I'm still testing this...might not work...

UPDATE: Tested the extension-pack and that worked.

Colengms (approved):

PR #14355, opened 2026-04-08
Fix cg.yml.

sean-mcmanus:
Follow up to https://github.com/microsoft/vscode-cpptools/pull/14354 .

It was bypassing the yarn.lock (package.json dependency overrides)l.


Build/cg/cg.yml
tatafyry26, line 90: Renamed

sean-mcmanus: What?

bobbrow (approved):

PR #14366, opened 2026-04-12
Implement per-workspace C/C++ configurations (#14363)

### LINKED ISSUE #14363: There is no way to set c_cpp_properties.json (with multiple configurations) per-workspace
### Feature Request

1. Have 2+ workspace files that contain the same folder, which has a c_cpp_properties.json with 2+ configurations.

Bug: All the workspaces share the same c_cpp_properties.json so I can't make a change to c_cpp_properties.json that I only want to apply to one of the workspace folders. Also, if each workspace has a C_Cpp.default.* setting set (such as a database location), then I can't use the multiple configurations in a c_cpp_properties.json to switch between settings if I'm using the workspace (e.g. if I want to switch between using different configuration providers each with a different database). E.g. with 2 workspaces and 2 configurations in a c_cpp_properties.json I can't get 4 databases to be used, so switching either configurations or workspaces causes unnecessary re-tag parsing.

Subham-KRLX:
This PR introduces the `C_Cpp.configurations` setting, allowing users to define C/C++ configurations directly within VS Code workspace or folder settings. This resolves **#14363**, where multiple workspaces sharing the same folder were forced to share a single `c_cpp_properties.json` file.

## Key Changes

- **Precedence Logic**: Updated the configuration parser to prioritize `C_Cpp.configurations` from settings over the local `c_cpp_properties.json` file.
- **Full Parity**: The new settings-based configurations support the exact same schema, variable resolution (e.g., `${workspaceFolder}`), and sanitization logic as the traditional file-based approach.
- **Reactive Updates**: Integrated with `onDidChangeSettings` to ensure IntelliSense/Status Bar UI updates immediately when settings are modified.
- **Backward Compatibility**: If the new setting is not present, the extension falls back to existing file-based behavior with no disruption.

## Verification
- Verified that workspace settings override local `c_cpp_properties.json`.
- Verified fallback behavior when settings are cleared.
- Confirmed that variable resolution remains consistent across both configuration sources.

Fixes #14363


Code Samples/Fib/.vscode/launch.json
```diff
@@ -3,7 +3,6 @@
     "configurations": [
         {
             "name": "(gdb) Launch",
-            "preLaunchTask": "build",
```
sean-mcmanus, line 6: @Subham-KRLX Can you revert the Code Samples folder changes?

Extension/src/LanguageServer/configurations.ts
```diff
@@ -1452,24 +1453,46 @@ export class CppProperties {
...
+        const settings: CppSettings = new CppSettings(this.rootUri);
+        const settingsConfigs: any[] | undefined = settings.configurations;
+
         let success: boolean = true;
         const firstParse = this.configurationJson === undefined;
-        try {
-            const readResults: string = fs.readFileSync(this.propertiesFile.fsPath, 'utf8');
-            if (readResults === "") {
-                return false; // Repros randomly when the file is initially created. The parse will get called again after the file is written.
+
```
sean-mcmanus, line 1461: There's a "try" removed above, but the catch doesn't appear to be removed. That seems to cause a bunch of compiler errors.

Extension/package.json
```diff
@@ -1060,6 +1060,88 @@
                         "markdownDescription": "%c_cpp.configuration.default.recursiveIncludes.order.markdownDescription%",
                         "scope": "resource"
                     },
+                    "C_Cpp.configurations": {
+                        "type": "array",
+                        "description": "A list of configurations that will be used instead of c_cpp_properties.json when present in workspace settings.",
```
sean-mcmanus, line 1065: Can you reference localizable strings, e.g. %c_cpp.configuration.default.recursiveIncludes.order.markdownDescription% in package.nls.json? Or did you want us to do that?

sean-mcmanus (changes requested):
After feedback from Bob, this PR appears unnecessary, and the current design is sufficient via using additional configurations with different names corresponding to which workspace is being used.

heartacker:
how

sean-mcmanus:
> how

@heartacker What are you asking exactly? I added a longer comment in the original issue.

PR #14370, opened 2026-04-13
Add a script to copy language server and debugger binaries to the Extension folder

bobbrow:
I frequently want to get the latest binaries without building in Release mode or tracking down a debugger build drop.

*Co-authored by Copilot*

### DIFF
--- a/Extension/.scripts/copyExtensionBinaries.ts
+++ b/Extension/.scripts/copyExtensionBinaries.ts
@@ -80,7 +80,9 @@ async function findLatestInstalledExtension(providedPath?: string): Promise<stri
 
     const searchRoots: string[] = [
         join(homedir(), '.vscode', 'extensions'),
-        join(homedir(), '.vscode-insiders', 'extensions')
+        join(homedir(), '.vscode-insiders', 'extensions'),
+        join(homedir(), '.vscode-server', 'extensions'),
+        join(homedir(), '.vscode-server-insiders', 'extensions')
     ];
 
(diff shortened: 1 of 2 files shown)


sean-mcmanus (changes requested):
Can you add the LLVM/bin folder too for clang-tidy and clang-format binaries?

sean-mcmanus:
FYI, on mac, I think a command like `xattr -dr com.apple.quarantine <file>` might also be required, but I could poentially update the script later to handle that...unless you wanted to try to handle that now.

bobbrow:
> Can you add the LLVM/bin folder too for clang-tidy and clang-format binaries?

done.

bobbrow:
> FYI, on mac, I think a command like `xattr -dr com.apple.quarantine <file>` might also be required, but I could poentially update the script later to handle that...unless you wanted to try to handle that now.

I can't really test on mac, so if you want to do it after, that would be good.

sean-mcmanus (approved):

sean-mcmanus:
@bobbrow It seems like `xattr -dr com.apple.quarantine` isn't required when that script is used to copy it? I'll re-evaluate if I ever see the binary is blocked popup again. Or it's possible the xattr is saved and persists.

bobbrow:
> @bobbrow It seems like `xattr -dr com.apple.quarantine` isn't required when that script is used to copy it? I'll re-evaluate if I ever see the binary is blocked popup again. Or it's possible the xattr is saved and persists.

ok

PR #14382, opened 2026-04-15
Set context flag for when C++ language server is activated

### LINKED ISSUE #14392: DevTools don't recognize the C/C++ extension after Restart Extensions
When DevTools is used for a C/C++ tool ("cpp-devtools.enableCppCodeEditingTools": true), and then the C/C++ extension is updated and Restart Extensions is used, then the DevTools tools stop working until a Reload Window happens.

Fixed by https://github.com/microsoft/vscode-cpptools/pull/14382

DavidARaygoza:
The primary purpose of this is for this flag to be consumed by the C++ devtools extensions. This way the C++ specific Copilot tools will only appear in tools list and register when the C++ language server is actually activated.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -275,6 +275,7 @@ export async function activate(): Promise<void> {
     }
 
     await registerRelatedFilesProvider();
+    await vscode.commands.executeCommand('setContext', 'cpptools.languageServerActivated', true);
 }
 
 export function updateLanguageConfigurations(): void {


bobbrow (dismissed):

Extension/src/LanguageServer/extension.ts
```diff
@@ -169,6 +169,7 @@ export async function activate(): Promise<void> {
         }
     }
 
+    await vscode.commands.executeCommand('setContext', 'cpptools.languageServerActivated', true);
```
sean-mcmanus: Shouldn't this be moved to the end of activate()? The C_Cpp commands aren't registered yet at this point.

sean-mcmanus (changes requested):

Extension/src/LanguageServer/extension.ts
bobbrow: For context, this is used to register the models for DevTools. It doesn't mean any cpptools work is going to be immediately required. But I'm ok with moving it to the end of the function too if you prefer. I had that same internal debate with myself but ultimately decided I didn't mind where this line went.

DavidARaygoza: Actually that does sound better, that might help with the issue Carson was having where devtools is loading commands before cpptools registers them. Just re-pushed

sean-mcmanus (approved):

sean-mcmanus (changes requested):
Should it be set to `false` if the language server is disabled?

sean-mcmanus (approved):

sean-mcmanus:
> Should it be set to `false` if the language server is disabled?

Oh, it appears we require a Reload Window in that case so it's not an issue.

sean-mcmanus:
Well, actually, I don't know what happens if the C/C++ is disabled via the Extension window....but I can try that out later.

PR #14386, opened 2026-04-16
Add a progress indicator with cancel for Switch Header/Source.

### LINKED ISSUE #14379: Using the shortcut to get to the header files the loading animation doesn't show up
### Environment

- OS and Version: tried on windows, rhel9, ubuntu.
- VS Code Version: 1.115.0
- C/C++ Extension Version: 1.31.4
- If using SSH remote, specify OS of remote machine: experienced both ssh and local instances.

If I use the shortcut to get the header file of the cpp I'm currently exploring, the loading animation doesn't show up. Unfortunately, this process can last very long and sometimes (I think maybe because of a timer?) it expires. What's even worse is that, unlike any other vscode feature, this can't be stopped by any user interactions. As a result, if I get bored about the waiting times I keep doing something else but all of a sudden the header shows up. That happens very often in my environments.

I would love to contribute to this project by solving this by my own and opening the PR if I'm allowed to do so. Thank you in advance.

Bug Summary:
The shortcut to get to the header files could result in unexpected behavior. You probably need to work on a big repository for this action to experience long enough waiting times.

Steps to reproduce:
1. Go to a cpp file.
2. Use the shortcut to get to the header file of that cpp.

Expected behavior:
1. Loading animation won't show up.
2. The action can't be stopped.
3. If you keep doing something else, the header file could show up all of a sudden.

### Configuration and Logs

```shell
-------- Diagnostics - 4/15/2026, 3:40:43 PM
Version: 1.31.4
Current Configuration:
{
    "name": "Linux",
    "includePath": [
        "/root/git/*******"
    ],
    "defines": [],
    "compilerPath": "/usr/bin/gcc",
    "cStandard": "c17",
    "cppStandard": "c++98",
    "intelliSenseMode": "linux-gcc-x64",
    "compilerPathInCppPropertiesJson": "/usr/bin/gcc",
    "intelliSenseModeIsExplicit": false,
    "cStandardIsExplicit": false,
    "cppStandardIsExplicit": false,
    "mergeConfigurations": false,
    "recursiveIncludes": {},
    "recursiveIncludesReduceIsExplicit": false,
    "recursiveIncludesPriorityIsExplicit": false,
    "recursiveIncludesOrderIsExplicit": false,
    "compilerPathIsExplicit": false,
    "browse": {
    }
}
Additional Tracked Settings:
{
    "editorTabSize": 4,
    "editorInsertSpaces": true,
    "editorAutoClosingBrackets": "languageDefined",
    "filesEncoding": "utf8",
    "filesAssociations": {},
    "filesExclude": {
        "**/.git": true,
        "**/.svn": true,
        "**/.hg": true,
        "**/.DS_Store": true,
        "**/Thumbs.db": true
    },
    "filesAutoSaveAfterDelay": false,
    "editorInlayHintsEnabled": true,
    "editorParameterHintsEnabled": true,
    "searchExclude": {
        "**/node_modules": true,
        "**/bower_components": true,
        "**/*.code-search": true
    },
    "workbenchSettingsEditor": "ui"
}
```

### Other Extensions

_No response_

### Additional context

Unfortunately I can't share much of the logs because I'm on a company environment.

sean-mcmanus:
Fixes https://github.com/microsoft/vscode-cpptools/issues/14379 .

Implemented by Copilot (took a few iterations though).

Screenshot 2026-04-16 at 10 56 04 AM

Waits 2 seconds before showing the notification (same as Find All References).


Extension/src/LanguageServer/extension.ts
```diff
@@ -478,19 +478,28 @@ async function onSwitchHeaderSource(): Promise<void> {
...
+        let targetFileNameReplaced: boolean = false;
+        clients.forEach(client => {
+            if (!targetFileNameReplaced && client.RootRealPath && client.RootPath !== client.RootRealPath
+                && targetFileName.startsWith(client.RootRealPath)) {
+                targetFileName = client.RootPath + targetFileName.substring(client.RootRealPath.length);
+                targetFileNameReplaced = true;
+            }
+        });
+        const document: vscode.TextDocument = await vscode.workspace.openTextDocument(targetFileName);
+        await vscode.window.showTextDocument(document).then(undefined, logAndReturn.undefined);
```
bobbrow: Can we make it wait at least 1 second before appearing? Generally switch header/source should be pretty fast, right? I don't want to start popping UI all the time if it's not needed.

sean-mcmanus: Yeah, good idea (looks like the Find All References notification uses 2 seconds so I'll try that).

Colengms (approved):

PR #14394, opened 2026-04-17
Update MIEngine changelog entries.

sean-mcmanus:


sean-mcmanus:
We'll try building the vsix with the older MIEngine instead.

PR #14397, opened 2026-04-18
Fix stuck "Find All References" progress notifications on overlapping requests

### LINKED ISSUE #14381: 1.32.0 pre-release: Multiple Find All References progress UIs get stuck if DevTools invokes a FAR operation
With 0.5.4 DevTools. Doesn't repro with 1.31.4 of the C/C++ extension.

I think the repro requires invoking it on a symbol that occurs in enough files to cause the progress UI to appear.

Subham-KRLX:
Closes #14381

Fixed a regression where overlapping "Find All References" requests (e.g., from DevTools) caused notifications to get stuck.
The fix ensures ReferencesManager properly clears previous timers and resolves active promises before starting a new request, preventing orphaned background loops. Verified via simulation.

### DIFF
--- a/Extension/src/LanguageServer/references.ts
+++ b/Extension/src/LanguageServer/references.ts
@@ -346,6 +346,7 @@ export class ReferencesManager {
     }
 
     private handleProgressStarted(referencesProgress: ReferencesProgress): void {
+        this.resetProgressBar();
         this.referencesStartedWhileTagParsing = this.client.IsTagParsing;
 
         let mode: ReferencesCommandMode = ReferencesCommandMode.None;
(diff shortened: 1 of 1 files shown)


carsonRadtke (commented):
Just want to add some additional context here:

Note: Everything I'm describing is when pre-release (0.5.4) devtools is paired with pre-release (1.32.0) cpptools.

Requests from devtools should not be interacting with the UI; they should be using the new, silent, `C_Cpp.FindAllReferences` command meaning the UI progress dialog never appears. In an effort to make new devtools compatible with older cpptools, devtools asks vscode at extension startup if the `C_Cpp.FindAllReferences` command exists. If not, it falls back to `vscode.executeReferenceProvider` which issues UI progress updates.

What was happening is that vscode was loading devtools before cpptools. This was causing devtools to determine that `C_Cpp.FindAllReferences` didn't exist (because the command was not yet registered). This bug has since been fixed internally on the devtools side.

---

Now there is a question of what happens when devtools _correctly_ identifies that it should fallback to `vscode.executeReferenceProvider` -- devtools may fire off several concurrent find all references requests which will trigger UI progress updates and some of them may get stuck. I believe this PR is still valuable in that scenario where new devtools are paired with old cpptools.

@bobbrow, WDYT - is this backwards compatibility something we wish to maintain?
@DavidARaygoza, is any additional context relevant on the internal devtools fix?
@sean-mcmanus, the parts of code that this diff touches are not familiar to me, but it seems OK. Requesting further review from you.

PR #14407, opened 2026-04-24
Adding additional parse checking logic

sergior2:
New handler to help determine if parsing has reached an idle state. Meaning no pending calls and finished workspace parsing, file parsing, and intellisense updates.


Extension/src/LanguageServer/client.ts
```diff
@@ -1008,6 +1010,56 @@ export class DefaultClient implements Client {
         };
     }
 
+    private isIdle(): boolean {
+        return !this.model.isInitializingWorkspace.Value
+            && !this.model.isIndexingWorkspace.Value
+            && !this.model.isParsingWorkspace.Value
+            && !this.model.isParsingFiles.Value
+            && !this.model.isUpdatingIntelliSense.Value;
```
carsonRadtke: Can this just be?

```suggestion
        return !this.IsTagParsing() && !this.model.isUpdatingIntelliSense.Value;
```

Nit: you didn't introduce this, but since all these properties being checked belong to `this.model`, I might expect there to be a `ClientModel.IsIdle(): bool` or something.

sergior2: Will add the suggested change, although I'm not sure I follow when you say there to be a `ClientModel.IsIdle()`?

Is that not what `isIdle()` currently fulfilling?

Extension/src/LanguageServer/extension.ts
```diff
@@ -977,6 +978,10 @@ function onGetActiveConfigCustomVariable(variableName: string): Thenable<string>
     return clients.ActiveClient.getCurrentConfigCustomVariable(variableName);
 }
 
+async function onWaitForIdle(timeout?: number): Promise<boolean> {
+    return clients.getDefaultClient().waitForIdle(timeout);
```
sean-mcmanus: It appears our status indicator is per-workspace folder, so it seems like this won't work for workpace folders > than the 1st.

sean-mcmanus: This seems strange though...might be a pre-existing problem. It seems like the status indicator should be shared among all workspace folders.

sean-mcmanus: Maybe you can wait for all for all the clients? Instead of just the default.

sean-mcmanus: Maybe something like this generated by Copilot:

```
    const waitForIdlePromises: Promise<boolean>[] = [];
    clients.forEach(client => {
        waitForIdlePromises.push(client.waitForIdle(timeout));
    });

    const waitForIdleResults: boolean[] = await Promise.all(waitForIdlePromises);
    return waitForIdleResults.every(result => result);
```

sean-mcmanus: But having the status be per-workspace folder doesn't really make sense -- but no one has complained yet as far as I know. i.e. there could be bugs/issues with that, not sure, when switching to file in a different workspace, I see the status be "incorrect" (i.e. idle instead of processing).

sergior2: I originally had this implemented as the following:

```
    const pending: Promise<boolean>[] = [];
    clients.forEach(client => pending.push(client.waitForIdle(timeout)));
    const results: boolean[] = await Promise.all(pending); 1
    return results.every(result => result);
```

But was told that only reports status notifications to the 1st client, and advised instead to use getDefaultClient. Would this not bind the notifications across?

sergior2: Sorry for the late reply, comments were not updating on my end for some reason

Extension/src/LanguageServer/client.ts
bobbrow: He's saying add a public method on ClientModel instead of DefaultClient.  Though there is already `DefaultClient.IsTagParsing` which I think is what you care about.  `isUpdatingIntelliSense` wasn't what I thought you wanted to block on.  So it looks like the property you want already exists.

bobbrow: You don't actually need to check for "idle" right? I thought you just wanted to know when tag parsing is done.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: What? Yeah, it seems like you should waitForIdle across all clients. I'm seeing them per-client, although that is probably a pre-existing issue as mentioned previously.

Extension/src/LanguageServer/client.ts
sean-mcmanus: Yeah, I would think the IsTagParsing should be sufficient.

Extension/src/LanguageServer/client.ts
```diff
@@ -1008,6 +1010,56 @@ export class DefaultClient implements Client {
...
+    }
+
+    public async waitForIdle(timeout?: number): Promise<boolean> {
+        if (this.isIdle()) {
+            return true;
+        }
+
+        if (timeout !== undefined && timeout <= 0) {
+            return false;
+        }
```
carsonRadtke: Consider moving these checks "up" to the command handler and changing the signature to `waitForIdle(timeout: number)`. 

Also, if devtools (or whoever calls this) supplies an invalid timeout, maybe we raise an exception instead of returning false?

sergior2: Fixed

Extension/src/LanguageServer/extension.ts
```diff
@@ -430,6 +430,7 @@ export async function registerCommands(enabled: boolean): Promise<void> {
     commandDisposables.push(vscode.commands.registerCommand('cpptools.activeConfigName', enabled ? onGetActiveConfigName : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('cpptools.activeConfigCustomVariable', enabled ? onGetActiveConfigCustomVariable : onDisabledCommand));
     commandDisposables.push(vscode.commands.registerCommand('cpptools.setActiveConfigName', enabled ? onSetActiveConfigName : onDisabledCommand));
+    commandDisposables.push(vscode.commands.registerCommand('C_Cpp.waitForTagParsing', enabled ? onWaitForTagParsing : onDisabledCommand));
```
Colengms: It looks like we're using `cpptools.` as a prefix for 'hidden' commands.  Not sure if that's a consistent rule or not.  @sean-mcmanus  ?

bobbrow: Yes, please use `cpptools.` for the command prefix.

bobbrow: You don't want to use `onDisabledCommand` here because that can pop UI.  Probably just use something like `() => true` ("true" because you shouldn't wait for tag parsing when the language server is disabled)

sergior2: Updated with correct prefix

sergior2: Updated with replacement

Extension/src/LanguageServer/client.ts
```diff
@@ -4360,6 +4430,7 @@ class NullClient implements Client {
     setCurrentConfigName(configurationName: string): Thenable<void> { return Promise.resolve(); }
     getCurrentConfigName(): Thenable<string> { return Promise.resolve(""); }
     getCurrentConfigCustomVariable(variableName: string): Thenable<string> { return Promise.resolve(""); }
+    waitForTagParsing(timeout: number, token: vscode.CancellationToken): Promise<boolean> { return token.isCancellationRequested ? Promise.reject(new vscode.CancellationError()) : Promise.resolve(true); }
```
bobbrow: null client doesn't do tag parsing. You can just always resolve to true.

Extension/src/LanguageServer/client.ts
```diff
@@ -1008,6 +1010,64 @@ export class DefaultClient implements Client {
...
+    public async waitForTagParsing(timeout: number, token: vscode.CancellationToken): Promise<boolean> {
+        // On initialization, the client has UI bools all set to false which could cause an early return. We want to ensure it's ready first.
+        await this.ready;
+
+        if (!this.IsTagParsing) {
+            return true;
+        }
+
+        if (token.isCancellationRequested) {
+            throw new vscode.CancellationError();
```
bobbrow, line 1045: Does throwing/rejecting here give you the response you want in DevTools, or would you rather just return false for cancellation events?  IOW, do you do anything different in DevTools if this is cancelled vs. if it times out?

Extension/src/LanguageServer/client.ts
```diff
@@ -919,6 +920,7 @@ export class DefaultClient implements Client {
 
     private configStateReceived: ConfigStateReceived = { compilers: false, compileCommands: false, configProviders: undefined, timeout: false };
     private showConfigureIntelliSenseButton: boolean = false;
+    private pendingTagParsingCalls: { promise: ManualPromise<boolean>; timer?: NodeJS.Timeout; cancellationListener?: vscode.Disposable }[] = [];
```
bobbrow: I would recommend defining a named interface for this instead of inlining `{ promise: ManualPromise<boolean>; timer?: NodeJS.Timeout; cancellationListener?: vscode.Disposable }` in several places.

Extension/src/LanguageServer/client.ts
sergior2: Cancelling is the only exception we actively try to catch on the DevTools side, where we can throw our own "User Canceled" exception. Any other caught exceptions are expected to always return true to continue letting the tool run in the case the command didn't exist. Timeouts are the one exception where we want it to return false so that we can notify the agent that parsing is still underway.

But considering we're version checking + command checking, it might be good to start catching those exceptions and throwing for good reason now.

Extension/src/LanguageServer/client.ts
sergior2: Fixed in next commit

Extension/src/LanguageServer/client.ts
bobbrow: ok.

Extension/src/LanguageServer/client.ts
sergior2: Agreed, added in next commit.

sergior2:
@microsoft-github-policy-service agree company="Microsoft"

Extension/src/LanguageServer/client.ts
```diff
@@ -1039,16 +1045,14 @@ export class DefaultClient implements Client {
             throw new vscode.CancellationError();
         }
 
-        const pendingCall: { promise: ManualPromise<boolean>; timer?: NodeJS.Timeout; cancellationListener?: vscode.Disposable } = {
-            promise: new ManualPromise<boolean>()
-        };
+        const pendingCall = { promise: new ManualPromise<boolean>() } as PendingTagParsingCall;
```
bobbrow: This breaks your interface definition and now you have no protection against using undefined fields. Please don't do this. Add a `?` to the fields in your interface which actually matches what you had before.

sergior2: Ah sorry, I removed those because we are requiring timeouts and cancellation tokens on the devtools side and wanted to follow that same structure here

bobbrow: If they are required, then you need to set them here at initialization and not do a type assertion here. It looked to me like they were still optional.

bobbrow: Oh, I see what's going on. You have the data and you were setting the fields below. Just set it at initialization and then you don't need the `?`.

sergior2: Setting now done within initialization of pendingCall. New commit will have these changes.

Extension/src/LanguageServer/extension.ts
```diff
@@ -977,6 +978,10 @@ function onGetActiveConfigCustomVariable(variableName: string): Thenable<string>
     return clients.ActiveClient.getCurrentConfigCustomVariable(variableName);
 }
 
+async function onWaitForTagParsing(timeout: number, token: vscode.CancellationToken): Promise<boolean> {
```
sean-mcmanus, line 981: Should it say somewhere that the timeout is in milliseconds?

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Ah, I see -- it looks like checking the 1st is sufficient. When the non-default workspaces are used, the status quickly changes to idle, so it seems like the main thing to wait for is the default.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: Well, there's existing code that doesn't, so I guess it's fine.

sean-mcmanus (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -2893,6 +2957,8 @@ export class DefaultClient implements Client {
         } else if (message.includes("/")) {
             this.lastInvokedLspMessage = message;
         }
+
+        this.resolvePendingTagParsingCallsIfReady();
```
DavidARaygoza, line 2961: I assume updateStatus is called multiple times? So, if resolvePendingTagParsingCallsIfReady does not resolve the pending calls on the first try due to parsing still going on, then it will get around to it on a later call to updateStatus?

DavidARaygoza (approved):

Extension/src/LanguageServer/client.ts
bobbrow: Yes. updateStatus is called multiple times.

bobbrow (approved):

bobbrow:
Wait, why did auto-merge not wait for the checks to complete?

PR #14419, opened 2026-04-30
Add Native Strings

kuchungmsft:
Add Native Strings

### DIFF
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -485,5 +485,163 @@
     "license_terms": {
         "text": "The C/C++ extension may be used only with Microsoft Visual Studio, Visual Studio for Mac, Visual Studio Code, Azure DevOps, Team Foundation Server, and successor Microsoft products and services to develop and test your applications.",
         "hint": "{Locked=\"C/C++\"} {Locked=\"Visual Studio\"} {Locked=\"Mac\"} {Locked=\"Visual Studio Code\"} {Locked=\"Azure DevOps\"} {Locked=\"Team Foundation Server\"}"
+    },
+    "help_title": {
+        "text": "Microsoft C++ Language Server",
+        "hint": "{Locked=\"Microsoft\"} {Locked=\"C++\"}"
+    },
+    "help_usage": {
+        "text": "Usage: {0} [options]",
+        "hint": "{0} is the executable name. {Locked=\"{0}\"}"
+    },
+    "help_options_header": "Options:",
+    "help_help": "Show this help message and exit.",
+    "help_version": "Show version information and exit.",
+    "help_accept_eula": {
+        "text": "Permanently accept the End User License Agreement (EULA).",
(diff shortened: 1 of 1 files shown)


sean-mcmanus (dismissed):
I made some comments about the strings in the other PR.

sean-mcmanus (approved):

Extension/src/nativeStrings.json
```diff
@@ -485,5 +485,154 @@
...
         "text": "The C/C++ extension may be used only with Microsoft Visual Studio, Visual Studio for Mac, Visual Studio Code, Azure DevOps, Team Foundation Server, and successor Microsoft products and services to develop and test your applications.",
         "hint": "{Locked=\"C/C++\"} {Locked=\"Visual Studio\"} {Locked=\"Mac\"} {Locked=\"Visual Studio Code\"} {Locked=\"Azure DevOps\"} {Locked=\"Team Foundation Server\"}"
+    },
+    "help_title": {
+        "text": "Microsoft C++ Language Server",
+        "hint": "{Locked=\"Microsoft\"} {Locked=\"C++\"}"
+    },
+    "help_usage": {
+        "text": "Usage: {0} [options]",
+        "hint": "{0} is the executable name. {Locked=\"[options]\"}"
```
sean-mcmanus: Shouldn't "options" not be locked, i.e. translated, since it's not a literal?

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
```diff
@@ -485,5 +485,154 @@
...
+        "hint": "{Locked=\".github\"}"
+    },
+    "help_disable_telemetry": "Disable sending telemetry data.",
+    "auth_device_flow_prompt": {
+        "text": "To authenticate with GitHub, please copy the code {0} and then visit {1}. Waiting for authorization...",
+        "hint": "{0} is the user code to enter. {1} is the verification URL to visit. {Locked=\"GitHub\"}"
+    },
+    "auth_timed_out": "Timed out waiting for authorization.",
+    "auth_success": {
+        "text": "Successfully authenticated with GitHub!",
```
sean-mcmanus: I don't think the `!` is supposed to be used.

Extension/src/nativeStrings.json
sean-mcmanus: It seems safer to lock all `{0}` etc. in the strings below.

kuchungmsft: Fixed, thanks.

Extension/src/nativeStrings.json
kuchungmsft: Removed, thanks.

Extension/src/nativeStrings.json
kuchungmsft: Locked all `{n}`, thanks.

sean-mcmanus (approved):

PR #14426, opened 2026-05-05
Database population unification

Colengms:
- Removed the recursive includes 'reduce' feature, recursive includes would no longer be expanded, so wouldn't need to be reduced.
 -  Added some status strings.
 - Changed didChangeCppProperties from a request to a notification.


bobbrow (approved):

bobbrow:
Are we removing `workspaceParsingPriority` as well? Or will we repurpose it or replace it with a setting to control the number of parsing processes?

sean-mcmanus (commented):
Was workspace parsing priority also supposed to be removed or deprecated?

Extension/c_cpp_properties.schema.json
```diff
@@ -213,17 +213,6 @@
                     "recursiveIncludes": {
                         "type": "object",
                         "properties": {
-                            "reduce": {
```
sean-mcmanus, line 216: Should this be deprecated instead or before it's removed? i.e. internally it would have no effect but users won't get a settings error?

bobbrow: I don't think we're going to have another release where this property will still exist, and I don't think it's possible to deprecate a deleted property. Simply deleting the property should show a squiggle in the settings.json. We could potentially log a warning to the Output window about the unused setting if it's still set though.

PR #14432, opened 2026-05-07
Copy extension binaries for CI tests (targeting older branch)

bobbrow:
Not for checkin. The other PR is trying to merge with TS changes that need new extension binaries and we haven't shipped those yet, so I'm targeting an older version (`vs` branch).


bobbrow:
CI doesn't run on this branch :(

PR #14443, opened 2026-05-11
bump xmldom version by updating plist dependency

bobbrow:
CVE-2026-41675 wants us to update `xmldom`. It is only used by `plist` and updating that module updates the other one.

Also adding a try/catch in the code because `plist.parse` can throw.


sean-mcmanus:
@bobbrow We're using 0.8.13 already, which has the fix.

PR #14449, opened 2026-05-13
cpptools-srv2 crash handler.

sean-mcmanus:
Linux Test results:

cpptools-srv2
5/13/2026, 3:05:33 PM
SIGABRT
…
abort+18
Main(std::vector<std::string>&)+1877
****

I had problems testing the crash on mac.

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1195,7 +1195,7 @@ function handleMacCrashFileRead(err: NodeJS.ErrnoException | undefined | null, d
     data = data.replace(/0x1........ \+ 0/g, "");
 
     // Get rid of the process names on each line and just add it to the start.
-    const processNames: string[] = ["cpptools-srv", "cpptools-wordexp", "cpptools",
+    const processNames: string[] = ["cpptools-srv2", "cpptools-srv", "cpptools-wordexp", "cpptools",
         // Since only crash logs that start with "cpptools" are reported, the cases below would only occur
         // if the crash were to happen before the new process had fully started and renamed itself.
         "clang-tidy", "clang-format", "clang", "gcc"];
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -1255,8 +1255,9 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, cr
 
     const lines: string[] = data.split("\n");
     let addressData: string;
+    const isCppToolsSrv2: boolean = crashFile.startsWith("cpptools-srv2-");
```
bobbrow: is there supposed to be a dash at the end of `cpptools-srv2-`?

bobbrow: I just saw the other PR and this aligns with that, but still looks weird. I guess it can go in as-is, but it also doesn't strictly need to end with a dash.

sean-mcmanus: I had to add the `-` because the cpptools-srv version 1 has a guid attached on the right which may contain a 2.

sean-mcmanus: But I guess I can just fix it by changing cpptools-srv to `cpptools-srv-<guid>` then I can just check the for the 2 here.

sean-mcmanus: I added the `-` before the guid in all cases so I removed the `-` here.

bobbrow (approved):

sean-mcmanus:
@bobbrow You approved the TypeScript but not the cpptools side change?

bobbrow:
> @bobbrow You approved the TypeScript but not the cpptools side change?

It's been a busy day. Sorry... I can take a look at the updates tomorrow.

PR #14457, opened 2026-05-16
Stop automatically cancelling the previous Find All References based request when a new one comes in.

### LINKED ISSUE #316783: vscode_listCodeUsages reports 0 usages to Copilot if a user invoked find all references is invoked
Version: 1.120.0
Commit: 0958016b2af9f09bb4257e0df4a95e2f90590f9f
Date: 2026-05-12T20:17:22Z
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0

Steps to Reproduce:

1. Install the Microsoft C/C++ extension which provides a find all references implementation that is slow enough to easily repro this.
2. Tell Copilot to invoke Find All References with some C/C++ project that is large enough on some symbol that has enough references to allow step 3 to be done without step 2 finishing (confirming C/C++ references can be slow, i.e. many seconds/minutes). Some GitHub repo like LLVM or bullet3 repos might work (I used the closed source VS repo)....or tell Copilot to generate 10 .cpp files that include all the standard C++ headers and some test class that is referenced in all 10 files should probably work (but I haven't tried).
3. Manually invoke find all references on any symbol before step 2 is done.

Bug: The Copilot invoked find all references is canceled and reports 0 references. Copilot doesn't appear to be aware that it was canceled, which could lead to incorrect processing by Copilot...although one would hope it would understand that 0 references is by definition a bug for Find All References, since it should find the source reference.

Image

Bug2: The opposite problem also occurs -- Copilot invoking a textdocument/references will cancel a pre-existing user invoked Find All References, return 0 results to the user in the UI.

Our C/C++ extension implementation of find all references will cancel the previous find all references request when a 2nd is invoked under the assumption that the human user doesn't want to wait for the 1st find all references result because it will get thrown away by the 2nd one (also, it's possible switching to not cancelling the previous request could cause like 10 find all references request to be queued (e.g. by Copilot) and block a user invoked find all references for 10 minutes for example).

If VS Code could add some "sender" parameter along with the find all references request then our C/C++ extension could check that and respond appropriately depending of it's for Copilot or a human user using the references UI.

Bug3: There is another bug in that a user can use the cancel button in the bottom right progress notification that appears after a couple seconds and Copilot will incorrectly think there were 0 references. I could file a new bug on this if you want. Ideally, there would be for some way for us to not show that progress UI with the cancel when invoked by Copilot, since users in that case already have the Cancel in the Copilot Chat UI.

The Microsoft C/C++ DevTools extension which implements the same feature with GetSymbolReferences_CppTools has these bugs too...I'm not sure what the status is on a fix from them or if they require changes to our C/C++ extension (or why we/Microsoft are implementing two tools for Copilot that do the same thing).

sean-mcmanus:
To fix https://github.com/microsoft/vscode/issues/316783 (where Copilot and user requested find all references based operations would cancel one other due to the code cancelling the previous request), assuming VS Code can't provide a better fix.

The problem with this change is that the previous requests are "lost" and never shown when manually invoked, so users have to know they should manually cancel the previous one now or it's a waste of time waiting for it to finish.

### DIFF
--- a/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
+++ b/Extension/src/LanguageServer/Providers/callHierarchyProvider.ts
@@ -223,7 +223,6 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
     public async prepareCallHierarchy(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.CallHierarchyItem | undefined> {
         await this.client.ready;
 
-        workspaceReferences.cancelCurrentReferenceRequest(CancellationSender.NewRequest);
         workspaceReferences.clearViews();
 
         const range: vscode.Range | undefined = document.getWordRangeAtPosition(position);
@@ -262,7 +261,6 @@ export class CallHierarchyProvider implements vscode.CallHierarchyProvider {
 
     public async provideCallHierarchyIncomingCalls(item: vscode.CallHierarchyItem, token: vscode.CancellationToken): Promise<vscode.CallHierarchyIncomingCall[] | undefined> {
         await this.client.ready;
-        workspaceReferences.cancelCurrentReferenceRequest(CancellationSender.NewRequest);
 
         const CallHierarchyCallsToEvent: string = "CallHierarchyCallsTo";
         if (item === undefined) {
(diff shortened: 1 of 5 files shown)


sean-mcmanus:
This doesn't work and break Find All Reference such that it enters a state where all requests get cancelled -- the current cancel implementation assumes there is only 1 in-flight FAR request that can be singular canceled.

sean-mcmanus:
I still need to test/review it more (it seemed to fix the case that was broken previously).

sean-mcmanus:
I'm not sure if we want this or not, so I'll just close it for now.

PR #14467, opened 2026-05-20
Remove compiler argument translation JSON files

kuchungmsft:


bobbrow:
I don't think we are ready for this yet. We still depend on these.

sean-mcmanus:
> I don't think we are ready for this yet. We still depend on these.

@bobbrow Glen's cpptools PR removes the dependency on them.

sean-mcmanus (approved):

sean-mcmanus:
But don't check this in until the C++ side gets checked in.

bobbrow:
@sean-mcmanus ok. I haven't had the chance to look at @kuchungmsft's PR yet and didn't realize that this would be done as part of that feature.

PR #14473, opened 2026-05-22
Add "important" doxygen.sectionTags.

sean-mcmanus:
Copilot was mentioning it thought these were missing while investigating another issue on doxygen.sectionTags.

See https://www.doxygen.nl/manual/commands.html


sean-mcmanus:
There's a cpptools side string change too.

Not sure yet if there's a reason these were omitted too...

bobbrow (approved):

PR #14475, opened 2026-05-22
Reset status UI if DB fails to open

Colengms:
I recently added a `C_Cpp: Parser Initialize Failed` on DB open failure.  This PR checks for that and resets status UI.


heartacker:
go

sean-mcmanus (approved):

PR #14500, opened 2026-06-01
Improve diagnostics.source.

sean-mcmanus:
This is primarily for if/when https://github.com/microsoft/vscode/issues/319418 gets fixed so Copilot can more easily know it's a not an actual C/C++ compiler error.

...but I figured it might help new users as well.


Colengms (approved):

sean-mcmanus:
I want to make sure @bobbrow is okay with this before checking it in.

sean-mcmanus:
Abandoned in favor or some alternative fix which allows us to tell Copilot specifically that the error is not a compiler error and is from our IntelliSense parser (which could be invalid/etc.).

PR #14516, opened 2026-06-15
add a fallback message for unsupported RWD configs

### LINKED ISSUE #14515: "Run Without Debugging is not supported" message, even though it works
### Environment

- OS and version: Linux 7.0.11-arch1-1
- VS Code: 1.124.0
- C/C++ extension: 1.32.2
- GDB / LLDB version: gdb 17.2

An annoying message:
```
Run Without Debugging is not supported for configurations with 'miDebuggerServerAddress' set.
```
appears in the `output` tab of vscode, focusing it, when launching a remote debug session with the `Debug: Start without debugging` vscode command. This message appears despite the 'run without debugging' feature working as expected (it does not stop at breakpoints).

To me, the issue is the combined, annoying fact that:
- run without debugging works fine
- the "output" tab gets focused, hiding the terminal output

If run without debugging actually didn't work, it would make sense to show this tab. But it actually works how intended (or at least for what I need it for)

The issue was introduced with this commit: 291e5a322822131e8ad1b99c1d67112513ae6bf7

Steps to reproduce:
1. `mkdir -p test/.vscode`
2. `echo 'int main(){return 0;}' >test/main.c && gcc -g -O0 test/main.c -o test/main`
3. `code test`
3. Write to `test/.vscode/launch.json`:
```json
{
    "version": "0.2.0",
    "configurations": [
		{
		"type": "cppdbg",
		"request": "launch",
		"name": "reproduce",
		"cwd": "${workspaceFolder}",
		"program": "${workspaceFolder}/main",
		"MIMode": "gdb",
		"miDebuggerServerAddress": "127.0.0.1:12345"
		}
    ]
}
```
4. Start a fake GDB (TCP) server: `nc -l 12345`
5. `code test`
6. Open command palette, run `Debug: Start without Debugging`, selecting the `reproduce` target

You can see the warning in the `OUTPUT` panel, which has been focused.
The connection will timeout, because there is no actual gdb server. But if there is, debugging works absolutely fine. I'm sorry i can't provide a more accurate reproduction, without effectively providing my entire project

### Debugger Configurations

```shell
{
    "version": "0.2.0",
    "configurations": [
		{
		"type": "cppdbg",
		"request": "launch",
		"name": "my_wonderful_reproducable_bug",
		"cwd": "${workspaceFolder}",
		"program": "${workspaceFolder}/main",
		"MIMode": "gdb",
		"miDebuggerServerAddress": "127.0.0.1:12345"
		}
    ]
}
```

### Debugger Logs

```shell
Run Without Debugging is not supported for configurations with 'miDebuggerServerAddress' set.
```

### Other Extensions

_No response_

### Additional Information

_No response_

bobbrow:
Fixes: #14515


Extension/package.nls.json
```diff
@@ -1026,6 +1026,7 @@
     "c_cpp.debuggers.VSSymbolOptionsModuleFilter.excludedModules.description": "Array of modules that the debugger should NOT load symbols for. Wildcards (example: MyCompany.*.dll) are supported.\n\nThis property is ignored unless 'mode' is set to 'loadAllButExcluded'.",
     "c_cpp.debuggers.VSSymbolOptionsModuleFilter.includedModules.description": "Array of modules that the debugger should load symbols for. Wildcards (example: MyCompany.*.dll) are supported.\n\nThis property is ignored unless 'mode' is set to 'loadOnlyIncluded'.",
     "c_cpp.debuggers.VSSymbolOptionsModuleFilter.includeSymbolsNextToModules.description": "If true, for any module NOT in the 'includedModules' array, the debugger will still check next to the module itself and the launching executable, but it will not check paths on the symbol search list. This option defaults to 'true'.\n\nThis property is ignored unless 'mode' is set to 'loadOnlyIncluded'.",
+    "c_cpp.debuggers.ignoreRunWithoutDebuggingWarnings.description": "If true, no warning will be logged when run without debugging fails to launch the program in the terminal.",
```
sean-mcmanus: Can you lock the string? I've been seeing new strings "randomly" having true be translated. We could do that for pre-existing strings, but it's only an issue if we modify a string and it has to be translated again.

   "c_cpp.debuggers.ignoreRunWithoutDebuggingWarnings.description": {
        "message": "If true, no warning will be logged when run without debugging fails to launch the program in the terminal.",
        "comment": [
            "{Locked=\"false\"}"
        ]
    },

e.g. https://github.com/microsoft/vscode-cpptools/pull/14520/changes

sean-mcmanus (changes requested):

Extension/package.nls.json
bobbrow: ok

sean-mcmanus (approved):

PR #14537, opened 2026-06-22
Remove Take Survey command

Colengms:
Removing the exposed, manually invokable 'Take Survey' command.

Alternatively, we could keep it and update the URL to use an aka.ms link.  (The survey is being moved from SurveyMoney to customervoice.microsoft.com)


bobbrow:
Can you also update `README.md` with the new link?  https://aka.ms/vcvscodesurvey

sean-mcmanus (commented):
But how does moving the survey relate to removing the command for it? It seems separate?

bobbrow:
> But how does moving the survey relate to removing the command for it? It seems separate?

We were talking about both things together offline. It didn't seem like we needed a separate PR for it, but I don't have a strong opinion about that.

Issue #14549, opened 2026-06-26
Extension causes high cpu load


moonl1ghtstar:
- Issue Type: `Performance`
- Extension Name: `cpptools`
- Extension Version: `1.32.2`
- OS Version: `Windows_NT x64 10.0.19045`
- VS Code version: `1.99.3`

:warning: Make sure to **attach** this file from your *home*-directory:
:warning:`file:///C:/Users/%USERNAME%/AppData/Local/Temp/ms-vscode.cpptools-unresponsive.cpuprofile.txt`

Find more details here: https://github.com/microsoft/vscode/wiki/Explain-extension-causes-high-cpu-load

[ms-vscode.cpptools-unresponsive.cpuprofile.txt](https://github.com/user-attachments/files/29375566/ms-vscode.cpptools-unresponsive.cpuprofile.txt)

sean-mcmanus:
@moonl1ghtstar Do you have any more info on this, such as a description or repro steps?

I'm not sure how to open the .cpuprofile file currently due to https://github.com/microsoft/vscode/issues/323137

sean-mcmanus:
@moonl1ghtstar We don't need more info. It appears to be caused by a ~1 sec load on startup from our vscode-nls dependency, see https://github.com/microsoft/vscode-cpptools/issues/13145 .

PR #14550, opened 2026-06-26
Migrate localization from vscode-nls to vscode.l10n

### LINKED ISSUE #13145: Migrate from `vscode-nls`/`vscode-nls-dev` to `vscode-l10n`
The underlying libraries that facilitate string localization in VS Code extensions (`vscode-nls`/`vscode-nls-dev`) have been replaced with a new library: `vscode-l10n`.

https://github.com/microsoft/vscode-nls
https://github.com/microsoft/vscode-nls-dev
https://github.com/microsoft/vscode-l10n

Opening this issue to track migrating to the new library.

sean-mcmanus:
Fix by Copilot with Claude Opus 4.8 in VS Code (with review fixes from GPT-5.5).

Fixes https://github.com/microsoft/vscode-cpptools/issues/13145 and https://github.com/Microsoft/vscode-cpptools/issues/14549 .

### Why
Profiling an unresponsive extension-host session showed a single synchronous call — `loadNlsBundle` (vscode-nls reading the standalone NLS bundle from disk during module `require`) — blocking the extension host for ~897 ms at activation, about 90% of the captured CPU profile. VS Code's built-in `vscode.l10n` API loads the localization bundle itself (declared via the `l10n` field in package.json), so strings are no longer read synchronously on the activation path.

### What changed
- Replaced all 384 `localize(...)` call sites across 38 source files with `l10n.t(...)` (via `const l10n = vscode.l10n;`, or a named `l10n` import in files that already import named symbols from `vscode`), dropping the explicit string keys (the English message is now the key). The `{ key, comment }` form becomes `{ message, args, comment }`.
- Removed the per-file vscode-nls header (`nls.config(...)` + `loadMessageBundle()`); ensured `import * as vscode from 'vscode'`.
- Updated the native-strings generator (`generateNativeStrings.ts`) to emit `l10n.t`, and regenerated `nativeStrings.ts` / `localized_string_ids.h`.
- package.json: added `"l10n": "./l10n"`, swapped `vscode-nls-dev` → `@vscode/l10n-dev`, removed the runtime `vscode-nls` dependency.
- webpack.config.js: removed the `vscode-nls-dev` webpack loader (VS Code loads bundles at runtime now).
- gulpfile.js: rewrote `translations-export` / `-import` / `-generate` onto `@vscode/l10n-dev` (English bundle → `l10n/bundle.l10n.json`, per-language → `l10n/bundle.l10n.<lang>.json`), preserving the custom HTML / JSON-schema / walkthrough / package.nls flows.
- Migrated existing translations: the 13 language bundles were rebuilt from the legacy nls-keyed `i18n/*/src/**.i18n.json` into `i18n/<lang>/bundle.l10n.json` (743 strings each), keyed by the English message — or by `message/comment` for the 81 strings that carry a translator comment, matching how `@vscode/l10n-dev` keys the bundle so those strings resolve in every locale.
- Updated CONTRIBUTING.md and the prep check; `.gitignore` now ignores the generated `l10n/bundle.l10n*.json`.

### Verification
- `tsc --noEmit` clean; `eslint src .scripts` clean; `node --check` on gulpfile + webpack.
- Three call sites needed `String(...)` coercion because `l10n.t` types args as `string | number | boolean` (stricter than nls's `any`); this preserves nls's formatting behavior.

### Localization call sites must be a literal `l10n.t` / `vscode.l10n.t`
`@vscode/l10n-dev` extracts strings at build time by statically scanning source for `l10n.t(...)` / `vscode.l10n.t(...)` calls — it does not execute code or follow aliases. An alias like `const localize = vscode.l10n.t` compiles and works at runtime, but the extractor never sees `localize(...)`, so those strings are silently dropped from `bundle.l10n.json` and render in English in every locale. That is why each file calls `l10n.t` directly via `const l10n = vscode.l10n;` rather than a custom-named alias.

### Notes
- The legacy `i18n/*/src/**.i18n.json` files are superseded for source strings and are safe to delete once a loc cycle validates the new bundles.

### Migration approach
- Source call sites were converted with an AST-based codemod (TypeScript compiler API) that
  located every `localize(...)` call and spliced in the equivalent l10n call, preserving
  surrounding formatting, argument expressions, and message/escape sequences:
  - `localize('key', 'Message {0}', a)` → `l10n.t('Message {0}', a)`
  - `localize({ key, comment: [...] }, 'Message {0}', a)` →
    `l10n.t({ message: 'Message {0}', args: [a], comment: [...] })`
- Placeholders remain `{0}`-style, so message text is unchanged across the migration.

### Call-site form (l10n.t)
Every call is a literal `l10n.t(...)`:
- Files that already import named symbols from `vscode` merge `l10n` into that named import
  (e.g. `targetNodes.ts`, `cppBuildTaskProvider.ts`).
- All other files keep `import * as vscode from 'vscode'` and add `const l10n = vscode.l10n;`.
`@vscode/l10n-dev` extracts strings by a static AST scan of `l10n.t(...)` / `vscode.l10n.t(...)`
calls; it does not follow aliases, so a custom-named alias (`const localize = vscode.l10n.t`)
would compile and run but silently drop those strings from the bundle — hence the literal `l10n.t`.

### Translation recovery
The existing translations were keyed by the old vscode-nls string IDs and would otherwise be
orphaned (the new model keys by the English string). A one-time remap joined the pre-migration
`key → English` map (from git HEAD, plus `nativeStrings.json` for the generated file) with the
legacy `key → translation` files, producing English-keyed `i18n/<lang>/bundle.l10n.json`:
- 13 languages: chs, cht, csy, deu, esn, fra, ita, jpn, kor, plk, ptb, rus, trk
- 743 unique entries per language (763 mapped, ~20 identical-English strings deduped)
- 4 stale keys (`debugger.noDebug.*.not.supported`) were dropped — verified absent from source,
  so never looked up at runtime.
- The per-language bundles were then re-keyed to match `@vscode/l10n-dev`'s scheme: plain strings
  stay keyed by the English message, while the 81 strings that carry a translator comment are
  keyed `message/comment`. The initial remap had keyed everything by message only, which would
  have made commented strings fall back to English in every locale; all 13 bundles were rebuilt so
  every key matches the generated English bundle.

At build time `generateSrcLocBundle` copies `i18n/<folder>/bundle.l10n.json` to
`l10n/bundle.l10n.<lang>.json` (e.g. `chs` → `zh-cn`).

### Type-coercion fixes
`vscode.l10n.t` types args as `string | number | boolean`, stricter than vscode-nls's `any`.
Two args were `undefined`-typed under the new signature and one file had a pre-existing duplicate
`vscode` import surfaced by the change:
- `common.ts`: `String(exitCode)`
- `LanguageServer/configurations.ts`: `String(dotConfigPath)`
- `Debugger/debugAdapterDescriptorFactory.ts`: removed a duplicate `import * as vscode`
`String(...)` reproduces vscode-nls's internal formatting exactly.

### Other fixes surfaced in review
- Corrected a user-facing typo `ProduceVersion` → `ProductVersion` (the literal
  `SystemVersion.plist` field name) in `platform.ts` and in both the key and value of all 13
  recovered bundles.
- Closed unmatched single quotes around interpolated paths in the `.scripts/common.ts` prep
  warnings.

### Follow-ups / validation
- Run `yarn prep` (invokes `translations-generate`) and confirm `l10n/bundle.l10n.*.json` are
  produced; verify a non-English VS Code display language shows the recovered strings.
- The XLF round-trip (`translations-export` / `-import`) now targets `@vscode/l10n-dev`; validate
  it against the MLCP loc pipeline on the next cycle.
- After the new bundles are validated, the legacy `i18n/*/src/**.i18n.json` files can be deleted.


Extension/src/Debugger/attachQuickPick.ts
```diff
@@ -22,7 +18,7 @@ class RefreshButton implements vscode.QuickInputButton {
     }
 
     get tooltip(): string {
-        return localize("refresh.process.list.tooltip", "Refresh process list");
+        return vscode.l10n.t("Refresh process list");
```
bobbrow: I'm not a fan of this new API for localization. Does it work to do something like `const localize = vscode.l10n.t;` at the top of every file so we can keep using an intuitively named function ("localize")?  Or if there's another way to make an alias...

sean-mcmanus: @bobbrow FYI, this is not really ready for human review.

Just a sec...

bobbrow: Yes, I saw it was draft, but I wanted to influence the direction, so I made the comment now before you locked in.

sean-mcmanus: @bobbrow No, you can't use a "localize" alias. The shortest it can get is

```TypeScript
const l10n = vscode.l10n;

quickPick.title = l10n.t("Attach to process");
```

But Copilot recommends using vscode.l10n.t and not reverting to vscode-nls — VS Code's documented, recommended localization path is vscode.l10n. Reverting reintroduces the perf bug and moves onto a deprecated dependency.

bobbrow: My comment wasn't about the length of the API. It's about the naming. `l10n.t` is not intuitively named, while `localize` is. I'm not asking you to revert. I just want to see what our options are here.

bobbrow: VS Code doesn't squiggle this:
image

sean-mcmanus: @bobbrow It fails silently. The localization doesn't work, no squiggle. However, it now recommends using l10n.t instead of vscode.l10n.t.

Extension/src/SSH/TargetsView/targetNodes.ts
```diff
@@ -3,15 +3,12 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 
+import * as vscode from 'vscode';
 import { constants } from 'fs';
 import { TreeItem } from "vscode";
```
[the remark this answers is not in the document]
sean-mcmanus: I'm fixing it, but there's no actual linter error.

Extension/src/LanguageServer/cppBuildTaskProvider.ts
```diff
@@ -3,11 +3,11 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 
+import * as vscode from 'vscode';
 import * as cp from "child_process";
 import * as os from 'os';
 import * as path from 'path';
 import { CustomExecution, Disposable, Event, EventEmitter, ProcessExecution, Pseudoterminal, ShellExecution, Task, TaskDefinition, TaskEndEvent, TaskExecution, TaskGroup, TaskProvider, tasks, TaskScope, TerminalDimensions, TextEditor, window, workspace, WorkspaceFolder } from 'vscode';
```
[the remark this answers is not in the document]
sean-mcmanus: I'm fixing it, but there's no actual linter error.

sean-mcmanus:
Local Copilot found another bug it's working on fixing...

Extension/src/LanguageServer/localization.ts
```diff
@@ -68,8 +65,5 @@ export function getLocalizedHtmlPath(originalPath: string): string {
 }
 
 export function getLocalizedSymbolScope(scope: string, detail: string): string {
-    return localize({
-        key: "c.cpp.symbolscope.separator", comment:
-            ["{0} is an untranslated C++ keyword (e.g. \"private\") and {1} is either another keyword (e.g. \"typedef\") or a localized property (e.g. a localized version of \"declaration\""]
-    }, "{0}, {1}", scope, detail);
+    return vscode.l10n.t({ message: "{0}, {1}", args: [scope, detail], comment: ["{0} is an untranslated C++ keyword (e.g. \"private\") and {1} is either another keyword (e.g. \"typedef\") or a localized property (e.g. a localized version of \"declaration\""] });
```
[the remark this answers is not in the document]
sean-mcmanus: A fix is pending...

Extension/src/platform.ts
```diff
@@ -75,10 +72,10 @@ export class PlatformInformation {
             if (systemVersionData) {
                 productDarwinVersion = systemVersionData.ProductVersion;
             } else {
-                errorMessage = localize("missing.plist.productversion", "Could not get ProduceVersion from SystemVersion.plist");
+                errorMessage = vscode.l10n.t("Could not get ProduceVersion from SystemVersion.plist");
```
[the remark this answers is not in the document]
sean-mcmanus: A fix is pending...

Extension/.scripts/common.ts
```diff
@@ -305,7 +305,7 @@ export async function checkPrep() {
     failing = !await assertAnyFolder('dist/walkthrough') && (quiet || warn(`The walkthrough files are not in place.`)) || failing;
     failing = !await assertAnyFolder('dist/html') && (quiet || warn(`The html files are not in place.`)) || failing;
     failing = !await assertAnyFolder('dist/schema') && (quiet || warn(`The schema files are not in place.`)) || failing;
-    failing = !await assertAnyFile('dist/nls.metadata.json') && (quiet || warn(`The extension translation file '${$root}/dist/nls.metadata.json is missing.`)) || failing;
+    failing = !await assertAnyFile('l10n/bundle.l10n.json') && (quiet || warn(`The extension localization bundle '${$root}/l10n/bundle.l10n.json is missing.`)) || failing;
```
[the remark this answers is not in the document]
sean-mcmanus: A fix is pending...

sean-mcmanus:
I didn't want to check this in until after we get our pending 1.33.x updated loc strings, in case there are any issues...maybe 1.33.4.

PR #14576, opened 2026-07-10
upgrade language client to version 10

bobbrow:
I'm seeing some strange behavior in the crashing telemetry. It's hard to tell why the client is disconnecting from cpptools. In some cases it looks like cpptools is orphaned, not crashed, so I'm starting with upgrading the client module.

A follow up PR will add additional telemetry when an error happens in case that can provide us additional information.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -7252,7 +7252,7 @@
         "ssh-config": "^4.4.4",
         "tmp": "^0.2.7",
         "vscode-cpptools": "^7.1.1",
-        "vscode-languageclient": "^9.0.1",
+        "vscode-languageclient": "^10.1.0",
         "vscode-nls": "^5.2.0",
         "vscode-tas-client": "^0.1.84",
         "which": "^2.0.2"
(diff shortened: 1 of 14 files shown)


bobbrow (commented):
There are a lot of files changed because the module upgrade resulted in some types being moved and I asked copilot to fix the compile errors.

sean-mcmanus (changes requested):
Are you sure this is possible without breaking compatibility with older VS Code version? I had updated to 9 recently and I seemed to recall that was the latest version possible at the time, but maybe 10 got released later?

UPDATE: Copilot says: 
vscode-languageclient version | Minimum VS Code (engines.vscode)
-- | --
10.1.0 (latest) | ^1.91.0

Our current min supported version is 1.77.0, and requiring 1.91 drops Ubuntu 16-18 and macOS 10.13/14/15.

We just recently upgraded to 9 and dropped support for Windows 7/8 etc. so it seems a little early to drop more OS support.

bobbrow:
> Are you sure this is possible without breaking compatibility with older VS Code version? I had updated to 9 recently and I seemed to recall that was the latest version possible at the time, but maybe 10 got released later?
> 
> UPDATE: Copilot says:
> 
> vscode-languageclient version	Minimum VS Code (engines.vscode)
> 10.1.0 (latest)	^1.91.0
> Our current min supported version is 1.77.0, and requiring 1.91 drops Ubuntu 16-18 and macOS 10.13/14/15.

Ok. I didn't look for a VS Code dependency change. I'll abandon this for now.

PR #14578, opened 2026-07-11
Add localizable strings for --check headless validation

kuchungmsft:
Add localizable strings for --check headless validation

### DIFF
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -713,7 +713,7 @@
     },
     "check_requires_source": {
         "text": "--check requires a source file: --check=<file>",
-        "hint": "{Locked=\"--check\"} {Locked=\"--check=<file>\"}"
+        "hint": "{Locked=\"--check\"}"
     },
     "check_source_not_found": {
         "text": "source file not found: {0}",
@@ -725,7 +725,7 @@
     },
     "check_compile_commands_not_discovered": {
         "text": "could not find compile_commands.json in any parent directory of {0}; pass --check-compile-commands=<path> to specify it explicitly",
-        "hint": "{0} is the starting directory path. {Locked=\"compile_commands.json\"} {Locked=\"{0}\"} {Locked=\"--check-compile-commands=<path>\"}"
+        "hint": "{0} is the starting directory path. {Locked=\"compile_commands.json\"} {Locked=\"{0}\"} {Locked=\"--check-compile-commands\"}"
     },
     "check_engine_init_failed": "failed to initialize the language engine",
     "check_no_workspace_folder": {
--- a/Extension/src/nativeStrings.json
+++ b/Extension/src/nativeStrings.json
@@ -698,5 +698,54 @@
     "status_indexing_files_with_text": "Indexing files {0}",
     "help_allow_missing_lsp_config": "Allow the server to start even if the specified --lsp-config file does not exist.",
     "initialize_failed_during_engine_setup": "Initialization failed during engine setup.",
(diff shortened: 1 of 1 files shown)


Extension/src/nativeStrings.json
```diff
@@ -698,5 +698,54 @@
...
+        "text": "Path to a specific compile_commands.json (or its directory) to use with --check. Defaults to auto-discovery.",
+        "hint": "{Locked=\"compile_commands.json\"} {Locked=\"--check\"}"
+    },
+    "check_not_authorized": {
+        "text": "not authorized; sign in is required to run --check",
+        "hint": "{Locked=\"--check\"}"
+    },
+    "check_requires_source": {
+        "text": "--check requires a source file: --check=<file>",
+        "hint": "{Locked=\"--check\"} {Locked=\"--check=<file>\"}"
```
sean-mcmanus: @kuchungmsft File should not be locked. It is a placeholder and not a literal, so it should be translated.

Extension/src/nativeStrings.json
```diff
@@ -698,5 +698,54 @@
...
+        "text": "source file not found: {0}",
+        "hint": "{0} is the source file path. {Locked=\"{0}\"}"
+    },
+    "check_compile_commands_not_found": {
+        "text": "compile_commands.json not found: {0}",
+        "hint": "{0} is the file path. {Locked=\"compile_commands.json\"} {Locked=\"{0}\"}"
+    },
+    "check_compile_commands_not_discovered": {
+        "text": "could not find compile_commands.json in any parent directory of {0}; pass --check-compile-commands=<path> to specify it explicitly",
+        "hint": "{0} is the starting directory path. {Locked=\"compile_commands.json\"} {Locked=\"{0}\"} {Locked=\"--check-compile-commands=<path>\"}"
```
sean-mcmanus: same here

sean-mcmanus: Feel free to use Copilot to check for any other cases were we might be incorrectly locking a placeholder string which should be translated.

sean-mcmanus (changes requested):

dhe-cruzer69 (approved):

Extension/src/nativeStrings.json
kuchungmsft: Fixed, thanks.

Extension/src/nativeStrings.json
kuchungmsft: Fixed, confirmed only two cases, thanks.

Extension/src/nativeStrings.json
```diff
@@ -698,5 +698,54 @@
...
+        "text": "Path to a specific compile_commands.json (or its directory) to use with --check. Defaults to auto-discovery.",
+        "hint": "{Locked=\"compile_commands.json\"} {Locked=\"--check\"}"
+    },
+    "check_not_authorized": {
+        "text": "not authorized; sign in is required to run --check",
+        "hint": "{Locked=\"--check\"}"
+    },
+    "check_requires_source": {
+        "text": "--check requires a source file: --check=<file>",
+        "hint": "{Locked=\"--check\"}"
```
sean-mcmanus: This should be `--check=`

sean-mcmanus (changes requested):

Extension/src/nativeStrings.json
kuchungmsft: Added, thanks.

sean-mcmanus (approved):

PR #14579, opened 2026-07-13
Normalize JSON trailing newlines.

sean-mcmanus:
Normalize JSON trailing newlines.

Fixed by Copilot (Claude Opus 4.8) in VS Code.

## Summary

Makes JSON trailing newlines consistent
across the extension and stops them from being silently stripped on save.

## JSON trailing-newline consistency

Two settings in `.vscode/settings.json` disagreed: `[jsonc]` inserted a final
newline on save, but `[json]` did not. Because `[json]` also formats on save,
saving any `.json` file reformatted it and then dropped the trailing newline,
producing spurious diffs whenever a JSON file was opened and saved. Several
files had already lost their final newline this way.

Changes:

- `.vscode/settings.json`: set `files.insertFinalNewline` to `true` for `[json]`
  (it was `false`), matching `[jsonc]`. Saving a JSON file no longer strips the
  trailing newline.
- Normalized all tracked `.json` files to end with a single trailing newline,
  including the generated `i18n/**/*.i18n.json` translation files.
- `gulpfile.js` (`translations-import`): append a trailing newline to each
  generated `.i18n.json` after it is written, so future translation imports keep
  the files consistent with the convention above. The existing generators for
  `bin/messages` and `package.json` already write a trailing newline, so no
  change was needed there.

## Notes

- No behavioral change to the extension beyond preventing the localized failure
  described above; the remaining changes are whitespace-only and tooling.


sean-mcmanus:
These shouldn't get checked in until the corresponding cpptools side PR 757614 gets approved.

UPDATE: The IntelliSense string changes were moved to https://github.com/microsoft/vscode-cpptools/pull/14607 .

Colengms (approved):

Colengms (approved):

PR #14607, opened 2026-07-26
Update IntelliSense loc strings for PR 757614.

sean-mcmanus:
The C++ IntelliSense language server can now emit an additional diagnostic
message. Its localized message tables under Extension/bin/messages/<locale>/messages.json
are positional arrays indexed by message ID, so every locale's array must be at
least as long as the highest message ID the server can produce. A locale whose
array was shorter than the new message's index would hit an out-of-range lookup
when that diagnostic was emitted, failing the language server for that locale.

This adds the new English string to every locale's table so all locales stay in
sync. The entry is intentionally left in English for now; localized translations
will arrive later through the normal translation-import pipeline.

Message table update
Appended the new string to all 13 locale files in Extension/bin/messages:

"too many nested user-defined conversions (possible cycle of converting constructors)"
It is added at the same positional index in every locale, matching the new
message ID, so lookups line up across all languages.


sean-mcmanus:
This PR is in draft until have time to get back to addressing the comments in PR 757614.

Colengms (approved):

Colengms (dismissed):

bobbrow (dismissed):

sean-mcmanus:
This should get checked in with 757614.

bobbrow (approved):

PR #14608, opened 2026-07-26
Fix si_code=, si_addr=undefined.

sean-mcmanus:
I noticed this in the crash call stack data.

Fixed by Copilot with Claude Opus 4.8 (in VS Code).

### DIFF
--- a/Extension/src/LanguageServer/extension.ts
+++ b/Extension/src/LanguageServer/extension.ts
@@ -1362,7 +1362,16 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, cr
     }
     if (lines[crashStackStartLine].startsWith("SIG")) {
         signalType = `${lines[crashStackStartLine]}\n`;
-        signalInfo = `si_code=${lines[crashStackStartLine + 1]}, si_addr=${bucketSignalAddress(lines[crashStackStartLine + 2])}\n`;
+        const siCode: string = lines[crashStackStartLine + 1] ?? "";
+        const siAddr: string = lines[crashStackStartLine + 2] ?? "";
+        const signalInfoParts: string[] = [];
+        if (siCode.length > 0) {
+            signalInfoParts.push(`si_code=${siCode}`);
+        }
+        if (siAddr.length > 0) {
+            signalInfoParts.push(`si_addr=${bucketSignalAddress(siAddr)}`);
+        }
+        signalInfo = signalInfoParts.length > 0 ? `${signalInfoParts.join(", ")}\n` : "";
         crashStackStartLine += 3;
     } else {
         // The signal type may fail to be written.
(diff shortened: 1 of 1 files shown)


Extension/src/LanguageServer/extension.ts
```diff
@@ -1362,8 +1362,21 @@ async function handleCrashFileRead(crashDirectory: string, crashFile: string, cr
...
+        if (siCode.length > 0) {
+            signalInfoParts.push(`si_code=${siCode}`);
+        }
+        if (siAddr.length > 0) {
+            signalInfoParts.push(`si_addr=${bucketSignalAddress(siAddr)}`);
+        }
+        signalInfo = signalInfoParts.length > 0 ? `${signalInfoParts.join(", ")}\n` : "";
+        // Only advance past the header lines that actually exist so a missing si_code/si_addr
+        // line does not cause the first stack frame to be skipped.
+        crashStackStartLine += 1 + (siCodeRaw !== undefined ? 1 : 0) + (siAddrRaw !== undefined ? 1 : 0);
```
bobbrow, line 1379: Could you also do `1 + signalInfoParts.length` and get the same result?

sean-mcmanus: @bobbrow Copilot says: No — they're not equivalent. `signalInfoParts.length` counts only the parts that survived trimming (non-empty), whereas the advance needs to count the header *lines that physically exist* in the file, including present-but-empty/whitespace-only ones.

Example: the crash file has a `SIG…` line followed by an empty `si_code` line and a valid `si_addr` line. Then:
- `siCodeRaw !== undefined` → true, `siAddrRaw !== undefined` → true → advance by `1 + 1 + 1 = 3` (correct — all three header lines get skipped).
- But `siCode` is empty so it's dropped, giving `signalInfoParts.length === 1` → `1 + 1 = 2`. That would leave the empty `si_code` line in the stack region and misalign parsing.

That present-but-empty case is exactly the scenario the reviewer's earlier comment flagged, so `1 + signalInfoParts.length` would reintroduce the bug. The current `siCodeRaw`/`siAddrRaw` undefined checks are what's needed.

bobbrow: Why would we have "present but empty" lines in some cases and undefined (missing) lines in other cases?

sean-mcmanus: @bobbrow The signal code and address are not always available.

sean-mcmanus: @bobbrow Oh, wait -- that was the case previously, but my active cpptools side PR changes that, just a sec...

sean-mcmanus: @bobbrow I'm changing my cpptools side PR, i.e. this PR is still valid -- it correctly captures the "missing" cases.

Colengms (approved):

PR #14648, opened 2026-08-07
Fix C++ agent definition lookup returning declarations

carsonRadtke:
Tracks [DevDiv 2860151](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2860151) and [DevDiv 3030687](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/3030687).

Agent lookup currently routes through the human definition request, which may return declarations. This adds an internal command backed by the strict native request while preserving standard navigation behavior.

### DIFF
--- a/Extension/src/LanguageServer/Providers/goToDefinitionProvider.ts
+++ b/Extension/src/LanguageServer/Providers/goToDefinitionProvider.ts
@@ -3,9 +3,15 @@
  * See 'LICENSE' in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import * as vscode from 'vscode';
-import { Definition, DefinitionLink, DefinitionRequest, Position, ResponseError, TextDocumentPositionParams } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


Extension/src/LanguageServer/Providers/goToDefinitionProvider.ts
```diff
@@ -55,3 +67,21 @@ export async function sendGoToDefinitionRequest(client: DefaultClient, uri: vsco
...
+}
+
+export function sendGoToDefinitionForAgentRequest(
+    client: DefaultClient,
+    uri: vscode.Uri,
+    position: vscode.Position,
+    token: vscode.CancellationToken
+): Promise<vscode.Location[] | undefined> {
+    return sendGoToDefinitionRequest(client, uri, position, token, DefinitionForAgentRequest);
+}
```
DavidARaygoza, line 87: Why have these helper functions when you could just keep `sendGoToDefinitionRequest` as `export` and call it directly from both `OnGoToDefinition` callers?

PR #14663, opened 2026-08-11
Validate Yarn lockfile selectors

sean-mcmanus:
## Summary

Add static, pre-install validation that checks dependency and resolution selectors from `Extension/package.json` are represented in `Extension/yarn.lock`. Run focused fixture tests and current-manifest validation before dependency installation in the shared compile-and-test workflow, catching stale resolution selectors that Yarn Classic's `--frozen-lockfile` accepts.

> This PR was investigated and created by GitHub Copilot in VS Code. Any message starting with `✨Copilot:` was sent by Copilot.

## Testing

- `yarn test-yarn-lock`
- `yarn verify-yarn-lock`
- `yarn eslint --report-unused-disable-directives src test ui .scripts`
- Verified the validator reports exactly `fast-uri@^3.1.5` for the historical malformed state from 77773532755f35a5d99ee2b0878d71ddbe4b8b00
- `git diff --check origin/main...HEAD`

### DIFF
--- a/.github/workflows/job-compile-and-test.yml
+++ b/.github/workflows/job-compile-and-test.yml
@@ -34,6 +34,10 @@ jobs:
         with:
           node-version: 24
 
+      - name: Validate Yarn lockfile
+        run: yarn test-yarn-lock && yarn verify-yarn-lock
+        working-directory: Extension
+
       - name: Install Dependencies
         run: yarn install ${{ inputs.yarn-args }}
         working-directory: Extension
--- a/Extension/.scripts/verifyYarnLock.mjs
+++ b/Extension/.scripts/verifyYarnLock.mjs
@@ -1,0 +1,97 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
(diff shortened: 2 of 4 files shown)


.github/workflows/job-compile-and-test.yml
```diff
@@ -34,6 +34,10 @@ jobs:
         with:
           node-version: 24
 
+      - name: Validate Yarn lockfile
```
bobbrow, line 37: Does this mean our builds will randomly fail as packages update, or do they only change when we're explicitly making changes to package.json?

sean-mcmanus: ✨Copilot: @bobbrow It only changes when the checked-in files change inconsistently. This validator is static and offline: it compares selectors in the repository's `package.json` and `yarn.lock` bytes before `yarn install`; it does not resolve packages or inspect a registry. Publishing a newer upstream package therefore cannot make an unchanged commit start failing.

bobbrow (approved):

PR #14666, opened 2026-08-11
Respect inactive region column offsets when applying decorations

### LINKED ISSUE #12882: Incorrect inactive region when preprocessor directive is preceded by a comment
The C/C++ Extension currently deals with inactive regions using a per-line granularity, but the IntelliSense engine provides inactive regions with line/column (offset) granularity.  Trying to reduce that granularity to per-line results in some unusual scenario, such as:

Image

Using this issue to track propagating the full granularity of the inactive region We're using `createTextEditorDecorationType` to render regions as inactive, which should support line/column granularity.

Various features check inactive regions to determine if operating within an active region or not, which makes this a non-trivial change.

8prashant:
## Summary

Fixes #12882

This change updates inactive-region decoration ranges to preserve the column offsets reported by the IntelliSense engine.

Previously, inactive regions were converted to whole-line VS Code ranges by forcing both start and end columns to `0`. That caused lines containing active text before a preprocessor directive, such as a leading comment before `#if 0`, to be dimmed incorrectly.

The decoration range now uses `startColumn` and `endColumn` when they are present, while preserving the existing line-based fallback behavior for responses that do not include column data.

## Changes

- Added optional `startColumn` and `endColumn` fields to `InputRegion`.
- Updated inactive-region decoration creation to use the reported start/end columns.
- Kept `0` as the fallback column value for compatibility with line-only inactive region data.

## Testing

Tested locally on Windows.

- `yarn compile`
- `yarn lint`
- `yarn test`
- `git diff --check`

Also manually verified in an isolated Extension Development Host using a repro case where a comment precedes `#if 0` on the same line. The comment remains active while the inactive block is dimmed.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -223,7 +223,9 @@ interface FileChangedParams extends WorkspaceFolderParams {
 
 interface InputRegion {
     startLine: number;
+    startColumn?: number;
     endLine: number;
+    endColumn?: number;
 }
 
 interface DecorationRangesPair {
@@ -3054,7 +3056,11 @@ export class DefaultClient implements Client {
             this.inactiveRegionsDecorations.set(uriString, currentSet);
         }
 
-        Array.prototype.push.apply(currentSet.ranges, inactiveRegions.map(element => new vscode.Range(element.startLine, 0, element.endLine, 0)));
+        Array.prototype.push.apply(currentSet.ranges, inactiveRegions.map(element => new vscode.Range(
+            element.startLine,
+            element.startColumn ?? 0,
+            element.endLine,
+            element.endColumn ?? 0)));
 
         // Apply the decorations to all *visible* text editors
         const editors: vscode.TextEditor[] = vscode.window.visibleTextEditors.filter(e => e.document.uri.toString() === uriString);


Colengms:
Hi @8prashant .  Thanks for the contribution!  The native component of the C/C++ Extension doesn't currently pass down `startColumn` and `endColumn` in `InputRegion`.  I went ahead and made the corresponding changes to the native side.  We can use this PR, if you'd like to remove the `?` and undefined checks. (They will always be present, with my native-side changes).  Or I could make the TypeScript-side adjustments myself in a new PR and you could abandon this one.

Colengms:
> Hi Prashant Kumar Rai (@8prashant) . Thanks for the contribution! The native component of the C/C++ Extension doesn't currently pass down `startColumn` and `endColumn` in `InputRegion`. I went ahead and made the corresponding changes to the native side. We can use this PR, if you'd like to remove the `?` and undefined checks. (They will always be present, with my native-side changes). Or I could make the TypeScript-side adjustments myself in a new PR and you could abandon this one.

Actually, I needed to make another minor change, so I created: https://github.com/microsoft/vscode-cpptools/pull/14667

We can abandon this PR.  This change should make it into 1.34.0

sean-mcmanus:
@8prashant Fixed in https://github.com/microsoft/vscode-cpptools/releases/tag/v1.34.0 (pre-release).

PR #14667, opened 2026-08-12
Update inactive regions to support column granularity

Colengms:
Added column fields.
Broke out a separate line-only object for code folding, which doesn't currently support column offsets.


8prashant:
Thanks for the update, @Colengms .
That makes sense. Since the complete fix needs the native-side changes as well, I'm fine with abandoning this PR #14666  in favor of #14667 .
If any part of this TypeScript-side change or investigation was useful as a reference, would you mind adding a **Co-authored-by** trailer if appropriate?
Either way, thank you for taking this forward. I'll abandon this PR #14666 .

sean-mcmanus (approved):

PR #14681, opened 2026-08-14
Set C/C++ go-to fallback defaults between definition and declaration

### LINKED ISSUE #14509: Add a command "Go To Definition/Declaration "
### Feature Request

I would love to be able to apply "Go To Declaration" when I am requesting it on a definition, and "Go To Definition" when I am requesting it on a declaration.

I foolishly thought this was the standard F12 behaviour before finding out I need to bind both and use them contextually! I imagine its trivial to implement the context switching internally and apply the correct one given the context of the request.

i.e. 
IF requested on a declaration -> Go To Definition
IF requested on a definition -> Go To Decleration

My workaround currently is binding the behaviour to the "when" field inside Keybinds.json based on file endings:
```
    {
        "key": "F12",
        "command": "editor.action.revealDeclaration",
        "when": "editorHasDeclarationProvider && editorTextFocus && !isInEmbeddedEditor && resourceExtname =~ /\\.(cpp|cxx|cc|c\\+\\+|c)$/"
    },
    {
        "key": "F12",
        "command": "editor.action.revealDefinition",
        "when": "editorHasDefinitionProvider && editorTextFocus && !isInEmbeddedEditor && resourceExtname =~ /\\.(hpp|hxx|hh|h\\+\\+|h|inl|ipp)$/"
    },
```

This fails in nuanced cases such as F12 on `#include` lines when in a source file, this will attempt "Go To Declaration" and fail but should be using "Go To Definition". Another reason a context based command would be useful!

A native command provided by the extension would be nice.

8prashant:
This PR improves C/C++ navigation defaults so definition and declaration commands fall back to each other automatically.

## Linked issue
Fixes https://github.com/microsoft/vscode-cpptools/issues/14509

## What changed
- Updated C/C++ language-specific configuration defaults to set:
  - `editor.gotoLocation.alternativeDefinitionCommand` = `editor.action.revealDeclaration`
  - `editor.gotoLocation.alternativeDeclarationCommand` = `editor.action.revealDefinition`
- Applied for:
  - `[c]`
  - `[cpp]`
  - `[cuda-cpp]`

## Why
This provides smoother F12 behavior across declaration/definition contexts without requiring users to create custom keybindings by file extension.

## Test coverage
- Added a unit test that validates these defaults are present in `package.json` for `[c]`, `[cpp]`, and `[cuda-cpp]`.
- Test executed:
  - `./node_modules/.bin/mocha.cmd dist/test/unit/gotoLocationDefaults.test.js --timeout 30000`
- Result: passing (1 test).


bobbrow:
For the most part we avoid changing VS Code defaults for languages because they are more difficult for the average user to override. We have only approved very few changes to the default settings that negatively impact the experience.

I don't think this is a change we want to take unless we get more feedback that confirms this is a widespread problem with the experience. I believe our language server is supposed to return the declaration when on the definition and vice versa already so I'd rather we fix any bugs in the language server itself than apply this policy for all users.

sean-mcmanus:
@bobbrow My impression was that is a widespread problem. I recall the default fallback is find all references which is almost never what a user wants. I think these defaults are what users actually want for C/C++, but I haven't tried it out yet.

I think there is one case where were return the declaration for the definition but not the other way around or something. I don't think there's a cpptools side bug.

sean-mcmanus:
@8prashant Sorry, I forgot about this issue -- I'll try to look into it for 1.34.3.

Extension/test/unit/gotoLocationDefaults.test.ts
sean-mcmanus, line 1: ✨Copilot (agent151): [Nit] This file is committed with CRLF line endings, unlike the existing TypeScript unit tests, so `git diff --check` reports trailing whitespace on every added line. Please normalize it to LF.

8prashant: Hi @sean-mcmanus,

I have updated the PR addressing the feedback:

**Normalized Line Endings**: Converted `Extension/test/unit/gotoLocationDefaults.test.ts` to standard LF line endings so `git diff --check` passes cleanly without trailing whitespace/CRLF warnings.

Could you please take another look when you have a chance? Thank you!

8prashant:
Hi @sean-mcmanus, when you have some time, could you please review this PR [#14681](https://github.com/microsoft/vscode-cpptools/pull/14681) also?

Thank you!

sean-mcmanus:
@8prashant FYI, I've been busy with other stuff. I still plan to review this, not 100% sure when yet.

UPDATE: This is still on my radar to review, but I have to investigate what's going on to trigger the Find All References fallback, and I've been very busy with other stuff.

sean-mcmanus:
@8prashant This was a bug on the cpptools side. A PR with a fix is pending (maybe 1.35.3). It's possible this PR might still be useful in the future if there's some other special case we're not able to fix.

PR #14684, opened 2026-08-14
Add processFilter for remote attach process selection

### LINKED ISSUE #14682: Support attaching to a process by name/pattern for remote attach (useExtendedRemote / pipeTransport
### Feature request

When attaching to a process on a **remote** target, the process to attach to always has to be selected manually from the process picker. Please allow the configuration to name the process, so that the attach can be fully automated.

### Why the current behaviour is a problem

I generate `launch.json` files automatically from build metadata (the Yocto Project's `devtool ide-sdk`, which creates VS Code configurations for cross-debugging an embedded target over `gdbserver`). Each generated configuration is bound to exactly one executable: `program`, `sourceFileMap`, `additionalSOLibSearchPath` and the `setupCommands` (sysroot, pretty printers) all belong to that one binary.

The PID cannot be put into the generated configuration, because it changes on every boot and every restart of the service. So the only option today is the picker, which lists *every* process on the device. That is unfortunate for two reasons:

- The user has to pick the right process manually on every debug session, although the configuration already knows which executable it is for.
- Picking anything else silently produces a session with mismatched symbols and source mappings.

### The extension already does this for local attach

`Extension/src/Debugger/configurationProvider.ts`:

```ts
// If program is specified and not remote attach, try to find the matching process by name
if (config.program && !config.pipeTransport && !config.useExtendedRemote) {
    processId = await this.findProcessByProgramName(config.program, token);
}
```

`findProcessByProgramName()` filters the process list by the basename of `program` and attaches directly when exactly one process matches. The remote cases are excluded, apparently because the function uses `NativeAttachItemsProviderFactory.Get()`, which enumerates local processes only.

For remote there is already an equivalent process source: `RemoteAttachPicker.getRemoteProcessesExtendedRemote()` (`info os processes` via `target extended-remote`) and `getRemoteOSAndProcesses()` for `pipeTransport`.

### Proposal

When `request: attach` and no `processId` is given, apply a name/pattern match to the remote process list as well, and only fall back to the picker if the match is not unique:

- match count `1` → attach directly
- match count `> 1` → show the picker, prefiltered to the matches
- match count `0` → show the full picker (or report an error)

Two possible shapes, either would solve it:

1. Simply extend the existing `program`-based matching to the remote paths.
2. Add an explicit property, e.g. `processFilter` (regular expression), matched against the fields already shown in the picker. This is more flexible for the remote case, because for `useExtendedRemote` the `AttachItem.label` is not a bare process name but the user plus the whole command line, built in `parseProcessesFromInfoOsProcesses()`:

```ts
processes.push({ label: userCommand, id, description: id });
```

so a regex like `/usr/bin/my-daemon` matches reliably where a basename comparison would not.

An example of the configuration I would like to generate:

```json
{
    "name": "attach my-daemon",
    "type": "cppdbg",
    "request": "attach",
    "program": "/path/on/build/host/to/unstripped/my-daemon",
    "MIMode": "gdb",
    "miDebuggerPath": "/path/to/aarch64-poky-linux-gdb",
    "miDebuggerServerAddress": "192.168.7.2:1234",
    "useExtendedRemote": true,
    "processFilter": "/usr/bin/my-daemon"
}
```

### Additional benefit

`request: attach` also detaches from the process instead of killing it when the debug session is stopped, unlike `request: launch`, which is what #4166 is about. Being able to use `request: attach` without a manual picker step would make that the natural configuration for attaching to a running service on a remote target.

### Environment

- C/C++ extension version: 1.32.2
- Target: aarch64 Linux (QEMU), `gdbserver --multi`, connected via `miDebuggerServerAddress` + `useExtendedRemote`

### Related

- #9505 — added `useExtendedRemote` (thanks, it works well)
- #4166 — detach instead of kill when stopping a `gdbserver` session

afreof:
## Problem

When attaching to a process on a remote target (`useExtendedRemote` or `pipeTransport`), the process to attach to always has to be picked manually from the full process list on every debug session, even though a generated `launch.json` already knows which executable it targets. The PID can't be hard-coded because it changes on every restart/boot, and picking the wrong entry silently produces a session with mismatched symbols and source mappings.

This mirrors the existing `program`-based auto-selection already used for **local** attach (`findProcessByProgramName`), which does not apply to remote attach because the remote process list comes from a different source (`RemoteAttachPicker`) and its `label`/`detail` shapes differ per transport.

## Solution

Adds an optional `processFilter` (regular expression) to the `cppdbg` attach configuration:

- Matched against `label`, `description`, and `detail` of each remote process entry (the exact fields differ by transport: `useExtendedRemote` puts the user + full command line in `label`; `pipeTransport` puts the process name in `label` and the command line in `detail`).
- **Exactly one match** → attach to it directly, no picker.
- **Multiple matches** → show the picker, pre-filtered to only the matching entries.
- **No match** → fall back to the full picker, unchanged from today.
- An invalid regular expression throws a clear error instead of silently matching nothing.

Scope: this only affects remote attach (`pipeTransport` / `useExtendedRemote`). Local attach is unaffected by this PR.

Example `launch.json`:
```jsonc
{
    "name": "attach my-daemon",
    "type": "cppdbg",
    "request": "attach",
    "program": "/path/on/build/host/to/unstripped/my-daemon",
    "MIMode": "gdb",
    "miDebuggerPath": "/path/to/aarch64-poky-linux-gdb",
    "miDebuggerServerAddress": "192.168.7.2:1234",
    "useExtendedRemote": true,
    "processFilter": "/usr/bin/my-daemon"
}
```

Closes #14682.

## Open question for maintainers

Should `processFilter` also apply to **local** attach, in addition to (or instead of) the existing `program`-based basename matching? Since `program` is a required field for `cppdbg` attach, both would typically be present locally at the same time, so the semantics need a decision — I'd lean towards intersecting with the `program` match (never letting `processFilter` alone select a process unrelated to `program`) rather than giving it precedence, but wanted to keep this PR scoped to the originally requested remote-only case and get feedback before extending it.

## Testing

- Added unit tests for the matching helper (empty/non-string filter, label/description/detail matching, multiple matches, invalid regex) — see `Extension/test/unit/processFilter.test.ts`.
- Full existing unit test suite passes (208 passing).
- Manually verified against a real `gdbserver --multi` target with `useExtendedRemote: true`:
  - unique `processFilter` match attaches directly without showing the picker
  - a filter matching multiple processes shows the picker pre-filtered to those entries
  - a filter matching nothing falls back to the full, unfiltered picker (existing behavior)

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -5260,6 +5260,11 @@
                                     }
                                 ]
                             },
+                            "processFilter": {
+                                "type": "string",
+                                "description": "%c_cpp.debuggers.processFilter.description%",
+                                "default": ""
+                            },
                             "filterStdout": {
                                 "type": "boolean",
                                 "description": "%c_cpp.debuggers.filterStdout.description%",
--- a/Extension/src/Debugger/processFilter.ts
+++ b/Extension/src/Debugger/processFilter.ts
@@ -1,0 +1,32 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as nls from 'vscode-nls';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+export interface ProcessFilterItem {
+    label?: string;
+    description?: string;
+    detail?: string;
+}
+
+export function filterProcessItems<T extends ProcessFilterItem>(items: T[], processFilter?: unknown): T[] | undefined {
+    // The value comes from launch.json, so it is not guaranteed to be a string.
+    if (typeof processFilter !== 'string' || !processFilter.trim()) {
+        return undefined;
+    }
+
+    let processRegex: RegExp;
+    try {
+        processRegex = new RegExp(processFilter);
+    } catch {
+        throw new Error(localize("invalid.processFilter.regex", "Invalid {0} regular expression: {1}", "processFilter", processFilter));
+    }
+
+    return items.filter((item: T) => [item.label, item.description, item.detail]
+        .some((value: string | undefined) => typeof value === "string" && processRegex.test(value)));
+}
--- a/Extension/test/unit/processFilter.test.ts
+++ b/Extension/test/unit/processFilter.test.ts
@@ -1,0 +1,61 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ * See 'LICENSE' in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { deepStrictEqual, strictEqual, throws } from 'assert';
+import { describe, it } from 'mocha';
+import { filterProcessItems } from '../../src/Debugger/processFilter';
+
+interface TestProcessItem {
+    label?: string;
+    description?: string;
+    detail?: string;
+    id: string;
+}
+
+describe('Remote attach process filter', () => {
+    const processes: TestProcessItem[] = [
+        { id: '101', label: 'root /usr/bin/my-daemon --serve', description: '101' },
+        { id: '102', label: 'root /usr/bin/other-service', description: '102', detail: 'worker' },
+        { id: '103', label: 'app /usr/bin/my-daemon --once', description: '103' }
+    ];
+
+    it('returns undefined when filter is empty', () => {
+        strictEqual(filterProcessItems(processes, ''), undefined);
+        strictEqual(filterProcessItems(processes, '   '), undefined);
+        strictEqual(filterProcessItems(processes, undefined), undefined);
+    });
+
+    it('returns undefined when filter is not a string', () => {
+        strictEqual(filterProcessItems(processes, 1234), undefined);
+        strictEqual(filterProcessItems(processes, true), undefined);
+        strictEqual(filterProcessItems(processes, {}), undefined);
+    });
+
+    it('matches by label and description and detail', () => {
+        deepStrictEqual(filterProcessItems(processes, 'other-service')?.map(p => p.id), ['102']);
+        deepStrictEqual(filterProcessItems(processes, '^101$')?.map(p => p.id), ['101']);
+        deepStrictEqual(filterProcessItems(processes, 'worker')?.map(p => p.id), ['102']);
+    });
+
+    it('preserves edge whitespace in the regular expression', () => {
+        deepStrictEqual(filterProcessItems(processes, '^root /usr/bin/my-daemon --serve ')?.map(p => p.id), []);
+    });
+
+    it('returns multiple matches when regex matches more than one process', () => {
+        deepStrictEqual(filterProcessItems(processes, 'my-daemon')?.map(p => p.id), ['101', '103']);
+    });
+
+    it('throws for invalid regular expression', () => {
+        throws(() => filterProcessItems(processes, '['), /Invalid processFilter regular expression/);
+    });
+
+    it('does not treat missing fields as empty strings', () => {
+        const items: TestProcessItem[] = [
+            { id: '201' }, // label, description, and detail are all missing
+            { id: '202', detail: '' } // detail is explicitly an empty string
+        ];
+        deepStrictEqual(filterProcessItems(items, '^$')?.map(p => p.id), ['202']);
+    });
+});
--- a/Extension/package.nls.json
+++ b/Extension/package.nls.json
@@ -982,6 +982,7 @@
             "{Locked=\"`${command:pickProcess}`\"}"
         ]
     },
+    "c_cpp.debuggers.processFilter.description": "Optional regular expression used to match remote attach candidates by label, description, or detail. If exactly one process matches, the debugger attaches automatically. If multiple processes match, the process picker is shown with only matching entries. If no process matches, the full process picker is shown. An invalid regular expression reports an error.",
     "c_cpp.debuggers.program.attach.markdownDescription": {
         "message": "Full path to the program executable. The debugger will search for a running process matching this executable path and attach to it. If multiple processes match, a selection prompt will be shown. This field is required to load debug symbols for the attached process.",
         "comment": [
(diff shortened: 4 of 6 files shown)


afreof:
@microsoft-github-policy-service agree company="Siemens AG"

Extension/src/Debugger/processFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface ProcessFilterItem {
+    label?: string;
+    description?: string;
+    detail?: string;
+}
+
+export function filterProcessItems<T extends ProcessFilterItem>(items: T[], processFilter?: unknown): T[] | undefined {
+    // The value comes from launch.json, so it is not guaranteed to be a string.
+    const trimmedFilter: string | undefined = typeof processFilter === 'string' ? processFilter.trim() : undefined;
```
sean-mcmanus: ✨Copilot (agent135): [Moderate] Please preserve the configured regex here. Leading and trailing whitespace is valid regex syntax: for example, `^my-daemon ` intentionally does not match `my-daemon-helper`, but trimming changes it to `^my-daemon`. If that broadened expression has exactly one match, the code silently auto-attaches to the wrong PID instead of falling back to the picker. Trimming can still be used to detect an empty/all-whitespace value, but construct `RegExp` from the original string and add an edge-whitespace test.

Extension/package.json
```diff
@@ -5259,6 +5259,11 @@
                                     }
                                 ]
                             },
+                            "processFilter": {
+                                "type": "string",
```
sean-mcmanus, line 5264: ✨Copilot (agent135): [Moderate] Please add this property to `Extension/tools/OptionsSchema.json` under `CppdbgAttachOptions` and regenerate this file. `OptionsSchema.json` is the documented schema source, and `generateOptionsSchema.ts` replaces the complete `cppdbg` attach schema in `package.json` from it. As written, the next `yarn generate-options-schema` removes `processFilter`, including its completion, validation, and documentation. This same source/generated drift previously required #14523 to resynchronize.

Extension/src/Debugger/processFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!trimmedFilter) {
+        return undefined;
+    }
+
+    let processRegex: RegExp;
+    try {
+        processRegex = new RegExp(trimmedFilter);
+    } catch {
+        throw new Error(`Invalid processFilter regular expression: ${trimmedFilter}`);
```
sean-mcmanus: ✨Copilot (agent135): [Minor] This newly introduced exception is user-facing through both remote-picker invocation paths, but its text is hard-coded English. Please route it through the existing `vscode-nls`/`package.nls.json` localization path (locking the `processFilter` identifier as appropriate) so localized installations receive a translated diagnostic.

sean-mcmanus (commented):
✨Copilot (agent135): [Moderate] Two correctness findings and one localization finding are noted inline.

Extension/src/Debugger/processFilter.ts
sean-mcmanus: ✨Copilot (agent135): The diagnostic now goes through `vscode-nls`, but `processFilter` is still literal translatable text. Please either use the object-form localization key with a `{Locked="processFilter"}` comment or pass the identifier through a placeholder, matching the nearby debugger diagnostics. Then this thread is fully addressed.

afreof: Thank you for your patience with an embedded developer :roll_eyes:

Extension/test/unit/processFilter.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        strictEqual(filterProcessItems(processes, {}), undefined);
+    });
+
+    it('matches by label and description and detail', () => {
+        deepStrictEqual(filterProcessItems(processes, 'other-service')?.map(p => p.id), ['102']);
+        deepStrictEqual(filterProcessItems(processes, '^101$')?.map(p => p.id), ['101']);
+        deepStrictEqual(filterProcessItems(processes, 'worker')?.map(p => p.id), ['102']);
+    });
+
+    it('preserves edge whitespace in the regular expression', () => {
```
sean-mcmanus, line 42: ✨Copilot (agent135): [Minor] Please add a regression case for the absent-field behavior fixed in this revision. With this fixture, `filterProcessItems(processes, '^$')` should return no entries because the missing `detail` fields are not empty strings; an explicitly empty field can be included to preserve that distinction. This test would fail the prior `?? ""` implementation and directly protect the fix.

sean-mcmanus (commented):
✨Copilot (agent135): [Minor] The two Copilot findings are fixed; one focused regression-test gap remains inline.

Extension/test/unit/processFilter.test.ts
sean-mcmanus: @afreof FYI, my review agent can potentially add this regression test itself in a follow up PR if you don't want to.

afreof: Let's try to have it as one consistent and complete PR. Thank you.

Extension/src/Debugger/processFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function filterProcessItems<T extends ProcessFilterItem>(items: T[], processFilter?: unknown): T[] | undefined {
+    // The value comes from launch.json, so it is not guaranteed to be a string.
+    if (typeof processFilter !== 'string' || !processFilter.trim()) {
+        return undefined;
+    }
+
+    let processRegex: RegExp;
+    try {
+        processRegex = new RegExp(processFilter);
```
sean-mcmanus, line 25: ✨Copilot (agent135): [Moderate] This compiles an unrestricted JavaScript regular expression and then evaluates it synchronously against process names and command lines supplied by the remote target. When the configured expression has catastrophic backtracking, a remote process owner can provide matching input that blocks the extension-host event loop during attach; catching syntax errors does not bound match time. Please use a linear-time regex implementation with a documented supported subset, or isolate the complete batch evaluation in a worker with a hard timeout. Pattern or input length limits alone are not a reliable bound.

sean-mcmanus (commented):
✨Copilot (agent135): [Moderate] A conditional regular-expression denial-of-service risk remains inline.

Extension/src/Debugger/processFilter.ts
afreof: Wouldn’t that be going a bit too far and being a bit too paranoid? It would also have a negative impact on performance, after all.

I’ll wait to hear your thoughts before I implement this.

sean-mcmanus: @afreof Yeah, I had a chat with my agent -- it's okay.

sean-mcmanus: ✨Copilot (agent135): You’re right. After reconsidering the trust model, the user explicitly supplies the regex and initiates attach, and can replace a pathological expression after reloading. Remote text can trigger a vulnerable user-authored pattern, but that is better treated as an inherent robustness/defense-in-depth concern for this regex option, not a Moderate security requirement. No change is requested for this PR.

sean-mcmanus (approved):

PR #14688, opened 2026-08-17
Add Select a Translation Unit command

sean-mcmanus:
## Summary
Adds a **Select a Translation Unit** command for choosing which source file provides IntelliSense context for the active file. The command displays matching translation-unit candidates in a quick pick and marks the current selection.

> This PR was investigated and created by Copilot with `GPT-5.6 Sol` (in VS Code). Any message starting with `✨Copilot:` was sent by Copilot.

## Details
- Registers the command palette action.
- Requests candidate paths and the current selection from the language server.
- Maps candidate paths into the active workspace and sends the selected path back.
- Labels the current translation unit with a check icon and description.

## Validation
- Extension ESLint checks
- TypeScript compilation


Extension/src/LanguageServer/extension.ts
```diff
@@ -468,6 +469,61 @@ async function onRestartIntelliSenseForFile() {
...
+}
+
+async function onSelectTranslationUnit(): Promise<void> {
+    const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
+    if (!activeEditor || !util.isCpp(activeEditor.document)) {
+        return;
+    }
+
+    const client: Client = clients.ActiveClient;
+    const result = await client.getTranslationUnitSourceCandidates(activeEditor.document.uri);
```
sean-mcmanus: ✨Copilot (agent22): [Minor] This request has no progress indication and no way to cancel it. Building the candidate list requires the language server to find the source files that reach the active file, so on a large project the command can sit here with no visible feedback after the user invokes it.

The sibling command right below makes a comparable server round-trip and treats it as potentially slow: `onSwitchHeaderSource` passes a `vscode.CancellationToken` into `requestSwitchHeaderSource`, then races the promise against a 2-second timer and, if it loses, shows a cancellable `vscode.window.withProgress` notification that forwards cancellation back through the token.

Consider reusing that same shape here, which would also mean threading a `CancellationToken` parameter through `Client.getTranslationUnitSourceCandidates` the way `requestSwitchHeaderSource` already does.

Extension/src/LanguageServer/extension.ts
```diff
@@ -468,6 +469,61 @@ async function onRestartIntelliSenseForFile() {
...
+        const isCurrent: boolean = editorPath === currentTranslationUnit;
+        return {
+            label: `${isCurrent ? "$(check) " : ""}${path.basename(relativePath)}`,
+            description: isCurrent ? localize('current.translation.unit', 'Current translation unit') :
+                directory === "." ? undefined : directory,
+            detail: editorPath,
+            translationUnit: editorPath
+        };
+    });
+    const selection: TranslationUnitQuickPickItem | undefined = await vscode.window.showQuickPick(items, {
```
sean-mcmanus, line 562: ✨Copilot (agent22): [Minor] Nothing guards `result.candidates` being empty, so `items` can be `[]` and still reach `showQuickPick`, which leaves the user with an empty picker rather than an explanation.

That is reachable without the server misbehaving. `NullClient.getTranslationUnitSourceCandidates` returns `{ candidates: [], currentTranslationUnit: "" }`, and `NullClient` becomes the active client once `ClientCollection.recreateClients(true)` switches to failsafe mode after a crash. That happens after activation, so this command is still registered in its enabled form — `registerCommands(false)` is only used on the `isIntelliSenseEngineDisabled` path in `main.ts` — and the `when` clause in `package.json` only tests `editorLangId` and `C_Cpp.intelliSenseEngine`, neither of which changes in failsafe mode.

`onSwitchHeaderSource` guards the equivalent case explicitly with `if (!targetFileName) { return; }`. An early return here when `result.candidates.length === 0`, optionally with a message, would match that.

sean-mcmanus (commented):
✨Copilot (agent22): Reviewed the TypeScript side at `65e388ec` as a companion to the language-server change. Two [Minor] findings on the new command, both about parity with the existing `onSwitchHeaderSource` sibling. The `getEditorPath` extraction itself is a genuine improvement — replacing the old `startsWith` prefix test with a `path.relative` containment check fixes a real false match (`/ws-other/x.cpp` matched a `/ws` root), and picking the longest matching root is correct for nested workspace folders.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: ✨Copilot: Addressed by showing an informational message and returning before opening the quick pick when no translation-unit candidates are available, including in failsafe mode.

Extension/src/LanguageServer/extension.ts
sean-mcmanus: ✨Copilot: Addressed by forwarding a cancellation token through candidate discovery, normalizing LSP cancellation, and showing a cancellable progress notification when the request takes longer than two seconds.

Extension/src/LanguageServer/client.ts
```diff
@@ -2759,7 +2773,7 @@ export class DefaultClient implements Client {
                 const isTrackedFile: boolean = hasNativeFileTypeMappings()
                     ? isTagParsableFile(uri.fsPath)
                     : isTagParsableFile(uri.fsPath) ||
-                        (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
+                    (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
```
sean-mcmanus, line 2791: ✨Copilot (agent22): [Nit] This is the whole content of the latest commit, and it reformats a pre-existing line that this PR does not otherwise touch — the `isTrackedFile` ternary belongs to the `onDidChangeTextDocument` watcher, unrelated to translation-unit selection.

Nothing in the repo asks for it. There is no `.editorconfig`, no Prettier config, and no `tsfmt.json`, and `Extension/eslint.config.js` configures no indentation rule at all, so this is an editor Format Document artifact rather than a lint fix.

It also reads slightly worse. The `||` continuation now sits at the same column as the two ternary arms:

```ts
const isTrackedFile: boolean = hasNativeFileTypeMappings()
    ? isTagParsableFile(uri.fsPath)
    : isTagParsableFile(uri.fsPath) ||
    (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
```

so the last line looks like a third branch instead of the continuation of the `:` branch, which is what the extra indent on `main` conveyed.

Reverting this hunk would keep the PR's diff limited to the feature and avoid a needless conflict for anything else in flight against `client.ts`.

sean-mcmanus (commented):
✨Copilot (agent22): Re-reviewed at `ececda78`. The only change since `360056d5` is a one-line reformat of pre-existing code in `client.ts`; the feature itself is unchanged and CI is green. One [Nit] on that hunk.

Extension/src/LanguageServer/client.ts
sean-mcmanus: ✨Copilot: Reverted the unrelated formatting hunk in a separate append-only commit. The `isTrackedFile` expression now matches `main`, while the translation-unit feature changes remain unchanged.

sean-mcmanus: ✨Copilot (agent22): Retracting this — it was wrong, and the revert it prompted was unnecessary.

Sean has clarified that because formatting is not enforced on check-in, pre-existing incorrect formatting should be cleaned up by whoever's PR happens to notice it. So the original commit was doing the right thing, and "this line is unrelated to the feature" is not a reason to leave it alone in this repo.

My second point was also wrong. I argued `main`'s deeper continuation indent read better, but with no `.editorconfig`, Prettier config, or ESLint indentation rule, the TypeScript formatter's output is the correct form — which is what the reverted commit had produced. I inverted which version was the defect.

Please feel free to re-apply the formatting fix; no need to keep the revert on my account. Sorry for the churn.

Colengms (approved):

Issue #12537, opened 2024-08-02
Warning in `launch.json`: Property `env` is not allowed


obsgolem:
### Environment

- OS and version: Windows
- VS Code: Version: 1.92.0 b1c0a14de1414fcdaa400695b4db1c0799bc3124
- C/C++ extension: 1,21,5
- OS and version of remote machine (if applicable):
- GDB / LLDB version:

### Bug Summary and Steps to Reproduce

Bug Summary:
Add the following to any C++ project's `launch.json`:
```
{
    "type": "cppvsdbg",
    "request": "launch",
    "name": "INSERT NAME HERE",
    "program": "INSERT PROGRAM PATH HERE",
    "args": [
    ],
    "internalConsoleOptions": "openOnSessionStart",
    "cwd": "${workspaceFolder}",
    "sourceFileMap": {},
    "preLaunchTask": "Build Server",
    "env": {
        "PATH": ""
    }
},
```

Note that `env` has a warning `Property env is not allowed.`

I am aware that `environment` is the officially recommended way to specify environment variables, but `env` is works too, and is significantly more convenient, as `environment` has extremely wacky and unintuitive syntax.

### Debugger Configurations

```shell
{
    "type": "cppvsdbg",
    "request": "launch",
    "name": "INSERT NAME HERE",
    "program": "INSERT PROGRAM PATH HERE",
    "args": [
    ],
    "internalConsoleOptions": "openOnSessionStart",
    "cwd": "${workspaceFolder}",
    "sourceFileMap": {},
    "preLaunchTask": "Build Server",
    "env": {
        "PATH": ""
    }
},
```
```


### Debugger Logs

```shell
placeholder
```

PR #14691, opened 2026-08-18
Fix: Add `env` object syntax to `launch.json` schema for `cppdbg` and `cppvsdbg`

8prashant:
### Problem

[#12537](https://github.com/microsoft/vscode-cpptools/issues/12537) : Using the shorthand `env` object syntax in `launch.json` produces a false validation warning:

> Property `env` is not allowed.

```json
"env": {
    "PATH": "/usr/local/bin"
}

### CODE REVIEW

**Comment by bobbrow** [comment]
I don't think we should have two properties that do the same thing (also, this PR does not update the optionsSchema.json which means the next time package.json is generated, this will be deleted).

Adding @WardenGnaw for comment.

**Comment by WardenGnaw** [comment]
Thanks for the contribution. This change will work for `cppvsdbg` because it can also deserialize `env` and `env` has higher precedence than `environment`. 

One issue with this change is that `cppdbg` does not currently support it. [MIEngine](https://github.com/microsoft/MIEngine) only reads the `environment` array. 

Aside from updating the [OptionsSchema.json](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/tools/OptionsSchema.json). There needs to be a change to MIEngine to support this or limit the new schema entry to `cppvsdbg`.

The other option is to add a cpptools configuration-provider conversion from `env` to `environment` but that adds an additional layer of transforms / errors that can happen when p;umbling the environment variables to the target process.

**Comment by 8prashant** [comment]
Thanks [Andrew Wang (WardenGnaw)](https://github.com/WardenGnaw) for the review. I’ve pushed updates to address this by fixing the no-debug path so the inline adapter now merges both environment and env with env taking precedence, adding cppdbg normalization in the configuration provider to convert env to environment so [MIEngine](https://github.com/microsoft/MIEngine) receives the format it currently supports, updating **OptionsSchema.json** so the new env schema entries are preserved during regeneration, and adding Run Without Debugging integration coverage to verify env is applied and correctly overrides duplicate environment entries. I intentionally kept this fix in cpptools rather than making [MIEngine](https://github.com/microsoft/MIEngine) changes in this PR because it resolves both normal launch and Run Without Debugging immediately, whereas a [MIEngine](https://github.com/microsoft/MIEngine)-only change would not cover no-debug, and this approach avoids cross-repo dependency and release-coordination risk while keeping the change scoped and low risk; native [MIEngine](https://github.com/microsoft/MIEngine) env support can still be added later as a separate follow-up enhancement.

File: Extension/package.json [hunk 1]
```diff
@@ -4378,6 +4378,14 @@
                                 },
                                 "default": []
                             },
+                            "env": {
+                                "type": "object",
+                                "description": "%c_cpp.debuggers.env.description%",
```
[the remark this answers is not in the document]
8prashant: Comment has been addressed in this commit: https://github.com/microsoft/vscode-cpptools/pull/14691/changes/13f598c3620592c63466d1cdaefa84b542abe05a

8prashant:
Hi @sean-mcmanus 
Can u please review the changes again!!

sean-mcmanus:
@8prashant Yeah, sorry, I've been busy with other stuff and forgot about these PRs.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -270,6 +270,10 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
         // Add environment variables from .env file
         this.resolveEnvFile(config, folder);
 
+        // cppdbg (MIEngine) consumes the legacy `environment` array, not `env`.
+        // Convert here so both syntaxes work while preserving `env` precedence.
+        this.resolveEnvObjectForCppdbg(config);
```
sean-mcmanus: ✨Copilot (agent117): [Minor] The added scenario never exercises this conversion. It starts with `{ noDebug: true }`, which selects `RunWithoutDebuggingAdapter`; that adapter independently consumes `env`, so the test would still pass if `resolveEnvObjectForCppdbg` were removed. Please cover the normal `cppdbg` path or directly assert the provider's resolved configuration, including `env` precedence, so the MIEngine compatibility behavior introduced here is protected.

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
```diff
@@ -175,6 +200,37 @@ suite('Run Without Debugging Test', function (): void {
...
+                program: envExecutablePath,
+                args: [testVarName, envResultFilePath],
+                cwd: workspacePath,
+                environment: [{ name: testVarName, value: fallbackValue }],
+                env: envConfig,
+                externalConsole: debugType === 'cppdbg' ? false : undefined,
+                console: debugType === 'cppvsdbg' ? 'internalConsole' : undefined
+            },
+            { noDebug: true });
+
```
sean-mcmanus, line 308: ✨Copilot (agent117): [Minor] This can complete before the no-debug session has terminated: `envTest.cpp` writes its file before exiting, while the inline adapter emits `terminated` only after the terminal-shell execution ends. The test neither subscribes to `onDidTerminateDebugSession` before launch nor stops the session on failure, so it can leave an active session for later cases; the current Linux run has both following cases timing out while waiting for debug events. Please await this session's termination and clean it up in `finally`.

Extension/package.nls.json
```diff
@@ -927,6 +927,7 @@
             "{Locked=\"[ { \\\"name\\\": \\\"\"} {Locked=\"\\\", \\\"value\\\": \\\"\"} {Locked=\"\\\" } ]\"} {Locked=\"[ { \\\"\"} {Locked=\"\\\": \\\"\"} {Locked=\"\\\" } ]\"}"
         ]
     },
+    "c_cpp.debuggers.env.description": "Object of environment variables to add to the environment for the program. Example: { \"MY_VAR\": \"value\" }. Use `environment` for the array-of-objects form.",
```
sean-mcmanus: ✨Copilot (agent117): [Nit] Please use the structured `{ message, comment }` form and lock the JSON example plus the literal `environment` key. The adjacent `c_cpp.debuggers.environment.description` does this; without locks, localization can translate schema syntax or property names and leave a misleading `launch.json` example.

Extension/test/scenarios/RunWithoutDebugging/assets/envTest.cpp
```diff
@@ -0,0 +1,21 @@
+#include <cstdlib>
```
sean-mcmanus: ✨Copilot (agent117): [Nit] This new asset was committed with CRLF while the neighboring C++ assets use LF, so every added line is reported as trailing whitespace by `git diff --check`. Please normalize the file to LF.

sean-mcmanus (commented):
Reviewed the current head; see the inline comments.

8prashant:
Hi @sean-mcmanus, thanks for the thorough review! I've updated the PR to address all comments:

1. **Direct test coverage for `resolveEnvObjectForCppdbg`**
   - Added a dedicated test in [Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts](Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts#L211) that directly invokes `DebugConfigurationProvider.resolveDebugConfigurationWithSubstitutedVariables()` for `cppdbg`.
   - Asserts that `env` is deleted after transformation and that entries from `env` take precedence over `environment` when key collisions occur.

2. **Normalized line endings in `envTest.cpp`**
   - Converted [Extension/test/scenarios/RunWithoutDebugging/assets/envTest.cpp](Extension/test/scenarios/RunWithoutDebugging/assets/envTest.cpp) from CRLF to LF to match neighboring C++ test assets and resolve `git diff --check` warnings.

3. **Resolved test race condition and session lifecycle in `runWithoutDebugging.integration.test.ts`**
   - Updated `waitForResultFileText()` in [Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts](Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts#L155) to wait until the trimmed contents length is `> 0`, preventing early return while `std::ofstream` opens the file before writing.
   - Pre-subscribed `createSessionTerminatedPromise()` before starting the debug session in the env test, awaited session termination, and added a `finally` block to stop any lingering debug sessions and clean up temp files.

4. **Structured localization comment in `package.nls.json`**
   - Updated `c_cpp.debuggers.env.description` in [Extension/package.nls.json](Extension/package.nls.json#L930) to the object format with `{ message, comment }` and added `{Locked}` rules for `{ "MY_VAR": "value" }` and `` `environment` ``.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -700,6 +704,35 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        const mergedEnvironment = new Map<string, string>();
+
+        for (const entry of environment) {
+            if (util.isString(entry?.name) && util.isString(entry?.value)) {
+                mergedEnvironment.set(entry.name, entry.value);
+            }
+        }
+
+        for (const [name, value] of Object.entries(envObject)) {
+            if (util.isString(value)) {
```
sean-mcmanus, line 729: ✨Copilot (agent117): [Moderate] `env` follows the standard VS Code launch contract where a `null` value undefines an inherited variable, but this filter drops that entry and line 733 then deletes the only record of it. For example, `"env": { "FOO": null }` leaves an inherited `FOO` set. MIEngine's legacy `EnvironmentEntry.Value` is nullable and explicitly uses null to delete, so the conversion can preserve this operation. Please allow null in both new `env` schemas, carry it through this merge (including precedence over `environment`/`envFile`), and cover deletion in both the adapter and no-debug paths.

sean-mcmanus (commented):
Re-reviewed the current head; see the inline comment.

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -95,12 +120,17 @@ export class RunWithoutDebuggingAdapter implements vscode.DebugAdapter {
      */
     private async launchIntegratedTerminal(program: string, args: string[], cwd: string | undefined, env: NodeJS.ProcessEnv): Promise<void> {
         const terminalName = path.normalize(program);
-        const existingTerminal = vscode.window.terminals.find(t => t.name === terminalName);
+        let existingTerminal = vscode.window.terminals.find(t => t.name === terminalName);
+        if (existingTerminal && !this.environmentsEqual(terminalEnvironments.get(existingTerminal), env)) {
+            existingTerminal.dispose();
```
sean-mcmanus, line 132: ✨Copilot (agent117): [Minor] This now disposes any same-name terminal that has no `terminalEnvironments` entry. The map is rebuilt when the extension host reloads while terminals can survive, and an unrelated terminal can also have this name, so the next no-debug launch may close a terminal it does not own and terminate its process. Please only select/dispose terminals tracked by this module; when no tracked terminal matches, create a managed terminal without touching untracked ones.

sean-mcmanus (commented):
Re-reviewed the current head; see the inline comment.

Extension/src/Debugger/configurationProvider.ts
sean-mcmanus: ✨Copilot (agent117): The schema and provider now preserve `null`, but the integrated-terminal path still does not remove a variable inherited when VS Code started. `TerminalOptions.env` is an overlay on the terminal host environment; deleting the key from the cloned `process.env` omits the tombstone, so the base value is inherited again. The new test sets `process.env` only inside the extension host after startup, so it cannot reproduce that case. Please retain `null` for `createTerminal` (or supply a strict complete environment) and use key deletion only for the `spawn` paths; also add normal-adapter coverage for the provider conversion.

Extension/test/scenarios/RunWithoutDebugging/tests/runWithoutDebugging.integration.test.ts
sean-mcmanus: ✨Copilot (agent117): This lifecycle race is reintroduced by the two new scenarios. The null case returns as soon as `ofstream` creates the empty file, and the reused-terminal helper returns as soon as content is written; neither awaits `onDidTerminateDebugSession`. The reuse test can therefore start its second launch while the first session still owns the terminal. Please create and await a termination promise for each launch, with failure cleanup as in the existing env-precedence test.

sean-mcmanus:
@8prashant I see some older unresolved threads -- can you resolve those if you already resolved them (I assume you did).

8prashant:
Hi @sean-mcmanus,

Thanks again for the detailed review. I’ve pushed updates to address the remaining Copilot review feedback from https://github.com/microsoft/vscode-cpptools/pull/14691#pullrequestreview-5117191993.

Summary of the latest fixes:
- Updated Run Without Debugging terminal reuse so only terminals tracked by the adapter module are reused or disposed, addressing `discussion_r3937293282`.
- Preserved `null` tombstones for `TerminalOptions.env` in the integrated-terminal path while keeping deleted keys omitted for spawned process environments.
- Fixed Windows case-insensitive env replacement/deletion for both `environment` and `env`.
- Aligned `OptionsSchema.json` and generated `package.json` so `env` and environment values support `null` consistently.
- Updated localization text to document the `null` removal behavior.
- Tightened the Run Without Debugging tests so each scenario waits for session termination and cleans up on failure.

Could you please re-review when you have a chance?

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -13,6 +13,21 @@ import { isWindows } from '../constants';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize = nls.loadMessageBundle();
+type TerminalEnvironment = NonNullable<vscode.TerminalOptions['env']>;
+const managedTerminals = new Map<string, vscode.Terminal>();
```
sean-mcmanus, line 17: ✨Copilot (agent117): [Minor] Closed terminals remain strongly referenced here until the same executable is launched again. Since each distinct program adds an entry and the per-session close listener is disposed when the session terminates, manually closing retained terminals can accumulate stale `Terminal` handles for the extension-host lifetime. Please install module-lifetime `onDidCloseTerminal` cleanup that deletes an entry only when its value is the closed terminal.

sean-mcmanus (commented):
Re-reviewed the current head; see the inline comment.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -700,6 +704,35 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        for (const entry of environment) {
+            if (util.isString(entry?.name) && (util.isString(entry?.value) || entry?.value === null)) {
+                mergedEnvironment.set(entry.name, entry.value);
+            }
+        }
+
+        for (const [name, value] of Object.entries(envObject)) {
+            if (util.isString(value) || value === null) {
+                mergedEnvironment.set(name, value);
+            }
```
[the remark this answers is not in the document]
sean-mcmanus: ✨Copilot (agent117): Confirmed on `dc91405e`: this commit now permits nullable `environment[].value`, but it does not change the normal adapter. MIEngine passes `EnvironmentEntry.Value` to `GetSetEnvironmentVariableCommand(string name, string value)`; GDB formats `set env NAME VALUE`, and LLDB calls `EscapeQuotes(value)`. A null therefore still cannot express deletion in normal `cppdbg`. Please keep this thread active until that path is implemented or nullable syntax is limited to supported paths, with a normal-debug test.

8prashant:
Hi @sean-mcmanus , Could you please re-review when you have a chance?

Extension/src/Debugger/runWithoutDebuggingAdapter.ts
```diff
@@ -208,10 +253,60 @@ export class RunWithoutDebuggingAdapter implements vscode.DebugAdapter {
...
+    private buildMacOSExternalTerminalCommand(cmdLine: string, env: TerminalEnvironment): string {
+        const envArgs = Object.entries(env).flatMap(([name, value]) => {
+            if (value === null) {
+                return ['-u', this.escapeShellArg(name)];
+            }
+
+            return value === undefined ? [] : [this.escapeShellArg(`${name}=${value}`)];
+        });
+
+        return envArgs.length === 0 ? cmdLine : `/usr/bin/env ${envArgs.join(' ')} ${cmdLine}`;
```
sean-mcmanus, line 265: ✨Copilot (agent117): [Moderate] macOS `env(1)` recognizes `-u` only before any `name=value` operands, but this preserves the object's insertion order. The added test builds `QUOTE_VAR=…` before `-u REMOVE_ME`, so `/usr/bin/env` treats `-u` as the utility to execute and the target never launches. Collect all `-u <name>` options before appending assignments (or remove the unsupported nullable path), and assert the resulting operand order rather than only checking substrings.

sean-mcmanus (commented):
Re-reviewed the current head; see the inline comment.

Extension/src/Debugger/configurationProvider.ts
```diff
@@ -700,6 +704,36 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
+        }
+
+        const envObject = config.env;
+        if (!util.isObject(envObject)) {
+            return;
+        }
+
+        const environment: Environment[] = util.isArray(config.environment) ? config.environment : [];
+        const mergedEnvironment = new Map<string, Environment>();
+        const getEnvironmentKey = (name: string): string => isWindows ? name.toLowerCase() : name;
```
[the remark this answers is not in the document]
sean-mcmanus: ✨Copilot (agent117): The common explicit remote fields are handled now, but this still assumes a local Windows target when a supported remote connection is established through `setupCommands` or `customLaunchSetupCommands` alone (for example, `target remote localhost:1234`). In that configuration `PATH` and `Path` are still collapsed before reaching the case-sensitive target. Please treat command-driven/unknown targets conservatively, and extend the remote test beyond `pipeTransport`.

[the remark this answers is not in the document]
sean-mcmanus: ✨Copilot (agent117): The remote examples are fixed, but this now overclassifies every nonempty `customLaunchSetupCommands` array. The documented local example is `target-run`; on Windows, adding that command makes `PATH` and `Path` remain separate instead of applying `env` precedence. Please use `hasRemoteCommands(config.customLaunchSetupCommands)` as you do for `setupCommands`, and add a local `target-run` case alongside the remote test.

sean-mcmanus:
@8prashant I'll try to get to this...

PR #14711, opened 2026-08-27
Use native file type mappings for TypeScript-side classification

Colengms:
## Summary

Use the effective mappings published by the native language server as the primary source of truth for TypeScript file classification.

This prevents hardcoded extension lists from drifting from native behavior and avoids forwarding changes for non-C/C++ files to the browse database.

`package.json` only augments VS Code’s language-association registry. VS Code’s built-in C++ language contribution already registers the standard extensions, so this package need only add associations that are missing or that cpptools wants to guarantee independently.

## Changes

- Add an atomic file type registry for extension and exact-filename mappings.
- Retain bootstrap defaults for startup and compatibility with older native binaries.
- Prefer exact filename mappings over extension mappings.
- Support native extensionless headers and newly recognized extensions such as `.cppm`.
- Preserve the existing uppercase `.C` compatibility behavior.
- Use the active document language for otherwise unregistered paths.
- Route common source/header predicates, build/debug checks, and file-watcher filtering through the registry.
- Refresh active-file state and invalidate cached build tasks when mappings change.
- Add focused tests for bootstrap behavior, native replacement, exact mappings, extensionless headers, watcher classification, and language fallback.

## Validation

- `yarn compile`
- 3 focused Mocha tests
- Focused ESLint
- `git diff --check`

Depends on the corresponding native protocol PR. The extension remains compatible with older native binaries through its bootstrap mappings.


Extension/src/fileType.ts
```diff
@@ -0,0 +1,117 @@
...
+    filenames: []
+};
+
+let extensionMappings: ReadonlyMap<string, FileTypeMapping>;
+let filenameMappings: ReadonlyMap<string, FileTypeMapping>;
+let nativeMappingsAvailable: boolean = false;
+let caseSensitiveFileSupport: boolean = !isWindows;
+
+function getMappingKey(name: string): string {
+    return caseSensitiveFileSupport ? name : name.toLowerCase();
```
sean-mcmanus, line 41: ✨Copilot (agent57): [Moderate] Native-provided mappings use case-insensitive association matching, but this key preserves path case whenever case-sensitive file support is enabled (always on Linux/macOS). With a native `.hpp` mapping, `FOO.HPP` misses it and the active `cpp` language fallback classifies the header as a source file; an exact `vector` mapping likewise turns `VECTOR` into a source. This can offer build/debug tasks for headers and miss mapped closed-file watcher updates. Please normalize native mapping keys/lookups independently of filesystem case sensitivity, while keeping the explicit `.C` check ahead of extension lookup, and update the native-mapping test that currently expects these misses.

sean-mcmanus (commented):
Reviewed together with the corresponding native protocol change.

Extension/src/fileType.ts
Colengms: (copilot)
Good catch. VS Code applies file associations case-insensitively regardless of platform or filesystem mode. I removed the `caseSensitiveFileSupport` coupling and now always normalize mapping keys and lookups. The explicit `.C` C++ check remains ahead of normalized extension lookup. Tests now cover uppercase native extension and exact-filename mappings with active-language fallback.

sean-mcmanus: ✨Copilot (agent57): Verified the fix at [f4bbb630](https://github.com/microsoft/vscode-cpptools/commit/f4bbb630ed3c19d42ce6966edc33eeaf63c7a09a) in a clean merge with current `main`. The extension compiles, all three focused file-type tests pass, and direct checks now classify `FOO.HPP` and `VECTOR` as headers while preserving uppercase `.C` as C++. This addresses the finding.

sean-mcmanus (approved):

PR #14719, opened 2026-08-30
Implement session state tracking for "Run and Debug" button when Inte…

### LINKED ISSUE #13001: Disabling intellisense disables debugging
### Environment

- OS and Version: MacOS Sequoia 15.1.1
- VS Code Version: 1.95.3
- C/C++ Extension Version: 1.22.11
- If using SSH remote, specify OS of remote machine:  N/A

Bug Summary:

Disabliing C_Cpp.intelliSenseEngine, removes the ability to run files with the extensions debugger.

Steps to reproduce:
1. set C_Cpp.intelliSenseEngine to disabled
2. reload vscode

Expected behavior:

The run button should not disappear and I should be able to use the extensions debugging features whilst having intellisense disabled.

### Configuration and Logs

```shell
* N/A
* Command does not exist
```

### Other Extensions

_No response_

### Additional context

_No response_

8prashant:
Fixes #13001

When `C_Cpp.intelliSenseEngine` is set to `"disabled"`, the language server client is not started. Previously, `cpptools.buildAndDebug.isSourceFile` and `cpptools.buildAndDebug.isFolderOpen` context keys (which control the visibility of the "Run / Debug C/C++ File" and "Add Debug Configuration" editor action buttons) were only updated within the language server client (`client.ts`). Consequently, disabling IntelliSense would cause the Run/Debug shortcut buttons to never appear in the editor title bar.

This PR updates the Debugger extension initialization (`Debugger/extension.ts`) to track active text editor changes and keep `buildAndDebugIsSourceFile` and `buildAndDebugIsFolderOpen` up to date regardless of whether IntelliSense is enabled or disabled.

### Changes
1. Added `updateBuildAndDebugSessionState` and registered an `onDidChangeActiveTextEditor` listener in `DebuggerExtension.initialize`.
2. Added unit tests in `Extension/test/scenarios/SingleRootProject/tests/buildAndDebug.test.ts` verifying session state updates for source files, header files, non-C/C++ files, and `undefined` editors.

### Verification
- Compiled with TypeScript (`yarn compile`) with 0 errors.
- Ran tests via `yarn test --scenario=SingleRootProject` and verified all `BuildAndDebug SessionState Tests` pass.


Extension/src/Debugger/extension.ts
```diff
@@ -129,6 +130,20 @@ export async function initialize(context: vscode.ExtensionContext): Promise<void
...
+    // Track active editor changes so "Run and Debug" button session state is kept up to date
+    // even if the language server (IntelliSense) is disabled.
+    updateBuildAndDebugSessionState(vscode.window.activeTextEditor);
+    disposables.push(vscode.window.onDidChangeActiveTextEditor(editor => updateBuildAndDebugSessionState(editor)));
+}
+
+export function updateBuildAndDebugSessionState(editor?: vscode.TextEditor): void {
+    if (editor && isCpp(editor.document)) {
+        void SessionState.buildAndDebugIsSourceFile.set(isCppOrCFile(editor.document.uri, editor.document.languageId));
+        void SessionState.buildAndDebugIsFolderOpen.set(isFolderOpen(editor.document.uri));
```
sean-mcmanus: ✨Copilot (agent146): [Moderate] Please keep the folder context key current on every transition that can change `isFolderOpen`. This listener recomputes it only when the active editor is C/C++, and there is no debugger-side `onDidChangeWorkspaceFolders` listener. With IntelliSense disabled, opening an in-workspace C++ file sets the key to `true`; switching to a non-C/C++ editor leaves it `true`, so the editor context menu still offers `C_Cpp.AddDebugConfiguration` because that contribution checks only `cpptools.buildAndDebug.isFolderOpen`. Likewise, adding or removing the active file's workspace folder leaves the key stale until another editor change. Clear/recompute this key for non-C++/undefined editors and recompute it when workspace folders change; the tests should assert the folder key for those transitions as well.

sean-mcmanus (commented):
Found one session-state lifecycle gap that should be addressed before merging.

Extension/src/Debugger/extension.ts
sean-mcmanus: ✨Copilot (agent146): Thanks for adding the workspace-folder listener and folder-state tests. One correction to my original comment: the contribution gated only by this key is the Command Palette entry, not the editor context menu. The latest implementation still leaves the active-file case incorrect, though: if an unrelated workspace folder is open while the active C++ source is outside it, `workspaceFolders !== undefined` sets this key to `true`, so the editor title/context actions are shown even though `AddDebugConfiguration` immediately warns/returns because `getWorkspaceFolder(textEditor.document.uri)` is undefined. Removing the active file's folder from a multi-root workspace while another root remains has the same result. It also makes enabled and disabled IntelliSense disagree because `LanguageServer/client.ts` still writes the URI-specific `isFolderOpen(editor.document.uri)` value. Please preserve active-URI membership for the editor actions and use a separate context key if the Command Palette needs global workspace existence; add an external-file or multi-root-removal test for this distinction.

8prashant:
Hi @sean-mcmanus,
I have addressed all the review feedback
Could you please re-review the updated changes?
Thanks.

Extension/test/scenarios/SingleRootProject/tests/buildAndDebug.test.ts
sean-mcmanus, line 1: ✨Copilot (agent146): [Minor] Please add focused regression coverage for the event-driven behavior rather than testing only the exported helper. These four tests call `updateBuildAndDebugSessionState` directly, so they would still pass if any of the new `onDidChangeActiveTextEditor`, `onDidOpenTextDocument`, or `onDidChangeWorkspaceFolders` registrations were absent. They also never use a C++ URI outside the open workspace, so they would have passed the prior global-folder implementation as well. At minimum, add an external C++ source case that asserts `isSourceFile === true` and `isFolderOpen === false`, plus one listener-driven transition (ideally an active-document language change while IntelliSense is disabled) to cover the behavior this PR fixes.

sean-mcmanus (commented):
The correctness fixes now look sound. One focused test-coverage gap remains for the new event-driven paths.

Extension/test/scenarios/SingleRootProject/tests/buildAndDebug.test.ts
sean-mcmanus: @8prashant FYI, my agents can add the regression coverage for you in a follow up PR if you do not want to....however, I don't think my GitHub agents have those instructions so I might have to directly inform them.

Also, I've been really busy -- I'll try to review your PRs as soon as I can.

sean-mcmanus:
@8prashant FYI, the full Run/Debug feature doesn't work since it relies on compiler info that is returned from the cpptools process, but if you have your tasks.json/launch.json already generated (with the IntelliSense engine enabled for example), then it works.

Extension/test/scenarios/SingleRootProject/tests/buildAndDebug.test.ts
sean-mcmanus: @8prashant My agent will just add this so we can check in now for 1.34.3.

sean-mcmanus (approved):

8prashant:
Hi @sean-mcmanus, thanks for reviewing the PR and for adding the regression coverage. I have confirmed that the requested changes are included, and the latest checks are clean.

The PR is approved, but I do not have the merge permission for the upstream repository, so the Merge button is not available to me. Could you please merge PR #14719 when you have a chance?

Thanks!

PR #14725, opened 2026-09-02
Fix custom configuration provider regression

### LINKED ISSUE #14724: 1.33.5 broke custom configuration providers
### Environment

- OS and Version: macOS 26.6.2
- VS Code Version:  1.134.0
- C/C++ Extension Version: 1.33.5

Bug Summary: When using a custom configuration provider, the compiler listed by the configuration provider is no longer used after 1.33.5. Instead, the host compiler is always used. Additionally, include paths are not picked up from the configuration either

Steps to reproduce:
Create a custom configuration provider. Have some include paths that are in folders outside of the workspace. Try to open a file using those include paths, and see that 

[1_33_4Output.txt](https://github.com/user-attachments/files/31713461/1_33_4Output.txt)
[1_33_5Output.txt](https://github.com/user-attachments/files/31713462/1_33_5Output.txt)

Expected behavior:

### Configuration and Logs

```shell
No c_cpp_properties.json, since we're explicitly using the configuration provider, provided in settings.json
```

### Other Extensions

WPILib extension, which provides the configuration provider.

### Additional context

Our provider can be found here

https://github.com/wpilibsuite/vscode-wpilib/tree/main/vscode-wpilib/src/cppprovider

sean-mcmanus:
## Summary

Restore `canProvideConfiguration` calls for each requested URI before invoking the batched `provideConfigurations` request. This preserves provider initialization and filtering behavior while retaining the multi-file batching introduced in #14582.

Add integration coverage for a provider that prepares its file configuration during `canProvideConfiguration`, and run the containing `SimpleCppProject` scenario in CI on Windows, Linux, and macOS.

Fixes #14724.

> This PR was investigated and created by Copilot with `GPT-5.6 Sol` (in VS Code). Any message starting with `✨Copilot:` was sent by Copilot.

### DIFF
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2774,7 +2774,7 @@ export class DefaultClient implements Client {
                 const isTrackedFile: boolean = hasNativeFileTypeMappings()
                     ? isTagParsableFile(uri.fsPath)
                     : isTagParsableFile(uri.fsPath) ||
-                        (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
+                    (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
                 if (isTrackedFile) {
                     // VS Code has a bug that causes onDidChange events to happen to files that aren't changed,
                     // which causes a large backlog of "files to parse" to accumulate.
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2316,7 +2316,7 @@ export class DefaultClient implements Client {
                 supportedUris.push(uri);
             }
             if (supportedUris.length === 0) {
-                return [];
+                return undefined;
             }
 
             let configs: util.Mutable<SourceFileConfigurationItem>[] = [];
--- a/Extension/src/LanguageServer/client.ts
+++ b/Extension/src/LanguageServer/client.ts
@@ -2304,9 +2304,24 @@ export class DefaultClient implements Client {
 
         // Wrap the provider lookup in a single task, so we can apply a timeout to the entire duration.
         const provideConfigurationAsync: () => Thenable<SourceFileConfigurationItem[] | undefined> = async () => {
+            const supportedUris: vscode.Uri[] = [];
+            for (const uri of docUris) {
+                try {
+                    if (!await provider.canProvideConfiguration(uri, tokenSource.token)) {
+                        continue;
+                    }
+                } catch {
+                    console.warn("Caught exception from canProvideConfiguration");
+                }
+                supportedUris.push(uri);
+            }
+            if (supportedUris.length === 0) {
+                return [];
+            }
+
             let configs: util.Mutable<SourceFileConfigurationItem>[] = [];
             try {
-                configs = await provider.provideConfigurations(docUris, tokenSource.token);
+                configs = await provider.provideConfigurations(supportedUris, tokenSource.token);
             } catch {
                 console.warn("Caught exception from provideConfigurations");
             }
--- a/Extension/test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts
+++ b/Extension/test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts
@@ -133,6 +133,8 @@ async function changeCppProperties(cppProperties: config.ConfigurationJson, _dis
 suite("extensibility tests v3", function(): void {
     let cpptools: apit.CppToolsTestApi;
     let lastResult: api.SourceFileConfigurationItem[];
+    let configurationProvidedBeforeCanProvide: boolean = false;
+    const supportedUris: Set<string> = new Set<string>();
     const defaultConfig: api.SourceFileConfiguration = {
         includePath: [ "${workspaceFolder}", "/v3/folder" ],
         defines: [ "${workspaceFolder}" ],
@@ -156,14 +158,19 @@ suite("extensibility tests v3", function(): void {
     const provider: api.CustomConfigurationProvider = {
         name: "cpptoolsTest-v3",
         extensionId: "ms-vscode.cpptools-test3",
-        canProvideConfiguration(_document: vscode.Uri): Thenable<boolean> {
+        canProvideConfiguration(document: vscode.Uri): Thenable<boolean> {
+            supportedUris.add(document.toString());
             return Promise.resolve(true);
         },
         provideConfigurations(uris: vscode.Uri[]): Thenable<api.SourceFileConfigurationItem[]> {
             const result: api.SourceFileConfigurationItem[] = [];
             uris.forEach(uri => {
+                const uriString: string = uri.toString();
+                if (!supportedUris.has(uriString)) {
+                    configurationProvidedBeforeCanProvide = true;
+                }
                 result.push({
-                    uri: uri.toString(),
+                    uri: uriString,
                     configuration: defaultConfig
                 });
             });
@@ -220,10 +227,15 @@ suite("extensibility tests v3", function(): void {
             disposables.push(testHook.IntelliSenseStatusChanged(result => {
                 result = result as apit.IntelliSenseStatus;
                 if (result.filename === "main3.cpp" && result.status === apit.Status.IntelliSenseReady) {
-                    const expected: api.SourceFileConfigurationItem[] = [ {uri: uri.toString(), configuration: defaultConfig} ];
-                    assert.deepEqual(lastResult, expected);
-                    assert.deepEqual(lastBrowseResult, defaultFolderBrowseConfig);
-                    resolve();
+                    try {
+                        const expected: api.SourceFileConfigurationItem[] = [ {uri: uri.toString(), configuration: defaultConfig} ];
+                        assert.strictEqual(configurationProvidedBeforeCanProvide, false);
+                        assert.deepEqual(lastResult, expected);
+                        assert.deepEqual(lastBrowseResult, defaultFolderBrowseConfig);
+                        resolve();
+                    } catch (error) {
+                        reject(error);
+                    }
                 }
             }));
             setTimeout(() => { reject(new Error("timeout")); }, testHelpers.defaultTimeout);
--- a/.github/workflows/job-compile-and-test.yml
+++ b/.github/workflows/job-compile-and-test.yml
@@ -97,6 +97,11 @@ jobs:
         run: yarn test --scenario=SingleRootProject
         working-directory: Extension
 
+      - name: Run SimpleCppProject tests (Windows)
+        if: ${{ inputs.platform == 'windows' }}
+        run: yarn test --scenario=SimpleCppProject
+        working-directory: Extension
+
       - name: Run E2E IntelliSense features tests (Windows)
         if: ${{ inputs.platform == 'windows' }}
         run: yarn test --scenario=MultirootDeadlockTest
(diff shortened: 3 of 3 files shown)


Extension/test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts
```diff
@@ -220,10 +227,15 @@ suite("extensibility tests v3", function(): void {
             disposables.push(testHook.IntelliSenseStatusChanged(result => {
                 result = result as apit.IntelliSenseStatus;
                 if (result.filename === "main3.cpp" && result.status === apit.Status.IntelliSenseReady) {
-                    const expected: api.SourceFileConfigurationItem[] = [ {uri: uri.toString(), configuration: defaultConfig} ];
-                    assert.deepEqual(lastResult, expected);
-                    assert.deepEqual(lastBrowseResult, defaultFolderBrowseConfig);
-                    resolve();
+                    try {
+                        const expected: api.SourceFileConfigurationItem[] = [ {uri: uri.toString(), configuration: defaultConfig} ];
+                        assert.strictEqual(configurationProvidedBeforeCanProvide, false);
```
sean-mcmanus, line 232: ✨Copilot (agent50): [Moderate] This assertion is the only automated guard against the #14582 regression returning, but the `SimpleCppProject` scenario is never run by CI. `.github/workflows/job-compile-and-test.yml` only invokes `yarn test --scenario=SingleRootProject`, `--scenario=MultirootDeadlockTest`, and `--scenario=RunWithoutDebugging`; the earlier bare `yarn test` step falls through to `unitTests()` because no scenario is specified. So a future change that drops the `canProvideConfiguration` calls again would still pass every check on the PR.

Minimal fix: either add `SimpleCppProject` to the scenario list in that workflow, or place the ordering assertion in a scenario CI already runs.

Smaller related gap: this provider's `canProvideConfiguration` always returns `true`, so the new filtering behavior (a URI being excluded from `supportedUris`) is not exercised — only "was it called at all" is.

Extension/src/LanguageServer/client.ts
```diff
@@ -2304,9 +2304,24 @@ export class DefaultClient implements Client {
 
         // Wrap the provider lookup in a single task, so we can apply a timeout to the entire duration.
         const provideConfigurationAsync: () => Thenable<SourceFileConfigurationItem[] | undefined> = async () => {
+            const supportedUris: vscode.Uri[] = [];
+            for (const uri of docUris) {
```
sean-mcmanus, line 2308: ✨Copilot (agent50): [Minor] These per-URI checks are awaited one at a time, and the caller intentionally skips `callTaskWithTimeout` when `docUris.length > 1`. That unbounded multi-file path predates this change, but until now it contained exactly one provider call; it now contains one call per file. So provider latency scales with batch size, and a single `canProvideConfiguration` that never settles stalls the whole batch — `FinishedRequestCustomConfig` is sent only from the `finally` in `provideCustomConfigurations`, so the language server never learns the batch finished.

Worth considering issuing the checks concurrently (`Promise.all` over `docUris`, then filtering by the positional results) and/or applying a bound to the multi-file path.

sean-mcmanus (commented):
✨Copilot (agent50): Reviewed source `e7240568` merged over `main` at `2f25880c` (provider merge `79b663e2`; parents confirmed).

Root cause verified rather than assumed: #14582 removed the `canProvideConfiguration` call, and the provider named in #14724 populates its `foundFiles` cache as a side effect of that call, so its `provideConfigurations` returned an empty array and IntelliSense fell back to the host compiler with no provider include paths. Restoring the call before the batched request fixes that, and the restored semantics match the pre-#14582 code exactly — including the subtle case where a thrown `canProvideConfiguration` still leaves the URI treated as supported. Filtering to `supportedUris` is safe for partial batches, since `sendCustomConfigurations` is keyed per URI.

Two non-blocking comments below. I could not run the `SimpleCppProject` scenario locally (extension dependencies and native test binaries are not deployed in this worktree), so the test change is reviewed statically; current-head CI covers compile, lint, and unit tests only.

Extension/test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts
sean-mcmanus: ✨Copilot: Addressed in the latest commit by adding `SimpleCppProject` to both the Windows and Linux/macOS CI paths, so this regression test now runs on every supported platform. The provider fixture intentionally focuses on the reported ordering regression; filtering a rejected URI would require a multi-file request and can be covered alongside future bulk-request work.

Extension/src/LanguageServer/client.ts
sean-mcmanus: ✨Copilot: I considered concurrent checks and a bulk timeout, but kept the calls sequential to preserve predictable behavior for stateful providers. The multi-file `provideConfigurations` path was already intentionally unbounded; changing its timeout or introducing concurrent provider entry would be a separate compatibility decision that needs dedicated design and tests rather than being folded into this regression fix.

Extension/test/scenarios/SimpleCppProject/tests/languageServer.integration.test.ts
sean-mcmanus: ✨Copilot (agent50): Correcting the "reviewed statically" caveat in my review body — I was able to build and run the scenario locally after all (`yarn install-and-copy-binaries-for-test` supplies the native binaries, and WSLg provides a display so `xvfb` is not needed).

- At `e7240568`: `SimpleCppProject` is green — 21 passing, including `extensibility tests v3 › Check provider - main3.cpp`.
- Probe with base `2f25880c` plus *only* this test file (no `client.ts` change): 20 passing, 1 failing — `AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: true !== false`, i.e. `configurationProvidedBeforeCanProvide`.

So the assertion is a genuine regression guard, and the `try`/`catch`/`reject` wrapper is what turns it into a clear assertion failure instead of an opaque timeout rejection. That makes the CI gap above more consequential, not less: the guard works, it just never runs automatically.

sean-mcmanus: ✨Copilot (agent50): Re-reviewed at `910bee68` (merge `67e8448d` over `main` `2f25880c`; parents confirmed). This resolves the finding, and I verified it from the Actions job API rather than the aggregate check conclusions — the new step really ran and passed on every platform:

- `windows-2025` — `Run SimpleCppProject tests (Windows)`: success
- `ubuntu-24.04` — `Run SimpleCppProject tests (linux/macOS)`: success
- `macos-15` — `Run SimpleCppProject tests (linux/macOS)`: success

The `if:` gating is correct too: each job skips the opposite platform's step. `client.ts` and this test file are byte-identical to `e7240568`, so the earlier pass-at-head / fail-at-base evidence still applies unchanged.

Agreed on deferring the rejected-URI filtering case — it needs a genuine multi-file request to mean anything.

Extension/src/LanguageServer/client.ts
sean-mcmanus: ✨Copilot (agent50): Conceding — your call is right, and worth recording why for the next reader: the provider in #14724 mutates shared state inside `canProvideConfiguration` (`findMatchingBinary` appends to `foundFiles`), so making these concurrent would be a provider-visible behavior change, not a free optimization. Agreed it does not belong in a regression fix.

Extension/src/LanguageServer/client.ts
```diff
@@ -2304,9 +2304,24 @@ export class DefaultClient implements Client {
...
+                    if (!await provider.canProvideConfiguration(uri, tokenSource.token)) {
+                        continue;
+                    }
+                } catch {
+                    console.warn("Caught exception from canProvideConfiguration");
+                }
+                supportedUris.push(uri);
+            }
+            if (supportedUris.length === 0) {
+                return [];
```
bobbrow: returning undefined seems more appropriate and matches line 2395, but this also works.

bobbrow (dismissed):
One optional comment

Extension/src/LanguageServer/client.ts
sean-mcmanus: ✨Copilot: Addressed in the latest commit by returning `undefined`, matching the existing no-configuration sentinel. TypeScript compilation, focused ESLint, and the full `SimpleCppProject` scenario (21 passing) succeed with the change.

bobbrow (dismissed):

bobbrow (approved):

Extension/src/LanguageServer/client.ts
```diff
@@ -2759,7 +2774,7 @@ export class DefaultClient implements Client {
                 const isTrackedFile: boolean = hasNativeFileTypeMappings()
                     ? isTagParsableFile(uri.fsPath)
                     : isTagParsableFile(uri.fsPath) ||
-                        (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
+                    (ext !== undefined && this.associations_for_did_change?.has(ext.toLowerCase()) === true);
```
sean-mcmanus, line 2777: ✨Copilot (agent50): [Nit] Non-blocking scope note: this hunk is whitespace-only and unrelated to the fix. The line comes from #14711's `hasNativeFileTypeMappings` code, so `client.ts` now carries a second hunk against `main` that has nothing to do with `canProvideConfiguration`.

It also was not lint-required — `Run Linter` passed on all three platforms at `75a3078c` with the original indentation, and passes again here, so ESLint accepts both forms.

Only worth a moment's thought because the approval on this PR was given at `75424b8a`, two commits before this one. Your call; nothing here blocks the merge.

sean-mcmanus (commented):
✨Copilot (agent50): Re-reviewed at `3e8d1092` (merge `71576369` over `main` `2857a071`; parents confirmed). The delta since `75a3078c` is whitespace-only — `git diff -w` between the two revisions is empty — so there is no behavioral change to re-validate.

Current-head CI is fully green on all three platforms, and I confirmed the individual steps rather than the job conclusions: `Compile Sources`, `Run Linter`, `Run unit tests`, and `Run SimpleCppProject tests` all succeeded on `windows-2025`, `ubuntu-24.04`, and `macos-15`. That is stronger evidence than my earlier single-platform local runs, and it means the regression guard added here now demonstrably runs and passes everywhere.

One non-blocking nit below about the scope of the new hunk. The fix itself remains correct and, from my side, ready to merge.

Extension/src/LanguageServer/client.ts
sean-mcmanus: ✨Copilot: This whitespace-only hunk was explicitly requested by the PR author after reviewing the current file. It is isolated in commit `3e8d1092`, and current-head compile, lint, unit, and `SimpleCppProject` checks pass on all three platforms. No further change is needed.

sean-mcmanus: ✨Copilot (agent50): Withdrawing my inference — you're right, and my reasoning was the flawed part. I used "lint passes both ways" to imply the change was unnecessary, but `yarn lint` does not enforce this repo's pre-check-in formatting step, so a green linter says nothing about whether a file was correctly formatted.

Checking the surrounding convention confirms the new form: multi-line `||`/`&&` continuations in this file sit one level in from the statement, not two — see lines 1841-1842, 2354-2355, 3041-3042, 3384-3390, and 3462-3466. The previous double indent was the outlier, so this is a genuine pre-existing formatting fix rather than added scope. No objection.

PR #14729, opened 2026-09-02
Remove unused gulp-sourcemaps dependency

sean-mcmanus:
## Summary

Remove the unused `gulp-sourcemaps` initialization from the localization pipelines and drop the dependency. This removes the vulnerable transitive `decode-uri-component@0.2.2` package while preserving localization behavior because the metadata generator does not consume source maps.

## Validation

- `yarn install --frozen-lockfile --ignore-scripts`
- `yarn compile`
- `yarn translations-generate`
- `yarn gulp translations-export`

> This PR was investigated and created by GitHub Copilot in VS Code. Any message starting with `✨Copilot:` was sent by Copilot.

### DIFF
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -7262,8 +7262,6 @@
         "which": "^2.0.2"
     },
     "resolutions": {
-        "postcss": "^8.5.23",
-        "nanoid": "^3.3.18",
         "gulp-typescript/**/glob-parent": "^5.1.2",
         "serialize-javascript": "^7.0.5",
         "@azure/msal-browser": "^4.29.1",
--- a/Extension/package.json
+++ b/Extension/package.json
@@ -7221,7 +7221,6 @@
         "gulp": "^5.0.0",
         "gulp-env": "^0.4.0",
         "gulp-filter": "^7.0.0",
-        "gulp-sourcemaps": "^3.0.0",
         "gulp-typescript": "^5.0.1",
         "minimist": "^1.2.8",
         "mocha": "^10.4.0",
--- a/Extension/gulpfile.js
+++ b/Extension/gulpfile.js
@@ -12,7 +12,6 @@ const nls = require('vscode-nls-dev');
 const path = require('path');
 const minimist = require('minimist');
 const es = require('event-stream');
-const sourcemaps = require('gulp-sourcemaps');
 const ts = require('gulp-typescript');
 const typescript = require('typescript');
 const tsProject = ts.createProject('./tsconfig.json', { typescript });
@@ -216,7 +215,6 @@ gulp.task("translations-export", (done) => {
 
     // Transpile the TS to JS, and let vscode-nls-dev scan the files for calls to localize.
     let jsStream = tsProject.src()
-        .pipe(sourcemaps.init())
         .pipe(tsProject()).js
         .pipe(nls.createMetaDataFiles());
 
@@ -308,7 +306,6 @@ const generateAdditionalLocFiles = () => {
 const generateSrcLocBundle = () => {
     // Transpile the TS to JS, and let vscode-nls-dev scan the files for calls to localize.
     return tsProject.src()
-        .pipe(sourcemaps.init())
         .pipe(tsProject()).js
         .pipe(nls.createMetaDataFiles())
         .pipe(nls.createAdditionalLanguageFiles(languages, "i18n"))
--- a/Extension/yarn.lock
+++ b/Extension/yarn.lock
@@ -4323,11 +4323,6 @@ mute-stream@~0.0.4:
   resolved "https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/mute-stream/-/mute-stream-0.0.8.tgz#1630c42b2251ff81e2a283de96a5497ea92e5e0d"
   integrity "sha1-FjDEKyJR/4HiooPelqVJfqkuXg0= sha512-nnbWWOkoWyUsTjKrhgD0dcz22mdkSnpYqbEjIm2nhwhuxlSkpywJmBo8h0ZqJdkp73mb90SssHkN4rsRaBAfAA=="
 
-nanoid@^3.3.16, nanoid@^3.3.18:
-  version "3.3.18"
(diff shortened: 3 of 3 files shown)


Extension/package.json
```diff
@@ -7221,7 +7221,6 @@
         "gulp": "^5.0.0",
         "gulp-env": "^0.4.0",
         "gulp-filter": "^7.0.0",
-        "gulp-sourcemaps": "^3.0.0",
```
sean-mcmanus, line 7224: ✨Copilot (agent165): [Minor] Removing `gulp-sourcemaps` also orphans the `postcss` and `nanoid` entries in `resolutions` (lines 7265-7266).

`postcss` entered the graph only through `@gulp-sourcemaps/identity-map@2.0.1`'s `postcss@^7.0.16`, which `resolutions.postcss` upgraded to 8.5.25, and `postcss` was in turn the only consumer of `nanoid` and `source-map-js`. After this change nothing in the graph requires any of the three: reachability analysis from `package.json` finds 0 unreachable `yarn.lock` entries at the base commit and 3 at this head (`postcss@8.5.25`, `nanoid@3.3.18`, `source-map-js@1.2.1`), and `yarn install --frozen-lockfile` no longer creates `node_modules/postcss`, `node_modules/nanoid`, or `node_modules/source-map-js`.

They survive in the lockfile because `.scripts/verifyYarnLock.mjs` requires a lock selector for every `resolutions` pattern, so `yarn verify-yarn-lock` keeps them pinned. That leaves the lockfile advertising three versions that are never installed, which is the same lockfile-only reporting surface this PR is trying to shrink. Consider dropping the now-unused `postcss` and `nanoid` resolutions in this PR and regenerating, so the three entries go away with them.

sean-mcmanus (commented):
✨Copilot (agent165): Reviewed at `a7bd689` against base `2f25880`, with local verification.

**The core change is correct.** `vscode-nls-dev@4.0.4`'s `createMetaDataFiles()` calls `processFile(file.contents.toString('utf8'))` with no source-map argument, so the maps were never consumed — only `rewriteLocalizeCalls()` reads `file.sourceMap`, and this gulpfile does not use it. The transpiled JS is also filtered out of both pipelines before `gulp.dest`, so it never reaches disk.

**Empirical A/B:** with `node_modules` installed from the base lockfile and only `gulpfile.js` swapped, `gulp translations-generate` and `gulp translations-export` produce byte-identical output (180 files under `dist/`, and the 312,756-byte `vscode-cpptools.xlf`). Re-running both at this head after `yarn install --frozen-lockfile` reproduces the same outputs. `yarn test-yarn-lock` and `yarn verify-yarn-lock` pass, and `gulp-sourcemaps`, `source-map-resolve`, and `decode-uri-component` are gone from `node_modules`.

**No debugging regression:** `gulp-typescript` already forces `inlineSourceMap: false` / `sourceMap: true` at project creation and then sets `sourceMap` from whether inputs carry a map, so this only stops an unused map from being generated. Shipped source maps still come from `tsc --build` (`inlineSourceMap`) and webpack (`devtool: 'source-map'`).

**Security rationale checks out:** `decode-uri-component@0.2.2` is still covered by GHSA-vcc3-ghjq-m6fr (`<= 0.4.2`, patched in 0.5.0), so calling it vulnerable is accurate even though it postdates the older GHSA-w573-4hg7-7wgq fix.

One [Minor] follow-through comment on the leftover `resolutions` entries. Build-time-only change with no product-code impact; low release risk.

Extension/package.json
sean-mcmanus: ✨Copilot: Fixed. The `postcss` and `nanoid` resolution pins were removed and the lockfile was regenerated, which also removed the unreachable `postcss`, `nanoid`, and `source-map-js` entries. `yarn install --frozen-lockfile --ignore-scripts`, `yarn test-yarn-lock`, and `yarn verify-yarn-lock` pass.

sean-mcmanus: ✨Copilot (agent165): Verified at `368f0b34`; agreeing with the resolution.

Both resolution pins are gone and the incremental lockfile delta from `a7bd689` is exactly the three orphaned entries and nothing else. Reachability from `package.json` is now 903 entries / 903 reachable / 0 orphans (was 3), and every remaining `resolutions` pattern still has a lock selector, so `verifyYarnLock` stays satisfied. `picocolors` was correctly kept — it still has two other dependents besides `postcss`.

Re-validated the whole change at this head rather than just the delta: `yarn install --frozen-lockfile --ignore-scripts` succeeds, `yarn test-yarn-lock` and `yarn verify-yarn-lock` pass, and `gulp-sourcemaps`, `source-map-resolve`, `decode-uri-component`, `postcss`, `nanoid`, and `source-map-js` are all absent from `node_modules`. Re-ran the base-versus-head A/B with identical generated inputs on both sides: `gulp translations-generate` (180 files) and `gulp translations-export` (313,156-byte `vscode-cpptools.xlf`) are still byte-identical to the base commit. All three `job / build` legs are green at this head.

No remaining findings from me.

sean-mcmanus (commented):
✨Copilot (agent165): Re-reviewed at `363428dc` after the two merges from `main`. No findings; nothing changed that affects my earlier conclusions.

The two new commits are clean merges of `main` with no conflict resolutions — the combined diff of each is empty. The PR's three files are byte-identical to the revision I fully validated at `368f0b34` (same blob SHAs for `Extension/gulpfile.js`, `Extension/package.json`, and `Extension/yarn.lock`), and against the advanced base `9e4a5024` the head still differs in exactly those three files and nothing else.

The target advance is `Remove ignored network isolation policy` (#14728) and `Bump fast-uri from 3.1.5 to 3.1.6 in /ExtensionPack` (#14731), which touch only `Build/*.yml` and `ExtensionPack/package-lock.json`. `Extension/` is untouched between the old and new base, so there is no interaction with this change — including with the `fast-uri` resolution here, which governs a different manifest.

Re-validated the merged result rather than assuming it: `yarn install --frozen-lockfile --ignore-scripts` succeeds, `yarn test-yarn-lock` and `yarn verify-yarn-lock` pass, lockfile reachability is 903/903 with 0 orphans, every `resolutions` pattern still has a lock selector, and `gulp-sourcemaps`, `source-map-resolve`, `decode-uri-component`, `postcss`, `nanoid`, and `source-map-js` remain absent from `node_modules` while `picocolors` is correctly retained. I did not re-run the localization A/B, because the pipeline inputs are provably byte-identical to the revision where it already produced byte-identical output. All three `job / build` legs are green at this head.

The branch is up to date with `main`, so `mergeable_state: blocked` reflects the pending required approval rather than anything in the change.

bobbrow (approved):

PR #14735, opened 2026-09-03
Fix fast-uri dependency.

sean-mcmanus:


Extension/package.json
```diff
@@ -7271,6 +7271,7 @@
         "flatted": "^3.4.1",
         "uuid": "^11.1.1",
         "undici": "^7.29.0",
-        "fast-uri": "^3.1.5"
+        "fast-uri": "^3.1.6",
+        "@humanfs/node": "^0.16.8"
```
bobbrow, line 7272: Why is this added for this package.json but not the other one?

sean-mcmanus: @bobbrow What are you talking about?

sean-mcmanus: It's only an issue for the Extrension and Themes.

bobbrow: I'm talking about `@humanfs/node`

sean-mcmanus: ExtensionPack already uses 3.1.6.

sean-mcmanus: We only got a dependabot warning for Extension.

sean-mcmanus: This PR is 2 separate issues in 1.

bobbrow: Ok. The title just said fast-uri, so I wondered where this one came from.

bobbrow (approved):

PR #11262, opened 2023-07-28
Test restructuring and cleanup

fearthecowboy:
In This PR:
- pulled in the scenarios from the `Vcls-vscode-test` repository and integrates them into this repository to make running the tests as streamlined as possible.
- removes use of webpack for day-to-day coding, instead direct use of the typescript compiler which offers a streamlined work process.
- created a `.scripts` folder to make it simpler to add more repository scripts/actions, and use `ts-node` to allow direct execution of `.ts` script files
- refactored out some of the gulp tasks into simpler `.ts` scripts
- added `readme.developer.md` markdown file with instructions and descriptions of all the available `yarn <command>`  scripts  
  see: [readme.developer.md](https://github.com/fearthecowboy/vscode-cpptools/blob/test-restructuring/Extension/readme.developer.md)

- removed use of an `out` folder, and have all compiling go into `dist` when using the tsc compiler
- adjusted the output of webpack to go to `dist/src` so that the paths are the same regardless of using tsc or webpack 

- notable added/modified yarn commands:
  - `yarn clean [all] [reset]`  - cleans files from the repository directory
  - `yarn test [--scenario=...] [all] `  - command line for running unit tests and scenario tests in an isolated environment
  - `yarn code` - command to run an isolated vscode instance with the extension
  - `yarn compile` and `yarn watch` - compile the code to the `dist` folder (not using webpack)
  - `yarn webpack` compile the extension using webpack.

 - removed many debug configurations, and streamlined it to three primary configurations. 
  - F5 with "Launch Extension" will prompt the user to select the workspace and launch the extension with the debugger 
  - F5 with "VSCode Tests" will prompt the user to select the scenario to run the tests for.
  - Debugging unit tests is handles thru the VSCode test explorer (if you have mocha test explorer installed) 
  
  - enabled linting on `ui`, `src`, 'test` and `.scripts` typescript files 
  
  - Added some of the reusable Utility code (most of it is what is needed for the next PR with compiler querying, but I needed much of it now)

Submitting as Draft - I still want to 
 - verify any impact to CI build scripts for building the extension binaries
 - verify impact of running tests on linux/osx
 - verify that the CI tests can be run 
 
 TODO in a future PR:
 - Add more `.scripts` to handle acquiring native binary components from 
   - local private build 
   - existing published vscode binary
   - a specific artifact from a build in ADO 

  - make a script to make adding a new scenario super easy
  - clean up remaining gulp scripts
  - maybe add more error checking in `.scripts`

PR #14743, opened 2026-09-05
Fix localization string import

sean-mcmanus:
## Summary
Skip unrelated source directories during EDGE string import and refresh the translated diagnostic for excessive nested user-defined conversions in 13 locales.

## User impact
[Minor] A localization refresh can silently stop before processing supported locales when the source checkout contains an unrelated directory. A later extension release can consequently retain stale or missing localized IntelliSense diagnostics until the import is rerun from a different checkout layout.

## Root cause
The importer added in #11262 returned from the entire operation when it encountered the first directory outside its fixed locale table. A standalone Git checkout exposes `.git` as a directory, so the command could exit successfully before importing any locale.

## Fix
Continue past unrecognized directories instead of ending the import. The generated update adds the newly available translations while preserving the reviewed corrections from #14676.


Extension/bin/messages/cs/messages.json
```diff
@@ -2999,7 +2999,7 @@
     "Přiřazení k anachronizmu this není kompatibilní s režimem C++11.",
     "Došlo k neznámé chybě. Soubor PCH IntelliSense se nevygeneroval.",
     "Nejméně jednu ze vzdálených cest nejde namapovat na stávající místní cesty. Funkce IntelliSense možná nebude fungovat zcela podle očekávání.",
-    "too many nested user-defined conversions (possible cycle of converting routines)",
+    "Příliš mnoho vnořených uživatelsky definovaných převodů (možný cyklus převodu rutin)",
```
sean-mcmanus: ✨Copilot (agent40): [Minor] The parenthetical reverses the relationship in the English source: `cyklus převodu rutin` says that the routines themselves are being converted, while the diagnostic means that conversion routines form a cycle. Please use wording equivalent to `možný cyklus převodních rutin`. The same semantic inversion appears in `es/messages.json` (`ciclo de conversión de rutinas`; the intended order is `ciclo de rutinas de conversión`) and `pl/messages.json` (`cykl konwertowania procedur`; the routines should be the participants in the cycle), so please correct all three generated values.

sean-mcmanus: ✨Copilot: @sean-mcmanus Addressed in the latest commit. The Czech text now says `cyklus převodních rutin`, Spanish says `ciclo de rutinas de conversión`, and Polish says `cykl procedur konwersji`, so each parenthetical describes a possible cycle formed by conversion routines.

bobbrow (approved):

Colengms (approved):

PR #14771, opened 2026-09-14
Batch native debug Output writes

sean-mcmanus:
## Summary

This PR is an attempt to work around VS Code running out of memory (OOM) when `C_Cpp.loggingLevel` is set to `7` and C/C++ generates a large volume of diagnostic output. It batches native debug Output writes to reduce per-append update pressure while preserving all localized text, line feeds, message order, logging levels, and Output-panel behavior.

## Details

- Use a 65,536 UTF-16-code-unit flush threshold and a single 50 ms timer anchored to the first buffered input. Keep oversized inputs intact and flush them synchronously.
- Flush pending native diagnostics before ordinary Output operations and before channel disposal; cancel timers cleanly, including when language-server shutdown rejects.
- Add deterministic buffer and logger tests for localization, LF behavior on both platform EOLs, ordering, size/timer flushes, reentrant writes, oversized inputs, delayed/rejected shutdown, and no duplicate flushes.

## Validation

- 21 focused unit tests passed.
- Project TypeScript compile/typecheck and ESLint for all touched TypeScript files passed.
- An isolated injected-writer comparison preserved all 53,258,890 bytes from 300,000 messages while reducing append calls from 300,000 to 812 (99.73% fewer), with matching input/output SHA-256 hashes.

## Scope

Batching reduces Output append frequency; it does not reduce native notification count or bound VS Code's total retained Output text. The synthetic measurement does not establish that every renderer out-of-memory scenario is eliminated.


Colengms:
It looks like we're working around an issue in VS Code.  OutputChannel.append() is fire-and-forget: no backpressure, completion signal, queue limits, or documented throughput constraints. If valid high-volume calls can build an unbounded internal queue and OOM, VS Code should coalesce/bound that work or expose a streaming/backpressure API.  Rather than working around this problem ourselves, perhaps we could instead propose a fix for VS Code itself?

sean-mcmanus:
✨Copilot: @Colengms Agreed that the underlying Output scalability issue belongs in VS Code. This PR is only a client-side mitigation, not an alternative to an upstream fix: it reduces append/update frequency while retaining the complete diagnostic stream, but it neither adds backpressure nor bounds the final Output model. The reported append-count reduction is from an injected writer, not a renderer OOM test. Whether to keep this as an interim mitigation or defer it in favor of an upstream fix is a maintainer decision; no upstream fix is included in this PR.

sean-mcmanus:
> It looks like we're working around an issue in VS Code. OutputChannel.append() is fire-and-forget: no backpressure, completion signal, queue limits, or documented throughput constraints. If valid high-volume calls can build an unbounded internal queue and OOM, VS Code should coalesce/bound that work or expose a streaming/backpressure API. Rather than working around this problem ourselves, perhaps we could instead propose a fix for VS Code itself?

I have a VS Code PR -- it's been un-reviewed for 6+ weeks, so I'm reluctant to submit a PR for it, but I could file an issue and see if anyone can agree to review it.

sean-mcmanus:
✨Copilot: @Colengms Sean would like to keep this PR as an interim mitigation: it appeared to avoid the OOM in his high-volume logging repro with `C_Cpp.loggingLevel` set to `7`. That observed benefit is useful without depending on acceptance or a review timeline for an upstream fix. An upstream solution would still be welcome; this remains a mitigation, not a guarantee against OOM from total retained Output text.

sean-mcmanus:
This appears to have been fixed by https://github.com/microsoft/vscode/issues/234393 .

PR #14785, opened 2026-09-18
Fix Actions Octokit response types

sean-mcmanus:
## Summary
Fix the 11 pre-existing `.github/actions` TypeScript errors noted in #14780 without changing issue-triage behavior or dependencies. Derive response types from the actual Actions Octokit client, accept both issue-get and search response shapes, and narrow timeline events by their available fields.

Regenerate the affected adapter JavaScript and add offline tests through the adapter's existing methods. The utility changes in #14781 remain separate; this PR does not change those files.

Add a path-scoped `CI (Actions)` workflow that installs locked development dependencies with lifecycle scripts disabled, runs `npm run typecheck` without emitting JavaScript, and runs `npm test`. The job uses read-only repository permissions and does not retain checkout credentials.

## Validation
- Locked TypeScript 4.9.5 build and typecheck pass, replacing the reproduced baseline of 11 errors.
- ESLint, JavaScript syntax checks, and whitespace checks pass.
- All 22 offline adapter tests pass. An isolated merge with current `main` passes all 39 adapter and utility tests.
- Stale adapter JavaScript makes the npm test command fail only its generated-output parity assertion; typechecking does not overwrite that evidence.
- Workflow validation with actionlint and ShellCheck passes, and the exact install/typecheck/test commands pass locally.
- Tests stub the GitHub client, logging, timers, and subprocesses; no issue-mutating Actions are executed.


.github/actions/api/octokit.test.js
sean-mcmanus, line 1: ✨Copilot (agent228): [Minor] Please wire this suite into PR validation with the Actions dev dependencies. `.github/actions/package.json` has no test script, and `job-compile-and-test.yml` runs the extension harness, which only discovers `Extension/dist/test/unit/**/*.test.js`; neither executes this file. As a result, future adapter/generated-JavaScript drift can pass the current PR checks without running these assertions. I verified that the parity assertion rejects stale `octokit.js` when invoked directly, so an Actions typecheck and `node --test api/octokit.test.js` step would make the new coverage effective.

sean-mcmanus: ✨Copilot: @sean-mcmanus Added a path-scoped `CI (Actions)` workflow with locked development dependencies, `npm run typecheck` (`tsc --noEmit`), and `npm test` (`node --test`). It uses read-only permissions and does not run issue-mutating Actions. The new check is running on the current revision.

The exact commands pass locally: 22 adapter tests on the source branch and all 39 tests with the merged utility changes. A stale `octokit.js` control still passes typechecking but fails the generated-output parity test, confirming CI does not overwrite the evidence before testing.

Colengms (approved):

